PR #10, opened 2019-02-21
Add REST Docs

wborn:
Adds the REST Docs from openhab2-addons. 

If we'd change the feature from `openhab-misc-restdocs` to `openhab-ui-restdocs` we also need to update openhab-distro for this change and users might need to update their configuration.

See also https://github.com/openhab/openhab-core/pull/586#issuecomment-465954881.


cweitkamp:
https://github.com/openhab/openhab2-addons/pull/4954 has been merged.

wborn:
Yes that makes more sense. So I've renamed the bundle @kaikreuzer.
I've also renamed the feature and also created https://github.com/openhab/openhab-distro/pull/881 accordingly.

wborn:
Should the restdocs also be added to the p2-maven-plugins configuration in the [repo/pom.xml](https://github.com/openhab/openhab-webui/blob/master/repo/pom.xml#L37) @kaikreuzer?

kaikreuzer:
Yes, then it would be available in the p2 repo and also available in the TP of the Addons IDE.

PR #31, opened 2019-03-03
Add missing gitignore files

wborn:
I have this strange feeling of déjà vu: https://github.com/martinvw/openhab-webui/pull/1


martinvw:
@wborn AFAIK there is no reason why they should still be generated there? Isn't that some misconfiguration? What should generate them there?

martinvw:
@kaikreuzer the good thing is that the performance looks better this time, I am 90% sure that its not related to the gitignores though.

martinvw:
@wborn maybe we can just gitignore all OSGI-INF directories, given we have 100% coverage (currently)? 

I always disliked the empty folder & gitignores everywhere :-)

wborn:
I once tried using the target directories instead of OSGI-INF but had issues in Eclipse to pick up the generated XML files in plugin unit tests. It's also easier to check if the generated XML is as expected if you can easily see the OSGI-INF content in your IDE which is usually not the case for target directories.

PR #37, opened 2019-03-12
Travis configuration improvements

wborn:
Similar to https://github.com/openhab/openhab-core/pull/643, I've also improved the openhab-webui Travis configuration.

---

This PR makes it possible again to follow progress of a Travis build while it's running. While building it shows the reactor progress information while making sure the build doesn't timeout by appending zero width characters every 3 minutes.

While I was at it I've made some other configuration improvements as well:

* Uses the new Xenial Travis images which easily allows for running Java 11 builds
* Builds using OpenJDK 8 instead of Oracle JDK 8. Oracle JDK 8 is no longer available in the Xenial images. I've done a lot of builds already with it and it works equally well.
* Adds a new OpenJDK 11 build which is allowed to fail. This makes it easy to see what needs to be done for successfully compiling the code with Java 11. It disables the enforcer for the Java 11 build.
* On successful builds only the reactor summary is printed (instead of the most recent 200 lines)
* Removes JVM options that are no longer supported in Java 8 and which result in warnings:
  `-XX:PermSize=512m -XX:MaxPermSize=1g`
* Fixes/adds missing NPM caches for the various UIs

The build times remain the same with this configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,57 @@
 sudo: required
-dist: trusty
+dist: xenial
 
 language: java
-jdk: oraclejdk8
+
+jdk:
+  - openjdk8
+  - openjdk11
 
 cache:
   directories:
-    - $HOME/.m2
-    - $HOME/.p2
-    - extensions/ui/org.eclipse.smarthome.ui.paper/npm_cache
+  - $HOME/.m2
+  - $HOME/.p2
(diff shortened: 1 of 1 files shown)


wborn:
Can you also enable Travis for this repository? Looks like there also isn't a Jenkins PR build yet.

PR #46, opened 2019-04-05
separate fixes for lost function on layout change

maggu2810:
rebase and don't squash if you want to make e.g. integration of SAT separately visible


maggu2810:
@wborn As the configuration of the directory maven plugin has been added to the addons POM tabs has been used instead of two spaces for indentation. To keep the POMs clean and have a consistent indentation that looks nice on every editor, this should be fixed.

Issue #53, opened 2019-04-09
Paper UI: missing patch / customization support


maggu2810:
At the time we migrate ESH to openHAB Core the decision has been taken to move the Paper UI not to openHAB Core but to this repository.
IIRC we agreed to keep the patch / customization support so it will still be able to use another layout.

Now this support has been removed.

Is my memory deceiving me?
Why has the customization support been removed without any public discussion?

kaikreuzer:
No, there was no decision to remove this support, it just happened to have disappeared by some "fixing the build issues".

wborn:
It now shows the following stacktrace again on the snapshot build:

```
22:43:30.755 [INFO ] [org.openhab.ui.paper                 ] - FrameworkEvent INFO - org.openhab.ui.paper
org.osgi.framework.BundleException: The bundle class path entry "patch/" could not be found for the bundle "osgi.identity; osgi.identity="org.openhab.ui.paper"; type="osgi.bundle"; version:Version="2.5.0.201904112024""
	at org.eclipse.osgi.internal.loader.classpath.ClasspathManager.findClassPathEntry(ClasspathManager.java:174) ~[?:?]
	at org.eclipse.osgi.internal.loader.classpath.ClasspathManager.buildClasspath(ClasspathManager.java:152) ~[?:?]
	at org.eclipse.osgi.internal.loader.classpath.ClasspathManager.<init>(ClasspathManager.java:81) ~[?:?]
	at org.eclipse.osgi.internal.loader.EquinoxClassLoader.<init>(EquinoxClassLoader.java:43) ~[?:?]
	at org.eclipse.osgi.internal.loader.BundleLoader.createClassLoaderPrivledged(BundleLoader.java:316) ~[?:?]
	at org.eclipse.osgi.internal.loader.BundleLoader$2.run(BundleLoader.java:239) ~[?:?]
	at org.eclipse.osgi.internal.loader.BundleLoader$2.run(BundleLoader.java:1) ~[?:?]
	at java.security.AccessController.doPrivileged(Native Method) ~[?:?]
	at org.eclipse.osgi.internal.loader.BundleLoader.getModuleClassLoader(BundleLoader.java:236) ~[?:?]
	at org.eclipse.osgi.internal.framework.EquinoxBundle$1.run(EquinoxBundle.java:595) ~[?:?]
	at org.eclipse.osgi.internal.framework.EquinoxBundle$1.run(EquinoxBundle.java:1) ~[?:?]
[...]

PR #54, opened 2019-04-11
Paper UI: restore customization support

maggu2810:
Fixes: https://github.com/openhab/openhab-webui/issues/53


kaikreuzer:
Did you check that this works? I remember that I had seen startup issues of this bundle in the distro since no "patch" folder could be found at runtime.

maggu2810:
I tested it on two systems, one with a fragment bundle and one without a fragment bundle.

kaikreuzer:
Ok, thanks, then let's go for it 👍.

PR #59, opened 2019-04-15
add BOM for the UI bundles

maggu2810:


maggu2810:
No, we don't need it. Removed.

PR #81, opened 2019-06-16
added webaudio feature to HABPanel

kaikreuzer:
Related https://github.com/openhab/openhab-core/issues/821


ghys:
FYI the HABot app doesn't play audio sent to the webaudio sink.

kaikreuzer:
Not or not yet? ;-) If it does not make sense for it, I can remove that line again.

ghys:
Simply completely forgot about it, but it's feasible and can make sense, cannot be ported over straight from HABPanel or Paper UI though.

kaikreuzer:
Ok, I have removed it from HABot again then.

PR #85, opened 2019-07-09
HABot: fix nullness

maggu2810:
After adding the nullness annotations to the automation the type
argument for the base module handler must be a non nullable one.
To apply the non-null annotation to the given type argument we mark the
whole class as non null by default.

Related to: https://github.com/openhab/openhab-core/pull/910


bjoernbrings:
Could it be the case that this change is causing build failure in [sandbox-openhab2-release](https://ci.openhab.org/view/Sandbox/job/sandbox-openhab2-release/1186/execution/node/433/log/ )? Or is it just a problem resulting from openhab/openhab-core#910 ?


[ERROR] /var/jenkins_home/workspace/sandbox-openhab2-release/openhab-webui/bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/notification/WebPushNotificationActionHandler.java:[36,1190] Null constraint mismatch: The type 'Action' is not a valid substitute for the type parameter 'T extends @NonNull Module'

PR #94, opened 2019-08-07
[BasicUI] Render switch element with slide switch for switch item with state options

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc]@free.fr>


kaikreuzer:
Just to help me understand what this is fixing: What do you mean with "slide switch"?

lolodomo:
This is in relation to issue #952.
This PR restores the classic rendering (little slider) for a switch element when linked to a **switch item**, even if the linked item has state options. The current behavior is to render a button for each option.

PR #98, opened 2019-08-14
[BasicUI] Ignore too big item state when updating widget icon

lolodomo:
When reloading a widget icon, Basic UI frontend was building an URL that included the item state without any limit on the size of this state.
This could be a problem when the state was big (like the image data of an Image item).
Item state is now ignored when building the icon URL if its size is greater or equal to 200.


lolodomo:
It fixes the following issue:
https://github.com/openhab/openhab-core/issues/456

PR #112, opened 2019-09-02
Providing modified icons for <motion>

curlyel:
Provides modified icons for <motion> suitable for both, bright _dark_ themes.

The current `motion.svg` and `motion.png` arecolored black which is obviosly not usable on black or dark backgrounds. _And_: These are not dynamic to reflect the current state.

The provided icons are using a light grey (#A0A0A0) for the `closed `state and a red (#CC6666) for the `open `state. So it should be visible on all the standard themes within the WebUI's and and the App's:

MOTION


curlyel:
Gerne doch ;-)

servim:
Dynamic icons only work for open-closed states, can't they work for on / off?

PR #130, opened 2019-10-19
Paper UI: NG rule sort criteria change from uid to name

Martin-Stangl:
Sort criteria of the rule list for NG rules changed to name.

**Reason for change:**

When developing NG rules, it is annoying that they move around in the rule list every time a rule file is updated and therefore reloaded by openHAB.
The same happens for all rules, when openHAB is restarted.

Actually the above statements might not be true in all circumstances, but practically, it is. Namely, when the UID is a randomly generated ID. This is the case when a rule is created from scratch in the Paper UI or when using the class SimpleRule (as done by the openHAB Helper Libraries and probably the Paper UI), which are likely the most common ways to create NG rules. Therefore sorting by UID is sorting by random , which basically is like not sorting at all (or more precise giving it a good shuffle once and then remembering the order).

### DIFF
--- a/bundles/org.openhab.ui.paper/web-src/js/rules/rules.html
+++ b/bundles/org.openhab.ui.paper/web-src/js/rules/rules.html
@@ -13,7 +13,7 @@
 		<div class="container">
 			<p class="text-center" ng-show="data.rules.length == 0">No rules defined yet.</p>
 			<div class="groups">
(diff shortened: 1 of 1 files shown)


hmerk:
I think this is a „won‘t fix“ as it is no critical bug and code for openHAB 2.5.x is feature freezed.
@kaikreuzer WDYT ?

kaikreuzer:
@hmerk Yes, indeed.

PR #142, opened 2019-11-20
norwegian translation of HABot

jmelhus:
Signed-off-by: jmelhus <post@jmelhus.com>

### DIFF
--- a/bundles/org.openhab.ui.habot/src/main/resources/answers_no.properties
+++ b/bundles/org.openhab.ui.habot/src/main/resources/answers_no.properties
@@ -1,0 +1,87 @@
+greeting =              Velkommen! Hva kan jeg hjelpe deg med? |\
+                        Hei, hva kan jeg hjelpe deg med? |\
+                        Hallo, hva kan jeg gjøre for deg? |\
+                        Klar til å hjelpe!
+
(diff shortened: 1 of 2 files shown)


jmelhus:
What about translation of openhab-webui\bundles\org.openhab.ui.habot\src\main\resources\train, are those generated or do they need manual translation as well?

ghys:
> What about translation of openhab-webui\bundles\org.openhab.ui.habot\src\main\resources\train, are those generated or do they need manual translation as well?

This is the training data, arguably the most important part, it needs to be done manually for every language, and doesn't need to be an exact translation from another language but is ideally tailored to the particularities of the language. It is also validated through a series of unit tests to make sure common queries are properly categorized and the named entities properly extracted. The maintainer of a particular language (that would be you, for Norvegian ;)) is trusted to come up with a sensible test suite and adjust the training until the tests pass.

In short, yes, you need to write these files, as well as a test suite similar to, but not necessarily exactly matching those for the other languages: https://github.com/openhab/openhab-webui/tree/master/bundles/org.openhab.ui.habot/src/test/java/org/openhab/ui/habot/test

florian-h05:
Closing as this PR went stale.

PR #154, opened 2019-12-21
Removed Paper UI, Classic UI and HABmin

kaikreuzer:
As decided in https://github.com/orgs/openhab/teams/architecture-council/discussions/1?from_comment=5#discussion-1-comment-5:

> - We deprecate Dashboard / Paper UI / HABmin / Classic UI
> - The aim of the new UI should be to also replace Basic UI / HomeBuilder and HABot on the long run; I'd leave this open for now and leave it to how progress goes. But this should imho be our vision.


kaikreuzer:
FTR: I didn't yet remove the dashboard, because the existing UIs have a dependency on it (by providing a dashboard tile).

@ghys Once https://github.com/openhab/openhab-core/issues/1290 is done, it would be great if you could move the code of the new UI (refactored to use the `org.openhab` namespace in core) to this repo (and remove the dashboard in this process).

@peuter The CometVisu wasn't in scope of the decision, mainly because nobody had it on the radar. I have to admit that I have no clue how many users this has for openHAB and whether it makes sense to continue supporting it for openHAB 3. What's your opinion on this?

Hilbrand:
Dashboard seems to be removed from the bundle pom.xml, and the classic pom.xml is still present.

wborn:
Can you also update the CODEOWNERS?

wborn:
If will be easier to replace the old JAX-RS implementation without cometvisu. :wink:

See: https://github.com/openhab/openhab-core/issues/726#issuecomment-561832261

ghys:
> @ghys Once openhab/openhab-core#1290 is done, it would be great if you could move the code of the new UI (refactored to use the org.openhab namespace in core) to this repo (and remove the dashboard in this process).

Will do, after the code gets in good shape, I hope to get some work done during the holiday season. Any suggestions for the bundle's name?
The dashboard tiles will also be featured in the UI after a REST endpoint is provided to expose them.

> The aim of the new UI should be to also replace Basic UI / HomeBuilder and HABot on the long run; I'd leave this open for now and leave it to how progress goes. But this should imho be our vision.

Still having a different opinion on simply removing these, as opposed to leaving them be mature, dedicated & more lightweight takes on specific use cases, but I guess it'll depend on the evolution of sitemaps and the HABot card system. I think we should open separate issues for discussion of the future of these concepts. The main UI will nonetheless be a sitemap viewer and an HABot client so using Basic UI or the current HABot UI would be optional.

peuter:
Well there are cometvisu users out there and I am one of them. I don't know how many they are, but I obviously have a strong intererest in continuing the openHAB<->CometVisu support.

kaikreuzer:
@Hilbrand Thanks, my Git client caused some issues there - should be fixed now.
@wborn CODEOWNERS file is also updated.

wborn:
This Travis cache can also be removed:

https://github.com/openhab/openhab-webui/blob/185180715389b499579ee83b4eef54cc91049514/.travis.yml#L16

kaikreuzer:
Thanks @wborn, both done.

kgoderis:
@kaikreuzer @wborn This PR breaks https://github.com/openhab/openhab-distro/blob/master/launch/app/app.bndrun

bwosborne2:
Should Home Builder be removed too?

kaikreuzer:
I didn't dare to suggest that yet, but if we had some simple wizard in the new UI for the initial setup, I'd be all for it. Wdyt, @ghys?

bwosborne2:
I thought its main purpose was to create sitemaps for the now defunct Basic UI. 
I now notice I guess it has not yet been removed though.

ghys:
Yes, since its purpose is pretty much to help when beginning with a blank system, that was my idea as well - the UI would offer some kind of (optional, skippable) setup wizard on first run to configure some common things, maybe install some common add-ons to get the user started, and also eventually build an initial "skeleton" of items (Home Builder's main use case) would be one of its steps. Home Builder's code and other assets - translations etc. - should be reusable rather easily since the tech is similar.
As for sitemaps, it depends what this concept will become so I won't comment on these for now.

PR #163, opened 2020-01-01
make sure that dashboard is installed as an UI add-on

kaikreuzer:
fixes https://github.com/openhab/openhab-distro/issues/1031


kaikreuzer:
Hm, testing this showed some unwanted side-effects: When installing a distro package, the UIs won't show up on the dashboard anymore...

PR #178, opened 2020-01-18
Add default UI bundle

ghys:
This add the UI in its current state of development,
the intent is to test the integration in the distro
early. The other apps are listed on the home page by
calling the new REST endpoint introduced in
https://github.com/openhab/openhab-core/pull/1329

The onboarding wizard implementation is not complete
so it isn't possible to select a startup package with
the UI yet.  The wizard steps will be discussed in a
separate issue; it should however be possible to install
addons as needed from the settings screen.


ghys:
Since the Travis build looks good, I think we can merge this one now so we can test https://github.com/openhab/openhab-distro/pull/1065.
I will continue the development on another branch.

kaikreuzer:
Give me a day to roughly review >33.000 loc before merging it unseen 😎.

bundles/org.openhab.ui/README.md
```diff
@@ -0,0 +1,3 @@
+# Default UI
```
kaikreuzer: We will definitely have to come up with some proper naming.
"Default" sounds too technical to me, I'd actually prefer to just call it the "openHAB UI". Or maybe to better distinguish it from the apps, the "Web UI". Or maybe the "Standard UI"...

Also, we have to decide whether this README gets consumed by the documentation build or if we have all infos about this UI in the docs repo (as it is spread across many sections, this is probably the better choice).

bundles/org.openhab.ui/web/README.md
```diff
@@ -0,0 +1,68 @@
+# openHAB
```
kaikreuzer: this might want to have a better title

bundles/org.openhab.ui/web/cordova/config.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version='1.0' encoding='utf-8'?>
+<widget id="org.openhab.ui" version="3.0.0" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
+    <name>openHAB</name>
+    <description>
+        A sample Apache Cordova application that responds to the deviceready event.
```
kaikreuzer: Err, no.

bundles/org.openhab.ui/web/cordova/config.xml
```diff
@@ -0,0 +1,92 @@
...
+    </description>
+    <author email="dev@cordova.apache.org" href="http://cordova.io">
```
kaikreuzer: Nope.

bundles/org.openhab.ui/web/src/index.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <!--
```
kaikreuzer: Can we remove this comment?

bundles/org.openhab.ui/README.md
ghys: Agreed that's why I simply named the package "org.openhab.ui" instead of "org.openhab.ui.default" or something like this (it wouldn't accept "default" in a package name anyway). I'll come up with a proper README later.

> Also, we have to decide whether this README gets consumed by the documentation build or if we have all infos about this UI in the docs repo (as it is spread across many sections, this is probably the better choice).

Since this would be the go-to UI I think would be indeed be described in the docs as part of openHAB instead of a specific feature.

bundles/org.openhab.ui/web/README.md
ghys: This file was actually auto-generated when scaffolding the project, not sure if it's really useful since there would already be a README in the parent folder. Some information on how to start the development server and contribute etc. could appear in a separate CONTRIBUTING.md.

bundles/org.openhab.ui/web/cordova/config.xml
ghys: Good catch - I left the Cordova stuff even if we're not sure yet whether we would publish Cordova builds in app stores and so on - we'll discuss it when the time comes, but definitely could, I regularly test iOS and Android builds, there's a additional login screen but overall it's a working full-screen version of the web app without the browser UI. Adding to the home screen from Safari or Chrome works too.

PR #202, opened 2020-03-24
Use latest node & npm versions

ghys:
Harmonize the node/npm versions for bundles using frontend-maven-plugin
Fix the Basic UI build


wborn:
While you're at it, you might also want to bump the frontend-maven-plugin to something newer.
Version 1.0 is ancient (Apr 13, 2016) , see also the [changelog](https://github.com/eirslett/frontend-maven-plugin/blob/master/CHANGELOG.md).

ghys:
The build now succeeds again.

PR #208, opened 2020-03-29
Adds an additional zoom visibility property to markers to show/hide b…

digitaldan:
…ased on current zoom level.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-marker.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-marker.vue
@@ -136,7 +136,13 @@ export default {
           type: 'TEXT',
           context: 'color',
           description: 'Color of the tooltip'
-        }
+        },
+        {
+          name: 'zoomVisibility',
+          label: 'Zoom Visibility',
+          type: 'NUMBER',
+          description: 'Visible only when zoomed to at least this level (empty by default)'
+        },
       ]
     }
   },
--- a/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-marker.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-marker.vue
@@ -141,8 +141,9 @@ export default {
           name: 'zoomVisibility',
           label: 'Zoom Visibility',
           type: 'NUMBER',
-          description: 'Visible only when zoomed to at least this level (empty by default)'
-        },
+          description: 'Visible only when zoomed to at least this level (empty by default)',
+          advanced: true
+        }
       ]
     }
   },
--- a/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
@@ -24,7 +24,7 @@
         :bounds="bounds"
       />
       <l-feature-group ref="featureGroup" v-if="context.component.slots && ready">
-        <component v-for="(marker, idx) in context.component.slots.default" :key="idx"
+        <component v-for="(marker, idx) in markers" :key="idx"
           :is="markerComponent(marker)" :context="childContext(marker)" @update="onMarkerUpdate" />
       </l-feature-group>
       <l-control v-if="context.editmode != null" position="topright">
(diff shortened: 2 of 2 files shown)


digitaldan:
zoomVisibility

digitaldan:
Hi @ghys was playing around with a way to be able to zoom into a map and display more markers based on the zoom level.  Not sure if this is the best way to handle that. or if i event did it right :-)

ghys:
Thanks @digitaldan, that's a nice feature! My only (small) concern is that users would need to figure out by themselves what the zoom level is, since it's neither shown anywhere nor retrieved easily, so it's not an option for everyone. Therefore I'd like to suggest making it an "advanced" option, meaning it will only be shown after checking a "show advanced options" box. Doing so is easy, the widgets have the same config parameter description format as other parts in openHAB so you just have to add `advanced: true` in your description, the checkbox will appear automatically.

bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-marker.vue
```diff
@@ -136,7 +136,13 @@ export default {
...
           context: 'color',
           description: 'Color of the tooltip'
-        }
+        },
+        {
+          name: 'zoomVisibility',
+          label: 'Zoom Visibility',
+          type: 'NUMBER',
+          description: 'Visible only when zoomed to at least this level (empty by default)'
+        },
```
ghys: Please remove the trailing comma here, and add `advanced: true,` above (3rd line or so) as discussed above.

bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
```diff
@@ -170,6 +170,14 @@ export default {
       zoomControl:false
       } : {})
     },
+    markers(){
```
ghys: Please add spaces before and after the braces.

ghys:
@digitaldan actually there quite a few linting errors in this and your previous PR (I overlooked it last time) 😅

image

I have a laxer approach than most of the other openHAB projects when it comes to enforcing them, in that they would not fail the build, but still it would be great if could sort them out while you're at it 😄.

If you're using VS Code as your IDE, searching for ESLint extension and installing `dbaeumer.vscode-eslint` should do the trick. When developing Vue files you'll want to install Vetur (`octref.vetur`) as well.

Then it should only be a matter of looking for the little lightbulb icons when hovering over code or in the problems window, and selecting 'Fix all auto-fixable problems':

image

Let me know how this works for you so we can eventually add this to the CONTRIBUTING file.

digitaldan:
>My only (small) concern is that users would need to figure out by themselves what the zoom level is

I was thinking the same thing.  I had played with displaying the zoom level when editing, but i think making it an advanced property makes sense. I may still play with displaying something.

>I have a laxer approach than most of the other openHAB projects when it comes to enforcing them

I would not mind make making this stricter actually. There is a lot of code, and it really helps to have consistency.  I don't mind making sure the code is properly linted before checking in.. 

> extension pane and installing dbaeumer.vscode-eslint

Nice, that worked perfectly.  I will start paying attention to the "problems" tab now :-)

digitaldan:
zoomVisibility

I added a zoom display when editing, thoughts?

ghys:
> I added a zoom display when editing, thoughts?

I like it! ;)

> I would not mind make making this stricter actually. There is a lot of code, and it really helps to have consistency. I don't mind making sure the code is properly linted before checking in..

Yes, that could be valuable, you could go further and add the `vue/recommended` rules for [stricter rules](https://vuejs.github.io/eslint-plugin-vue/rules/) (right now there's only `vue/essential` and `@vue/standard` in .eslintrc), also why not trying out [Prettier](https://prettier.io), which is widely praised, as well and checking a default VS Code settings file to fix problems on save.
Some details here: https://medium.com/@gogl.alex/how-to-properly-set-up-eslint-with-prettier-for-vue-or-nuxt-in-vscode-e42532099a9c.
For now I was happy with only having to fix basic problems highlighted in the editor but if there are more contributions over time we should probably have a plan to keep the code clean. I just don't want to deal with the backlog right now 😄

digitaldan:
FYI, there was an issue with markers flickering when zooming, my last push fixes this, but may seem counterintuitive at first glance

digitaldan:
Ignore my last comment and commit, there was something else goofy going on with my browser.

PR #214, opened 2020-04-07
Add null annotations to SubscriptionProvider

wborn:
Without these annotations the build will fail when https://github.com/openhab/openhab-core/pull/1412 gets merged.

### DIFF
--- a/bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/notification/internal/webpush/Subscription.java
+++ b/bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/notification/internal/webpush/Subscription.java
@@ -36,13 +36,9 @@ public class Subscription implements Identifiable<Keys> {
     }
 
     public class Keys {
-        public @NonNullByDefault({}) String p256dh;
+        public final String p256dh;
 
-        public @NonNullByDefault({}) String auth;
-
-        public Keys() {
-            // No-args constructor
-        }
+        public final String auth;
 
         public Keys(String p256dh, String auth) {
             this.p256dh = p256dh;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/notification/internal/webpush/Subscription.java
```diff
@@ -24,21 +24,21 @@
...
         this.endpoint = endpoint;
         this.keys = keys;
     }
 
     public class Keys {
-        public String p256dh;
+        public @NonNullByDefault({}) String p256dh;
 
-        public String auth;
+        public @NonNullByDefault({}) String auth;
```
wborn, line 41: Do you know if the default constructor is unnecessary @ghys?
If so, we can remove it and make these fields final without this annotation.

ghys: Hi @wborn, yes it appears to be unnecessary. It actually comes from https://github.com/web-push-libs/webpush-java/blob/master/src/main/java/nl/martijndwars/webpush/Subscription.java, I don't remember the exact reason why I didn't simply use the library, I think I needed some changes.

wborn: Gone it is! :+1:

wborn:
You might also want to merge this @openhab/webui-maintainers because builds will likely fail since https://github.com/openhab/openhab-core/pull/1412 got merged.

ghys:
Thank you @wborn!

PR #217, opened 2020-04-14
[basicui] Use command options for rendering Switch and Selection elements

cweitkamp:
- Use command options for rendering Switch and Selection elements

Related to https://github.com/openhab/openhab-core/issues/1016

grafik

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SelectionRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SelectionRenderer.java
@@ -12,16 +12,15 @@
  */
 package org.openhab.ui.basic.internal.render;
 
-import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.emf.common.util.EList;
 import org.eclipse.smarthome.core.items.Item;
 import org.eclipse.smarthome.core.items.ItemNotFoundException;
 import org.eclipse.smarthome.core.library.items.NumberItem;
 import org.eclipse.smarthome.core.library.types.QuantityType;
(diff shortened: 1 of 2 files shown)


kaikreuzer:
As this is in some way breaking existing sitemaps of users: Should we maybe keep that change for 3.0 and commit it on master only?

cweitkamp:
Yes, should be okay to merge it against master. I am not aware of any existing binding using command descriptions.

On the other hand I do not see a breaking change because the implementation of the `getCommandDescription()` method converts all existing `StateOption`s to `CommandOption`s in case no dedicated `CommandDescription` is provided.

https://github.com/openhab/openhab-core/blob/fb7a7ac42173d3228b7e4447582e827aa541e9ac/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java#L415-L430

ghys:
What if there are more than a couple of command options and the buttons don't fit?

kaikreuzer:
> On the other hand I do not see a breaking change

Ok, fair point, so merging it here is ok for me.
But still: It'll be more important on master, so I hope you will port it and provide another PR there :-)

cweitkamp:
@kaikreuzer Done (see https://github.com/openhab/openhab-webui/pull/220). Easy going - thanks to cherry-pick.

kaikreuzer:
That was quick, thanks :-)
Did the cherry picking work even though the namespace of classes has changed? I guess you needed to help this a bit to resolve!

cweitkamp:
It worked in general. Yes, I had to solve merge conflicts due to namespace change. But only for the imports - which is not a big deal.

lolodomo:
I think that the default should remain a selection. The buttons are not adapted when you have 50 options !

cweitkamp:
The decision for the default element is done in OHC (https://github.com/openhab/openhab-core/pull/1422).

PR #226, opened 2020-04-16
Items: metadata & group management, improve default rendering

ghys:
Add the ability to edit metadata from the items details page:

- Provide interactive forms for the alexa, ga, synonyms, widget,
  listwidget, stateDescription, commandDescription namespaces
- Fallback to YAML view for other namespaces (or in a tab for
  supported namespaces as an alternative).

Add "options" widget action to open an action sheet with a list
of command options.

Add default representation for items: look in the widget for
a preconfigured widget (custom or from the standard library),
or try to build a default live representation of the item's state
(WIP), considering its type, state/command descriptions and
other properties.

Change the "current state" part of the items details & link edit
pages with the default representation.

Add a convenient way to add or remove members to/from a group item.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/metadata/alexa.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/metadata/alexa.js
@@ -36,6 +36,43 @@ const categories = [
   'WEARABLE'
 ]
 
+const labels = {
+  'Switchable': [],
+  'Lighting': [],
+  'Blind': [],
+  'Door': [],
+  'Lock': [],
+  'Outlet': [],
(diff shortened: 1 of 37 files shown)


digitaldan:
Just playing around with this and WOW!

ghys:
Yep, I don't use neither Alexa nor Google so the metadata description was done by browsing and grepping the docs, there might be some mistakes here and there, but they're in their own files in the assets folder - no need to dive in the code.
So I wouldn't mind delegating the review/maintenance of these :)
If there are more metadata namespaces in use that I'm not aware of they can be added there as well.

PR #228, opened 2020-04-18
avoid rendering markers on every view change

digitaldan:

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
@@ -153,7 +153,8 @@ export default {
       zoom: -0.5,
       crs: CRS.Simple,
       showMap: true,
-      mapKey: this.$f7.utils.id()
+      mapKey: this.$f7.utils.id(),
+      markers: []
     }
   },
   computed: {
@@ -172,20 +173,6 @@ export default {
         scrollWheelZoom: false,
         zoomControl: false
       } : {})
-    },
-    markers () {
-      var newMarkers = this.context.component.slots.default.filter((e) => {
-        const zoomVisibilityMin = parseFloat(e.config.zoomVisibilityMin)
-        const zoomVisibilityMax = parseFloat(e.config.zoomVisibilityMax)
-        const isVisibleMin = isNaN(zoomVisibilityMin) || zoomVisibilityMin < this.currentZoom
-        const isVisibleMax = isNaN(zoomVisibilityMax) || zoomVisibilityMax > this.currentZoom
-        return this.context.editmode != null || (isVisibleMin && isVisibleMax)
-      })
-      // Only return new markers if the list has changed to avoid rendering on every zoom change
-      return this.context.component.slots.default.length !== newMarkers.length ||
-        JSON.stringify(this.context.component.slots.default) !== JSON.stringify(newMarkers)
-        ? newMarkers
-        : this.context.component.slots.default
     }
   },
   mounted () {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
```diff
@@ -174,13 +174,18 @@ export default {
...
+      var newMarkers = this.context.component.slots.default.filter((e) => {
         const zoomVisibilityMin = parseFloat(e.config.zoomVisibilityMin)
         const zoomVisibilityMax = parseFloat(e.config.zoomVisibilityMax)
         const isVisibleMin = isNaN(zoomVisibilityMin) || zoomVisibilityMin < this.currentZoom
         const isVisibleMax = isNaN(zoomVisibilityMax) || zoomVisibilityMax > this.currentZoom
         return this.context.editmode != null || (isVisibleMin && isVisibleMax)
       })
+      // Only return new markers if the list has changed to avoid rendering on every zoom change
+      return this.context.component.slots.default.length !== newMarkers.length ||
+        JSON.stringify(this.context.component.slots.default) !== JSON.stringify(newMarkers)
```
digitaldan, line 186: So this feels dirty to me, but short of pulling something like lodash into the project, seems like the simplest way to compare objects, considering the arrays in question will be small.  Open to other suggestions :-)

digitaldan:
FYI, this is to address flicking when zooming in or out, i noticed it again when on my mobile since i introducing marker visibility .

bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
ghys: maybe something like:
```js
const markers = this.context.component.slots.default
if (!markers.find((m) => !isNaN(parseFloat(m.config.zoomVisibilityMin) ||
    !isNaN(parseFloat(m.config.zoomVisibilityMax)) {
  return markers
}
return markers.filter(...)
```
?

digitaldan: I think you are suggesting that we return the default set of markers if non have zoom visibility set. If at least one marker does have this set, we will always return a new filtered array, which is what is happening now and causing flickering.  So we would still have the issue if zoom visibility is set on at least one marker. 

The problem i have observed is that if your return a new instance of the markers object , Vue will re-render the list, causing flickering issues when zooming.  By returning the same reference to the markers object when nothing has changed, Vue seems to not trigger a render.  This goes against how i thought Vue uses a virtual DOM, but nevertheless is what i am observing. 

If each marker had a unique id, then iterating on the list of marker objects would be trivial , but I don't think components have such an id?

ghys: > I think you are suggesting that we return the default set of markers if non have zoom visibility set. If at least one marker does have this set, we will always return a new filtered array, which is what is happening now and causing flickering. So we would still have the issue if zoom visibility is set on at least one marker.

Right! 🤦

I didn't experience any flickering when zooming, but if you change the zoom level then it's computed properties which are using it will be reevaluted, so that makes sense.

Instead of using JSON.stringify maybe you could compare with something like:
image
in your case:
```js
const allMarkers = this.context.component.slots.default
return (newMarkers.every((e) => allMarkers.indexOf(e) >= 0) ? allMarkers : newMarkers
```
it's slightly more elegant ;)
(and maybe rename newMarkers to visibleMarkers or something like that)

digitaldan: 🤦 So for some bizarre  reason i was not thinking about object equality in terms of object references, i had convinced myself i needed to deep compare objects, oof. 

So your suggestion (and much more elegant btw) is much appreciated. I just updated the PR with a slightly modified approach as i need to compare the current displayed list to the newly computed list to know what has changed.  This works nicely.

digitaldan:
Before this change and after FYI 
zoomVisibility

PR #229, opened 2020-04-18
Color picker control

ghys:
Make toggle controls compatible with Color items/HSB values


bundles/org.openhab.ui/web/src/components/widgets/system/oh-colorpicker.vue
```diff
@@ -0,0 +1,115 @@
...
+            }
+          }, 200)
+        }
+        setTimeout(() => {
+          this.updateValue(this.pendingUpdate)
+          this.delayUpdate = false
+        }, 2000)
+      }
+    },
+    commandFromHSB (hsb) {
```
digitaldan, line 97: Question about common functions like this, I could see many widgets needing to do similar conversions, should there be a mix-in or common class that would contain common conversion like this?  For example, i can see both a switch and slider needing to share common util functions as both can operate on a Dimmer item and may need to convert ON, OFF, 0 , 100, etc..... I'm more curious what your opinion here is for others (like me).

ghys: Yes you're absolutely right, there's some refactoring to do in the widgets and in other places to share common code, mixins might be a good choice... there's a widget-mixin already which does the evaluation of expressions in the config and a few other things, there could be some utility functions in there, or maybe another mixin dedicated to conversions.

PR #237, opened 2020-04-23
nearley items parser: avoid recognizing item types in names

### LINKED ISSUE #233: [Default-UI] Cannot parse group names if the name starts with 'Group'
Hey @ghys , i tried looking into this myself, but could not decipher the nearly parser.  

If I try to import items that have a Group definition who's group name starts with the word "Group", the parser throws and error:

```
Group Group_Sonos
Group Group_Sonos_GR (Group_Sonos)
Group Group_Sonos_Patio (Group_Sonos)
Group Group_Sonos_MasterBed (Group_Sonos)
Group Group_Sonos_GR_Control (Group_Sonos_GR)
Group Group_Sonos_Patio_Control (Group_Sonos_Patio)
Group Group_Sonos_MasterBed_Control (Group_Sonos_MasterBed)

String Sonos_GR_currentTrack  	"Track [%s]"  		(Group_Sonos_GR)   {channel="sonos:CONNECT:RINCON_000E58AF01BE01400:currenttrack"}
String Sonos_GR_currentArtist  	"Artist [%s]"  		(Group_Sonos_GR)   {channel="sonos:CONNECT:RINCON_000E58AF01BE01400:currentartist"}
String Sonos_GR_currentAlbum  	"Album [%s]"  		(Group_Sonos_GR)   {channel="sonos:CONNECT:RINCON_000E58AF01BE01400:currentalbum"}
```

and the error:

```
Error: Syntax error at line 1 col 7:

  Group Group
        ^
Unexpected itemtype token: "Group". Instead, I was expecting to see one of the following:

A comment token based on:
    __ → __ ● %comment
    Item → Type ● __ Name Label Icon Groups Tags Metadata
    Items →  ● Item
    Main → _ ● Items _
A NL token based on:
    __ → __ ● %NL
    Item → Type ● __ Name Label Icon Groups Tags Metadata
    Items →  ● Item
    Main → _ ● Items _
A WS token based on:
    __ → __ ● %WS
    Item → Type ● __ Name Label Icon Groups Tags Metadata
    Items →  ● Item
    Main → _ ● Items _
A identifier token based on:
    Name →  ● %identifier
    Item → Type __ ● Name Label Icon Groups Tags Metadata
    Items →  ● Item
    Main → _ ● Items _
```

image

ghys:
Fixes #233.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/items-lexer.nearley
+++ b/bundles/org.openhab.ui/web/src/assets/items-lexer.nearley
@@ -1,26 +1,27 @@
 @{%
-	const moo = require('moo')
+  const moo = require('moo')
 
-	let lexer = moo.compile({
-		WS:         /[ \t]+/,
-		comment:    /\/\/.*?$/,
-		number:     /0|[1-9][0-9]*/,
-		string:     { match: /"(?:\\["\\]|[^\n"\\])*"/, value: x => x.slice(1, -1) },
-		itemtype:   ['Group', 'Number', 'Switch', 'Rollershutter', 'String', 'Dimmer', 'Contact', 'DateTime', 'Color', 'Player', 'Location', 'Call', 'Image'],
-		identifier: /[A-Za-z0-9_-]+/,
-		lparen:  	'(',
-		rparen:  	')',
-		colon:   	':',
-		lbrace: 	'{',
-		rbrace: 	'}',
-		lbracket: 	'[',
-		rbracket: 	']',
-		lt: 		'<',
-		gt: 		'>',
-		comma: 		',',
-		equals: 	'=',
-		NL:      	{ match: /\n/, lineBreaks: true },
-	})
+  let lexer = moo.compile({
+    WS:         /[ \t]+/,
+    comment:    /\/\/.*?$/,
+    number:     /0|[1-9][0-9]*/,
+    string:     { match: /"(?:\\["\\]|[^\n"\\])*"/, value: x => x.slice(1, -1) },
+    itemtype:   ['Group ', 'Number ', 'Switch ', 'Rollershutter ', 'String ', 'Dimmer ', 'Contact ', 'DateTime ', 'Color ', 'Player ', 'Location ', 'Call ', 'Image '],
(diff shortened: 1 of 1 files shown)


ghys:
@digitaldan had a second look since I remembered having the same problem with the sitemap parser and solved it that way. It seems to work with the huge items declaration snippet here https://www.openhab.org/addons/bindings/ecobee/#items (which wasn't working before)
without a significant drop in performance, can you confirm it solves your issue as well?

The caveat is that now the character after an item type must be a space - i.e. not a tab or newline.
Also things like `Switch Switch "My Switch"` or `String Location (Group)` still don't work but I hope nobody does that.

digitaldan:
>without a significant drop in performance, 

Do you have a version that parses completely, but is less performant?  Is performance an issue here?

>The caveat is that now the character after an item type must be a space - i.e. not a tab or newline

Could we replace tabs with newlines before parsing the file ?  

>Also things like Switch Switch "My Switch" or String Location (Group) still don't work but I hope nobody does that.

I know we will run into the label containing "Switch" a lot ("Light Switch", "On/OFF Switch", etc...) 

> can you confirm it solves your issue as well?

I will give it a try today.  Thanks!

digitaldan:
> > Also things like Switch Switch "My Switch" or String Location (Group) still don't work but I hope nobody does that.
> 
> I know we will run into the label containing "Switch" a lot ("Light Switch", "On/OFF Switch", etc...)

Nm, i didn't read that right, you meant the Item name, not label.  I'm not sure how much that would be an issue, i would hope no one does that in a real (non test/demo) system as well.

digitaldan:
>can you confirm it solves your issue as well?

Yep, my items now import, thanks!

ghys:
> Do you have a version that parses completely, but is less performant? Is performance an issue here?

Writing and adjusting Nearley grammars is quite sensitive because if you have the slightest ambiguity, it will consider every alternative and yield every possible result, and the parsing performance will deteriorate exponentially, or even go into an infinite recursion. It doesn't seem to be the case with this change though :)

> Could we replace tabs with spaces before parsing the file ?

Yes probably a good idea. Will consider that if the problem surfaces.

> Yep, my items now import, thanks!

Good, let's merge then!

PR #243, opened 2020-05-03
Better support for non-editable rules

ghys:
Depends on https://github.com/openhab/openhab-core/pull/1451.

Change the rules list to a "contacts" list with initials in
headers and a list index.
Add a lock icon to denote rules which cannot be edited with the
rule editor.

Strip all editing features in the rule editor when the rule is
not editable.

Move the module title/description suggestions to a mixin and add
the new group member triggers.

Fix bug, look & feel in tag editor.


ghys:
I'll merge this because it doesn't actually depend on https://github.com/openhab/openhab-core/pull/1451, if the editable flag isn't present it assumes that it is.

PR #249, opened 2020-05-16
Add null annotations to REST resources

wborn:
Depends on https://github.com/openhab/openhab-core/pull/1475.

There were still missing whiteboard annotations on `HABPanelResource` which are also added.

### DIFF
--- a/bundles/org.openhab.ui.cometvisu/src/main/java/org/openhab/ui/cometvisu/internal/backend/rest/ChartResource.java
+++ b/bundles/org.openhab.ui.cometvisu/src/main/java/org/openhab/ui/cometvisu/internal/backend/rest/ChartResource.java
@@ -39,6 +39,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.io.rest.RESTConstants;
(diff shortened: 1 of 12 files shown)


wborn:
I've removed required from PathParams because they are always required:

```
    /**
     * Specifies if the parameter is required or not.
     * <p>
     * Path parameters will always be set as required, whether you set this property or not.
     */
    boolean required() default false;
```

wborn:
I also have a question for you @ghys:

The language HeaderParams in HABotResource seem to be currently ignored.

https://github.com/openhab/openhab-webui/blob/586cea4444684c5c1e1f78a9c9e8c5bb9ad899b8/bundles/org.openhab.ui.habot/src/main/java/org/openhab/ui/habot/rest/internal/HABotResource.java#L125-L127

Is that because it still has to be implemented in the UI before it can be used?

ghys:
> The language HeaderParams in HABotResource seem to be currently ignored.

Yes I intended to use it at first, the browser's configuration would define the language, but it was impractical - some browsers would be set up in a unsupported language, or some would be in English and other in another language, forcing a retraining on each change... In the end it was more logical to use the locale defined in openHAB's i18n service, so this parameter can be removed.

wborn:
I already thought so and will then remove it before someone will create an issue that it doesn't work. :wink:

ghys:
Great, be my guest ;)

ghys:
Thanks!

PR #255, opened 2020-05-18
Move the login hint in the sidebar

### LINKED ISSUE #254: [defaultui] "Click on the shield" help text confusing when shield does not show
The side menu isn't necessarily showing when using a small browser window or accessing the UI on a smartphone

As a result users may get confused when they see this otherwise very helpful text:

shield

ghys:
Displayed only when there are no pages - if there are, it means the admin
user already found the button.
Closes #254.

Change some label texts & empty overview icon.

Make the whole logo area clickable.

Display an indicator when there are no pages to display on the sidebar.


ghys:
New look:

image

image

PR #257, opened 2020-05-19
REST docs: use the redirect URI as client ID

ghys:
Depends on https://github.com/openhab/openhab-core/pull/1496


kaikreuzer:
> Depends on openhab/openhab-core#1483

Which is closed now. Do you plan to do a new PR with a rewrite of the class in question? Or will you first do a change in the upstream project?

ghys:
May catch some heat with that, but while experimenting I ended removing the library completely 🤷‍♂️. See https://github.com/openhab/openhab-core/pull/1496.

kaikreuzer:
Wow, excellent, sounds like an even better solution 😄.

ghys:
No harm in merging this early I guess, it'll sit there until the server-side comes.

PR #258, opened 2020-05-20
Use constructor injection and update null annotations

wborn:
Depends on https://github.com/openhab/openhab-core/pull/1487


wborn:
Without this PR webui builds will fail @openhab/webui-maintainers.

PR #259, opened 2020-05-20
[WIP] Format relative URLs for cordova/mobile builds.

digitaldan:

I noticed that when using the cordova version of the app on IOS, my images for pages were not loading. I have these as static files with relative links (much like habpanel) on OH,for example,  `/static/ground_floor.svg` 

This was a quick and dirty fix on my part to see my pages working, so i'm not actually suggesting the current PR for submission, but i am curious if this should be expanded across the whole app (like in a mixin)  to widgets, or is there another/better way?

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
@@ -20,7 +20,7 @@
     @update:center="centerUpdate"
     @update:zoom="zoomUpdate">
       <l-image-overlay
-        :url="config.imageUrl"
+        :url="prepareRequest(config.imageUrl)"
         :bounds="bounds"
       />
       <l-feature-group ref="featureGroup" v-if="context.component.slots && ready">
@@ -216,6 +216,10 @@ export default {
       this.$nextTick(() => {
         this.fitMapBounds()
       })
+    },
+    prepareRequest (url) {
+      // format the URL for codova if not a relative link
+      return !url || url.indexOf('http') === 0 ? url : this.$oh.api.prepareRequest(url)
     }
   }
 }
--- a/bundles/org.openhab.ui/web/src/js/openhab/cordova/api.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/cordova/api.js
@@ -19,6 +19,7 @@ function prepareRequest (uri) {
 }
 
 export default {
+  prepareRequest,
   get (uri, data) {
     const fullUri = prepareRequest(uri)
     if (fullUri) {
--- a/bundles/org.openhab.ui/web/src/js/openhab/openhab.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/openhab.js
@@ -60,6 +60,9 @@ export default {
     accessToken = token
   },
   api: {
+    prepareRequest (uri) {
+      return uri
+    },
     get (uri, data) {
       return wrapPromise(Framework7.request.promise.json(uri, data))
     },


ghys:
I had no idea you built Cordova apps 😄 the Cordova support is currently not in a good shape, I let it fall behind lately and didn't even bother implementing the API for the SSE state updates (so that part should be broken, the current SSE implementation is also quite ugly), and I'm not sure how Cordova apps work overall as it's been some time since I last built one. Would love to have some feedback :)

On the relative URL part and general media handling I believe you're right, it's even more of an headache because there might be an extra layer of authentication (e.g. basic auth from openHAB Cloud or a reverse proxy), and adding the credentials can be quite tricky.

One solution is to load all images as blobs with XHR then convert them to data URLs with the image encoded as base64 in the URL - like I did before for HABot: https://github.com/openhab/openhab-webui/blob/2878a59ceae55daba1e4c4ba9fd4c84f917c0d45/bundles/org.openhab.ui.habot/web/src/components/HbImage.vue#L101-L123

But it's very slow...

Another better approach would be to add a method to https://github.com/openhab/openhab-webui/blob/master/bundles/org.openhab.ui/web/src/js/openhab/cordova/media.js to be used for all image downloads: it would use the http library to fetch the image to the device's cache and then use a local 'file' URL, much like `getIcon`, which you can see it in action in the developer tools' debug tab. We have to be careful about not letting the cache grow too much, when we'll have widgets for security cameras etc.

digitaldan:
>it's even more of an headache because there might be an extra layer of authentication

Yeah, i started thinking about this and how cordova loads assets and realized there are going to be a lot of corner cases. 

>and I'm not sure how Cordova apps work overall as it's been some time since I last built one

Me too, my last time was a few years ago, and in honesty it was clunky and not a good experience, which makes me even more impressed with how well you have done here. The app feels native, besides a few rotation glitches which can be fixed.  

> it would use the http library to fetch the image to the device's cache and then use a local 'file' URL

Ah, excellent, i was hoping there was a way to inject a root like listener for requests, that's definitely a better way of doing this. 

tld;
There is also the issue of coupling between a openHAB version and the web-ui version if we were to publish an app. How many variations would be out in the wild of backend OH versions that may contain changes that are tied to a specific web-ui version? 

But.... the mobile experience is really, really good, it feels native.  I love the idea of a consistent look and feel across our clients, and the thought of our mobile clients having the full power that our browser client has is very enticing, so i'm hoping there may be a way to still support a mobile build.  So i like the idea of keeping the cordova support as an option.  

Another alternative is we have our existing mobile clients load the app directly from a user's oh server in a webview.  I believe android is doing this for Habpanel.

ghys:
I got curious and built an iOS version of the app for the first time in months and it still works surprisingly okay, even though states updates don't work (as expected) and you can't go to the admin area because the OAuth2 flow & token are not supported either. But they can still be made to work.

With Cordova the app's assets are bundled in, so it's loading as fast as possible and there is no need to download them from your server, but it comes with some drawbacks: the fact that we have to make sure the server's API matches the app is one (we _could_ release the Cordova apps with the openHAB release and enforce the same version of the server and app, but that could certainly backfire), and then there's the fact it's not actually serving the app under the same origin, leading to all sort of web browser issues (CORS, authentication, cookies...). That's why the middleware layer to access the API and remote resources is different for Cordova, using native APIs for HTTP and even raw sockets for SSE... it's not ideal and an additional maintenance burden.

I've added the webapp to the home screen from Chrome on Android and Safari on iOS and they offer a very good fullscreen experience, similar to what you get from Cordova (especially with HTTPS so you get PWA features like caching), although you lose the ability to switch servers/have remote & local URLs.

> Another alternative is we have our existing mobile clients load the app directly from a user's oh server in a webview. I believe android is doing this for Habpanel.

That could be a great option as well - if we can combine that with efficient caching, that could be the best of both worlds, and we wouldn't have multiple openHAB-branded apps in the app stores, which would probably be a bit confusing.

TL;DR: I never intended a Cordova build to replace the current openHAB mobile apps, just keep it as an option we have, which it still is; it has a great look & feel, but there are easier to maintain alternatives for sure. That being said, what's currently broken could be fixed rather easily.

> But.... the mobile experience is really, really good, it feels native. I love the idea of a consistent look and feel across our clients, and the thought of our mobile clients having the full power that our browser client has is very enticing

Thanks, this was the idea - "mobile as a first class citizen" was a stated goal from the start, which led to the framework choices like Framework7 which is very strong in that regard. I wish it were better on accessibility and keyboard navigation though.

digitaldan:
I hacked our production IOS client a bit today to see how this would look

ios

I think this actually works quite well.  I would recommend that the default view for 3.0  mobile users be our new UI.  If however,  a user selects a sitemap to view in the side menu list (in the native menu) , the next time the app opens, it should remember and load that, this would allow the same functionality that users are used to today, but also give the new experience to everyone. 

I was thinking we would probably want to change the top right menu action in the web ui when loaded this way.  If we can expose our app on a global variable (like window.$ohwebui ) then the mobile clients can inject a call back to respond to events, which might include the top left icon triggering the native menu, rather then listing the other UI's as it does today.  

There is also some very aggressive caching we can do if needed, especially if we can detect the build version of the UI through the API.  

One extreme example would be to load the app normally,  but tell the client to always prefer the cache and never expire it, then save the current build identifier of the app that is cached.   If a different build version of the app is detected (check everytime our view is loaded) , invalidate the cache and start over.

ghys:
Awesome! Love how the transition from webapp from/to native seem barely noticeable!

> I was thinking we would probably want to change the top right menu action in the web ui when loaded this way. If we can expose our app on a global variable (like window.$ohwebui ) then the mobile clients can inject a call back to respond to events, which might include the top left icon triggering the native menu, rather then listing the other UI's as it does today.

Yes this has great potential - speech-to-text for HABot queries (which I intend to add to the home page has an option soon), reacting to push notifications, even using the phone's camera to take a picture to put on a card/floor plan background, you name it... kind of a Cordova-like API but very lightweight. The native app would describe in the global variable which features it supports and the functionality would unlock in the web app.

(btw: noticed that glitch with the sidebar account controls as well, it should be fixed with https://github.com/openhab/openhab-webui/pull/256/commits/ac41ea9350b943cacf2c9d1115277c07531029cf, unrelated to but part of #256 which will be merged when the build succeeds again)

digitaldan:
> kind of a Cordova-like API but very lightweight

I actually was thinking about this.  Cordova has an embed option for just this use case.  I think most people opt for a simple web view, without all the cordova baggage, but, if we started doing a lot of web -> native integration,  especially around bits like camera access, would we end up rebuilding a cordova like framework, aka reinventing the wheel?   I honestly don't know that much about cordova or embedding it, so i guess it's something that needs research. From the little that i have read, you would still inject/evaluate javascript ( which i think can pass and return only strings) for custom interactions, but then use cordova plugins for direct phone access from the web app (like to the camera).   Not sure if you have any experience here ?

>(btw: noticed that glitch with the sidebar account controls as well, it should be fixed with ac41ea9, unrelated to but part of #256 which will be merged when the build succeeds again)

Awesome! Was not aware the build was broken.

ghys:
>  Not sure if you have any experience here ?

No, can't say I have. But I'm wondering if embedding Cordova would be overkill or not, or a burden to maintain in the long term for the app developers... I really don't know, it depends on the native functionality we'd like to provide to the web app. Maybe we could start with a simple variable injection :)

digitaldan:
I played some more with this yesterday, the cordova embedding seems very heavyweight and without a whole lot of benefit, so let's not do that :-) 

I made a little test in my IOS fork to understand how communication might work from Web -> Native

in app.js I went ahead and exposed the f7 app object on the global window:
```js
const app = new Vue({
  el: '#app',
  render: (h) => h(App),

  // Register App Component
  components: {
    app: App
  }
})

window.$ohwebui = app
```

Then in IOS, through native code, i inserted some Javascript into the webview object at the end of the loaded app:
```js
window.$ohwebui.$f7.on('routeChanged', function (newRoute, prevRoute) {
        window.webkit.messageHandlers.myHandler.postMessage(newRoute.path)
})
```
This allows me to have a callback handler in IOS (myHandler) to log the route change messages above. You can of course always evaluate a javascript string from Native -> Web.  So was pretty straight forward to message back and forth. 

>reacting to push notifications, even using the phone's camera to take a picture to put on a card/floor plan background, you name it... kind of a Cordova-like API but very lightweight.

Yep, push notifications and being able to trigger the native right hand panel through the top right icon in the app would be useful for sure right now.  We might look at adding an additional payload to push messages in the cloud binding so compatible apps can navigate to a page or item accordingly .

mueller-ma:
When integrating the new UI into the Android app, there are three side menus:

1. Left side in the app bar the native menu. There the user can choose between his default sitemap (or all sitemaps, based on this settings), HABPanel (via Android Webview), app settings and about menu. In RTL languages this menu is on the right.
2. Left side in the webview the menu of the new UI
3. Right side in the webview the "Other apps" menu of the new UI

What about giving the browser (in this case the client) to ability to move the second menu to the other side and hiding the third menu? The third menu isn't required, as the user can choose between other UIs in the Android menu.
A similar thing could be done in HABPanel to avoid the double menu on the left.

IMO it would be nice to integrate all UIs in the mobile apps even if the website is a webapp and can act like a native app. This would allow quicker switches between the UIs and authentication data needs to be saved only once.

ghys:
> What about giving the browser (in this case the client) to ability to move the second menu to the other side and hiding the third menu? The third menu isn't required, as the user can choose between other UIs in the Android menu.

I can't say I like moving the new web UI menu to the right too much when it's run inside a native app, it doesn't feel natural. But I don't have an ideal solution either.

What I would suggest instead is that the native Android or iOS apps launch the web UI fullscreen, without a native title bar or anything, and the web UI would then feature a "return to app" button (which may be replacing the "other apps" button). There could also be a callback to trigger the "pin to home" functionality directly from the web app (on Android), and the native app may offer a setting to display either a sitemap or the main UI as the default i.e. upon launch.

mueller-ma:
> What I would suggest instead is that the native Android or iOS apps launch the web UI fullscreen, without a native title bar or anything, and the web UI would then feature a "return to app" button (which may be replacing the "other apps" button). There could also be a callback to trigger the "pin to home" functionality directly from the web app (on Android)

Sounds good!

> the native app may offer a setting to display either a sitemap or the main UI as the default i.e. upon launch.

This is something I wanted to add anyway.

mueller-ma:
I updated https://github.com/openhab/openhab-android/pull/2244. Now the app drawer cannot be opened while the oh 3 ui is open.

ghys:
So now we need to decide on a [JS interface](https://developer.android.com/guide/webapps/webview#BindingJavaScript) so that the web ui can detect it's running inside the Android (or iOS) app and replace the button to open the "other apps" drawer on the right with a "return to app" button that will return to the native ui.
It would be cool if it were a common API for both Android & iOS. :)
For instance `OHApp.returnToApp()` in both cases.

mueller-ma:
>  It would be cool if it were a common API for both Android & iOS. :)

👍 Definitely.

Things the web UI should call back to the app:

- Return to app
- Pin to home
- Start voice recognition?

And the other way:

- Set theme (Auto, Android, iOS, Desktop)
- Set bright/dark (Auto, dark, bright)

PR #263, opened 2020-05-24
Auto selects the color default color scheme based on the system preference

digitaldan:

This is useful on the desktop, and very useful when running on mobile.  Still allows the user to save a permanent preference. 

theme

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -247,7 +247,7 @@ export default {
         // theme: (document.documentURI && document.documentURI.indexOf('?theme=ios') > 0) ? 'ios'
         //   : (document.documentURI && document.documentURI.indexOf('?theme=md') > 0) ? 'md'
         //     : 'auto', // Automatic theme detection
-        autoDarkTheme: true,
+        autoDarkTheme: !(localStorage.getItem('openhab.ui:theme.dark')),
         // App root data
         data () {
           return {
@@ -414,10 +414,6 @@ export default {
       this.themeOptions.homeNavbar = localStorage.getItem('openhab.ui:theme.home.navbar') || 'default'
       this.themeOptions.expandableCardAnimation = localStorage.getItem('openhab.ui:theme.home.cardanimation') || 'default'
       this.themeOptions.pageTransitionAnimation = localStorage.getItem('openhab.ui:theme.home.pagetransition') || 'default'
-      // f7 adds theme-dark to the class list if autoDarkTheme is enabled and dark mode is true
-      if (this.themeOptions.dark === 'light') {
-        document.querySelector('html').classList.remove('theme-dark')
-      }
     }
   },
   created () {
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -247,8 +247,8 @@ export default {
         // theme: (document.documentURI && document.documentURI.indexOf('?theme=ios') > 0) ? 'ios'
         //   : (document.documentURI && document.documentURI.indexOf('?theme=md') > 0) ? 'md'
         //     : 'auto', // Automatic theme detection
-        // App root data
         autoDarkTheme: true,
+        // App root data
         data () {
           return {
           }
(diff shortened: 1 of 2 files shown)


ghys:
Cool, however there's a [`autoDarkMode` parameter on the App object](https://framework7.io/docs/app.html#app-parameters) and also an `app.darkTheme` property and `darkThemeChange` event introduced here: https://github.com/framework7io/framework7/blob/master/CHANGELOG.md#v560---april-18-2020

I think I would be better to use those, so that the framework can update its state properly (but also update the properties on the app component like you did for consistency).
The filled property for the top bars also has some logic for its default: it should be filled in light mode only, and not if the theme is iOS.

digitaldan:
>Cool, however there's a autoDarkMode

lol, i think i googled "freamwork7 prefers-color-scheme"  and nothing came up, so i assumed i had to do this the hard way.  Thats a far better approach, i'll go ahead and look at that.  Thanks!

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -406,14 +407,20 @@ export default {
         this.$f7.preloader.hide()
         this.$f7.dialog.alert('Error while signing out: ' + err)
       })
+    },
+    updateThemeOptions () {
```
digitaldan, line 411: Should themeOptions be a computed property vs me sticking them in a method?  Seem like i took the cheap way out here.  Since we rely on this.$f7 , this can not be called from the "created" function anymore

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -406,14 +407,20 @@ export default {
...
+      this.themeOptions.dark = localStorage.getItem('openhab.ui:theme.dark') || (this.$f7.darkTheme ? 'dark' : 'light')
+      this.themeOptions.bars = localStorage.getItem('openhab.ui:theme.bars') || ((!this.$theme.ios) || (!this.$f7.darkTheme) ? 'filled' : 'light')
+      this.themeOptions.homeNavbar = localStorage.getItem('openhab.ui:theme.home.navbar') || 'default'
+      this.themeOptions.expandableCardAnimation = localStorage.getItem('openhab.ui:theme.home.cardanimation') || 'default'
+      this.themeOptions.pageTransitionAnimation = localStorage.getItem('openhab.ui:theme.home.pagetransition') || 'default'
+      // f7 adds theme-dark to the class list if autoDarkTheme is enabled
+      if (this.$f7.darkTheme && this.themeOptions.dark === 'light') {
+        document.querySelector('html').classList.remove('theme-dark')
```
digitaldan, line 419: in order to use the built in dark detection,  f7 is going to inject the dark-theme into the class list, not sure if there's a better way of removing it using the framework (or css for that matter) See the code in question for more info:
https://github.com/framework7io/framework7/blob/080ea7e7cea5c3332ad211f263f1ac61dd37dfff/src/core/components/app/app-class.js#L107  ,

digitaldan:
@ghys i have this working using the built in detection.  Moving the theme options to a public method seem kludgy to me, any suggestions?  (see comments)

bundles/org.openhab.ui/web/src/components/app.vue
ghys: Maybe you could set autoDarkTheme to true when initializing the app only if `localStorage.getItem('openhab.ui:theme.dark')` returns nothing? Otherwise it means the user has overridden the theme.

digitaldan: that's a very good idea

ghys:
One last thing: could you clean up the two CSS rules with `.demo-theme-picker` in theme-picker.vue if they're not in use anymore? (these should also be converted to stylus and lose the demo prefix...)

digitaldan:
Done!

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -406,14 +407,16 @@ export default {
         this.$f7.preloader.hide()
         this.$f7.dialog.alert('Error while signing out: ' + err)
       })
+    },
+    updateThemeOptions () {
+      this.themeOptions.dark = localStorage.getItem('openhab.ui:theme.dark') || (this.$f7.darkTheme ? 'dark' : 'light')
+      this.themeOptions.bars = localStorage.getItem('openhab.ui:theme.bars') || ((!this.$theme.ios) || (!this.$f7.darkTheme) ? 'filled' : 'light')
```
ghys, line 413: Whoops that was actually if not iOS AND not dark then filled (I'm getting filled bars in dark mode by default on my desktop, this is a rather ugly combination). No worries I'll fix it with the next commit.

PR #271, opened 2020-06-12
Main UI HABot client on home page

ghys:
This implementation transforms certain cards issued
by HABot for its own GUI to compatible equivalents.
It will not try to transform cards saved on HABot's
"card deck"/CardRegistry because they may have been
modified and compatiblity cannot be guaranteed.

The main UI's client calls /rest/habot/chat with a
`?useCardRegistry=false` option which has to be added
to the API to prevent saved cards from being reused.
This will be done in a separate PR.

The chat input box appears on the home page only if
the "habot" endpoint is present, by calling the root
/rest early to determine the version of the API and
its links (it's stored in the Vuex store for the
remainder of the session for easy access). It's also
possible to disable the feature on a per-device basis
with an option on the about page.

The 5 most recent unique queries will be retained for
quick recalling (also per-device, stored in the local
storage).

For charts, it simply opens the Analyzer window since
since image-related support in widgets is not yet done.
(altered the analyzer to be able to open it with a
predefined initial period).

The "create rule" skill featured in the HABot app with
its dedicated card is not supported.


digitaldan:
I'm feeling a little guilty i have not played with HABot more, but will use this as a good reason to do so.  Couple thoughts initially:

- I did not have HABBot installed, i'm wondering if we should make it part of the default install?
- When i did install it i think i needed to reload the UI to get the chat box to show up (i'm not 100% sure) 
- would it be possible to add a speech entry (https://www.google.com/intl/en/chrome/demos/speech.html ) if on a supported browser to the text entry input box ?  

I'll play around a bit this weekend !

ghys:
> I did not have HABBot installed, i'm wondering if we should make it part of the default install?

That would be up to the @openhab/distro-maintainers but it only works well after you've made a well-designed semantic model, and added a couple synonyms to allow shortcuts and so on. So that's something that would not be functional initially unless the user starts from scratch with this in mind (the semantic model screen and emphasis on the semantics throughout the app was specifically to try to make this happen). We can certainly start a discussion.

> When i did install it i think i needed to reload the UI to get the chat box to show up (i'm not 100% sure)

Probably, there might be a couple reasons for that, probably not worth the extra requests ;)

> would it be possible to add a speech entry (https://www.google.com/intl/en/chrome/demos/speech.html ) if on a supported browser to the text entry input box ?

Sure, that was done in the [original client](https://github.com/openhab/openhab-webui/blob/master/bundles/org.openhab.ui.habot/web/src/components/speech/SpeechButton.vue) with the browser functionality and a couple of cloud services (Google and Azure) if you happen to have API  keys for these, but:
1. it requires secure origins i.e. HTTPS so it's not for everybody - unless they use e.g. https://home.myopenhab.org; by default, but then everything would transit on the cloud... hence my wishful thinking of a [Let's Encrypt integration](https://community.openhab.org/t/discussions-possible-improvments-to-cloud-service-for-openhab/99988/25) using a myopenhab.org DNS service
😄 
2. it won't be available in web views either, if I'm not mistaken - that's why I'd probably do a refactoring of the STT interface to make it interchangeable (`this.$oh.speech.*` like `this.$oh.api.*` etc) so if we're to implement a native-to-web interface like we discussed, it could hand over the speech-to-text to a native implementation for instance with the [SFSpeechRecognizer](https://developer.apple.com/documentation/speech/recognizing_speech_in_live_audio) on iOS and the equivalent on Android. Then 1. would become a moot point.

ghys:
Again about the speech recognition... Just realizing now: on mobile you're very likely to have dictation included in the keyboard you're using, whether first-party ([the native iOS keyboard](https://support.apple.com/guide/iphone/dictate-text-iph2c0651d2/ios) and [Google's Gboard](https://ai.googleblog.com/2019/03/an-all-neural-on-device-speech.html) will even do it offline in some cases, with presumably good results these days) or third-party like [Swiftkey](https://support.swiftkey.com/hc/en-us/articles/201467491-Can-I-use-voice-to-text-with-SwiftKey-Keyboard-for-iOS-). I tried it and while it's true you do have to tap a couple more times, it's not that bad, and you have an opportunity to fix mistakes before sending the query. It will also work no matter how your web app is served from and the visual audio feedback is prettier than what you'll get in a browser.

Even of desktop, there are similar easy-to-access features on both [Windows](https://support.microsoft.com/en-us/help/4042244/windows-10-use-dictation) (Win+H) and [macOS](https://support.apple.com/guide/mac-help/use-dictation-mh40584/mac) (Fn twice) - but in my experience I mostly ended up using HABot by typing small shortcuts I configured with a synonym anyway.



In other words, I'll still probably port over the browser-based speech button, but we should keep in mind dictation has widespread native support too on modern platforms now - which is sometimes even better -.

ghys:
No-frills implementation using the [`SpeechRecognition`](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition) API implemented, the button only appears if the feature is available - basically: Chrome, Edge with HTTPS (or on localhost).

ghys:
final look of the interim results, I think they're better in a chat bubble (and long sentences don't overflow the text box):

ghys:
In case you didn't know this existed (like me until now), you can tweak a flag in `about://flags` in Chrome and derivatives to treat some addresses as a secure origin even when if it's not HTTPS - just search for "insecure origins treated as secure" and don't forget to enable the flag on the right, editing the list is not enough:

image

Then it will allow you to use the microphone and perform speech recognition on your LAN instance 🙂.

bodiroga:
This is f***ing awesome!

ghys:
Let's merge then? ;)

PR #207, opened 2020-03-27
Optimize webpack bundling

ghys:
Split heavy dependencies away from the entrypoint
by using dynamic imports at critical places along with
some cleanup.

Also replace crypto-based `uuidv4()` by `f7.utils.id()`.

This reduces the main app.js by more than 25%
(from >2 MB to around 1.4 MB).

Add npm script to launch the Webpack bundle
analyzer (`npm run webpack-analyzer`) after a
production build.

Signed-off-by: Yannick Schaus <github@schaus.net>

PR #272, opened 2020-06-15
Check what REST services are available before requesting them

cdjackson:
@ghys I'm not sure that this is the best approach here so I welcome your feedback, but I would like to avoid assuming that REST services are available. Instead, we really should check the REST root to ensure that a particular service is available before making the request.

This PR resolves a startup issue with the UI by checking the REST root before issuing calls that may fail. Ideally, I would like to only show tabs/pages for services that are available - eg don't show the rules tab if rules aren't available - same for the pages and other such services. Also don't ask for authentication if authentication is not being used...  I know that's a wider issue than addressed here, but I thought I'd mention it in general to give an overview of the broader aim of this PR, and I can open another issue for this.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -327,23 +327,61 @@ export default {
   },
   methods: {
     loadSidebarPages () {
-      return Promise.all([
-        this.$oh.api.get('/rest/sitemaps'),
-        this.$oh.api.get('/rest/ui/components/ui:page'),
-        this.$oh.api.get('/rest/ui/components/ui:widget')
-      ]).then((data) => {
-        this.sitemaps = data[0]
-        this.$store.commit('setPages', { pages: data[1] })
-        this.$store.commit('setWidgets', { widgets: data[2] })
-        this.pages = data[1].filter((p) => p.config.sidebar && this.pageIsVisible(p))
-          .sort((p1, p2) => {
-            const order1 = p1.config.order || 1000
-            const order2 = p2.config.order || 1000
-            return order1 - order2
-          })
+      var that = this;
+      return new Promise(
+        function (resolve, reject) {
+          var restRoot = that.$oh.api.get('/rest').then((root) => {
+            that.$oh.servicesSupported = new Array();
+            for (var link in root.links) {
+              that.$oh.servicesSupported.push(root.links[link].type);
+            }
 
-        this.ready = true
-      })
+            var requestUris = new Array();
+            if(that.$oh.servicesSupported.includes('sitemaps')) {
+              requestUris.push('/rest/sitemaps');
+            }
+            if(that.$oh.servicesSupported.includes('ui')) {
+              requestUris.push('/rest/ui/components/ui:page');
+              requestUris.push('/rest/ui/components/ui:widget');
+            }
+
+            var requestPromises = new Array();
+            for (var uri in requestUris) {
+              requestPromises.push(that.$oh.api.get(requestUris[uri]));
+            }
+
+            Promise.all(
+              requestPromises
+            ).then((data) => {
+              var index;
+              index = requestUris.indexOf('/rest/sitemaps');
+              if(index != -1) {
+                that.sitemaps = data[index];
+              }
+              index = requestUris.indexOf('/rest/ui/components/ui:page');
+              if(index != -1) {
+                that.$store.commit('setPages', { pages: data[index] })
+                that.pages = data[index].filter((p) => p.config.sidebar && that.pageIsVisible(p))
+                  .sort((p1, p2) => {
+                    const order1 = p1.config.order || 1000
+                    const order2 = p2.config.order || 1000
+                  return order1 - order2;
+                })
+              }
+              index = requestUris.indexOf('/rest/ui/components/ui:page');
+              if(index != -1) {
+                that.$store.commit('setWidgets', { widgets: data[index] });
+              }
+          
+              that.ready = true;
+
+              resolve();
+            }).catch((error) => {
+              reject();
+            })
+          })
+        }
+      )
     },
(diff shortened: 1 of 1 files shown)


ghys:
@cdjackson thanks for having a look! It just so happens that I had the same requirement while working on #271 and ended up implementing pretty much the same thing (you can check app.vue in here, it's the first file shown: https://github.com/openhab/openhab-webui/pull/271/files).

The results of the root `/rest` request is stored as-is in Vuex and you'll be able to check whether an endpoint is present or not in a component's code with something like:

```js
this.$store.getters.apiEndpoints('habot')
```
([implementation of the getter here](https://github.com/openhab/openhab-webui/pull/271/files#diff-4b2a87639e63bcc7e3dae4d4af6604cb))

So checks on the existence of services can be added wherever relevant. It's true that in this implementation the call is made in parallel with the pages to slightly increase the initial rendering performance (I assumed they'll always exist since they're part of a core bundle, maybe wrongly).

I was also contemplating implementing the suggestion here: https://github.com/openhab/openhab-core/pull/738#issuecomment-485052184
i.e. adding fields like `runtimeVersion` and `runtimeBuildNumber` to this root response, which might be nice to display in the about page and maybe could be of use elsewhere.

Would appreciate your comments on this.

cdjackson:
Thanks @ghys - good to know we're on the same page :)

> I assumed they'll always exist since they're part of a core bundle, maybe wrongly

For some embedded systems, we've trimmed down the core bundles to remove "unnecessary" parts - this has allowed reduction in the core code size by around 50% by removing things like sitemaps etc to get rid of all XText models and dependencies.

Ideally I would prefer not to make the assumption that pretty much any services are available. In this case, it also meant that the UI just didn't load at all. I think we should read the REST root an tailor the UI to remove unimplemented services so they are not presented to the user.  This also extends to authentication which is not used in some cases where security is handled separately - we could check to see if the `/auth` endpoint is available and if not assume that we're logged on. There's no security issue with this since it's just the UI side.

> I was also contemplating implementing the suggestion here:

Yep - this is useful, and also something that I've implemented in my system. Probably this is largely superficial though - ie it's good for the user to be able to read this information, but I think it probably doesn't allow the UI to significantly change in respect to the server side, but I think resolving the rest services does support this.

I'm no expert on the frameworks used, but am happy to progressively help with this if you're happy with the general concept, or want to create an API that can be used to conditionally display parts of the UI based on REST services.

ghys:
Absolutely, we can discuss about a way to hide/disable features selectively, provided it's kept on the lighter side and not too much overhead - I'll admit my only stake is in the official, "standard" openHAB distribution, so I don't think much about anything else, but can certainly understand the needs.

With the store getter above, hiding for instance menu entries or other UI parts could be as simple as
`v-if="$store.getters.apiEndpoints('rules')`
in the template, it's clear and easy to understand, and can also be used in methods, computed properties etc. To disable the auth it's a little different, but it could also simply be a matter of setting a "fake" user with an admin role, and prevent the auth code from running - and tokens to be sent with every request.

As part of your trimming-down process you might want to consider reducing the amount of the frontend code that is preloaded as well - the so-called "entry point" of the app in webpack's jargon, here app.js and app.cs only - which I already reduced a bit in #207 by isolating the biggest non-essential dependencies (the charting, mapping, code editor libraries, and so on) in separate lazy loaded "chunks". Still, it remains quite big: a little over 2 MB. We might get a nice performance improvement by going further with the [code splitting](https://web.dev/reduce-javascript-payloads-with-code-splitting/) and isolating modules that aren't strictly necessary away from this entry point.

For example if you look at this second image from #207 (not representing the current codebase, you can use `npm run webpack-analyzer` to get a current version):



you can see all the settings pages are in app.js (the entry point), but it's clearly not necessary to preload and parse all the code for the admin area for every user since they might not go there, or even have access, or a large part of it could be unavailable for different reasons. We could certainly go even more granular than that and isolate smaller features to carve out of the main chunk; even individual Framework7 components - the main entry point dependency - could be lazy loaded in theory. On the other hand we shouldn't be _too_ granular either and end up with hundreds of chunks because lazy loading has a performance cost too... it's a balance.

cdjackson:
Thanks @ghys - very useful.

So, if I understand correctly, once #271 is merged we will have this `apiEndpoint` method available. I think this PR is still needed in that the sitemap etc is loaded during the `loadSidebarPages` call and that will currently prevent the UI running if these endpoints are not available. My PR here also saved the `/rest` response (or at least the list of services) so that is not needed.

I therefore propose to wait until #271 is merged (I assume that's not too far away?) and I'll then refactor this PR to use the `apiEndpoint` method.

Any subsequent changes will be through further PRs.

> I'll admit my only stake is in the official, "standard" openHAB distribution, so I don't think much about anything else, but can certainly understand the needs.

That's fair - thanks.

> it's a balance.

I completely agree with all your points here and it's certainly not my intention to make thing complex or inefficient etc - just a bit more tailored :). I will look into further updates once #271 is merged.

ghys:
I just merged #271 and included in https://github.com/openhab/openhab-webui/pull/271/commits/ec937212199c7a89cb0be37c9d07e1616f6a1766 an early (but benign and with very little overhead) example of how you can make the sidebar settings menu and the options on the main settings page dynamic, thanks to the endpoints being in the Vuex store.
Storing this in Vuex instead of doing your own thing has obvious benefits because you're getting reactivity for free (https://vuex.vuejs.org/guide/mutations.html#mutations-follow-vue-s-reactivity-rules) - so now when an addon is installed or uninstalled the settings page's menu entries and HABot chat input will be refreshed automatically.

cdjackson:
Thanks @ghys - much appreciated - I'll look to refactor this in the next day or two.

cdjackson:
Thanks @ghys - I see you've effectively included this in #271 and the startup and removed the sitemap dependency. Thanks. I'll close this.

PR #275, opened 2020-06-17
Add Items from Thing: use model picker for parent group

ghys:
Use a treeview-based semantic picker to select a parent
location or equipment group item when using the
Add Equipment to Model or Add Points to Model functions
in a Thing's Channels tab.
Replaces the current flat list of group items.

Add functionality to the model picker popup to support this
use case and adjust some texts and controls on the page.


ghys:
Before:



After:

PR #276, opened 2020-06-17
Config parameter options shown inline if 5 or less

ghys:
Fix parameter not saved when unset & absent from payload on save.


ghys:
Before:

image

After:

image

PR #279, opened 2020-06-17
Fix duplicate thing state being printed

cdjackson:
The `Thing` state is currently printed twice in the thing detail page.


cdjackson:
> Although it's not actually appearing twice on the page

It definitely is appearing twice on the page - that's how I found it. It shows ```ONLINEONLINE``` without this fix (at least on my Mac with Chrome - I guess other systems may vary).

ghys:
Interesting... I'd definitely have fixed it by now if it were the case for me, anyways thanks again!

PR #280, opened 2020-06-20
Fix conditional display of sections on settings page

cdjackson:
This fixes the display of sections on the settings page that are not available on the server.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
@@ -81,10 +81,11 @@
               <f7-icon slot="media" f7="calendar" color="gray"></f7-icon>
             </f7-list-item>
           </f7-list>
-          <f7-block-title v-if="addonsLoaded">Add-ons</f7-block-title>
+          <f7-block-title v-if="$store.getters.apiEndpoint('extensions') && addonsLoaded">Add-ons</f7-block-title>
           <f7-list media-list class="search-list"
             v-if="$store.getters.apiEndpoint('extensions')"
           >
+            <f7-list media-list class="search-list">
             <f7-list-item
               media-item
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
```diff
@@ -81,10 +81,10 @@
               <f7-icon slot="media" f7="calendar" color="gray"></f7-icon>
             </f7-list-item>
           </f7-list>
-          <f7-block-title v-if="addonsLoaded">Add-ons</f7-block-title>
-          <f7-list media-list class="search-list"
-            v-if="$store.getters.apiEndpoint('extensions')"
-          >
+        </f7-col>
+        <f7-col width="100" medium="50" v-if="$store.getters.apiEndpoint('extensions')">
```
ghys, line 85: This messes up the layout when the extensions are present. Could you please revert this back and simply change the v-if condition on f7-block-title to `$store.getters.apiEndpoint('extensions') && addonsLoaded`?

cdjackson: I thought it made more sense to break this block? Then you have the three sections in separate blocks (automation and rules, extensions, services). As it is at the moment, you have the "AddOns" section in the same block as the Configuration, but Services is separate.

Doesn't it make more sense to group them separately?

ghys: Logically, I suppose, but here you have 3 columns of size 50% on medium screens and larger, which means the third gets dropped to a second row and it looks like this:

image

Which is far less aesthetically pleasing than limiting to 2 columns to get a balanced look:

image

cdjackson:
Thanks @ghys  - I've updated the PR.

PR #288, opened 2020-06-22
[WIP] Add knob widget

ghys:


ghys:
Seems to work well enough!



Only a system library widget for now (i.e. no oh-knob-card) since by convention there are no descriptions for those.
However it will pass any config parameter to https://github.com/kramer99/vue-knob-control#properties

digitaldan:
This does work very nice!  

>Only a system library widget for now (i.e. no oh-knob-card) since by convention there are no descriptions for those.

How are you thinking about adding it as a card then ?  I could see this being an option to the slider card  ( options would be horizontal, vertical, round/knob) , as well as having a real thermostat card.

ghys:
> This does work very nice!

Noticed it's kind of hard to set the zero (or min) value, no problems with the max though - not much we can do about it.

> I could see this being an option to the slider card

Good idea, some parameters are different or accept different values (e.g. colors) but a number of them are shared between the slider and the knob, so it could be an option to merge it with the slider card (and perhaps call it "dimmer card" or similar).

ghys:
Going to self-merge this since it has little impact but would be nice to include on a broader widget-related PR I'm planning next.

PR #289, opened 2020-06-22
Fix HABPanel REST resource not registered

ghys:


ghys:
Is this good to merge? HABPanel's widget gallery API still 404s...

PR #294, opened 2020-07-16
Updated Swagger UI to version 3.28.0

kaikreuzer:


kaikreuzer:
@ghys Would you be able to help getting the OAuth authentication working again? I am not quite sure what exactly you did for the old swagger ui to make it work...

ghys:
All that was done in https://github.com/openhab/openhab-webui/pull/257 is replace the client Id to pass the specific checks enforced by our "simplified" OAuth2 implementation (it has to be equal to the redirect Uri so the user can verify easily what  origin is requiring access without registering clients in advance), but I see that the updated version has those changes already. So it might just work - provided https://github.com/openhab/openhab-core/pull/1496 or an equivalent is integrated to put the info in the spec.

kaikreuzer:
> So it might just work

It doesn't - that's why I asked for your help 😟.

kaikreuzer:
Feel free to merge first and we then look for the auth part, if that's easier.

wborn:
Thanks for the upgrade it was on my nice to haves list for OH3! :+1:
I gave it a test and it works equally well as the previous version.
It also allows for upgrading swagger-core in OH3. :smile:

kaikreuzer:
@ghys Shall we merge?

ghys:
Just back from holidays... Looks good, I wanted to give it a quick try before merging, I'll do it today ;)

kaikreuzer:
> I kinda preferred the old version's looks

Mee too! But I can live with it, if this is what the Swagger guys decided to be more modern...

>with openhab/openhab-core#1496 or similar we'll be able to perform real OAuth2 authorization flows

wborn:
Yes I've spent some time with/on people and plants since things are more normal now and there is nice summer weather. :sun_with_face: 

For most of my remaining OH3 plans it would be easier when the add-ons are migrated so I also work on non-OH projects that make more sense now. If we still want to release a polished OH3 this year we probably want to migrate the add-ons soon.

I only knew the new-look Swagger UI before I started using the old-look UI in openHAB so for me the Swagger UI upgrade finally makes it look "normal" again. :wink: 

I've retested openhab/openhab-core#1496 with the new Swagger UI and added my comments to the core PR.

kaikreuzer:
Hope you've had a good time @wborn! 🌞 

>  If we still want to release a polished OH3 this year we probably want to migrate the add-ons soon.

Why do you think so? I actually came to the impression that we could wait until autumn with that. My feeling is that there is a lot of activity on the 2.5.x bindings and people expect it those fixes and improvements to be available on their existing 2.5 installations. As soon as we end working on the 2.5.x branch, there's no evolution anymore until we release 3.0.
I haven't seen many PRs that are blocked by keeping the code on the 2.5.x branch. For the few bindings (like Nest), which require the 3.0 core bundles, it should be fine to move development to the master branch just now already (and not migrate them later anymore).

> For most of my remaining OH3 plans

As you are one of the very few core and distro maintainers, I'd hoped that you'll also still see enough stuff to work on on those parts of openHAB. If you are out of ideas, let me know 😉 .

wborn:
Most likely there will be a lot more regressions compared to when we only changed the build system. So it needs to be easy for users and developers to find these regressions and fix them. That's not the case now because you cannot run/test and debug all add-ons in the IDE with OH3. I think you've experienced this yourself firsthand with the tesla and lametric add-ons. If we wait until autumn and release OH3 in December, there will realistically only be ~2-3 months in which users/developers might seriously start using and testing OH3.

kaikreuzer:
My hope is that we won't see many regressions - but I might be too optimistic here.
That's why I try to move to OH3 productively asap (and thus push all bindings that I need for that) and hope the other maintainers do the same.

PR #297, opened 2020-07-22
Embed Swagger UI as a developer tool in the main UI

ghys:
This imports Swagger UI as a dependency to the main UI and embeds it in a new "API Explorer" page in the Developer Tools section.

Advantages of this include:
- It is only available to logged in admins, and replaces the standalone version provided by the "REST Docs" UI bundle which may eventually be removed; this removes that UI from the "Other Apps" side drawer (a good thing IMHO, those users who are likely to need such a technical/developer-oriented UI likely have an admin account already, and end users won't see it);
- There is no need to perform a separate OAuth2 authorization flow because the access token can directly be injected with the [requestInterceptor](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md#network) feature - in other words it makes https://github.com/openhab/openhab-core/pull/1496 unnecessary, API requests are made in the context of the current user (but it would still be good to have the security definitions in the spec);
- CSS can be (and has been) overridden to match the overall look of the app and make Swagger UI a little bit prettier, and give it a dark mode (screenshots below)
- Probably easier to manage in the end, since upgrading it can be done with npm.

Current issues or potential drawbacks are:
- Swagger UI is based on React and a bunch of other stuff so it's a rather heavy addition to the dependency graph (several MB), _but_ thanks to webpack's code splitting they are carved out in their own bundle(s) and don't get loaded until the user accesses the API Explorer page;
- The Swagger spec generation support is activated with the restdocs UI add-on so as of now it still has to be installed (otherwise the API Explorer page will report a 404 error on `/rest/swagger.json`);
- With this approach to authorization there's currently no way to get a different token, with different scopes (if we ever add more) than the one currently in use by the UI.


ghys:
image

image

image

image

image

kaikreuzer:
Wow, that looks awesome!
I agree that this is very nice for the user to have it fully integrated in the dev tools section.
I'd still prefer to have it as an optional add-on, though (especially the swagger parts in core).
Maybe we could introduce a "devtools" add-on category and the UI could hide/show the "API Explorer" entry based on the existence of that add-on? Or we define a "DevTools" add-on in the UI category and add whatever dev-related stuff we might have and toggle the "Developer Tools" section of the UI based on whether this add-on is installed or not?

ghys:
> toggle the "Developer Tools" section of the UI based on whether this add-on is installed or not

There is already some conditional visibility in menus based on the [RootResource ](https://github.com/openhab/openhab-core/blob/859da2e1c7ab5ce21f345782e54f593146de4240/bundles/org.openhab.core.io.rest/src/main/java/org/openhab/core/io/rest/internal/resources/RootResource.java#L87)response (`/rest`), which is requested and retained before the UI is rendered, so I think it would make sense if a reference to the Swagger document could appear there as well if it is available - since the purpose of this root resource is to describe the API - either in the `links` array (we can consider it's a regular resource after all) or give it special treatment i.e.  adding a field like `"spec": "http://localhost:8080/rest/swagger.json"` in the RootBean itself.

That way it doesn't depend on the way we activate the Swagger feature (`openhab-core-io-rest-swagger`) with an add-on if it remains outside the standard package; I don't really have a strong preference on that. We would only have a single case of a developer tool requiring a add-on for now however, so it might be overkill to introduce a new add-on category at this stage? Maybe start by cleaning up and renaming the `org.openhab.ui.restdocs` bundle to maybe `org.openhab.ui.devtools.apiexplorer`, and the `openhab-ui-restdocs` feature to `openhab-devtools-apiexplorer` (similar to the naming convention for iconsets, with only `classic` in it). To install it, I'd even to as far as arguing that a "developer" shouldn't have a problem installing features like that with `feature:install` at the console, provided it's properly documented...

kaikreuzer:
> I think it would make sense if a reference to the Swagger document could appear there as well if it is available

I've created https://github.com/openhab/openhab-core/pull/1568. This adds it as a "spec" link with url `/rest/spec` (url is constructed from the name of the link per convention).

>  it might be overkill to introduce a new add-on category at this stage

I agree.

> To install it, I'd even to as far as arguing that a "developer" shouldn't have a problem installing features like that with feature:install at the console

With https://github.com/openhab/openhab-core/pull/1568, I have added it to the runtime-base for now, so it will always be present on a standard distro. We can still decide whether we want to exclude it or package it in some kind of add-on or feature.

I assume that with this PR the existing restdocs UI becomes completely obsolete, correct? If so, you should directly remove it within this PR.

bundles/org.openhab.ui/web/src/pages/developer/api-explorer.vue
```diff
@@ -0,0 +1,145 @@
...
+    onPageAfterIn () {
+      const swaggerCss = import(/* webpackChunkName: "swagger-css" */ 'swagger-ui-dist/swagger-ui.css')
+      const swaggerModule = import(/* webpackChunkName: "swagger" */'swagger-ui-dist')
+      const refreshToken = this.refreshAccessToken()
+
+      Promise.all([swaggerModule, swaggerCss, refreshToken]).then((results) => {
+        const SwaggerUI = results[0].SwaggerUIBundle
+        const tokenResponse = results[2]
+        SwaggerUI({
+          url: '/rest/swagger.json',
```
kaikreuzer, line 127: Once https://github.com/openhab/openhab-core/pull/1568 is merged:
```suggestion
          url: '/rest/desc',
```

wborn:
It looks very nice integrated!

> replaces the standalone version

It may make debugging issues harder if you can't get into the UI due to some REST API issue. I guess we need to launch another Swagger UI then e.g. in a container?

ghys:
> I assume that with this PR the existing restdocs UI becomes completely obsolete, correct? If so, you should directly remove it within this PR.

Done, but @wborn's remark is also worth considering. If need be, I think copying a standalone version of Swagger UI in `conf/html/swagger` and access it at `/static/swagger/index.html` could work.

kaikreuzer:
To me, the swagger ui is not a tool for debugging openHAB itself - it rather helps learning about and exploring its REST API.
I thus do not really see a need to provide any additional standalone version.

wborn:
I used the Swagger UI extensively while testing the Jersey/Whiteboard migration. :-)

digitaldan:
I use the swagger interface quite a lot actually, so i love this feature.

PR #298, opened 2020-07-23
[Main UI] Include bridgeID in thingUID on manual creation

### LINKED ISSUE #290: [Default UI] BUG : Manually add Thing to Bridge => but Bridge uid NOT included in the Thing uid
There is a bug in PaperUI **_EDIT: and in Default UI_**. When you try to add a Thing manually to an existing Bridge, the created Thing's UID does not include the UID of the respective bridge; and it's respective status is shown as `ONLINE`.

Note: we have tested this with the Hue, HD PowerView, Velux and NeoHub bindings, and the problem is the same; so it is NOT a binding problem.

 The steps to create the bug are shown in the series of screenshots below. 
---

---

---

---

---

---

ghys:
This will include the ID of the bridge in the ThingUID if it is specified during the
manual creation of the thing.

The thingUID format will therefore be `binding:thingType:bridgeID:thingID` if a bridge is specified during the creation, otherwise it will be `binding:thingType:thingID`. This is consistent with the naming convention adopted when creating things by other means (files, inbox).

Note that the thingUID is immutable once the thing has been created, so the bridge ID part will be false or missing when the bridge is altered or assigned after the creation. Relying on the thingUID to determine the bridge should be discouraged.
See openhab/openhab-addons#7496 (comment)

Fixes #290.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
+++ b/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
@@ -47,7 +47,7 @@ export default {
   methods: {
     computedThingUid () {
       return (this.thing.bridgeUID)
-        ? [this.thing.thingTypeUID, this.thing.bridgeUID.substring(this.thing.bridgeUID.lastIndexOf(':')), this.thing.ID].join(':')
+        ? [this.thing.thingTypeUID, this.thing.bridgeUID.substring(this.thing.bridgeUID.lastIndexOf(':') + 1), this.thing.ID].join(':')
         : [this.thing.thingTypeUID, this.thing.ID].join(':')
     },
     changeUID (event) {
--- a/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
+++ b/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
@@ -44,8 +44,6 @@ export default {
   components: {
     ThingPicker
   },
-  computed: {
-  },
   methods: {
     computedThingUid () {
       return (this.thing.bridgeUID)
--- a/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
+++ b/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
@@ -26,7 +26,7 @@
             <f7-list v-if="thingType.supportedBridgeTypeUIDs.length" inline-labels no-hairlines-md>
               <thing-picker
                 title="Bridge" name="bridge" :value="thing.bridgeUID"
-                @input="(value) => { thing.bridgeUID = value; $emit('updated') }"
+                @input="updateBridge"
                 :filterType="thingType.supportedBridgeTypeUIDs" />
             </f7-list>
           </f7-col>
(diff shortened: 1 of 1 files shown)


openhab-5iver:
Why include the bridgeID at all?


- There does not seem to be any benefit to including it.
- The bridge can change and then be misrepresented in the thingUID.
- A bridge is not always present and included in the thingUID, even if a bridge is needed, which can be very confusing when comparing thingUIDs.

bobadair:
> Why include the bridgeID at all?
>     * There does not seem to be any benefit to including it.

Good question. Not sure what the original thinking behind it was, but I would guess that one big benefit is uniqueness. A lot of binding discovery services use a device address for the thingID that is only unique within the context of a bridge, so they need to include the bridgeID to create a unique thingUID. If consistency in thingUID structure is desirable, then that would imply that all things that are children of a bridge should include the bridgeID.

>     * The bridge can change and then be misrepresented in the thingUID.
>     * A bridge is not always present and included in the thingUID, even if a bridge is needed, which can be very confusing when comparing thingUIDs.

Is it a common occurrence for a child thing to be switched to a different bridge? Maybe this shouldn't be allowed (in the case of the second), or should force the creation of a new thingUID (in the case of the first).

ghys:
@openhab-5iver this PR is only a proposal following the discussion in #290, it's not going to be merged until we all agree on the principle. It's a matter of allowing a number of add-ons to work without changes when they didn't before (in this case) vs. fixing them.

andrewfg:
If you manually create a Thing with a Bridge via the default ui, then currently that Thing does not have the bridge instance name as part of its UID (even if you select a bridge instance in the drop down).

The problem is that such a manually created Thing is BROKEN — it is a zombie without any reference to a bridge instance, and so it simply DOES NOT WORK (try it yourself please).

It is all very well you guys being purists and saying that the bridge instance might be changed later (it is true). But surely you cannot use your purity as an excuse for having a UI that creates broken Things.

Either you must fix the UI as proposed. Or you must fix the core in some other way so that broken Things are not created. It’s up to you.

ghys:
> The problem is that such a manually created Thing is BROKEN — it is a zombie without any reference to a bridge instance, and so it simply DOES NOT WORK

Not always, manually created MQTT things seem to work even though they have a 3-part UID without the bridge ID (in this case the broker) - and in fact I can move them to another broker, they continue to work. In this case it's even more logical not to have the broker in the thing UID because of the fact that it does not in fact have a tight link to it, they can migrate to another one if the user chooses.
So it's clearly not a general rule, it depends on the add-on... and it's not being a purist to try to figure out the root cause why certain add-ons break. The fact is, the presence or not of the bridge in the UID should be irrelevant: Things have a [proper reference to their parent bridge](https://github.com/openhab/openhab-core/blob/61e17ce39e83daa7f1d8c41512dead8d95d6e5d9/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingImpl.java#L58) in a field so there shouldn't be any messing around to find it.

That being said, most things added through discovery (e.g. Hue etc.) do have a tighter link to their bridge, which is unlikely to change, and in fact are added with a reference to it in their UID, so it's perfectly acceptable that the UI would do the same. And since it fixes things immediately, I'd prefer to merge this fix ASAP and try to tackle the deeper design flaw later.

andrewfg:
> I'd prefer to merge this fix ASAP and try to tackle the deeper design flaw later.

Makes sense to me.

ghys:
As there doesn't seem to be any more objections, I'll merge this to work around the bugs described in #290.

Issue #266, opened 2020-05-26
[Main UI] Widgets to include in the standard library


ghys:
Let's discuss some ideas of additional built-in (standalone) widgets to include in the UI for layout pages. The library will also include some "list item" widgets, to discuss separately.

As a reminder it's not in the plans currently to let users inject their own Javascript code (including libraries) so we have to take into account what we want to offer. Webpack's code splitting allows to split heavy dependencies into separate chunks loaded on-demand so we should leverage that as much as possible - but still consider the impact of these libs in the UI bundle size. Also remember that the whole app is prefetched when on HTTPS for caching purposes (once it's completely cached a PWA "add to home screen/desktop" button appears).

### Gauges

- [ ] Framework7 has a gauge component which may be enough in most cases.
https://framework7.io/vue/gauge.html - note that it can already be used in a custom widget without a dedicated widget, using expressions.

Other alternatives:

- [ ] Canvas-Gauges: https://canvas-gauges.com/documentation/examples/
Bundle size:  42.4kB minified, 12.9kB minified+gzipped ([source](https://bundlephobia.com/result?p=canvas-gauges@2.1.7))

- [ ] Steel series: https://nicolas-van.github.io/steelseries/
Bundle size: 733kB minified, 100.8kB minified+gzipped (!) ([source](https://bundlephobia.com/result?p=steelseries@1.0.7))
(big, but this is the whole library, we can include only the components we need)

None of these are interactive but they cover a lot of potential usage - thermometers, compasses (for wind direction?), general-purpose gauges.
Note that ECharts, the library used for charting in the app, also has a decent gauge chart type (https://echarts.apache.org/examples/en/index.html#chart-type-gauge) but the gauge chart type is not included at the moment. Also worth mentioning is https://github.com/ecomfe/echarts-liquidfill which can be included to nicely visualize tank fill levels etc.

### Thermostat

- [ ] Nest thermostat clone: https://www.svidget.io/examples/nestthermostat
(Maybe a little too blatant. Obviously we would change it a little and drop the Nest logo)
I haven't found any other good alternatives to this one yet.

### Image

[...]

PR #301, opened 2020-08-08
Widget definition API overhaul

ghys:
This changes the way a widget component is described from a simple
object in the default export in SFCs to a function that will only
be evaluated when needed; this should improve performance a bit
when the widget isn't being edited, and allows to use helper functions
to build the parameters in a more concise way. More logic can be
introduced when building these definitions i.e. i18n: translating
them in the currently active language.

All widget definitions can also be considered to be assets rather than
code (even if there's some logic in their building) so they've been
moved to the assets/definitions/widgets folder and simply imported in
the component's SFC; this allows to re-use parameters in a cleaner
way - for instance, the toggle card and toggle list item of the
standard library accept the same parameters coming from the toggle
widget from the system library, also all cards/list items support a
common set of parameters related to the card/list item themselves.


ghys:
I believe this is ready for review (if there are any takers for this 73-file PR ;) and merge, I tested it and there doesn't seem to be any regressions, and it's a better foundation before we add more widgets.

In particular as discussed in #266 I'd like to add a whole new class of controls inspired by the new Android 11 power menu quick controls - a button/cell to put in a grid with the ability to open a drawer with more controls:

image

(by the way, this new Android power menu allows apps to register as control providers: https://developer.android.com/preview/features/device-control
Maybe that's something @openhab/android-maintainers  would like to investigate for the openHAB app 😉)

mueller-ma:
> Maybe that's something @openhab/android-maintainers would like to investigate for the openHAB app 

Thanks for the hint and the reminder that I need to adapt the app for Android 11 👍

PR #291, opened 2020-06-29
Fix corrupt SVG icon files and refactor image compression workflow 

philippwaller:
I noticed that some SVG icon files were corrupt and could not be processed with the available scripts nor opened with common SVG viewers. For example, executing the "_iconconvert" script generates the following error messages: 

```
…
presence.svg: Error: invalid color at 11:41.
recorder.svg: Error: invalid color at 19:124.
screen.svg: Error: invalid color at 15:41.
washingmachine_2-0.svg: Error: invalid color at 29:38
…
```

To avoid manual correction of the affected files and to further optimize the compression process, I have adapted the scripts and tooling. Here is a list of the changes I made:

-  Added viewBox property to the none.svg file. This property is necessary to ensure error-free processing in the subsequent process.
- Replaced the `svgcleaner` tool with `scour`, which in our case not only produces better lossless compression results, but also automatically corrects the corrupted files. (all svg files, `svgcleaner`: 837.092 Byte -> `scour`: 794.586 Byte, -5.07%)
- Replaced `pngout` by `optipng` since `pngout` is is closed source, it's license prohibits bundled distribution and `optipng` delivers better lossless compression results) (all png files, `pngout`: 514.752 Byte -> `optipng`: 494.134 Byte, -4.01%)
- Added the `pngquant` tool for lossy compression to further reduce the size of the png files without noticeable loss of quality. (all png files, before: 494.134 Byte -> after: 236.338 Byte, -109.08%)
- The conversion script has been refactored and renamed in order to use the new tooling.
- Dockerfile has been added to execute the script platform independent.

The script as well as the new tooling can be tested most easily with the provided docker container. Just execute the following command in the `/openhab-webui/bundles/org.openhab.ui.iconset.classic` directory:

`docker run --rm -it -v ${PWD}/src/main/resources/icons:/workdir $(docker build -q .) ./.convert.sh`

You should see following outcome:

```
...
(374/393): washingmachine.svg
====================================================================================
scour:          lossless svg compression        (1953 bytes -> 1763 bytes, -9.73%)
rsvg-convert:   convert svg to png              (1763 bytes -> 2266 bytes, 28.53%)
optipng:        lossless png compression        (2266 bytes -> 1296 bytes, -42.81%)
pngquant:       lossy png compression           (1296 bytes -> 588 bytes, -54.63%)

(375/393): washingmachine_2-0.svg
====================================================================================
scour:          lossless svg compression        (3618 bytes -> 3244 bytes, -10.34%)
[...]

PR #312, opened 2020-08-23
Fix time icons

philippwaller:
I just saw that the icons "time.svg" and "time-on.svg" are faulty (second hand incorrectly positioned). The files were already broken before PR #291, but the new workflow has now applied the bug to the PNG versions. This PR will fix the issue.


ghys (approved):
Thanks!

PR #317, opened 2020-08-30
[Main UI] Cell Widgets

ghys:
This introduces a new first-level class of widgets in the
standard library: cells.

They are heavily inspired by
https://developer.android.com/preview/features/device-control
and can either perform an action on a click/short tap, like
toggling a switch or perform something stateless like opening
a modal, navigating, running a rule...

It will also expand on a long tap (if the device doesn't support
touch, a button will have to be added). If no short tap/click
action is configured it will default to expanding the cell.

Ways to alter the design of the cell button (when not expanded)
and the expanded controls, eventually with personal widgets
configured in the developer tools, are also on the roadmap.
This will allow the community to develop specialized widgets
for e.g. thermostats, weather forecasts etc.

A series of cells tailored to common scenarios (toggle +
dimmer or color picker) and maybe a few more are envisioned.
There will be a way to define the default cell widget for an
item using the "cellWidget" metadata namespace, in the same
manner as the existing default standalone and list item widgets.

The container, oh-cells, can be added to a block in a layout
page, instead of a regular row. It will automatically arrange
the cells, which have a fixed height and are all of the same
size.


ghys:
image
Expanding on a mobile device fills up the entire screen:

Expanding on a large screen opens the detailed controls in a modal with a blurred backdrop (the animation and blurring will be configurable):
ezgif-4-a1617c3c9abc

ghys:
Examples of cells with controls!
(note the `title`, `footer`, `on`, and various other parameters are hardcoded here, but can be made dynamic with expressions. That will be simplified before merging though).
Slots can already be used to insert custom controls in the expanded cell, below the main control - or in the empty space for `oh-cell`).

```yaml
component: oh-slider-cell
config:
  title: Lights
  footer: Master Bedroom
  color: yellow
  on: true
  buttonIcon: lightbulb_fill
  header: On - 80%
  item: Color1
slots:
  afterSlider:
    - component: oh-button
      config:
        text: A button
        color: blue
        fill: true
        large: true
        class:
          - margin-top
```

image

```yaml
component: oh-knob-cell
config:
  title: Thermostat
  footer: First Floor
  color: blue
  buttonIcon: thermometer_fill
  on: true
  header: 23 °C - Cooling
  item: Dimmer1
slots:
  afterKnob:
    - component: oh-button
      config:
        text: Cool mode
        color: blue
        outline: true
        large: true
        class:
          - margin-top
        action: options
        actionOptions: COOL,HEAT,AWAY,OFF
```

image

```yaml
component: oh-colorpicker-cell
config:
  title: Hue Bar
  footer: Living Room
  header: Off
  item: Color1
```

image

```yaml
component: oh-cell
config:
  title: Fear Inoculum
  subtitle: TOOL
  footer: 3 rooms
  buttonIcon: music_house
  on: true
  color: teal
  header: Playing
slots:
  default:
    - component: oh-image
      config:
        item: Receiver_TrackPictureURL
    - component: oh-player-controls
      config:
        item: Receiver_PlaybackControl
```

image

ghys:
Quite happy with the design now, it's getting there; some more code to ease the configuration still necessary but we'll have something ready soon enough... Try it on https://tiny.cc/ohdemo (if you went there previously go to About first and click on the "purge caches" button!)

hmerk:
Just gave it a try, really looks amazing on my iPhone. Can't wati to get this merged and have my openHAB3 test environment up and running.

ghys:
- Components in the `background` slot replace the default background and are displayed behind the header
- Components in the `header` slot replace the default header - note that the width is not the width of the cell when not opened so you can't rely on things like `width: 100%` when styling.

Example:

```yaml
    - component: oh-knob-cell
      config:
        header: test
        title: A really long title to check the wrapping
        item: Spotify_Volume
      slots:
        background:
          - component: f7-gauge
            config:
              type: semicircle
              size: 180
              borderBgColor: "#33333370"
              borderColor: "#0000ff40"
              value: =Math.fround(items.Spotify_Volume.state)/100
              class:
                - margin
                - card-opened-fade-out
    - component: oh-slider-cell
      config:
        item: Spotify_Volume
      slots:
        header:
          - component: Label
            config:
              text: Volume
          - component: Label
            config:
              text: =items.Spotify_Volume.state
              style:
                font-size: 50px
```

image

ghys:
There's always room for improvement, but I think there's nothing pressing that would justify delaying this any longer, so...

ghys:
Last minute addition:

How to highlight the card based on an expression using the state of an item:
```yaml
component: oh-label-cell
config:
  action: analyze
  actionAnalyzerItems:
    - ElectricityMeter_Power
  item: ElectricityMeter_Power
  title: Apparent Power
  stateAsHeader: true
  expandable: false
  on: =Number.parseInt(items.ElectricityMeter_Power.state) >= 350
  color: red
```

If you don't specify the `on` parameter but there is an `item` parameter the widget will try to determine the highlight status from the state automatically, that is:
- highlight if ON
- number values: highlight if >0
- HSB values: get the third part (brightness) and highlight if >0

Highlighting only occurs when there's a `color` parameter set which is the color of the highlight.

hmerk:
Excellent 👍

digitaldan:
Looks awesome @ghys !

ghys:
Thank you.
It'll be interesting to know whether versatility (our community is passionate and can deal with complexity) vs. ease of use (pander to new users) is the best way to go, I'm trying to find a balance but time will tell.

A recent screenshot of my playground page while I worked on this, for illustration purposes:

image

digitaldan:
I'm not complaining about having choices, and advanced ones at that!   But, you have a valid point about complexity.  One strength of competing platforms is out of the box widgets that match things, so if I add thermostat to the system. i get a thermostat widget pre made and installed in the UI.  We offer endless customization to build your own thermostat widget, but requires a level of knowledge of our UI widget system that has a steep learning curve.  I don't have a good answer unfortunately .

hmerk:
Indeed a valid point. Introducing a first time setup wizard was a great step forward to get new users faster into openHAB, so I completely agree with @digitaldan that we would need a set of predefined widgets for easy building a functional UI. Additionally, there should be the possibility to extend the widget set with customized widgets, ideally with the possibility to get them from some kind of marketplace, similar to widget gallery for habPanel.

hmerk:
@ghys could you please explain how to use an items state as header ?
the following does not work: 
`component: oh-cell
config:
  stateAsHeader: true
  title: Beleuchtung
  subtitle: Arbeitszimmer
  color: yellow
  action: toggle
  actionItem: Ceiling_Light_Office
  actionCommand: ON
  actionCommandAlt: OFF`

ghys:
> I'm not complaining about having choices, and advanced ones at that! But, you have a valid point about complexity. One strength of competing platforms is out of the box widgets that match things, so if I add thermostat to the system. i get a thermostat widget pre made and installed in the UI. We offer endless customization to build your own thermostat widget, but requires a level of knowledge of our UI widget system that has a steep learning curve. I don't have a good answer unfortunately .

> Additionally, there should be the possibility to extend the widget set with customized widgets, ideally with the possibility to get them from some kind of marketplace, similar to widget gallery for habPanel.

Agreed, the UI only offers basic building blocks, usually to control a single item, but there is no built-in controls for real-world devices. That's partly explained by the fact that in openHAB items are largely independent of each other and there's no authoritative standard to relate them in a reliable way. The semantic tags are a step in the right direction but we're not quite there yet. Moreover these relations are not always the same, e.g. sometimes for a media player you have power/input/volume/mute/album art controls available, sometimes you don't.  For the weather the forecast period varies, sometimes you have the wind speed or humidity, sometimes you don't.
So users need to configure their equipment widgets, like thermostat, media player, etc., by manually specifying every relevant item, which is not ideal especially for a built-in widget.

Also design choices always end up being rather personal and subjective... You know how it is - while few people will care too much about which version of Karaf openHAB runs, when it comes to UX everyone and their grandmother will have an opinion. That's why I'm counting a lot about having a marketplace where users can have a choice among several options, and easily pick and choose the more specialized widgets that they like most.
@rgrollfitz has already done a good job of evaluating what can be done with this widget system here: https://community.openhab.org/t/wiki-building-pages-in-the-oh3-ui-documentation-draft-2-3/104392/11
and I'm sincerely hoping there will be more initiatives like this over time, so we'll eventually have a well-furnished marketplace :)

ghys:
> @ghys could you please explain how to use an items state as header ?
the following does not work:
component: oh-cell
config:
stateAsHeader: true
title: Beleuchtung
subtitle: Arbeitszimmer
color: yellow
action: toggle
actionItem: Ceiling_Light_Office
actionCommand: ON
actionCommandAlt: OFF

You need to specify the item with `item: Ceiling_Light_Office`, I'm realizing that the option is not described in the basic `oh-cell` (it is however in the derivatives like `oh-slider-cell`), that should be fixed.

hmerk:
Great, thanks @ghys 👍

PR #327, opened 2020-09-08
[habpanel] Use UI components for backend storage

### LINKED ISSUE #320: [habpanel] Can not load config in 3.x
Apologies if this has already been reported (or is a known issue).  Moving my main server to OH3.x this weekend.    HABPanel in 3.x can not load its config from the REST service due to our authentication changes. I know there was some talk about moving this out to not use the config service, but i don't know where that stands yet.  Thanks!

```
Request URL: http://oh:8080/rest/services/org.openhab.core.i18n/config
Request Method: GET
Status Code: 401 Unauthorized
Remote Address: 192.168.91.15:8080
Referrer Policy: no-referrer-when-downgrade
``

### LINKED ISSUE #323: [habpanel] Exception thrown using custom widgets in 3.x
After importing my habpanel export file to my new 3.x setup, none of my custom widgets work (nothing renders).  I don't think much should have changed in the web app, so i'm not clear what could be breaking this (maybe not 3.x related?) .  The console log shows this stack trace:

```
vendor.js:119 TypeError: Cannot read property 'undefined' of undefined
    at render (http://oh:8080/habpanel/app/widgets/template/template.widget.js:39:45)
    at Object.link (http://oh:8080/habpanel/app/widgets/template/template.widget.js:125:13)
    at http://oh:8080/habpanel/vendor/vendor.js:16:71
    at http://oh:8080/habpanel/vendor/vendor.js:82:187
    at sa (http://oh:8080/habpanel/vendor/vendor.js:82:244)
    at n (http://oh:8080/habpanel/vendor/vendor.js:68:6)
    at g (http://oh:8080/habpanel/vendor/vendor.js:59:428)
    at g (http://oh:8080/habpanel/vendor/vendor.js:59:445)
    at n (http://oh:8080/habpanel/vendor/vendor.js:67:452)
    at g (http://oh:8080/habpanel/vendor/vendor.js:59:428) <widget-template ng-model="vm.previewmodel" class="ng-pristine ng-untouched ng-valid ng-isolate-scope">
....
Show 3 more frames
vendor.js:119 TypeError: Cannot read property 'undefined' of undefined
    at render (template.widget.js:39)
    at template.widget.js:122
    at m.$broadcast (vendor.js:149)
    at settings.widgets.designer.controller.js:69
    at vendor.js:160
    at f (vendor.js:45)
    at vendor.js:48
```

ghys:
This PR moves the storage backend for HABPanel's panel registry
from the service configuration to the "habpanel:panelconf" UI
components namespace, so they're accessible without authorization.

The panel registry object format is converted back and forth to
the UI components format during loading/saving.

To save, it tries to find a refresh token from the main UI stored
in LocalStorage and uses it to get access tokens - a new access
token is requested for each save. Whether or not implementing a
complete OAuth2 authorization code flow (given the refresh token
if there for use as it's the same origin as the main UI, and users
can be taught to first sign in with the main UI to be able to edit
HABPanel) can be debated and is still t.b.d.

(The code to load the configuration with the service config, like
in OH2, has been left but it is not loaded. It will also work
because it tries to get access tokens as well.)

The other options that were formerly in the service configuration
have been changed as well:

- the "initial panel configuration" is now a "default"/"starred"
flag on the panel configuration itself, it can be set within
HABPanel itself (the "star" icon after the panel config name in
Settings), only one should normally be starred at any given time,
but it's not enforced - it's the admin responsibility to ensure
that.

- the "lock editing" is now tied to the presence of the refresh
token; this means, signing out in the main UI will automatically
disable the editing features in HABPanel because there's no more
possibility to get an access token to save.

Similarly, the locale is taken from /rest/ instead of the i18n
service directly.

Some missing translations have been added and some changed.

Fixes #320.
Fixes #323.


ghys:
I'd like to have this ASAP because it fixes severe regressions, I believe it's working OK.
There's no automatic upgrade path to migrate the registry in the service configuration to the new UI components, but I prefer to address that in another PR.
Manually exporting to .json files with the embedded "raw configuration editor" in OH2 and reimporting them again in OH3 will work, and is always an option, not ideal though.

PR #328, opened 2020-09-08
[Main UI] Pass access token in header if cookie flag set

ghys:
This makes a check for the presence of a non-HttpOnly
`X-OPENHAB-AUTH-HEADER` cookie; if found, pass the current
access_token in the `X-OPENHAB-TOKEN` HTTP header instead of
the `Authorization` header as Bearer credentials.

This is explicitely supported by the backend and mitigates
the collisions with authorization schemes set by proxies like
openHAB Cloud.


ghys:
I believe with this and https://github.com/openhab/openhab-cloud/pull/307, https://github.com/openhab/openhab-addons/pull/8429 and https://github.com/openhab/openhab-core/pull/1631 we have the OH3 auth/tokens working in myopenHAB!

PR #336, opened 2020-09-15
Rule editor: module type grouping

ghys:
Sort module types in rules editor and group them by scope.
(following the discussion in https://github.com/openhab/openhab-core/issues/1639#issuecomment-692874984)

Fix blank labels when a config parameter with has both options and a
"thing" context.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/config-parameter.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-parameter.vue
@@ -49,7 +49,7 @@ export default {
   computed: {
     control () {
       const configDescription = this.configDescription
-      if (configDescription.options && configDescription.options.length && configDescription.limitToOptions) {
+      if (configDescription.options && configDescription.options.length && configDescription.limitToOptions && !configDescription.context) {
         return ParameterOptions
       } else if (configDescription.type === 'INTEGER') {
         return ParameterInteger
--- a/bundles/org.openhab.ui/web/src/components/config/controls/parameter-thing.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/parameter-thing.vue
@@ -1,5 +1,5 @@
 <template>
-  <thing-picker :title="configDescription.label" :value="value" @input="updateValue" />
+  <thing-picker :title="configDescription.label" :value="value" @input="updateValue" :filter-uid="configDescription.options.map((o) => o.value)" />
 </template>
 
 <script>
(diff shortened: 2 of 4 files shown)


ghys:
Before

image

After 

image


Note: about the scopes order - core appears first, then alphabetically

Also I'm going to stop prefixing [Main UI] to PRs and issues, if there's no prefix we can assume it's related to the main UI.

wborn:
Certainly a nice improvement! :+1:
I think it's best to remove those add-on prefixes from actions labels when all add-ons are migrated to OH3.
They may still be useful for users of OH 2.5 UIs which probably don't use have such nice grouping. :wink:

kaikreuzer:
I'm not fully convinced about the visuals, it somehow makes it look more fragmented and considering that we'll have probably many small actions of different add-ons, it will be many single-line boxes.
Would it be possible to e.g. name the scope in grey right-aligned in the same box or something like that?

ghys:
Those are standard controls so in general I think it's best to leave them be wrt styling.
Also the UI when _adding_ a module is different than when you change the module type of an existing module - debatable whether you should be allowed to or not - which is a simple "smart select" (https://framework7.io/docs/smart-select.html#examples):

image

you cannot change it, and it's consistent with the module creation UI, but OTOH the creation could be completely different - like big boxes for common, essential modules at the top...

wborn:
For less fragmentation the UI could also prefix the scope to the label:

* core: enable or disable rules
* core: run rules
* core: send a command
* habot: send a web push notification to HABot
* hue: send a light command with a custom fade time
* media: play a sound
* media: say something
* mqtt: publish an MQTT message
* network: Send WoL packet
* script: execute a give script

I think you want the scope above or to the left of the labels so it is clear what is grouped on and you can easily find a specific group (which is less easy when adding the scope to the right).

kaikreuzer:
OK, let's use the standard, which the framework brings. 👍

kaikreuzer (approved):

ghys:
I still think we could tailor the "add module" experience and have for instance big buttons at the top for featured module types like "when an item receives a command". When modifying a module's type it doesn't really matter what the UI looks like because it's not supposed to be a common scenario.
Still I think merging this PR is an improvement ;)

kaikreuzer:
Yes, if you have good ideas to "feature" the most common of the core modules, please go for it!
Is changing a module type at all possible? I wouldn't even support this as people should rather delete the old and add a different type (afaik, that's how Paper UI dealt with it).

Issue #293, opened 2020-07-09
[Default UI] Default values for configuration parameters are not visible


cweitkamp:
A lot of configuration parameters have default values. Currently those are not visible when editing the configuration. Thus the user does not know that there is a default and what its value. I suggest to prefill or select those values. An example:

```
...
		<parameter name="retry" type="integer" min="30" unit="s">
			<advanced>true</advanced>
			<label>Retry</label> <!-- German "Wiederholungen" in Screenshot -->
			<description>The retry parameter specifies how often (in seconds) the Pushover servers will send the same
				notification to the user.</description>
			<default>300</default>
		</parameter>
		<parameter name="expire" type="integer" min="0" max="10800" unit="s">
			<advanced>true</advanced>
			<label>Expire</label> <!-- German "Verfall" in Screenshot -->
			<description>The expire parameter specifies how many seconds your notification will continue to be retried for.</description>
			<default>3600</default>
		</parameter>
...
```

![grafik](https://user-images.githubusercontent.com/5131747/87001800-8f449080-c1b8-11ea-92d0-c246f9ac7f85.png)

Additionally we can define the unit attribute (`unit="s"`) or a `<unitLabel>`-tag for configuration parameters. Would be nice to display them next to the configuration parameter so we can omit adding it in the description.

ghys:
You're absolutely right, I noticed it a while ago and this has been on my backlog ever since...

PR #339, opened 2020-09-16
Allow 0 as a valid integer param value

### LINKED ISSUE #338: [Main UI] Integer parameter input won't accept '0' and doesn't respect max/min
I have a parameter defined as follows -:

```
			<parameter name="zigbee_networkid" type="integer" min="0" max="3" required="true" groupName="ember">
				<label>Network Index</label>
				<default></default>
				<description>The Network Index used on the NCP</description>
				<limitToOptions>false</limitToOptions>
			</parameter>
```

When entering a value it will not accept 0. I can manually type this in, but it is removed as soon as the input looses focus. If I use the arrows to change the value, I can enter -1, then step up and it's blank, then step up and it's 1.

In the console I see the following when using the arrows to scroll up/down the value -:
```
app.js:12 {"zigbee_networkid":null}
12.app.js:1 Update zigbee_networkid to -1
app.js:12 {"zigbee_networkid":-1}
12.app.js:1 Update zigbee_networkid to -2
app.js:12 {"zigbee_networkid":-2}
12.app.js:1 Update zigbee_networkid to -3
app.js:12 {"zigbee_networkid":-3}
12.app.js:1 Update zigbee_networkid to -2
app.js:12 {"zigbee_networkid":-2}
12.app.js:1 Update zigbee_networkid to -1
app.js:12 {"zigbee_networkid":-1}
12.app.js:1 Update zigbee_networkid to 0
app.js:12 {"zigbee_networkid":null}
12.app.js:1 Update zigbee_networkid to 1
app.js:12 {"zigbee_networkid":1}
12.app.js:1 Update zigbee_networkid to 2
app.js:12 {"zigbee_networkid":2}
```

When it gets to 0, the value is "null".

I should say I'm probably running code that's a week or so behind master in case this was fixed recently.

I should also note that it allows me to enter values outside of the max/min, and I can also save with an out of range value (eg 6). If values are out of range there probably should be an error, and for sure it shouldn't be able to save the configuration.

ghys:
Respect the min, max, step in the config description.

Fixes #338.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
@@ -94,7 +94,7 @@ export default {
       this.showAdvanced = !this.showAdvanced // event.target.checked
     },
     updateParameter (parameter, value) {
-      if (value === undefined) {
+      if (isNaN(value) || value === '') {
         // deleting the parameter sometimes lead to saves not updating it, so set it explicitely to null
         this.$set(this.configuration, parameter.name, null)
       } else {
--- a/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
@@ -94,7 +94,7 @@ export default {
       this.showAdvanced = !this.showAdvanced // event.target.checked
     },
     updateParameter (parameter, value) {
-      if (isNaN(value) || value === '') {
+      if ((typeof value === 'number' && isNaN(value)) || value === '') {
         // deleting the parameter sometimes lead to saves not updating it, so set it explicitely to null
         this.$set(this.configuration, parameter.name, null)
       } else {
--- a/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
@@ -94,7 +94,7 @@ export default {
       this.showAdvanced = !this.showAdvanced // event.target.checked
     },
     updateParameter (parameter, value) {
-      if (!value) {
+      if (value === undefined) {
         // deleting the parameter sometimes lead to saves not updating it, so set it explicitely to null
         this.$set(this.configuration, parameter.name, null)
       } else {
--- a/bundles/org.openhab.ui/web/src/components/config/controls/parameter-integer.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/parameter-integer.vue
@@ -4,6 +4,9 @@
         :name="configDescription.name"
         :label="configDescription.label"
         :floating-label="$theme.md"
+        :min="configDescription.min"
+        :max="configDescription.max"
+        :step="configDescription.step"
         :value="actualValue"
         @input="updateValue"
         :required="configDescription.required" validate


bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
```diff
@@ -94,7 +94,7 @@ export default {
       this.showAdvanced = !this.showAdvanced // event.target.checked
     },
     updateParameter (parameter, value) {
-      if (!value) {
+      if (isNaN(value) || value === '') {
         // deleting the parameter sometimes lead to saves not updating it, so set it explicitely to null
```
ghys, line 98: Perhaps some more explanation is needed here.

When the user clears the input, the natural thing to do would be to remove the config parameter from the resulting config object, but there's a catch: I noticed at least for PUT /rest/services/<serviceId>/config, if you don't specify a parameter, it won't be unset. For example if you first do a PUT of:
`{ "param1": "A", "param2": "B" }`
then:
`{ "param2": "C" }`
the actual config will be:
`{ "param1": "A", "param2": "C" }`

However if the second time you PUT:
`{ "param1": null, "param2": "C" }`
when you GET the config again you will have:
`{ "param2": "C" }`
as expected - "param1" has been unset by setting it to null.

So here if the input is blank (empty string for `<input type="text">` or NaN for `<input type="number">`) it's set to null to unset it from the config when the API is called. Hope it makes sense.

cdjackson:
This may not be related to the changes here, but I'll mention it as I didn't notice this issue before (but I did pull down changes from the past few days)...

I now see an issue with a string input - this is a serial port entry, and I cut and paste the port name into the input field. This now doesn't work - it takes a few times to paste in the data. This is also true if I try and type in the data - the first couple of characters are ignored.

With a cut and paste -:

```
Update zigbee.serialport to /dev/tty.SLAB_USBtoUART72
app.js:12 {"zigbee.serialport":null}
```

And typing in `ssss`...

```
Update zigbee.serialport to /dev/tty.SLAB_USBtoUART72
app.js:12 {"zigbee.serialport":null}
12.app.js:1 Update zigbee.serialport to s
app.js:12 {"zigbee.serialport":null}
12.app.js:1 Update zigbee.serialport to s
app.js:12 {"zigbee.serialport":null}
12.app.js:1 Update zigbee.serialport to s
app.js:12 {"zigbee.serialport":null}
12.app.js:1 Update zigbee.serialport to ss
app.js:12 {"zigbee.serialport":null}
```

You can see above that the first two `s`'s typed in are ignored - only after the third one do we see the s's being appended - although the port is still showing `null`.

When saved, it is also null -:

image

bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
cdjackson: I'm not sure I understand this yet - I need to digest it a bit more, but I guess this is linked to the issue I just added below? Sorry that I didn't see this till after adding that comment.

ghys:
Whoops, this seems to be new... Good thing you tested it!
The latest commit fixed that.

I guess you can put that on the hall of fame of JavaScript oddities:
image

bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
ghys: When you think about it - it should probably fixed in core. If you use a HTTP PUT to update a config it should replace it completely, not merge it with the existing (that would actually be a PATCH - https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH)

cdjackson:
Thanks - I'll try this...

One other question. I thought from discussions we had in the past that your approach was to always save all configuration even if nothing changed (when it's saved). I'm not seeing that - if I set the serial port, I only see the serial port config being PUT - not the baud, flow control etc. These then end up not set at all when the service starts as the defaults aren't applied in the core (but that's another issue). I'm just wondering if that is expected behaviour that you're only sending the serial port in this situation?

cdjackson:
This seems to work now - all issues are working - allowing 0, respecting min/max, and I can add the strings again.

Thanks @ghys.

cdjackson (approved):
LGTM.

ghys:
> One other question. I thought from discussions we had in the past that your approach was to always save all configuration even if nothing changed (when it's saved). I'm not seeing that - if I set the serial port, I only see the serial port config being PUT - not the baud, flow control etc. These then end up not set at all when the service starts as the defaults aren't applied in the core (but that's another issue). I'm just wondering if that is expected behaviour that you're only sending the serial port in this situation?

Yes I still think the Save button on the Thing page should save everything together in one operation, instead of auto-saving immediately when a config parameter changes - among other things, it allows the user opportunities to cancel unwanted changes.

If a parameter has a default value, the field will be pre-filled with that default value however it won't be saved back if the user doesn't touch it. The rationale being that if a parameter is not set, it's probably logical that the default value should be considered. For instance, if the Json Storage service config never got changed, it remains empty, all parameters should be considered to be set to their default values (5 backups, 500ms write delay, 30s max write delay), but when visiting the page the default values will still be shown (this wasn't the case before #293).

If these assumptions lead to issues we can have another look and maybe save the default values.

cdjackson:
I guess it makes some sense, but it does lead to issues. I've not checked the Json service recently (although I originally wrote it!), but I've written another service, and the issue is that default parameters are not provided by the config system - unlike default parameters for thing configuration. This means that the service needs to hold the default in multiple places - the XML file for the UI, and in the Java file to be used if not provided by the UI/config system - this isn't especially nice to have this duplication.

Maybe @kaikreuzer can comment on this? Is it possible for the system to provide defaults for services, or since this is provided directly from OSGi is that not possible? If it's not possible, then maybe the UI should set these?

kaikreuzer:
@cdjackson Are you aware of the method `ConfigUtil.applyDefaultConfiguration()`? What you need at hand is the config description itself, but if you have it, there is no need to duplicate default information in Java.

cdjackson:
Ok, thanks. From looking through a few services in the core, all of them duplicate the configuration between the XML and Java. The `Configuration` is not readily available and I guess we'd have to go through the registry to get this which is why it's not used in the services I looked at at least.

kaikreuzer:
> The Configuration is not readily available

At what place at the code do you think specifically? If you have a `Map` of parameters, you can easily make a `Configuration` out of it.

cdjackson:
It’s within a bespoke service - I guess I can access the config registry, but is there a helper to do the heavy lifting?

PR #393, opened 2020-10-15
[Main UI] Sort addons case insensitive

Hilbrand:
The list of (installed and uninstalled) addons is not sorted.
As a result it depends on what the rest api returns.
In this case the bindings with lowercase first character are placed after the bindings with uppercase.
To fix this after the filter a sort is added.


ghys:
I noticed it too and wondered if it shouldn't be the API's job to sort them properly, but let's do it on the client anyway 👍

ghys:
https://github.com/openhab/openhab-webui/blob/c6eadd4d51b9c0d20843667b3b052cfc51acdab7/bundles/org.openhab.ui/web/src/pages/wizards/setup-wizard.vue#L388-L389

could you do it here as well please?

ghys (approved):
Great, thanks!

PR #405, opened 2020-10-16
[Main UI] Player control rewind and fast forward

### LINKED ISSUE #400: [Main UI] Extend player control to include fast forward and rewind buttons.
The player command class not only supports play/pause and previous/next, but also fast forward and rewind. The current player control does not have buttons for this.
It would be nice to have these buttons available to model players that support it in the UI. It might be a property of the control if these buttons are shown or not (not all bindings support fast forward and rewind).

mherwege:
Closes #400 

Add optional buttons for rewind and fast forward to player control.


ghys (approved):
LGTM, thank you very much!

PR #426, opened 2020-10-21
[Main UI] Created input list item and card control

mherwege:
- Extended oh-input to show an (optional) button to send an update command to the linked item. If used, it avoids having to create a separate button to update the item. The broader scoped variable will also still be available, but an internal variable is used inside the control when the button is present. So the broader functionality should not be impacted.
- Created oh-input-item for use in lists.
- Created oh-input-card

I use this with a binding I am currently developing that may need text input through the UI (UPnPControl). E.g. the binding allows to provide a search string as input to search for music.

I am not a web developer and these are my first steps in doing any web, vue, f7 and javascript development. Nevertheless, I hope it is useful. It certainly was a good exercise for me.


ghys:
@mherwege many thanks! This will certainly be useful to many.

I took the liberty of adding some changes of my own:

- some reorderings, styling & label changes

- if there is no button to be shown do not build the f7-row/f7-col structure, just render the f7-input component

- rename `showOK` option to `sendButton` (since there's already `clearButton`) and it sends a command

- add `sendButtonConfig` option (not in the widget description because object options cannot be defined yet, so a little bit undocumented) to allow altering the button design, all options will be forwarded to the f7-button, for example:
```yaml
component: oh-input-card
config:
  type: text
  item: Test
  sendButton: true
  sendButtonConfig:
    text: Send!
    fill: true
    color: blue
```
image

For list items I would have preferred using a `f7-list-input` like in https://framework7.io/vue/inputs.html to be honest, but this will do for now, we can always revisit this later.

Thanks again!

ghys (approved):

PR #452, opened 2020-10-26
Close #408 - Allow customizing styles of pages & popups

### LINKED ISSUE #408: [MainUI] Customize css of pages/popups
It would be nice to be able to change the background of pages (page-content class) via css.

ghys:
Expose the config of the root page component in page designer
code views.


rgrollfitz:
Thanks, again awesome work. Is it for page configs only atm or is there a way to interact with the popup config as well? Saw something like modalStyle in your commit but didn´t found a way to interact with it?

PR #458, opened 2020-10-26
Add .gitattributes

wborn:
Without this configuration there are Spotless issues with line endings on Windows.

See: openhab/openhab-addons#8712

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+.java text=auto
+.xml text=auto


ghys (approved):
I've been confronted with this multiple times 😄 
There are menus and commands in Eclipse & VS Code to convert line breaks though, so it's fine, but it's a good to have this sorted 👍

kaikreuzer (approved):

PR #481, opened 2020-11-01
Cosmetics

ghys:
Change equipment icon (to be different than the one for things)
Change layouts of item stats in semantic cards
Fix Equipment tab view title
Change labels during script creation, generate UID


ghys:
This is the last one that I'd like in the release, after that I'm good 🙂

PR #482, opened 2020-11-01
Blockly editor for scripts (experimental)

ghys:
Allow to create new ECMAScript-based scripts with a
Blockly editor: https://developers.google.com/blockly.

The undocumented "blockSource" configuration parameter
is added to the script action module and will trigger
the display of the Blockly editor instead of the code
editor if present.

3 openHAB-related blocks have been added to the standard
toolbox: get item state, send command, print.

image

image

### DIFF
--- a/bundles/org.openhab.ui/web/package-lock.json
+++ b/bundles/org.openhab.ui/web/package-lock.json
@@ -4587,6 +4587,120 @@
         "file-uri-to-path": "1.0.0"
       }
     },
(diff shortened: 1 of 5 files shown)


ghys:
This actually works very well, on phones too... Even if it's incomplete, it has no impact on the app's entry point, and probably would appeal to a lot of users, so I'm including this early version so that it can be battle-tested in M2 👍.

Issue #394, opened 2020-10-15
[Main UI] How to organize i18n source files?


Hilbrand:
To be able to internationalize the new ui all text should be put into the i18n directory (obviously). I want to create some of this. My question is. What is the best way to organize this? A js file per module? or 1 file? and what naming conventions for the keys should be used?

Regarding `empty-states.json`. In the `empty-state-placeholder.vue` it says `TODO: i18n`. Is that still relevant?

ghys:
Good question, getting the i18n infrastructure right the first time is actually a pretty complex matter, which is probably why I've been avoiding it and postponing it for so long ;)

The contents of the i18n folder and their format with `empty-states.json` and so on were temporary until we get that i18n infrastructure done, so they will likely change.

First we need to choose the framework, for Vue projects there's basically vue-i18n, vuex-i18n and vue-i18next.
- [vue-i18n](https://kazupon.github.io/vue-i18n/) is by far the most popular and adopted, it has a good feature set, good documentation and the best [tooling](https://kazupon.github.io/vue-i18n/guide/tooling.html#official-tooling) (I especially like [vue-i18n-extract](https://github.com/pixari/vue-i18n-extract))
- [vuex-i18n](https://github.com/dkfbasel/vuex-i18n) is interesting as well since it uses Vuex, but a little more niche;
- [vue-i18next](https://panter.github.io/vue-i18next/) is based on the larger [i18next](https://www.i18next.com) which is framework-agnostic and has the best features wrt. pluralization rules etc. and can work with gettext .po/.pot files which worked really well for HABPanel.

I'd tend to go with the safest and probably better supported choice - **vue-i18n**.

Now to organize the translation files, there are a lot of options.

First we need to make sure that whatever we decide plays nice with Crowdin, as I suppose we'll keep on using it to crowdsource the translations (no real need to change and the other projects already use it).

Second, since we can expect to have translations in 20+ languages and don't really need to dynamically switch (the locale is loaded from `/rest/` before anything is displayed),  I think it's obvious that we don't load them at once and do [lazy loading](https://kazupon.github.io/vue-i18n/guide/lazy-loading.html) at least to filter out the messages in the other languages. One exception could be the setup wizard where it could be nice to switch after you set the language on the first screen. Not sure if it's really necessary after saving the regional settings, a page refresh does the trick and isn't too much of a hassle. We can actually just do that - refresh the page - if the language/region has been changed.

[...]

PR #507, opened 2020-11-04
Removed HomeBuilder from the repo

kaikreuzer:
I am open to discuss this, but I think with the new UI there isn't really any need or use for the home builder ui anymore.


ghys:
Not disagreeing, but we could salvage a few things from it, like the existing i18n, or the code that translates a model into items definitions.

kaikreuzer:
The code is still on the 2.5.x branch, so I think we can pick anything we'd like at any time - but no need to continue shipping it with the 3.0 distro.

kaikreuzer:
@ghys Would you want to merge this one?

ghys (approved):
Yes - I was hoping for @kubawolanin's opinion, but I think it makes sense to remove unmaintained UIs (with no active code owners).

As said above the functionality would be eventually integrated in a different form, but hopefully reusing as much work as possible, in the main UI - like a "model bootstrapper".

I planned to consider it only after #394 in case there were a need for a localization infrastructure first. I doubt it will be ready for 3.0 though. But no need to include a deprecated UI in the distro even at the cost of temporarily losing this functionality, it would only add to the confusion.

Issue #419, opened 2020-10-19
[MainUI] Title of 'oh-list-card' is also visible as alt-tag on the whole content-section


rgrollfitz:
The title already describes the problem - its not that big of a deal, but I don´t think its meant ot be like that.

ghys:
Fixed in #434.

PR #532, opened 2020-11-18
Added tooltips for player control

mherwege:
By default, the control shows the control title when hovering over the buttons. I replaced it by the button function.


ghys:
Thanks, but I'd rather have it opt-in with a config parameter, disabled by default - those symbols on the buttons are already universally recognizable so it's not too useful for everyone.

image

mherwege:
I agree. The issue is that this is what I see when I hover over the buttons when I don't have a tooltip defined:
image

This default behaviour does make sense in many other cases, like the example below, but doesn't look nice for the player control with it's multiple buttons. I don't know how to remove this default tooltip, so opted for putting a proper one.
image

Do you see a way to remove them all together?

ghys:
Wow you're right, never noticed that. It's similar to #419. That's because the widget config is passed to the controls without modifications and if there's a `title` in some cases the browser will display it as a tooltip. 😡
I'll find a solution.

florian-h05:
Superseded by #2426.

PR #544, opened 2020-11-20
Add widgetOrder metadata, fixes

### LINKED ISSUE #489: [MainUI] Properties display: sorting of properties
How properties are sorted ?
I don't see any logic. It does not follow the ordering of definition in my config file. It is not sorted on item name, nor in item label.

### LINKED ISSUE #498: [MainUI] Back button closing the app
On Android phone, the back button of the phone is closing the app. Very annoying.

### LINKED ISSUE #533: [MainUI] Missing item name for the first property in Locations
## The problem

The item is displayed for each property displayed excepot for the first.
In the following example you can see "Temperature dans le bureau" without any item just below, while the second "Humidé dans le bureau" has the item name just below.

image

## Expected behavior

Display the item for the first element too.

## Steps to reproduce

Open any card showing properties in "Locations".

### LINKED ISSUE #542: [MainUI] No simple way to disable the default action of the default list widget
## The problem

The new way to edit the default list widget simplifies the thing, for example to enable the dynamic icon.
One thing I can't disable is the default action that open the chart page.
In the action field, "Analyze item(s)" is not selected and there is no entry like "No action":
image

## Expected behavior

A simple way to disable the default action without redefining a full new list widget, for example with a new action type "No action".

ghys:
Add a new widgetOrder metadata namespace editor to configure a
*default* index-based widget order. (The cards on the home page
will also feature another, more visual way to influence the order
when they'll be rewritten.)

Expose visible/visibleTo on default widget metadata configuration
pages.
Known bug: counters on home page cards are not accurate when some
items are hidden with visible/visibleTo.

Add a mixin to sort items; use it on cards, model treeview, group
popup, add from model in layout pages.

Close #489.

Fix #533 - wrong call to render properties in location cards.

Fix #498 - make cards routable (has glitches).
(but will be superseded with the cards rewrite).

Fix #542 - allow to unset the options set by the default widget
routines. Fix action name ("analyze" -> "analyzer").

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
@@ -5,8 +5,9 @@ export default [
   { name: 'widget', label: 'Default Standalone Widget' },
   { name: 'listWidget', label: 'Default List Item Widget' },
   { name: 'cellWidget', label: 'Default Cell Widget' },
+  { name: 'widgetOrder', label: 'Default Widget Order Index' },
   { name: 'autoupdate', label: 'Auto-update' },
-  { name: 'expire', label: 'Expiration timer' },
+  { name: 'expire', label: 'Expiration Timer' },
   { name: 'alexa', label: 'Amazon Alexa' },
   { name: 'homekit', label: 'Apple HomeKit' },
   { name: 'ga', label: 'Google Assistant' }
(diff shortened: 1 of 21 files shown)


lolodomo:
Bravo.
Sorting based on item label is a major improvement, now we have by default deterministic ordering of locations, equipments and properties.
We are getting very close to something really usable in practice now.

PR #565, opened 2020-11-28
Introduce Mobile App interface

ghys:
This introduces a few functions allowing apps embedding the web UI
in a WebView to interact with them and adjust part of its rendering.

More precisely, the UI will look for an `OHApp` object
(defined in `window`, for instance with `addJavascriptInterface`
for Android apps):
https://developer.android.com/guide/webapps/webview#BindingJavaScript

The `OHApp` object can define the following functions:

- `string preferTheme()`: if present, will be called during the
  initialization of the app, will force the theme - unless the user
  has made a choice on the About page. These values are allowed:
  `auto`, `ios`, `md` (Material Design/Android), `aurora` (desktop);
- `string preferDarkMode()`: if present, will be called during the
  initialization of the app, will force the color scheme - unless the
  user has made a choice on the About page. These values are
  allowed: `auto`, `light`, `dark`;
- `void pinToHome()`: if present, this will show an icon on the home
  page which will call this function when tapped;
- `void exitToApp()`: if present, this will show an icon on the home
  page which will call this function when tapped. It **replaces** the
  "Other Apps" icon showing the right-side panel.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -243,6 +243,11 @@ export default {
   },
   data () {
     let theme = localStorage.getItem('openhab.ui:theme')
+
+    if ((!theme || theme === 'auto') && window.OHApp && window.OHApp.preferTheme) {
+      theme = window.OHApp.preferTheme()
+    }
+
     // choose Aurora as default theme for desktops
     if ((!theme || theme === 'auto') && this.$device.desktop) {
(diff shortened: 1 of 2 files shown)


ghys:
@openhab/android-maintainers @openhab/ios-maintainers

here's a minimal interface as discussed here: https://github.com/openhab/openhab-webui/pull/259#issuecomment-716197988

I'm going to merge it as defined above so you can test it- let me know how it works for you.

PR #578, opened 2020-11-29
Model cards: consider the more specialized equipment class

### LINKED ISSUE #567: [MainUI] Missing cards in equipment view
## The problem

In Equipment view, it looks like a card is created only for "root/high-level" equipment.
First example, I was expecting a card "Television" but I get a card "Screen" containing my groups marked as "Television".
Second example, I get a card "Sensor" containing my groups marked as "MotionDetection"; I was expecting a card "Sensor" (as I have few of my items marked with "Sensor") buyt also a card "MotionDetector".

## Expected behavior

In the equipment view, create one card of each equipment present in the user model. As an example, if the user is tagging with "Television", a card "Television" should be present.

ghys:
Fix #567.


lolodomo:
Thank you, works as expected now.

PR #517, opened 2020-11-14
Update LastNPE EEAs to 2.2.1

wborn:
Updates the dependencies so the latest EEAs are used.

PR #580, opened 2020-11-30
cometvisu: fix nulltype mismatch

maggu2810:
Instead of returning null, we return an empty map so the caller does not need to care about.

This fixed the following problem:

| Description | Resource | Path | Location | Type |
| --- | --- | --- | --- | --- |
| Null type mismatch (type annotations): required '@NonNull Map<@NonNull String,@Nullable Class<? extends @NonNull State>>' but this expression has type '@Nullable Map<@NonNull String,@Nullable Class<? extends @NonNull State>>' | ReadResource.java | /org.openhab.ui.cometvisu/src/main/java/org/openhab/ui/cometvisu/internal/backend/rest | line 255 | Java Problem |


peuter (approved):

wborn:
There's also https://github.com/openhab/openhab-webui/pull/517 with similar fixes.

ghys:
What to do with this PR now that #517 is merged? There are conflicts now.

PR #601, opened 2020-12-07
basic ui: Properly handle escaped characters when parsing quantities

lucasec:
## Description

Quantity values can sometimes contain scientific notation. For example `7E+1 °F` (70 degrees Fahrenheit). The Basic UI backend URL encodes this value before rendering into the data attribute of the setpoint widget, so the value of the raw `data-value` attribute is `7E%2B1+%C2%B0F`.

To calculate the up/down steps, we of course have to parse this value. The current logic hands the URI-encoded blob directly to parseFloat, which (due to the escaping) fails to identify the scientific notation. In the example above, the 70 degrees gets parsed as 7 degrees--obviously an unintended result. This fixes the issue by decoding the escaped characters before attempting to parse the number.

## Changes

- Wrap value parsing in 'decodeURIComponent`

## Testing

I backported this to the `2.5.x` branch and validated on my setup, which consists of several Z-Wave thermostats (that are often set to 70 degrees). Validated that the number is parsed correctly and the up/down arrows behave as expected.

## Additional Notes

Note that the code that results in the state (data-value) attribute getting URL encoded is here: https://github.com/openhab/openhab-webui/blob/2.5.x/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java#L131

Whether or not encoding the state is the correct behavior in this situation I would leave up for debate. Additionally, passing the entire state string (with unit included) to `parseFloat`, does work, but feels somewhat like a hack since we rely on the function to ignore spaces and the unit characters (god forbid a unit abbreviation ever includes a digit or other character that might trip up the parser).

Issues handling scientific notation seem to be part of a wider class of bugs. The behavior is similarly broken in the iOS app (PR forthcoming).

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
@@ -106,11 +106,13 @@ public abstract class AbstractWidgetRenderer implements WidgetRenderer {
         snippet = StringUtils.replace(snippet, "%visibility_class%",
                 itemUIRegistry.getVisiblity(w) ? "" : "mdl-form__row--hidden");
 
-        String state = getState(w);
-        snippet = StringUtils.replace(snippet, "%state%", escapeURL(state));
+        String rawState = getRawState(w);
+        snippet = StringUtils.replace(snippet, "%state%", escapeHtml(rawState));
+        snippet = StringUtils.replace(snippet, "%stateUrlEnc%", escapeURL(rawState));
 
-        String category = getCategory(w);
-        snippet = StringUtils.replace(snippet, "%category%", escapeURL(category));
+        String rawCategory = getRawCategory(w);
+        snippet = StringUtils.replace(snippet, "%category%", escapeHtml(rawCategory));
+        snippet = StringUtils.replace(snippet, "%categoryUrlEnc%", escapeURL(rawCategory));
 
         return snippet;
     }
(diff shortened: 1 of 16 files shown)


lolodomo:
The good approach would be, I believe, a little update of the setpoint snippet and backend to provide data-value with the correct format and without unit. data-unit is available for the unit. 

Is there no problem when the page is opened and the item state is updated?

lolodomo:
The good news is that parseFloat in frontend is mainly used for the setpoint widget (also the map widget). 
Finally your solution with the decoding in frontend is not bad. But I suggest the little change in backend to remove the unit.

lucasec:
The state updates in the JSON event stream use the unencoded value, but still with the unit. For example:
```
{"widgetId":"0402","label":"Heat [70 °F]","visibility":true,"item":{"state":"7E+1 °F","stateDescription":{"pattern":"%.1f %unit%","readOnly":false,"options":[]},"type":"Number:Temperature","name":"AdditionThermostat_HeatSetpoint","label":"Addition Thermostat Heat Setpoint","tags":[],"groupNames":["AdditionThermostat"]},"sitemapName":"default","pageId":"default"}
```

So perhaps for consistency we should keep the unit in the data-value attribute, but avoid excessive escaping (only what is needed for HTML attribute). This will also keep consistency with the openHAB rest API.

I will revise the PR when I get a moment and push an amended commit.

lucasec:
Pushed up a new commit (plus rebase on latest master) that includes a more comprehensive look and re-thinking of the encoding of our template attributes. See the commit message for more details. This should at least bring about a bit more consistency and make frontend code easier to reason about.

For quantity states, is it guaranteed they are always formatted as `<quantity> <unit>` (separated by a single space)? If so I could still separate those out, probably a good idea. But there are some caveats such as we pass the %state% value in the icon URL as well, not sure if the backend expects to get the unit with the value when doing icon lookups.

@lolodomo what's the easiest way to get a 3.x.x environment set up for testing? The last change was simple enough I backported to my 2.5 install, but I should probably test this one in a proper environment.

lucasec:
Thanks for reminding me. I need to get a new dev environment set up (I guess I would build against latest 3.x release), so I need to set aside some time. Will see if I can get around to it over the upcoming holiday weekend here in the US.

lolodomo:
@lucasec : I close this PR, similar code is included in PR #1706

PR #613, opened 2020-12-09
Add a "pin" icon to remove the panel breakpoint

### LINKED ISSUE #523: OH3 side view only shows the menu symbol when the screen is narrow
It would be helpful we you could hide the menu bar on the left for pure operating computers or tablets when pages are displayed. With a narrow width thats possible, but not with 1920.

ghys:
Based on a experiment by @hubsif.
Save and restore the current status in localStorage.

Fixes #523.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -91,7 +91,7 @@
         </f7-list-item>
       </f7-list>
       <f7-link class="breakpoint-pin" @click="toggleVisibleBreakpoint">
-        <f7-icon slot="media" size="14" :f7="this.visibleBreakpointDisabled ? 'pin_fill' : 'pin'" color="gray"></f7-icon>
+        <f7-icon slot="media" size="14" :f7="this.visibleBreakpointDisabled ? 'pin_slash' : 'pin'" color="gray"></f7-icon>
       </f7-link>
 
       <div slot="fixed" class="account" v-if="ready">
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -90,7 +90,7 @@
           <f7-icon slot="media" ios="f7:question_circle_fill" aurora="f7:question_circle_fill" md="material:help" color="gray"></f7-icon>
         </f7-list-item>
       </f7-list>
-      <f7-link class="breakpointPin" @click="toggleVisibleBreakpoint">
+      <f7-link class="breakpoint-pin" @click="toggleVisibleBreakpoint">
         <f7-icon slot="media" size="14" :f7="this.visibleBreakpointDisabled ? 'pin_fill' : 'pin'" color="gray"></f7-icon>
       </f7-link>
 
@@ -206,14 +206,14 @@
       width 100%
       margin-bottom 0
       height calc(var(--f7-tabbar-labels-height) + var(--f7-safe-area-bottom))
-  .breakpointPin
+  .breakpoint-pin
     position fixed
     top calc(var(--f7-safe-area-top))
     right 0
-    margin 10px
+    margin 12px 10px
     opacity 0
   @media (min-width 960px)
-    .breakpointPin
+    .breakpoint-pin
       opacity 0.75
 
 .theme-dark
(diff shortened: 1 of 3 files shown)


ghys:
@hubsif I have played with your idea a little and reach something that should work.

I've moved the icon in the logo area, made it more discreet, and used localStorage to retain the status across page refreshes. I looked into the setup wizard problem and haven't experienced any bug.

image

image

I've also made one more CSS rule so the icon doesn't appear when there's no need (on smaller screens).

If you agree with this, please drop a comment with the sign-off line (see https://github.com/openhab/openhab-webui/blob/master/bundles/org.openhab.ui/CONTRIBUTING.md) since it's basically your code and I'll include it in the squashed commit.

hubsif:
Hey, I'm delighted how open-armed and quick such tiny (and probably low-priority) contributions are treated 👍 .
And all of your comments and changes were exactly what I meant when I said it requires a review. And great that you thought of storing the state, I actually forgot to mention that this was still missing.
Will have a quick look at it ...

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -90,6 +90,9 @@
           <f7-icon slot="media" ios="f7:question_circle_fill" aurora="f7:question_circle_fill" md="material:help" color="gray"></f7-icon>
         </f7-list-item>
       </f7-list>
+      <f7-link class="breakpoint-pin" @click="toggleVisibleBreakpoint">
+        <f7-icon slot="media" size="14" :f7="this.visibleBreakpointDisabled ? 'pin_fill' : 'pin'" color="gray"></f7-icon>
```
hubsif, line 94: Actually, for me the filled pin means "pinned". So, I think it should be the other way round: When the breakpoint is disabled the side panel will disappear, meaning the side panel is not-pinned = not filled = 'pin'. What do you think?

```suggestion
        <f7-icon slot="media" size="14" :f7="this.visibleBreakpointDisabled ? 'pin' : 'pin_fill'" color="gray"></f7-icon>
```

bundles/org.openhab.ui/web/src/pages/home.vue
```diff
@@ -156,7 +156,7 @@ export default {
       this.$store.dispatch('stopTrackingStates')
     },
     onPageInit () {
-      this.$f7.panel.get('left').enableVisibleBreakpoint()
+      // this.$f7.panel.get('left').enableVisibleBreakpoint()
```
hubsif, line 159: not sure how you and the OH team usually handle this, but I think with version control it's not required to keep this line as a comment 😉

hubsif (changes requested):

bundles/org.openhab.ui/web/src/components/app.vue
ghys: Yeah the thing is: I'd still like the pinned state to be the default normal state, and unpinned as a non-default alternative. And I'd want the less visible icon to be shown when it's pinned because it'll be always displayed.
And I also don't want a "negative" icon like pin_slash in the default state ;)
So no perfect solutions...
Maybe pinned=pin, unpinned=pin_slash_fill, so that the icon reflects the state, not the action and it's more prominent when it's on the hidden panel?

hubsif: > I'd want the less visible icon to be shown when it's pinned
I'm totally with you.

>  And I also don't want a "negative" icon like pin_slash in the default state ;)
You probably also got that the wrong way when you thought about the slash: pin_slash would "unpinned", so actually not in the default state 😉 
So, seems the accepted solution for us both would be to have "pin" as pinned (=default) and "pin_slash" for side panel hidden... 👍 

```suggestion
        <f7-icon slot="media" size="14" :f7="this.visibleBreakpointDisabled ? 'pin_slash' : 'pin'" color="gray"></f7-icon>
```

hubsif (changes requested):

hubsif:
ok, thinking twice about which icon to display when got me to the following thought:
What it's usually like with "buttons" is that their text is the following action (e.g. "save"). One could think that it's the same with an icon only, like in our case, but I think an icon only like this actually reflects the "current" state, not the "following" - if you understand what I mean (and hopefully agree 😉 )

ghys:
Yes, agreed! 👍

ghys:
@hubsif I'd still like a
```
Signed-off: your name <your email>
```
from you so we can merge this, thank you! :)

vanso-hubsi (approved):

hubsif:
> @hubsif I'd still like a
> 
> ```
> Signed-off: your name <your email>
> ```
> 
> from you so we can merge this, thank you! :)

oh, ok, I did not know.

Signed-off: hubsif <hubsif@gmx.de>

PR #614, opened 2020-12-10
[mainUI] add support for QR code context in configuration

yfre:
add support for context "QRCode" to binding configuration view. 
it can be used for example for HomeKit pairing. 

it looks like this
image

### DIFF
--- a/bundles/org.openhab.ui/web/package.json
+++ b/bundles/org.openhab.ui/web/package.json
@@ -71,6 +71,7 @@
     "moo": "^0.5.1",
     "nearley": "^2.19.6",
     "pkce-challenge": "^2.1.0",
+    "qrcode": "^1.0.0",
     "sprintf-js": "^1.1.2",
     "template7": "^1.4.2",
     "tern": "^0.23.0",
--- a/bundles/org.openhab.ui/web/package.json
+++ b/bundles/org.openhab.ui/web/package.json
@@ -81,6 +81,7 @@
     "vue-fragment": "^1.5.1",
     "vue-knob-control": "^1.6.0",
     "vue-magic-grid": "0.0.4",
+    "vue-qrcode": "^0.3.5",
     "vue2-leaflet": "^2.5.2",
     "vuetrend": "^0.3.4",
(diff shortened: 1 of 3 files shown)


ghys:
Interesting and clever use case for a config parameter!

Normally those are user-editable (and they are, in the code view) so you'd want to be careful, but I can see the clear benefit for the HomeKit users, and you did the implementation, so thanks!

I made 2 changes:
- use the webpack code splitting trick (https://webpack.js.org/guides/lazy-loading/) to put the vue-qrcode component (and the qrcode library in its separate bundle so that it's only loaded when needed (qrcode is [29kb minified](https://bundlephobia.com/result?p=qrcode@1.4.4), it's not that much but it doesn't hurt either!).
- there are some light linting rules in place, when you develop the UI make sure to use for instance the ESLint VSCode extension to see and fix those errors (it doesn't break the build but still):
image

ghys (approved):

yfre:
> Interesting and clever use case for a config parameter!
> 
> Normally those are user-editable (and they are, in the code view) so you'd want to be careful, but I can see the clear benefit for the HomeKit users, and you did the implementation, so thanks!
> 
> I made 2 changes:
> 
> * use the webpack code splitting trick (https://webpack.js.org/guides/lazy-loading/) to put the vue-qrcode component (and the qrcode library in its separate bundle so that it's only loaded when needed (qrcode is [29kb minified](https://bundlephobia.com/result?p=qrcode@1.4.4), it's not that much but it doesn't hurt either!).
> * there are some light linting rules in place, when you develop the UI make sure to use for instance the ESLint VSCode extension to see and fix those errors (it doesn't break the build but still):
>   image

@ghys  thank you for the review and feedback with explanation. i learned something. it was my first UI PR, im more the backend guy :)

ghys:
No problem 👍 
This should go into the CONTRIBUTING.md file...

These are the VSCode extensions that are most useful:
- Vetur (so useful it's almost required!)
- ESLint (probably required too)
- language-stylus (if you work on stylus)
- Nearley (if you work on the nearley grammars)

PR #638, opened 2020-12-14
Fix startlevel parameter casing

ghys:


kaikreuzer:
Did you see https://github.com/openhab/openhab-core/pull/1923/files#diff-a6a3c1453732714bbb77a8d6d87afcb0de30787cb082f93dc71ebb7310a3e580R14-R35?

Do you actually need to duplicate the information in the UI? If so, it should imho be aligned with what we have in the core description (texts as well as only levels >40).

ghys:
No - but I did it before you added the options... let me see if I can replace it quickly with the sheet for the module.

ghys:
Okay, that's the last one 💪

kaikreuzer:
Awesome, you used every minute before the RC1 merge stop ;-)

PR #648, opened 2020-12-16
webui: new Video Card widget.

tarag:
Creation of a new main UI card to display video in a page.

This relies on an embedded js player (video.js). The main advantage is that it allows play back of HLS streams in browsers where they would not be natively supported such as Chrome. Using the IP camera binding, HLS streams can be created by openhab without transcoding the camera H264 stream, avoiding some CPU work and quality degradation compared to MJPEG transcoding that allows playback in existing image cards. 

The player provides controls that allow seeking to the past and then going back to live (when streaming an ip camera this can be useful). It may also offer several features that I have not yet explored.

Signed-off-by: Gautier Taravella <tarag@mailbox.org>


ghys:
Thanks for your first contribution @tarag!
It should be a nice addition for some users, and the code looks pretty good already.

I have one major reluctance to accepting it as is, though.
According to https://bundlephobia.com/result?p=video.js@7.10.2, _video.js_ is an additional 512.1kB burden on the size of the app. All widgets are part of the "entry point" meaning the code for them is part of the initial "app.js" bundle that gets loaded immediately when the page is displayed.
That intiial bundle according to the latest build logs is 2.13 MiB, which is admittedly a lot already, and I'll probably look into reducing it. We can't afford an additional hit on performance with a 25% increase of JS code (not mentioning CSS) loading & parsing on top of that, that should only be required if you have a video widget displayed. Adding anything that significant to the initial bundle at this point should be only limited to features that profits the vast majority of users, which is not the case here.

So I'd ask you to look into webpack's [Lazy Loading guide](https://webpack.js.org/guides/lazy-loading/) and the related [application to Vue components](https://vuedose.tips/dynamic-imports-in-vue-js-for-better-performance/), as well as study how it's done throughout the code of this app, for example https://github.com/openhab/openhab-webui/pull/614/commits/a57ce704d8698de26edb06ddeeb41ff10cb5edb4, and make sure video.js is only loaded when the video widget is rendered. Thanks again!

PS: note that some may have succeeded in displaying HLS streams with the `photo` action opening a photo browser popup, available with many widgets: https://community.openhab.org/t/building-a-oh3-camera-widget/110069/11. You may also want to look into that and see if it matches your needs.

tarag:
Hi, thanks for the reply, I somehow expected your answer but was not aware of the possible solution. I will follow the pointers you sent and try to switch to lazy loading, with makes sense indeed.

Concerning the forum thread, I had seen it yesterday and tried, but unfortunately this does not work for all browsers. HLS is an Apple standard and except Safari where it works natively, you need the JS player. Furthermore this allows having the video directly in the page without a popup which may be more suitable depending on the situation.

ghys (approved):
It's looking good now, still quite a heavy dependency but doesn't have an impact anymore when you don't have use the video widget, so all good. I fixed the code style problems while testing, in the future make sure you have an extension to detect & fix them as you code, see https://github.com/openhab/openhab-webui/pull/614#issuecomment-742694629.

I also added some description about "type" as it was quite confusing, and a undocumented parameter (because the config sheet doesn't support objects) to pass an arbitrary set of options to video.js in YAML:
```
component: oh-video-card
config:
  url: https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/720/Big_Buck_Bunny_720_10s_1MB.mp4
  videoOptions:
    loop: true
    controls: false
```

As it doesn't impact any existing code, I don't have an objection to including it for the release.
Thanks again!

tarag:
Thanks,  I need to understand how to do this using vscode, I assumed this would happen automatically on save.

That's great to be able to pass options to the player directly, I thought I would have had to update the widget as needs would arise.

PR #670, opened 2020-12-19
Add a digital clock card widget to the standard library, closes #669

### LINKED ISSUE #669: Add a clock standard widget
## The problem

Some people would like to have a clock as widget (see [Community Forum](https://community.openhab.org/t/oh3-ui-show-clock/110369) and [a general Javascript issue](https://github.com/openhab/openhab-webui/issues/626)). As long as custom widgets don't support Javascript code, this needs to come as standard widget.

## Your suggestion

A simple digital clock card widget, including date display.

## Additional information

I added a clock widget to the standard library which I'm going to open a pull request for.

hubsif:
Hi!

As it seems that more people would like to have a clock widget, I tried to create one as standard widget.

This pull request adds a "digital clock as a card" widget to the standard library. I did not add it to the system library, as I found this to not be useful (a clock can't really be re-used in custom widgets, but perhaps I got something wrong).

The widget displays a digital clock whose format can be customized according to dayjs. It also supports displaying the date above or below the time and setting the font sizes. In settings, it "hints" some standard formatting for unexperienced users.

I tried to follow the coding standards, though I'm lacking experience with Vue.js and Framework7.

Signed-off-by: Hubert Nusser <hubsif@gmx.de>


ghys:
Thanks @hubsif - I tested it and it works great.

> I did not add it to the system library, as I found this to not be useful (a clock can't really be re-used in custom widgets, but perhaps I got something wrong).

I think I would have done it - a system widget - but there's room for improvements afterwards.

I made some changes related to parameters.

Since it's only an addition and doesn't touch existing code, and at first sight it looks okay, I'll allow it despite the code freeze.

hubsif:
Thank you, Yannick!

So, this is how you conditionally display settings - also wanted to do that, but wasn't able to figure that out ;-).

hubsif:
>> I did not add it to the system library, as I found this to not be useful (a clock can't really be re-used in custom widgets, but perhaps I got something wrong).

> I think I would have done it - a system widget - but there's room for improvements afterwards.

Ah, a good reason for having it as system widget just came to my mind: people could then fully individually style it to their needs using a custom widget, right? Argh, didn't think of that.
But like you said, can also be done later if required...

ghys:
> Ah, a good reason for having it as system widget just came to my mind: people could then fully individually style it to their needs using a custom widget, right? Argh, didn't think of that.

Yes but it's tied to my need to rework the cards and make them inherit from a standard "oh-card".
So no worries for now, these things can be addressed later.

Issue #673, opened 2020-12-19
Rules list: poor performance


morph166955:
I'm on 3.0.0.RC2.  I have (according to the UI) 202 rules, all are text based.  Some are quite long (thousands of lines).  I've noticed that if I pull up the rules window that it's effectively unusable.  Attempting to scroll can take as long as a minute (either with the scroll bar or by clicking a letter on the right).  It does eventually scroll.  I've tried multiple web browsers with the same results.  Chrome and Edge have their CPU go from 3-4% to 20% while the page is up (with no interactions).  I have attempted to enable DEBUG and TRACE logging for org.openhab.ui and org.openhab.core.ui but I don't seem to ever get any output.

schmidmuc:
I see the same behavior, when opening the Rules with many (210) text-based rules.

Doublechecking the network traffic it shows that http://192.168.0.10:8080/rest/events?topics=openhab/rules/*/* causes the problem, because this file/command loads over minutes (in my case 5 minutes), but doesn't receive data (100 kB) and shows status code 200. After some time it freezes.

Console shows only one error:
Uncaught TypeError: Cannot read property 'update' of undefined
    at app.js:27
(anonymous) @ app.js:27

ulijoergens:
Since my upgrade from 2.5 to 3.0 I cannot display the list of rules (about 15 or 20) in the main UI. It just displays the animation forever. Ican pull the list via REST API as a json, though. 
I tried to look into network traffic in the browser console.and logs on the server but I can’t find anything. I also did the usual stuff like clearing the cache.
I don’t know if it is related to this issue here, but I thought I give ot a try.

florian-h05:
Closing as I have never experienced such an issue on openHAB 3.4.x or 4.0.0 SNAPSHOTs.

I guess this has been solved by using `?summary=true`.

I have about 90 rules, all work as expected.

PR #675, opened 2020-12-20
Rules: add item state update to action module wizard

ghys:
Depends on https://github.com/openhab/openhab-core/pull/1970.

Also:
- Add abiilty to create a Blockly script directly.

- Only update statuses that changed in the rules list (might help with #673 and similar to #466/#439).


ghys:
Let's test this one before the release in a snapshot along with the core change.

PR #726, opened 2020-12-29
Fix #538 - Do not concatenate search for items

### LINKED ISSUE #538: [MainUI] Unexplained result of filtering items
## The problem

The results of filtering items include items having apparently no link with the filter.

image

## Expected behavior

Having only items matching the filter.

eikowagenknecht:
Fixes #538

Items parts (label, type) were concatenated together by mistake. Added a space between so searching for "Test" doesn't find e.g. Item "Contras**te**" with type "**St**ring" any more.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
@@ -11,7 +11,6 @@
           class="searchbar-items"
           :init="initSearchbar"
           search-container=".virtual-list"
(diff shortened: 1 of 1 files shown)


ghys:
I figured it was standard behavior of these f7-searchbars, turns out it was my code 😅
Thanks for the fix, may I ask you to rebase/fix the conflict with the remove-diacritics, tried to do it with the web editor on the phone but it doesn't work so well...

eikowagenknecht:
Sure, it is rebased now.

ghys (approved):
Thanks!

Issue #734, opened 2020-12-31
Habpanel event subscription does not work for GroupItemChanges


HolgerHees:
## The problem

in this pull request https://github.com/openhab/openhab-webui/pull/128 we filter event subscription to statechanged events. This does not work if I send a command to a group item.

If I change a group item by sending a command to one of its child items. Everything works fine.

```
2020-12-31 14:30:19.725 [ome.event.ItemCommandEvent] - Item 'pFF_Bedroom_Light_Ceiling_Powered' received command ON
2020-12-31 14:30:19.728 [nt.ItemStatePredictedEvent] - pFF_Bedroom_Light_Ceiling_Powered predicted to become ON
2020-12-31 14:30:19.734 [GroupItemStateChangedEvent] - gFF_Bedroom_Lights changed from OFF to ON through pFF_Bedroom_Light_Ceiling_Powered
``` 

The Group "gFF_Bedroom_Lights" is getting a GroupItemStateChangedEvent, because one i´of it child members (pFF_Bedroom_Light_Ceiling_Powered) was changed. But if I send a OFF command to this group item, I get the following events.

2020-12-31 14:25:16.929 [ome.event.ItemCommandEvent] - Item 'gFF_Bedroom_Lights' received command OFF
2020-12-31 14:25:16.930 [ome.event.ItemCommandEvent] - Item 'pFF_Bedroom_Light_Ceiling_Powered' received command OFF
2020-12-31 14:25:16.935 [ome.event.ItemCommandEvent] - Item 'pFF_Bedroom_Light_Hue_Left_Color' received command OFF
2020-12-31 14:25:16.941 [ome.event.ItemCommandEvent] - Item 'pFF_Bedroom_Light_Hue_Right_Color' received command OFF
2020-12-31 14:25:16.946 [nt.ItemStatePredictedEvent] - pFF_Bedroom_Light_Ceiling_Powered predicted to become OFF
2020-12-31 14:25:16.949 [nt.ItemStatePredictedEvent] - pFF_Bedroom_Light_Hue_Left_Color predicted to become OFF
2020-12-31 14:25:16.955 [nt.ItemStatePredictedEvent] - pFF_Bedroom_Light_Hue_Right_Color predicted to become OFF
2020-12-31 14:25:16.959 [vent.ItemStateChangedEvent] - pFF_Bedroom_Light_Ceiling_Powered changed from ON to OFF

[...]

PR #735, opened 2020-12-31
fix wrong event subscriptions

HolgerHees:
will fix, https://github.com/openhab/openhab-webui/issues/734

and should be backported to 3.x branch too


HolgerHees:
is handled with this pull request already 

https://github.com/openhab/openhab-core/issues/1330

PR #737, opened 2020-12-31
removed ItemStateEvent if case

HolgerHees:
this will never happend, because these events are filtered out before


ghys:
Cool, since the small patch exceptions can't apply for code, could you add a sign-off line (as a comment below) as described in https://github.com/openhab/openhab-core/blob/master/CONTRIBUTING.md#sign-your-work? Thanks!

HolgerHees:
commit is signed now

ghys (approved):
Thanks!

PR #790, opened 2021-01-11
Update bnd to 5.2.0

wborn:
See: https://github.com/bndtools/bnd/wiki/Changes-in-5.2.0

---

Related to https://github.com/openhab/openhab-core/pull/2105

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
     <maven.compiler.target>${oh.java.version}</maven.compiler.target>
     <maven.compiler.compilerVersion>${oh.java.version}</maven.compiler.compilerVersion>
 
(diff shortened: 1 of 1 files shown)


ghys (approved):
Since it was approved in core, I'll go ahead :)

ghys:
@wborn @kaikreuzer should this be cherry-picked for 3.0.x?

wborn:
It doesn't immediately fix end user issues, so I don't think it is useful to be cherry picked.

kaikreuzer:
I agree with @wborn.

ghys:
Ok - though in case a dependency upgrade aims at fixing a vulnerability and has low risk of breaking, like #761, I'd tend to include it in patch releases.

PR #795, opened 2021-01-13
Adding validation on channel id when adding a channel

bigbasec:
Adding validation and warnings when adding a channel with an invalid name as the Unique ID could have been blank which would fail and not let the user know.   

This change would only allow alphanumeric characters and numbers as well as an underscore character.   That should keep it valid when it's added.


ghys (approved):
Thanks! Fixed some indentation errors - don't use tab characters, and please use ESLint to make sure the code is correctly formatted in the future.

bigbasec:
Sorry for the indentation, was a quick and dirty done in notepad.   Won't happen in the future.

PR #800, opened 2021-01-13
Script editor: Basic hints for unknown modes

### LINKED ISSUE #783: [3.1.0-SNAPSHOT - Build #2127]: Make the segmentation of the code window movable (sitemap validator)
## The problem
When using the sitemap creator in the mainUI is being used, the "code" view offers two parts:
a) an upper part for editing the code
b) a lower part with the result of the parsing and possible error codes
Currently the position of the scroll bar is fixed; both parts have the same size:

image

## Your suggestion
Please make the scroll bar movable, so that the size of the two frames can be adjusted according to the current needs (at some point you need more space in the upper frame, at some other point you may need more space in the lower frame)

## Your environment
openHAB 3.1.0 - Build #2127

## Additional information
-

ghys:
(Slightly) less basic hints for Python

Make "add from textual definition" & sitemap code editor
splits resizable (closes #783).

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
@@ -62,6 +62,7 @@ import 'codemirror/addon/edit/closebrackets.js'
 // for autocomplete
 import 'codemirror/addon/hint/show-hint.js'
 import 'codemirror/addon/hint/show-hint.css'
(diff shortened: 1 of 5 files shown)


Nadahar:
The Python hints don't work because the MIME type used for Python isn't `application/python`. There seem to be multiple potentially in use: `application/x-python`, `application/x-python2`, `application/x-python3` and possibly others.

I don't know if is is considered a problem, it looks to me like these custom hints are effectively sidelined.

Issue #824, opened 2021-01-19
Front- and backend do not use a consistent locale format


philippwaller:
**Update**: The problem is caused by an incompatible locale format. 

--------

## The problem
In certain constellations, no link can be clicked/no page can be opened after pressing the back button:
![locale bug](https://user-images.githubusercontent.com/1090452/105193527-e13b7b00-5b38-11eb-95e9-64614a83d5b0.gif)

## Expected behavior

All links should be clickable after using the back button.

## Steps to reproduce

1. Open the page "http://localhost:8080/#!/settings/"
2. Click on "Regional Settings"
3. Click on "Settings" (Back button top left)
4. Try to click on "Regional Settings" again

## Your environment

Tested with Browser:
- Chrome Version 87.0.4280.141
- Firefox Version 84.0.1
- Safari Version 14.0.2 

Tested with OpenHAB Version:

- openHAB 3.0.0 Release Build
- openHAB 3.0.0 SNAPSHOT - Build #2154

```yaml
runtimeInfo:
  version: 3.0.0
  buildString: Release Build
locale: de_DE_#Latn
systemInfo:
  configFolder: /openhab/conf
  userdataFolder: /openhab/userdata
  logFolder: /openhab/userdata/logs
  javaVersion: 11.0.9
  javaVendor: Azul Systems, Inc.
  javaVendorVersion: Zulu11.43+21-CA
  osName: Linux
  osVersion: 4.19.121-linuxkit
  osArchitecture: amd64
  availableProcessors: 6
  freeMemory: 34185352
  totalMemory: 102760448
bindings: []
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: true
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: macos
    pixelRatio: 2
    prefersColorScheme: dark
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: 8
    hardwareConcurrency: 12
    language: de-DE
    languages:
      - de-DE
      - de
      - en-US
      - en
    onLine: true
    platform: MacIntel
  screen:
    width: 2560
    height: 1440
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
[...]

PR #830, opened 2021-01-21
Fix incompatible locale format

philippwaller:
This PR fixes the problem #824.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/store/index.js
+++ b/bundles/org.openhab.ui/web/src/js/store/index.js
@@ -25,7 +25,7 @@ const store = new Vuex.Store({
   getters: {
     apiEndpoint: (state) => (type) => (!state.apiEndpoints) ? null : state.apiEndpoints.find((e) => e.type === type),
     locale: (state, getters) => {
-      const segments = state.locale?.split('_') || []
+      const segments = (state.locale) ? state.locale.split('_') : []
       switch (segments.length) {
         case 3: return `${segments[0]}-${segments[2].substring(1)}-${segments[1]}`
         case 2: return `${segments[0]}-${segments[1]}`
--- a/bundles/org.openhab.ui/web/src/js/store/index.js
+++ b/bundles/org.openhab.ui/web/src/js/store/index.js
@@ -4,6 +4,7 @@ import Vuex from 'vuex'
 import components from './modules/components'
 import states from './modules/states'
 import user from './modules/user'
+import { convertJavaLocale } from '@/js/i18n'
 
 Vue.use(Vuex)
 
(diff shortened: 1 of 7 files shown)


ghys:
Thanks!
I see some other parts with `$store.state.locale`: https://github.com/openhab/openhab-webui/search?q=%24store.state.locale, not sure if those need to be updated as well (it seems vue-i18n handles it better, but you never know). Maybe it would be best to perform the transformation in the `setRootResource/setLocale` mutations in the Vuex store (`js/store/index.js`) or perhaps better, make a getter there to get the frontend compatible locale.

philippwaller:
> Maybe it would be best to perform the transformation in the `setRootResource/setLocale` mutations in the Vuex store (`js/store/index.js`) or perhaps better, make a getter there to get the frontend compatible locale.

Thanks for the food for thought! I totally agree. I’ll look for a more generic solution to address this issue.

bundles/org.openhab.ui/web/src/js/store/index.js
```diff
@@ -23,7 +23,15 @@ const store = new Vuex.Store({
     developerSidebar: false
   },
   getters: {
-    apiEndpoint: (state) => (type) => (!state.apiEndpoints) ? null : state.apiEndpoints.find((e) => e.type === type)
+    apiEndpoint: (state) => (type) => (!state.apiEndpoints) ? null : state.apiEndpoints.find((e) => e.type === type),
+    locale: (state, getters) => {
+      const segments = state.locale?.split('_') || []
```
ghys, line 28: Could you avoid the use of `?.` - I'm not strictly against starting using it at some point, but since it's the first and only occurrence, let's stick with old-school for a little while for consistency in the codebase :)

`const segments = (state.locale) ? state.locale.split('_') : []`

ghys (commented):
Thanks, one little comment left.

bundles/org.openhab.ui/web/src/js/store/index.js
philippwaller: Sounds good. Change is on the way. Should I create a separate PR and update the ESLint rule accordingly ?

ghys: >  Should I create a separate PR and update the ESLint rules accordingly ?

Oh, sure!
I'll merge the rest tomorrow, need to check if anything breaks :)

philippwaller:
@ghys: Out of curiosity, I read up on the Java locale representation. According to JavaDocs the locale sting can contain more information than just language country and script:
```
Returns a string representation of this Locale object, consisting of language, country, 
variant, script and extensions as below:

language + "_" + country + "_" + (variant + "_#" | "#") + script + "_" + extensions

Examples:
en
de_DE
_GB
en_US_WIN
de__POSIX
zh_CN_#Hans
zh_TW_#Hant_x-java
th_TH_TH_#u-nu-thai
```

Therefore, this pull request needs to be revised.

philippwaller:
The locale state getter should be fine now and cover all kind of variants. However, I think it could be further optimized by creating a separate state for the ui locale as the getter is called multiple times. What do you think @ghys?

ghys:
> The locale state getter should be fine now and cover all kind of variants. However, I think it could be further optimized by creating a separate state for the ui locale as the getter is called multiple times. What do you think @ghys?

Perhaps, yes. It's getting a little more complex now so making this code execute only when mutating the locale (and setting a frontend compatible locale in the state) would be better.

bundles/org.openhab.ui/web/src/js/store/index.js
```diff
@@ -23,7 +23,38 @@ const store = new Vuex.Store({
...
+      jsLocale += (language) || jsLocale
+      jsLocale += (script) ? `-${script}` : `${script}`
+      jsLocale += (country && jsLocale.length > 0) ? `-${country}` : `${country}`
+
+      // check whether the browser support the returned locale
+      try {
+        new Date().toLocaleDateString(jsLocale)
+        return jsLocale
+      } catch (e) {
+        return 'default'
```
ghys, line 55: Perhaps return something like `language[-region]` which should always work?

philippwaller:
@ghys: Thank you for your feedback. The new proposal is ready for review. I created a Vuex action for the transformation to keep the mutation functions as lean as possible. 

BTW: In the long run, we should consider to refactor the `app.vue` file and put all asynchronous state loaders into Vuex actions.

ghys (approved):
> BTW: In the long run, we should consider to refactor the app.vue file and put all asynchronous state loaders into Vuex actions.

Right!

PR #831, opened 2021-01-21
Modelview deletekey functionality

bigbasec:
As I was testing my OH setup I found myself always winding up with orphan items in the model view, and clicking to remove them was a bit of a hassle, so I came up with this.   

Might be a better way to do this, just really copied the code from the detail view and refactored for the main page.


ghys:
Since the delete logic is already in `item-details.vue` maybe it would be better to add the handler there instead of in model.vue?

bigbasec:
100% first time ever doing anything w/ F7 or Vue, so little bit of a learning curve here, appologies.   I'll take a look at that though and see what I can come up with.   And yes, this was literally a copy/paste from there and changing it to match the vue.

bigbasec:
@ghys Should be all set now.  Thanks

ghys (approved):
LGTM now!

PR #834, opened 2021-01-22
Adds a delete link to the bottom of the sitemap edit page

bigbasec:
Adds a simple delete link to the bottom of the sitemap edit page, as there is currently no way to delete a sitemap once it's created other than editing the jsondb.


bigbasec:
Removing this, as I found out if you hold down the control key while clicking in the Pages view you can select and remove whatever you want.

PR #814, opened 2021-01-16
Upgrade ECharts to v5, add support for ECharts gauge & chart widget

crnjan:
Based on [feedback](https://community.openhab.org/t/main-ui-echarts-gauge-and-pie-poc/113987) prepared a PR - implemented support for [gauge](https://echarts.apache.org/examples/en/index.html#chart-type-gauge). Adding it to chart page example:

```
config:
  label: Test Gauge
  sidebar: true
slots:
  series:
    - component: oh-data-series
      config:
        type: gauge
        data:
          - value: =Number.parseFloat(items.gBasement_BoilerRoom_HP_RAD_Return.state)
            name: HEATING
        detail:
          formatter: "{value} °C"
        axisLine:
          lineStyle:
            width: 8
            color:
              - - =Number.parseFloat(items.gBasement_BoilerRoom_HP_RAD_MinimalRadiatorReturn.state)
                  / 100
                - "#58D9F9"
              - - =Number.parseFloat(items.gBasement_BoilerRoom_HP_RAD_MaximalRadiatorReturn.state)
                  / 100
                - "#7CFFB2"
              - - 1
                - "#FF6E76"
```

renders

Screenshot 2021-01-16 at 21 03 15

Also added a widget wrapper for `oh-chart-page` so one can use all available ECharts goodies in layout pages, i.e.

```
config:
  label: Test Gauge
  sidebar: true
blocks: []
masonry:
  - component: oh-masonry
    slots:
      default:
        - component: oh-chart
          config: {}
          slots:
            series:
              - component: oh-data-series
                config:
                  type: gauge
                  data:
                    - value: 42
```

renders

Screenshot 2021-01-16 at 21 39 21

A bit heavier example with gauges, line chart and some other cards ...

Screenshot 2021-01-16 at 20 44 34

There are still some pending tasks, as adding ability to hide period controls, ... but would like first to get feedback if going into the right direction.

I look forward to any feedback or comments.

Signed-off-by: Boris Krivonog <boris.krivonog@inova.si>

PR #835, opened 2021-01-22
Add a HABPanel like layouting page (WIP)

hubsif:
So, I finally managed to create a pull request, following up on the [discussion on the forum](https://community.openhab.org/t/experiment-habpanel-like-dashboard-for-mainui/114770).

This PoC PR adds a design page (currently) called "dashboard", which allows to design a grid layout in a drag&drop/resize manner.

More notes to follow...


hubsif:
So, as mentioned in the forum the goal was to be able to create a HABPanel like fixed-size layout. The used library vue-grid-layout also supports responsive layouts.
Its default behaviour is to shrink and grow the columns to match the full width.
Enabling the "responsive" mode makes it lower the number of columns on certain breakpoints and rearrange the items.

While this is a cool thing and certainly helpful for a lot of people, my personal target is more of a fixed layout. I made this choosable in the layout config.
With "fixed screen layout" enabled, the behaviour is changed to:
- adjust and limit the number of rows to match the provided aspect ratio
- set the height of rows to match the width of column, so the grid is a square grid
- automatically do above on screen resizes, so the aspec ratio of grid items is always kept

With all this a layout should look quite alike on different screen resolutions with the same aspect ratio.

Styling:
One issue of this whole resize-layouting is that the current widgets are (from what I see) designed to only use their own required height. Due to this most cards don't really look good in a fixed height item. I think this can mostly be fixed by css - perhaps the existing cards can even be permanently modified to support this (both).

hubsif:
Thinking about where and how to integrate this drag&drop layouting into the UI, I came to another suggestion:
When creating a layout page the user could be asked which type of layouting he wants to use for that page. If you would like to "demote" it a little more, it could also be integrated somehow like in the screenshot below (perhaps with different/better terms), so that the user is always taken to the current layouting style first:

layouting_type

ghys:
> Due to the different behaviour with drag&drop/resize and support for fixed layouts, I meanwhile actually think that it's not a good idea to have it as a third variant in the existing layout page.

Can you elaborate why? I really think it should be seamlessly integrated in the existing layout page.
Looking at the code it seems that a lot is duplicated from the layout page (esp. the designer) and I don't understand why it couldn't be perhaps offered as an alternative to the masonry, as the last element of the page (and still allowing blocks above it).

I also still have reservations about having fixed sized layouts in this app, since its design goal was to be "universal" and as the pages/sidebar are shared, when on a phone for example you'll have access to pages that weren't designed for it and they will look bad. But I don't want to rule them out either because displaying the UI on a tablet à la HABPanel is still a valid use case.

I'm sure there are elegant solutions to this problem, like tagging pages for certain devices only (pages have tags even if the designers don't feature them) and having the ability to filter them according to those; i.e. display pages with a certain tag only with a per-device option and don't display tagged pages by default.

hubsif:
Let me try to re-define my use case:

I'd like to create a "panel page" for my tablet. By "panel page" I mean that I have a fixed screen resolution, no scrolling and can make use of the full space (and for aesthetic reasons I want all cards to be also bottom aligned). I also don't need any sidebars or navigation. Additionally, I'd like to be able to create a layout like this:

panel_layout

With the new UI this seems to not be possible right now. And at this point I didn't care about drag&drop/resize at all.

Now, from a UI that I'm able to create such a layout with I think for a good user experience it would be best if the UI would allow me to predefine my "screen". With the current approach I implemented that by letting the user choose a screen aspect ratio (it could also have been a fixed screen resolution, but at that point I thought the result might be fully scalable).
For best results, the UI would then allow me to put items only in a region that has exactly that ratio (or resolution).
If I had to implement this, I would put it in an extra page layout and on creation initially ask the user about the major parameters, like aspect ratio/screen resolution and number of columns and such, before he can eventually start layouting. I would also try to allow the user to layout in fullscreen mode, as this is how the result will be used.

With all this and for this purpose only(!), mostly due to the fixed region, I don't think it fits in the existing layouting.

<br />

Now, the above is quite different from the purpose of providing the ability to drag&drop/resize cards during layouting, targetting to create a responsive layout while allowing the user to position and size cards in a maybe more flexible or intuitive manner - which I think is more of what you are thinking of, isn't it?
For this purpose only, that new grid layouting would perfectly fit as third option within the current layouting.

<br />

From your comment in the forums I got the feeling you'd like to "demote" that "free-positioning-grid feature" (man, I really can't find a good and short expression for this 😄) a little, so that users wouldn't miss on the current layouting and its (unbreakable) responsiveness.
If I had my way, I would actually implement both: The fixed screen "panel" layouting as mentioned above, with the user being asked for inital parameters, as a new design page called sth. like "Create fixed/fixed screen/panel/fixed panel layout", and the third type of layouting under the current layout page, which will then be called "Create responsive layout".
The former having the fixed screen tweaks I added (like square grid and shrinking on resize), meaning to replace HABPanel sooner or later. The latter to have all the responsive features (and probably vertical compacting) enabled, perhaps even with a modification to support dynamic heights, so it mostly adds the ability to layout adjusting width and position with the mouse.
I think with all this users would be much aware that a "panel layout" will be limited to certain devices/screens.

<br />

About the mix of fixed and responsive pages:
I think it actually wouldn't be much of an issue, since even now you can simply hide the fixed pages from the sidebar. I wouldn't even mind if it wasn't possible at all to add them to the sidebar. On my tablet I probably wouldn't require a sidebar, I'd add buttons that take me to the desired pages instead.

egoist6:
from a user perspective I second this type of layout. This is ideal for wall mounted devices.

ghys:
> I'd like to create a "panel page" for my tablet. By "panel page" I mean that I have a fixed screen resolution, no scrolling and can make use of the full space (and for aesthetic reasons I want all cards to be also bottom aligned). 

Yes I got that!

> If I had to implement this, I would put it in an extra page layout and on creation initially ask the user about the major parameters, like aspect ratio/screen resolution and number of columns and such, before he can eventually start layouting. I would also try to allow the user to layout in fullscreen mode, as this is how the result will be used.

But I still struggle to understand why a separate type of page would be needed to achieve that, or why this couldn't simply be a layout container (oh-grid-layout) similar to the oh-block or a oh-masonry in the existing layout page. You would still configure the aspect ratio of the grid and other things like in your current implementation by configuring the oh-grid-layout widget.

Contrary to the other types, which are very specialized, the layout page is just an empty shell where you put widgets in. In this case, that's exactly what the dashboard/panel layout is, another way to arrange your widgets. Unless I'm missing something (being able to go full-screen even in the design mode can be allowed for oh-grid-layouts) I see no evident reason to give them special treatment. A layout page with a single oh-grid-layout added and configured as "fixed" seems very similar to what you want, as it will be the only thing on the page. And maybe there would even be interesting combinations of grid layouts in responsive mode added with e.g. cells on top. Maybe there would be some code to e.g. make sure full-screen stays on when you navigate to another page and other stuff like that.

> I think it actually wouldn't be much of an issue, since even now you can simply hide the fixed pages from the sidebar. I wouldn't even mind if it wasn't possible at all to add them to the sidebar. On my tablet I probably wouldn't require a sidebar, I'd add buttons that take me to the desired pages instead.

Still I'm warming up to my previous idea of introducing a "only on these devices" option and then have a way to identify the device (could be an arbitrary string, its hostname maybe, or even any criteria that you have in About > Technical Info > Details > clientInfo, including OS, screen resolution...). That way you could have a completely different UI for every tablet if you want - the sidebar, if you need it, would even change because the set of pages to be displayed changed.

hubsif:
> But I still struggle to understand why a separate type of page would be needed to achieve that

ok, I think I now get you, and the answer is actually quite clear to me: usability

Looking at what it's for, which is to create a widget layout, it fits to the current layout page. But at least the fixed screen type is quite different from the current layouting by being non-responsive. 
Now, while I personally would be able to get along with your approach, I think you generally probably never want (nor need) to mix a responsive with a non-responsive layout. I also think that in order to avoid confusion the user shouldn't even be able to do so.
Also, with an extra page I think it's much clearer to the user: "Create a fixed layout" or "Create a responsive layout" - the user has to make up his mind in advance and think about what he wants to achieve and that's exactly where he's taken to.
I my opinion it helps to lead the user to get his desired results.
Actually, the very same could also be achieved if the user was asked about which type of layout he wants to create, when entering the layout page. If he chooses "fixed" (and his ratio/resolution), he then can only use the fixed-screen oh-grid-layout. For "responsive" he can add all three types, blocks, masonry and (responsive) oh-grid-layout.

For me this feels less complex for the user than entering the layout page, having to find out about the three styling types to eventually see that there's a fixed-screen/panel like option hidden in the third style. And then perhaps only after having learned that it's not a good idea to mix those, having to start over again.
Even if users do know how to get there, "to design a panel page, create a new layout page, add one of the third styling types 'grid-layout', click 'configure' and select 'fixed screen mode' (and make sure to not add any other styling type or another grid-layout)" doesn't sound very straightforward, does it? 😉 

<br />

This has become quite some debate 😄. I mean that in a positive way, as of sharing our thoughts to get the best result. I just hope that I'm not stealing to much of your precious time to attend to the many other mainUI things! 😟

ghys:
Okay, so this I believe would work: when creating a layout page, have 2 big buttons "Responsive" and "Fixed size" (looking similar to the big Add Grid/Add Calendar in the chart page designer). When choosing "Responsive" the block+masonry layout appears like before, when "Fixed Size" is selected a grid layout is added in fixed mode.

Whether the page is responsive or fixed-size can be determined by the slots in the root component. If you have a `block` slot (even empty), it's responsive, if you have a `grid` slot (with either a single `oh-grid-layout` component, or maybe we don't even need it, just add the grid layout config at the page level and add the widgets directly in the `grid` slot), it's fixed-size.

hubsif:
sounds good 👍 
Had the same thought about the two big buttons and looked for a styling template. I didn't think of the charts page, so I came up with this for now:
choose_layout

Will adjust it to the charts styling and try to implement your guidelines.

ghys:
Your design looks good too actually!

vanso-hubsi:
Hi Yannick!
So, I've worked on this and while it's not fully finished yet, I already like it a lot.

This is what you're presented when adding a layout:

add_layout

I adjusted it to match the chart page designer.

This is a sample fullscreen layout with most available cards on it:

fullscreen

## Notes

### Integration
As discussed before, the grid is now part of the layout page. I'm sorry, I didn't fully follow your instructions regarding the page code integration. When I started I realized that your approach would not allow the responsive layout to be added to blocks and masonry, which I thought was desired.
Instead, when selecting a layouting style now, the config property `layoutType` gets set to either `responsive` or `fixedSize`. This way both types can be used. If the user chose the fixed-size layout and yet adds blocks or masonry manually in code, an error will be shown. 
The default for `layoutType` is `responsive`, which makes it compatible to existing pages. Hope that's ok for you like that.

### Layouting
To be adjustable in height and width, the items use a flexbox layout showing the first element (title) on top, the second (of class `card-content`) centered in the middle and the third (footer) at the bottom. With title and footer being optional, it requires the `card-content` to be present. This seems to be the case for all cards except the player card, when there's no artist or track, which I adjusted.

Trying a custom widget (an older and customized version of the weather widget in the forum), I think I found an issue with framework7 cards. In their docs regarding card slots they state:

> ### Card Slots
> &lt;f7-card&gt; component has the following slots:
> - default - element will be inserted as a child of card content element
> - content - same as default
> ...

But that doesn't seem to be true. When using the `content` slot, elements get wrapped (correctly) in a `card-content`. With the `default` slot this wrapping does not happen. In my weather widget after I changed the slot from `default` to `content` the weather widget was displayed ok.

I gave images a special css treatment so they are scaled to fit the available `card-content` in height as well.

I'm still struggling with the `oh-slider`. In editmode it doesn't get the correct sizes without resizing the grid item. Will have to look further into that.

### Screen / vue-grid-layout properties
For easier use I don't give the user the ability to adjust all properties of vue-grid-layout. The user can adjust the number of columns and the UI will adjust the number and height of rows automatically, to provide the most square-like grid.

For fixed-size layouts the user can choose if the layout gets adjusted to the available width, which of course can create styling issues, but generally works suprisingly well!

The responsive mode of vue-grid-layout actually supports adjusting all breakpoints and store a layout for each. I cut that to support only the default, since this would lead to very complex/large YAML notations.

### Naming
As suggested I called the component `oh-grid-layout`. I hesitated to call it like this before, as there already exists `oh-grid-row` and `oh-grid-col` which are actually unrelated and might confuse advanced users a little.

### Issues
#### setRowHeight
The method `setRowHeight` gets called on resize events and many times there is no $ref or $el available. I couldn't figure out why yet.
#### reload
I'm having the issue (in both Firefox and Chrome) that when I reload the edit page of a fixed layout, the height of `oh-grid-layout` doesn't get set due to the `setRowHeight` being called in the `mounted()` hook being called before `$el` is actually available. Some people having similar issues said it has to be run in `$nextTick` and production mode. I didn't manage yet to get production working for trying that, though.

<br />

I'm going to look into the outstanding issues and do more testing. Perhaps you'll somewhen find the time to also have a look at this and give your feedback.

hubsif:
Hi Yannick!
So, I've worked on this and while it's not fully finished yet, I already like it a lot.

This is what you're presented when adding a layout:

add_layout

I adjusted it to match the chart page designer.

This is a sample fullscreen layout with most available cards on it:

fullscreen

## Notes

### Integration
As discussed before, the grid is now part of the layout page. I'm sorry, I didn't fully follow your instructions regarding the page code integration. When I started I realized that your approach would not allow the responsive layout to be added to blocks and masonry, which I thought was desired.
Instead, when selecting a layouting style now, the config property `layoutType` gets set to either `responsive` or `fixedSize`. This way both types can be used. If the user chose the fixed-size layout and yet adds blocks or masonry manually in code, an error will be shown. 
The default for `layoutType` is `responsive`, which makes it compatible to existing pages. Hope that's ok for you like that.

### Layouting
To be adjustable in height and width, the items use a flexbox layout showing the first element (title) on top, the second (of class `card-content`) centered in the middle and the third (footer) at the bottom. With title and footer being optional, it requires the `card-content` to be present. This seems to be the case for all cards except the player card, when there's no artist or track, which I adjusted.

Trying a custom widget (an older and customized version of the weather widget in the forum), I think I found an issue with framework7 cards. In their docs regarding card slots they state:

> ### Card Slots
> &lt;f7-card&gt; component has the following slots:
> - default - element will be inserted as a child of card content element
> - content - same as default
> ...

But that doesn't seem to be true. When using the `content` slot, elements get wrapped (correctly) in a `card-content`. With the `default` slot this wrapping does not happen. In my weather widget after I changed the slot from `default` to `content` the weather widget was displayed ok.

I gave images a special css treatment so they are scaled to fit the available `card-content` in height as well.

I'm still struggling with the `oh-slider`. In editmode it doesn't get the correct sizes without resizing the grid item. Will have to look further into that.

### Screen / vue-grid-layout properties
For easier use I don't give the user the ability to adjust all properties of vue-grid-layout. The user can adjust the number of columns and the UI will adjust the number and height of rows automatically, to provide the most square-like grid.

For fixed-size layouts the user can choose if the layout gets adjusted to the available width, which of course can create styling issues, but generally works suprisingly well!

The responsive mode of vue-grid-layout actually supports adjusting all breakpoints and store a layout for each. I cut that to support only the default, since this would lead to very complex/large YAML notations.

### Naming
As suggested I called the component `oh-grid-layout`. I hesitated to call it like this as there already exists `oh-grid-row` and `oh-grid-col` which are actually unrelated and might confuse advanced users a little.

### Issues
#### setRowHeight
The method `setRowHeight` gets called on resize events and many times there is no $ref or $el available. I couldn't figure out why yet.
#### reload
I'm having the issue (in both Firefox and Chrome) that when I reload the edit page of a fixed layout, the height of `oh-grid-layout` doesn't get set due to the `setRowHeight` being called in the `mounted()` hook being called before `$el` is actually available. Some people having similar issues said it has to be run in `$nextTick` and production mode. I didn't manage yet to get production working for trying that, though.

<br />

I'm going to look into the outstanding issues and do more testing. Perhaps you'll somewhen find the time to also have a look at this and give your feedback.

ghys:
> But that doesn't seem to be true. When using the content slot, elements get wrapped (correctly) in a card-content. With the default slot this wrapping does not happen. In my weather widget after I changed the slot from default to content the weather widget was displayed ok.

That's good to know. Eventually I wanted to add a `oh-card` component, and "derive" all cards (`oh-slider-card`, `oh-toggle-card` etc.) from it i.e. use slots defined in `oh-card` for the contents/header/footers - like the cells (the cards were done before I had the idea).

> As suggested I called the component oh-grid-layout. I hesitated to call it like this as there already exists oh-grid-row and oh-grid-col which are actually unrelated and might confuse advanced users a little.

I agree it's confusing. I have no better suggestion, shouldn't have called them `oh-grid-row` and `oh-grid-col` but it's too late now.

> For easier use I don't give the user the ability to adjust all properties of vue-grid-layout. The user can adjust the number of columns and the UI will adjust the number and height of rows automatically, to provide the most square-like grid.

Then you can be sure in time someone will open an issue asking how to configure a missing option :)

> For fixed-size layouts the user can choose if the layout gets adjusted to the available width, which of course can create styling issues, but generally works suprisingly well!

I understand but find this very confusing - it's fixed size, but adapts to the width? Then what's responsive mode? I can imagine the average user getting very confused by this.

> The responsive mode of vue-grid-layout actually supports adjusting all breakpoints and store a layout for each. I cut that to support only the default, since this would lead to very complex/large YAML notations.

I think being able to adjust at least the number of columns __per breakpoint__ would be nice (but still provide sensible defaults) - like the width by breakpoint for the `oh-grid-col`.
https://jbaysolutions.github.io/vue-grid-layout/guide/06-responsive.html is what I would expect the responsive mode to be.

Some other notes:

- The animation should be removed, because when you remove an item, it's a little annoying.



- Why is it responsive only in Run mode? It should look exactly the same in both modes in my opinion.



- Moving/resizing doesn't work on touchscreens, which is a must and a design goal for this app that every screen is usable even on a small smartphone screen. You should be able to scroll the page AND move/resize without struggling.



I think you should add `touch-action: none` to your move and resize handles and it will work, like the interact.js warnings in the console suggest (I hope these warnings disappear in production mode!)

- Context menus too often become partially obstructed and unusable (this is also the case with the masonry layout, in particular, so I had to fix a lot of z-index orders)



Without all these problems fixed or mitigated, I don't think it's ready for prime time yet, but it's slowly getting there! 🚀 

Now a suggestion:

I think it might be best to introduce a special "rearrange" mode where you don't see the actual widgets, only placeholders, like in HABPanel, and the move/resize handles. When you're not in "rearrange" mode the widgets are shown and can be configured but you're not able to move or resize them. I think this will prevent a lot of problems.

When you click on the "Add Widget" (to rename from "Add Item" IMO) then you switch to rearrange mode, and you have a button on the right to exit the rearrange mode. Where the "Configure Grid" button is, you could have a little square button with a menu for:
- entering/exiting the rearrange mode
- configuring the grid
- removing the grid

You could also have a button to exit the rearrange mode directly outside the menu.

hubsif:
> I understand but find this very confusing - it's fixed size, but adapts to the width? Then what's responsive mode? I can imagine the average user getting very confused by this.

Well, "responsive" for me refers to: "adapt to the screen size by rearranging elements as required". That doesn't happen with the other layout, it's just scaled.
The idea behind this is to be able to view the fixed-size layout for different reasons on different screens, e.g. a 1280x720 layout on a second tablet's 1024x600 screen (non-native), which might actually work well.

Though you might be right, I also don't know in how far the term "responsive" is clear to the average OH user. While writing this I had the following idea: Instead of "Fixed size", what about calling it simply "Fixed" (-layout) - as opposed to "responsive"? And with scaling enabled the configured screen resolution is then just the "optimal" screen size.

Actually that average user was the reason I added those explanations on the choose-layout cards - so that everyone would be able to understand what each type is about. This can certainly be reworked.

> I think being able to adjust at least the number of columns per breakpoint would be nice (but still provide sensible defaults) - like the width by breakpoint for the oh-grid-col.

I will look into this. Actually, we could fully support the responsive configurations vue-grid-layout provides. The two major reasons I skipped that were the huge YAML configuration this would result in and the question how to explain this complexity to the user ...

> Why is it responsive only in Run mode? It should look exactly the same in both modes in my opinion.

You've got me there 😁 . Actually, this is a workaround related to above: When vue-grid-layout is in responsive mode, it stores the item arrangements in a layout for that breakpoint only. And then it's not possible to get the x,y-values for the large/main mode, which are the only ones stored right now ...

> Moving/resizing doesn't work on touchscreens, which is a must and a design goal for this app that every screen is usable even on a small smartphone screen.

Oh, I neither knew nor thought about this (like one doesn't always have the themes in mind)! From now on will also test with my phone and tablet 😉 .

> I think it might be best to introduce a special "rearrange" mode where you don't see the actual widgets, only placeholders, like in HABPanel, and the move/resize handles. When you're not in "rearrange" mode the widgets are shown and can be configured but you're not able to move or resize them. I think this will prevent a lot of problems.

Hm, my first thought on this is that if I can't see the widgets while resizing, I cannot see how good they fit into the grid item - which I'd actually like to, and actually think can cause problems if not visible. What's your goal behind the rearrange mode? Do you think it's generally a better UX or does it explicitly target at "touch" usability?

> "Add Widget" (to rename from "Add Item" IMO)

Hm, in my opinion to add a "widget" when you click the placeholder's "+". Before that I think you add an item/element/container/frame/component/... - but not a widget yet 😜 . But I'm not native English and thus don't feel too safe deciding on english terms 😄 .

ghys:
> Well, "responsive" for me refers to: "adapt to the screen size by rearranging elements as required". That doesn't happen with the other layout, it's just scaled.

I know, I just wanted to act like a beginner to these concepts with these questions "it's fixed size, but adapts to the width? Then what's responsive mode?"
There should be no confusion between them, in names and descriptions. One adapts the number of columns and their widths based on the page width, another adapts the column widths only, not their number, and then you have a fixed layout which might overflow the screen width.
I'm not sure yet how this should be presented, will think about it in the next days.

> Actually, we could fully support the responsive configurations vue-grid-layout provides.

Even if it's not documented - when possible - I pass the config object from the OH component to the underlying component (with `v-bind="config"`). Letting people play with esoteric options (eventually with help on the forum) seems to scale better than "it's not supported/ok then please support this" discussions.

> When vue-grid-layout is in responsive mode, it stores the item arrangements in a layout for that breakpoint only.

I think that's fine, you can design for whatever your "primary" screen is and let the library figure it out how to rearrange for the other ones.

> Hm, my first thought on this is that if I can't see the widgets while resizing, I cannot see how good they fit into the grid item - which I'd actually like to, and actually think can cause problems if not visible. What's your goal behind the rearrange mode? Do you think it's generally a better UX or does it explicitly target at "touch" usability?

Mostly to declutter the UI when there's a lot of moving around, since I suspect there will be problems (things become too small and cannot be drag and dropped for some reason... for example there's a problem with the swiper in touch mode. Or you could add a blurry-translucent layer in front of the widget to make sure it can't be interacted with, there are a lot of options.
The "rearrange mode" would be something like the "jiggle" mode that you have in the iOS home screen:
https://www.macworld.com/article/3566286/how-to-add-remove-and-customize-widgets-in-ios-14.html
You can only move widgets in that mode and you exit it with the "Done" button.
image

> Hm, in my opinion to add a "widget" when you click the placeholder's "+". Before that I think you add an item/element/container/frame/component/... - but not a widget yet 😜 . But I'm not native English and thus don't feel too safe deciding on english terms 😄 .

You're right but in that case I think you can abstract the "container" away, not in terms of actual components but in concepts at least - the widget container has no configuration and it's removed when you remove the widget anyways. And instead of the "transient" mode with the placeholder I would be fine with the widget menu appearing as soon as you click the "+" button to add a new one. Lastly, "Item" seems wrong here, it makes more sense in a list.

hubsif:
The issue with storing the responsive layout is that it requires to store index, position and size (5 values) for each widget and each breakpoint (default: 4). For one single widget this creates 24 lines of yaml code 😮 , like this:

```yaml
  - component: oh-grid-item
    config:
      lg:
        x: 5
        y: 4
        w: 2
        h: 2
        i: '0'
      md:
        x: 3
        y: 3
        w: 2
        h: 2
        i: '0'
      sm:
        x: 2
        y: 5
        w: 2
        h: 2
        i: '0'
      xs:
        x: 2
        y: 7
        w: 2
        h: 2
        i: '0'
```

For 10 items this would be 240 lines of yaml. Technically that's of course not an issue, but it makes manual editing quite hard. What do you think?

> I think that's fine, you can design for whatever your "primary" screen is and let the library figure it out how to rearrange for the other ones.

Well, I don't know if that works well. Would have to experiment with it.

hmerk:
Any Progress here ????
Would love to see this feature implemented...

hubsif:
Hi Hans-Jörg,
Actually, I am going to push an update later today...

hubsif:
So, it's been a while since the last update, but here it is!

Notes:

### Responsive mode

Since this whole idea originally was about adding a fixed layout for tablets and with the responsive mode turning out to be more complex than assumed, I dropped the responsive mode for now and focused on the fixed mode,
with still having the responsive mode in mind for later addition.

### Fixed mode changes/features

The most significant changes first:

#### Widget menus

When I looked into the menu overflow/z-index thing, new issues came up which made me search for alternatives. I actually never liked the framework7 menu design much and I get the feeling that framework7's maintainers don’t put much attention to it.
I decided to try with “popovers”, assuming to not like it much either. But big surprise: It turned out to be a cool, much better solution! Not only do the popovers (and their frameless menu icon) look better, they also provide a real good UX on all devices. And the best of it is that they detect and show up where there is space available, so no more fiddling with menus that drop down below the end of the page!

#### Fullscreen mode

I desperately wanted to support a fullscreen-editmode, so it’s possible to design the layout on the targeted device in WYSIWYG manner.
In editmode you can now go into fullscreen, which hides all panels and bars. It then shows a FAB menu at the bottom right for adding widgets and such. This also works for the “special” fullscreen modes, i.e. iOS devices in browser and homescreen mode.

For things like modals to still work, this requires to put the whole app (not just the oh-grid-layout div) into fullscreen. For this I did some tweaks which are not best-practice, e.g. the way I hide the navbar. I didn’t want to touch the parent pages (like “layout-edit.vue” or “page-view.vue”) before you approve that feature. Once you do, this can be cleaned up.

#### Hide navbar

Related to above is the option to hide the navbar when the page is displayed (in regular mode). Together with the sidebar pin this allows users to have a fullscreen-like appearance without having to use the browser’s fullscreen mode (which requires extra user-interaction in all browsers). This is interesting for users like me that use sth. like the Fully app on their tablet.
For navigation a small menu icon is shown overlayed on the top left corner, which opens the sidebar. This can be disabled in preferences. Additionally, a fullscreen icon on the top-right corner can be enabled in settings.

#### Detect free spot

When adding a widget the app now looks for a free spot to put it on (instead of just putting it in the middle, which placed the widget below existing, possibly exceeding the bottom).

#### Widget size limit

The minimum size of widgets is now limited to 50px. I think less doesn’t really make sense and this avoids styling issues with (unintentionally) tiny widgets, which cannot be grabbed properly anymore.

#### Display resolution

In editmode the currently defined screen size is displayed, together with the detected screen size. For Retina displays an extra hint/tooltip is displayed. This is for users to easily determine their “css” resolution. Many users with Retina displays probably don’t know about the difference between physical and css pixels. This is to aid them in finding their correct setting.

#### Rearrange mode

Well, this actually is not a change 😄. I followed your thoughts and tested on different devices and found this to actually not be required (anymore). I added the touch-action css and extended the handles’ size a little and IMHO it works just fine on all tested devices (well, except for iOS devices, where iOS claims to be smarter in some situations). Also, this way the UX is consistent with responsive layouts.
Please test the current version and let me know if you still see the need for it.


### Outstanding issues

#### Obstructed context menus

This is actually still an issue I haven’t found a solution for yet. The problem is that I think that it definitely should be avoided that widgets can overflow their grid-item. But with the widgets being children of grid-items (and a lot of css position styles on most elements by the framework), it can’t be solved with adjusting z-indexes. Perhaps you can think of a way to tackle this…

#### Safe areas

I just found another (minor) issue. The "safe areas" on iOS devices reduce the screen size, so the hint is wrong in these cases. Generally, I don't fully understand when the bars on iOS are displayed or not. I'll try to understand that better.

#### Non-reproducible issues

During development I have experienced miscellaneous issues which I wanted to attend to but cannot reproduce anymore (e.g. the navbar to be hidden when it shouldn’t). I can just hope they somehow solved themselves 😉.

<br />

While all this doesn’t look like big chances, it yet has been quite time consuming. Making it work on all devices requires a lot of fiddling with css and such, and every time you change a parameter, you have to check if it’s still ok on other devices.
I also had to learn about some framework7-vue characteristics. One being that it keeps the previous page loaded, which makes components not being destroyed and thus e.g. still listening to events. This also results in differences if you navigate to a page within the app or if you load the page directly via url. Another one is that pages are actually not ready in the Dom in mounted hooks, even not in nexttick. And then I stumbled across that 9 years old [firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=748518) that probably made you add a margin at the end of edit pages ... 🙂

BTW: I tested on Chrome, Firefox and Safari on Mac, Fire OS 7 (based on Android 9), iPhone iOS 13 and iPad 8th gen simulator running iOS 14. Though I didn’t test everything in-depth on all devices 😉.

<br />

All in all I'm quite satisfied already and for me this would be ready to go into "beta phase". Please let me know your opinion!

hmerk:
Great Job, thanks !

hubsif:
So, I decided to cleanup that fullscreen stuff and eventually touch the other components like layout-edit, and what can I say - it's so much cleaner! Less code, less error prone and no more ugly tweaks.
I dropped one feature, though, that's the pseudo-fullscreen mode for iOS homescreen apps, as you can achieve quite the same by choosing to hide the navbar and unpinning the side panel.

With this fullscreen support now covers layout pages in general, meaning you can also go fullscreen editing for responsive pages. And you could actually also set the navbar to be hidden for responsive pages, though that's not part of the responsive settings dialog (yet).

Additionally, with this comes a general change, which I think I should open an extra pull request for: In "Run mode" the tabbar (Design/Code) now gets hidden. I can't count how often I switched to Run mode, then clicked on "Code" and wondered why it didn't show up! I somehow accidentally stumbled across this and then realized that it IMHO actually doesn't make sense to be able to switch those tabs with the same still being displayed.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-grid-item.vue
```diff
@@ -0,0 +1,127 @@
...
+  <grid-item v-bind="$attrs" v-if="visible" class="oh-grid-item card no-margin" @moved="movedEvent" @resized="resizedEvent" dragAllowFrom=".drag-handle">
+    <template v-if="context.editmode">
+      <f7-link :popover-open="'.item-popover-' + _uid" class="configure-item-menu" ><f7-icon f7="menu"></f7-icon></f7-link>
+      <f7-popover :class="'item-popover-' + _uid">
+        <f7-list>
+          <f7-list-item v-if="context.component.slots.default.length > 0" title="Configure Widget" @click="context.editmode.configureWidget(context.component.slots.default[0], context)" link="#" popover-close no-chevron><f7-icon slot="media" f7="gear_alt" /></f7-list-item>
+          <f7-list-item v-if="context.component.slots.default.length > 0" title="Edit YAML" @click="context.editmode.editWidgetCode(context.component.slots.default[0], context)" link="#" popover-close no-chevron><f7-icon slot="media" f7="square_pencil" /></f7-list-item>
+          <f7-list-item title="Remove Item" @click="context.editmode.removeWidget(context.component, context.parent, 'grid')" link="#" popover-close no-chevron><f7-icon slot="media" f7="xmark_circle" /></f7-list-item>
+        </f7-list>
+      </f7-popover>
```
ghys, line 11:
```suggestion
      <f7-popover :class="'item-popover-' + _uid" :backdrop="false">
        <div class="display-flex margin justify-content-center">
          <f7-link href="#" v-if="context.component.slots.default.length > 0" class="text-color-blue display-flex flex-direction-column margin-right" :popover-close="'.item-popover-' + _uid" @click="context.editmode.configureWidget(context.component.slots.default[0], context)" icon-f7="square_pencil">Configure</f7-link>
          <f7-link href="#" v-if="context.component.slots.default.length > 0" class="text-color-blue display-flex flex-direction-column margin-right" :popover-close="'.item-popover-' + _uid" @click="context.editmode.editWidgetCode(context.component.slots.default[0], context)" icon-f7="doc_text">YAML</f7-link>
          <!-- <f7-link href="#" class="text-color-blue display-flex flex-direction-column margin-right" @click="context.editmode.copyWidget(context.component, context.parent)" icon-f7="scissors_alt">Copy</f7-link> -->
          <f7-link href="#" class="text-color-red display-flex flex-direction-column" :popover-close="'.item-popover-' + _uid" @click="context.editmode.removeWidget(context.component, context.parent, 'grid')" icon-f7="trash">Remove</f7-link>
        </div>
      </f7-popover>
```

It makes the popover of like the one on the floor plan pages, it's better for consistency sake I believe.
This needs some work, but it would be even better if the drop shadow and "angle" (callout) would resemble those (overriding CSS variables maybe, and additional classes). Leaving out the "copy" link because there's no paste.

image

ghys (commented):
Thanks, just tested it and it looks way more logical and useable now, thanks!

One small comment first, I'll review the rest soon, but it looks like there's not much left to nitpick :)

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-grid-item.vue
ghys:
```
    background-color: #fff;
    border-radius: 12px;
    box-shadow: 0 3px 14px rgb(0 0 0 / 40%);
    will-change: transform, opacity;
    font: 12px/1.5 Helvetica Neue,Arial,Helvetica,sans-serif;
```
on the popover makes it look quite similar (at least in the desktop theme).

hmerk:
@ghys Any chance to get this into M2 ???

hubsif:
> It makes the popover of like the one on the floor plan pages, it's better for consistency sake I believe.

ok, I also like consistency and that design is also nice. Though, I think the default framework7 popover design matches the theme better and I'd rather adjust the floor plan popover instead 😉.
Also, I just learned that the framework7 popovers are actually made for vertical layouts, as they have a fixed width! I tweaked that to be somewhat dynamic, since it didn't look good with the Remove item only (had a lot of white space on the left and right). And with those popovers being fixed in width I hope things don't break in certain situations.

hmerk:
Too late for M2...

hubsif:
so, I attended two issues:
One is the outstanding overflow issue with e.g. list item menus. Since there doesn't seem to be a possible css solution, I found another approach: overflow is now possible in edit mode, while it's hidden in run mode.
It's not perfect, but I think it's an acceptable compromise. Actually, in edit mode it somewhat helps to see an overflowing item's actual size, which eases resizing the widget 😉.

Another was an issue with "swipers": when added as a widget, also often on resizing, they didn't get the correct size. I figured that this is probably due to `setDimensions` being called in `$nextTick`, which makes grid-layout sizes being set after the swiper has been rendered.
I discovered that the swiper can observe its parent and activating this solved all issues. I found this to be the best solution and don't think it can cause any performance issues.

ghys:
Thanks, there are ESLint errors which will now result in a build failure, so could you have a look to fix them:
```
ERROR in 
I:\dev\openhab\git\openhab-webui\bundles\org.openhab.ui\web\src\components\widgets\layout\index.js
  8:63  error  Newline required at end of file but not found  eol-last

I:\dev\openhab\git\openhab-webui\bundles\org.openhab.ui\web\src\components\widgets\layout\oh-grid-item.vue
   4:83  error  Expected no space before '>', but found               vue/html-closing-bracket-spacing
  15:1   error  Expected indentation of 30 spaces but found 6 spaces  vue/html-indent
  15:66  error  Trailing spaces not allowed                           no-trailing-spaces
  16:1   error  Expected indentation of 30 spaces but found 6 spaces  vue/html-indent

I:\dev\openhab\git\openhab-webui\bundles\org.openhab.ui\web\src\components\widgets\layout\oh-layout-page.vue
   4:1  error  Expected indentation of 16 spaces but found 22 spaces  vue/html-indent
   5:1  error  Expected indentation of 16 spaces but found 22 spaces  vue/html-indent
   6:1  error  Expected indentation of 16 spaces but found 22 spaces  vue/html-indent
  30:1  error  Expected indentation of 18 spaces but found 10 spaces  vue/html-indent
  31:1  error  Expected indentation of 18 spaces but found 10 spaces  vue/html-indent
  32:1  error  Expected indentation of 18 spaces but found 10 spaces  vue/html-indent
  33:1  error  Expected indentation of 18 spaces but found 10 spaces  vue/html-indent

I:\dev\openhab\git\openhab-webui\bundles\org.openhab.ui\web\src\pages\page\page-view.vue
  155:23  error  Strings must use singlequote  quotes

✖ 13 problems (13 errors, 0 warnings)
  13 errors and 0 warnings potentially fixable with the `--fix` option.
```
Also heads up I'll probably merge #905 soon after I check for regressions so assuredly there will more of them with the additional template rules.

A couple quick cosmetic remarks - but we're getting closer!

- The fullscreen link on the bottom toolbar should be blue IMHO. Also not your fault, but the icon looks kind of inverted 😞 Maybe try `rectangle_arrow_up_right_arrow_down_left` and `rectangle_arrow_up_right_arrow_down_left_slash` for exiting, they seem to look nicer?
- The red background on the grid layout in edit mode looks a bit odd, would you like to change it to a subtle shade of gray instead?
- The FAB when in full screen mode should also be blue, like on other pages, and the icon replaced from `plus` to maybe `square_pencil` - since there are more functions than "add widget" in that menu.
- The upper right icon to call the popover could maybe be `gear_alt` since it's not a list-looking menu anymore.

ghys:
Another thing, I checked with `npm run webpack-analyzer` and this vue-grid-layout library is kind of on the heavier side. I'd like to keep the entry point (`app.js` & `app.css`) leaner, and strive extract the biggest dependencies to separate bundles when possible, unless the library is strictly necessary to warrant being in the entry point, or is small enough not to cause much overhead.

I wouldn't include vue-grid-layout in either category (size is more than a few kb and not everyone will use these layouts), so it would be good if you could make it so it's only loaded dynamically when needed, see https://github.com/openhab/openhab-webui/pull/814#issuecomment-761849886:

image

hubsif:
Linting cleaned up

> The fullscreen link on the bottom toolbar should be blue IMHO.

done.

> Also not your fault, but the icon looks kind of inverted 😞 

Yeah, they're mirrored. I also didn't like this and was very much surprised that there are no "correct" arrows in the f7 library.
I adjusted it to the rectangle, which is ok. Though I think I'd prefer the two non-mirrored icons I found in material (open_in_full and close_fullscreen), because they are less prominent.

> The red background on the grid layout in edit mode looks a bit odd, would you like to change it to a subtle shade of gray instead?

done.

> The FAB when in full screen mode should also be blue, like on other pages, and the icon replaced from plus to maybe square_pencil - since there are more functions than "add widget" in that menu.

I experimented a little and in the end did this: I adjusted the "Configure Layout" icon to square_pencil, to be consistent with the Configure for each widget. I didn't like the gear_alt on the fab too much, so I went for the menu icon instead 😬. But in the end it's your decision.

> The upper right icon to call the popover could maybe be gear_alt since it's not a list-looking menu anymore.

Done, though I somewhat like the menu icon better.

> so it would be good if you could make it so it's only loaded dynamically when needed

Never done this before. Please verify if what I did is sufficient.

<br />

Additionally, I fixed the Gauge CSS so it now behaves like images (instead of overflowing in height it now shrinks). I also added some modifications for the swiper, though that's not perfect yet.
I found that one of the issues is actually also present in the responsive layout: If a column is too small, the swiper menu gets cut like this:
cut_swiper_menu
Perhaps all the menus should be replaced by popovers 😉.

What I didn't take care of yet, is theming, meaning to show different icons for different themes. I've seen that you do that at various places.

ghys (approved):
> Never done this before. Please verify if what I did is sufficient.

There was an import in the `components/widgets/layout.index.js` file left, so the library was still part of the entry point.

> Perhaps all the menus should be replaced by popovers 😉.

Maybe, although the menus are kind of part of the design language throughout these screens. I also fixed a problem that the popover wouldn't be destroyed properly when you removed the widget from it, so it would remain invisible in front of other elements, preventing them from being clicked. Closing it manually with no animation seems to do the trick.

I also made some finishing touches.

Anyways, enough with the nitpicking, this is probably ready to be included at least as a "beta"/experimental feature.

Thanks a lot!

hubsif:
Thank your for reviewing and merging!
I will keep an eye on upcoming issues here in github and see if I can help solve them.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-grid-layout.vue
```diff
@@ -162,8 +159,8 @@ export default {
           for (x = 0; x < this.colNum; x++) {
             free = true
             for (let i = 0; i < this.layout.length; i++) {
-              if (!((this.layout[i].x + this.layout[i].w <= x) || (this.layout[i].x >= x + 1) ||
-                    (this.layout[i].y + this.layout[i].h <= y) || (this.layout[i].y >= y + 1))) {
+              if (!((this.layout[i].x + this.layout[i].w <= x) || (this.layout[i].x >= x + 2) ||
+                    (this.layout[i].y + this.layout[i].h <= y) || (this.layout[i].y >= y + 2))) {
```
hubsif, line 163: Hi @ghys, 
Unfortunately, it's not that simple to get 2x2 widgets: This adds a 2-width element to a column when there's only a 1-width spot left, same for rows. As a result widgets overlap or exceed the defined screen.

ghys: Heh- shouldn't be too hard to fix. The 1x1 defaults widgets were too small, I thought.
Also the documentation should be generated (and ideally manually completed) for the new components, and the layout page documentation page should mention the new features.

hubsif: > Heh- shouldn't be too hard to fix. The 1x1 defaults widgets were too small, I thought.

Surely possible, though one issue always remains: What if there only remains a 1x1 spot or 1x3 row somewhere in the middle of the layout and the user wants to fill that (e.g. on a 3x5 (button) grid created for a mobile phone)? The logic would then have to detect that there's only space for 1x1 widgets left and add such instead.

I aggree that a 1x1 widget can be quite small. Another way to fix this would be to simply default to less columns, e.g. 8 or 12 instead of currently 16.

For now I'd ask you to revert your change, as it results in this:
add_widgets_issue

hubsif:
> Also the documentation should be generated (and ideally manually completed) for the new components, and the layout page documentation page should mention the new features.

Ok, will see when I can find some time to check out how the docs are organized and will try to write up something. Though, I'm not native English, don't know how that's handled - if there are maybe native reviewers or if it just doesn't have to be perfect.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-grid-layout.vue
ghys: > Surely possible, though one issue always remains: What if there only remains a 1x1 spot or 1x3 row somewhere in the middle of the layout and the user wants to fill that (e.g. on a 3x5 (button) grid created for a mobile phone)? The logic would then have to detect that there's only space for 1x1 widgets left and add such instead.

Exactly, this doesn't seem that complex to implement.

> I aggree that a 1x1 widget can be quite small. Another way to fix this would be to simply default to less columns, e.g. 8 or 12 instead of currently 16.

Preferably not since there might be cases where the a widget thinner (or smaller) than the default could be desired.

ghys:
> Ok, will see when I can find some time to check out how the docs are organized and will try to write up something. Though, I'm not native English, don't know how that's handled - if there are maybe native reviewers or if it just doesn't have to be perfect.

The `node generate.js` has to be executed in https://github.com/openhab/openhab-webui/tree/main/bundles/org.openhab.ui/doc/components/src (but I'll do it along with the refactoring above). The layout page article is here: https://github.com/openhab/openhab-docs/blob/main/ui/layout-pages.md

hubsif:
Hi @ghys,

Don't know where to put this, so I just add it here, since it was initiated by this.

> > Perhaps all the menus should be replaced by popovers 😉.

> Maybe, although the menus are kind of part of the design language throughout these screens.

Actually, I started experimenting with this before you merged the pull request. I agree that designs should be somewhat consistent, so it again took me further than expected (though still only for layouts).
Here are the results, perhaps you like parts of it or it gives some ideas. I just didn't want to discard it unshared:

Popover buttons, icons only:
layouts_new

Popover buttons, with text:
layouts_new_text

Column options:
columns

Widget options:
widgets

Different coloring:
layouts_new_diffcols

jimtng:
Great work! Thanks very much for adding this. I was struggling to achieve the "full screen" layout and this solved it.

One request: I'm trying to make my entire screen to show a photo, however there's still a tiny bit of margin / padding around the edges of the screen. How can I make the little gap / padding disappear? Here's my entire yaml:

```yaml
config:
  layoutType: fixed
  screenWidth: 1920
  screenHeight: 1080
  scale: false
  hideNavbar: true
  hideSidebarIcon: true
  showFullscreenIcon: true
  colNum: 1
  margin: 0
  label: Main Photo Page
  order: ""
  sidebar: true
blocks: []
masonry: null
grid:
  - component: oh-grid-item
    config:
      x: 0
      y: 0
      h: 1
      w: 1
    slots:
      default:
        - component: widget:widget_photoframe
          config:
            photos: PhotoFrame_Data
```

hubsif:
Hi @jimtng,
You've found a bug that zero margins wouldn't apply! Just fixed it in #973. Thank you!

Issue #760, opened 2021-01-05
Clickable link to item and to channel from channel-item link view.


cinadr:
## The problem

There is a view where one can observe the relation of a channel and linekd items:
![chrome_NvqcxJLfpf](https://user-images.githubusercontent.com/16361576/103642152-09c94f80-4f53-11eb-9f7e-53af21753a57.png)
None of the items in the list are clickable.

## Your suggestion

It would be very useful to have either the channel or all the linked items to be clickable to make some changes or view additional details.
Furthermore there are some buttons on other pages where items can be _unlinked_ or _unlinked and removed_ with one click. This would be useful here as well. And on this page there is no furher link to item as well.
![chrome_WDnMEehlAm](https://user-images.githubusercontent.com/16361576/103643189-8ad51680-4f54-11eb-81e7-50ed8eb55969.png)

PR #841, opened 2021-01-24
Makes the item clickable in the Channel/Item link view

bigbasec:
Makes the item in the channel/link page clickable and go to the item in question.   

Resolves issue : https://github.com/openhab/openhab-webui/issues/760


ghys:
Thanks, my main concern is that it creates a flow with more than 2 pages when going to thing > link details > item details or model > link details > item details.
Only the previous page is kept hidden behind the current one, so with these flows when going to the item details the original page will get destroyed, and therefore loses its state when coming back. That was the main reason why there were no links on these pages.
But let's revisit this and try it anyway.

ghys (approved):

bigbasec:
@ghys Would just changing it to a href work better?   I know that would break the flow a bit, but it might be a better way to do it.   

I was also considering making a popup for that and just showing the item details, so that's also on the table too.   I see the need, though not sure the best way to solve it.   This was honestly the least effort for something I didn't know if it was going to fly anyway.

PR #855, opened 2021-01-28
Evaluate `oh-time-series` too so one can bind it in widgets

crnjan:
Please see [this](https://community.openhab.org/t/main-ui-echarts-gauge-and-pie-poc/113987/18) for details.

Also seen `dataSeriesId` sometimes pops up in yaml editor, polluting yaml, saving ids separately is a better idea, so not changing component itself.

Maybe evaluating `markArea` part of `oh-time-series` would make sense too ...


ghys (approved):
LGTM, thanks!

PR #864, opened 2021-01-29
Add generated component reference

ghys:
First version of script
Add System Library, Standard Library


ghys:
I'll merge the autogenerated stuff which seems to work and try to have them on the website early, and then we can add more documentation in subsequent PRs.

The part between comments like `<!-- GENERATED ... -->` and `<!-- GENERATED /... -->` in the articles should not be touched as it is subject to being overwritten when the pages are regenerated.

However what's outside those sections will be preserved.

What I'd like to see in other manually written sections is:

## Description

For some components, a more detailed description than the provided generated description is desirable - it can even replace the generated section which is usually only a one-liner.

## Inherited Configuration

_(below Configuration)_
Indication, when applicable, that the configuration will be passed to an underlying component (from Framework7, ECharts, Leaflet...) and either provide a link to their documentation (https://v5.framework7.io, https://echarts.apache.org/en/option.html, https://vue2-leaflet.netlify.app/components/, etc.) or eventually replicate it in the same format as the autogenerated config.

## Examples

_(last section)_
Complete examples of (simple) widgets illustrating how the component can be used, or one of its features - should have a title, a YAML code snippet (ready to parse in Developer Tools > Widgets) and a screenshot/GIF.
Some of these examples can be sourced from the forum.

PR #868, opened 2021-01-30
[iconset] Fixed pantry.svg to avoid iOS issues

cweitkamp:
- Fixed pantry.svg to avoid iOS issues

@timbms I changed the icon according to your replacement version. Are you sure it will work? Because I saw a second ref in line 8 and 9 on an object defined in line 10. Shouldn't those be switched too?

Closes https://github.com/openhab/openhab-core/issues/2113


ghys (approved):
Thanks for the investigation!

timbms:
> @timbms I changed the icon according to your replacement version. Are you sure it will work? Because I saw a second ref in line 8 and 9 on an object defined in line 10. Shouldn't those be switched too?
I am confident that the proposed fix works on iOS app that makes use of SVGKit as there is unfortunately no native SVG rendering on iOS, at least none that is easily accessible.

PR #867, opened 2021-01-30
Add control-s to settings page

bigbasec:
Very simply adds the control-S to the page to make it more in line with some of the other pages.

PR #883, opened 2021-02-06
Fixes ctrl+save on settings pages

crnjan:
When navigating to settings page, i.e. Basic UI, the following is logged:

```
vue.esm.js?a026:628 [Vue warn]: Property or method "onPageAfterIn" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.
```


ghys (approved):
Wow, I'm the one to blame for reviewing #867 a little too fast...
Thanks!

PR #889, opened 2021-02-07
Activation of additional ESLint rules

philippwaller:
This PR activates the following ESLint rules and provides the necessary code changes:
- indent
- no-trailing-spaces
- quotes
- vue/html-indent
- vue/html-quotes
- vue/no-multi-spaces


ghys (approved):
Awesome, thanks!

philippwaller:
@ghys: Thanks for the review. The next batch is on its way. I guess we need to rebuild the pending PR's to identify ESLint errors.

ghys:
You're right. In the end I think it would be a good thing to make Prettier run as well (possibly on save in dev mode too).

philippwaller:
> I think it would be a good thing to make Prettier run as well (possibly on save in dev mode too).

Sounds good. I will take care of it after the next ESLint PR is merged.

PR #890, opened 2021-02-07
Allow customisation of location-card badges.

crnjan:
Was using OH 3 before https://github.com/openhab/openhab-webui/pull/771 was merged - at that time blinds badges were displayed on location card(s) when rollershutters were closed (in my case position > 0) - and after fix was merged, really still cannot get used seeing badge when blinds are open (up) :)

So instead of arguing what is better, let's make it configurable so each can decide what suits best.

For example - Living Room, 3 blinds, 2 closed - default configuration:

If we want to see blind's badge when `position > 0` we can write

```
config:
  label: Home Page
locations:
  - component: oh-locations-tab
    config:
      cardOrder:
       ...
      excludedCards: []
    slots:
      gGF_Living:
        - component: oh-location-card
          config:
            badges:
              blinds:
                expression: Number.parseInt(state) > 0          
```

result

We can override icon too

```
        - component: oh-location-card
          config:
            badges:
              blinds:
                expression: Number.parseInt(state) > 0       
                badge:
                  icon: f7:logo_windows
```

Badges can be overridden per card (as above example) or for all cards

```
config:
  label: Home Page
locations:
  - component: oh-locations-tab
    config:
      badges:
        blinds:
          badge:
            icon: f7:logo_windows
          expression: Number.parseInt(state) > 0
```

If both are specified, per-card definition takes precedence.

All badges can be overridden, i.e. both lights&blinds

```
config:
  label: Home Page
locations:
  - component: oh-locations-tab
    config:
      badges:
        blinds:
          expression: Number.parseInt(state) > 0
        lights:
          badge:
            icon: f7:sun_max
```

result

I look forward to any feedback or comments.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/location/measurement-badge.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/location/measurement-badge.vue
@@ -73,14 +73,14 @@ export default {
       }
     },
     map () {
-      return this.query.map((item) => this.store[item.name].state).filter((state) => Number.isFinite(Number.parseFloat(state.split(' ')[0])))
+      return this.query.map((item) => this.store[item.name].state).filter((state) => Number.isFinite(Number.parseFloat(state)))
     },
     mapAux () {
-      return this.queryAux.map((item) => this.store[item.name].state).filter((state) => Number.isFinite(Number.parseFloat(state.split(' ')[0])))
+      return this.queryAux.map((item) => this.store[item.name].state).filter((state) => Number.isFinite(Number.parseFloat(state)))
     },
     reduce () {
       const ret = this.map.reduce((avg, state, arr, { length }) => {
-        const value = Number.parseFloat(state.split(' ')[0])
+        const value = Number.parseFloat(state)
         if (Number.isFinite(value)) {
           return avg + value / length
         }
@@ -92,7 +92,7 @@ export default {
     reduceAux () {
       if (this.type !== 'temperature') return undefined
       const ret = this.mapAux.reduce((avg, state, arr, { length }) => {
-        const value = Number.parseFloat(state.split(' ')[0])
+        const value = Number.parseFloat(state)
         if (Number.isFinite(value)) {
           return avg + value / length
         }
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
@@ -34,9 +34,10 @@
 
 <script>
 import { findEquipment, allEquipmentPoints, findPoints } from '../glance-helpers'
+import expr from 'expression-eval'
 
 export default {
-  props: ['element', 'type', 'customConfig', 'invertColor', 'store'],
+  props: ['element', 'type', 'badgeOverrides', 'invertColor', 'store'],
   data () {
     return {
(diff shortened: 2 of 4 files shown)


eikowagenknecht:
Just wanted to say I really like this idea. Having all "open" Rollershutters shown makes no sense for me either, but for other types of Blinds I like it, so being able to configure it would indeed be nice.

ghys:
Thanks @crnjan and sorry for the delay.
First about #771 - it seems it was indeed a regression and feels less intuitive for most, so it might be wise to revert it altogether - and adjust the documentation to make it clear that for blinds/rollershutters the icons is shown for those which are _closed_, whereas for doors/windows it's shown for those which are _open_, as this also might be considered less consistent - but as long as it's documented, it's fine.

Part of the reason I sat on this PR is that while we're at it, it could be good to go all the way and add some parameters to also specify not only the "reduce" part but also the "map" part: arrays of equipment classes to consider, with or without subclasses, possibly other criteria too.
This would unlock the ability to maybe offer the ability to add custom badges, not only the hardcoded list of those which already exist - this could be in a `badges` slot of the `oh-location-card`?

I'd be tempted to approve this PR as-is, as it looks good, to be honest, that being said, if we have to change the API afterwards again it's probably best to figure out what we aim to do before introducing new options.
Wdyt?

bundles/org.openhab.ui/web/src/components/cards/glance/location/measurement-badge.vue
```diff
@@ -73,14 +73,14 @@ export default {
       }
     },
     map () {
-      return this.query.map((item) => this.store[item.name].state).filter((state) => Number.isFinite(Number.parseFloat(state.split(' ')[0])))
+      return this.query.map((item) => this.store[item.name].state).filter((state) => Number.isFinite(Number.parseFloat(state)))
```
ghys, line 76: Can you explain why these `.split(' ')|0]` were removed? They seemed to be useful when you had math operations to perform on quantity types (represented as strings with the unit in the API, you you can't perform e.g. averages with them).

crnjan: ```If parseFloat encounters a character other than a plus sign (+), minus sign (- U+002D HYPHEN-MINUS), numeral (0–9), decimal point (.), or exponent (e or E), it returns the value up to that character, ignoring the invalid character and characters following it.```

So IMHO it's not needed ... from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat

```
The following examples all return 3.14:

parseFloat(3.14);
parseFloat('3.14');
parseFloat('  3.14  ');
parseFloat('314e-2');
parseFloat('0.0314E+2');
parseFloat('3.14some non-digit characters');
```

crnjan:
Sorry for delayed responses, I hope I'll find some spare time over the weekend to make some progress, if not before ...

ghys (approved):
Sorry Boris for the long delay. I think your changes are valuable and if we're ever to change the API, as long as we agree on them for the long-term and revisit them, eventually, they're okay to include for now.

cweitkamp:
@crnjan Nice work. May I ask you to add this to the [documentation](https://next.openhab.org/docs/ui/components/oh-list-item.html#badge). That would be really much appreciated. Thanks in advance.

PR #900, opened 2021-02-09
Component docs

AndyXMB:
Hi @ghys,

I thought I would try to 'do may part' and help with the component documentation.  This pull request is a means to start the conversation on the layout / headings etc.  None of this is ready for a merge, but I thought it sensible to share what I've done before I go too far.

I took the liberty of trying out a table layout for the automatically generated sections.  I was hoping that this would result in a clearer layout, taking the inspiration from the way the Framework 7 documentation is laid out.  If I'm honest, it's not as great as I hoped (due to the limitations of markdown tables).  I think we could do better with HTML table and styles, but I'm not sure if this is allowed (maybe it would be okay as this is script generated)??  **Happy to revert to current layout if you think that's better.**

I have added bookmark link generation in the script on each of the properties and property options.  The idea was to allow you to click on the property/option and go to a relevant example (if one exists). Just used `<div id="[property]">` or `<div id="[property]-[option]">` as the targets.  I think this is an idea worth adding.  

[Try it out](https://github.com/AndyXMB/openhab-webui/blob/component-docs/bundles/org.openhab.ui/doc/components/oh-button.md).  The links for any of the general properties and `variable` option are functional.

Created `img` folder for image storage and `oh-button` sub-directory. Propose generate sub-directory for each page to simplify organisation.  Common images in parent. 

Used summary / details sections for the example code.  Could see the page getting very long and cluttered otherwise...

Think we should add a table of contents after the header image. Not sure how this is done in the OH environment (I have tools locally that can generate).

- Any headings missing ?
- Is the **Usage / reference documents** section what you were anticipating (also technically correct?) ?
- Header image what you were expecting ?
- Anything else

Let me know what you think.

Known issues:

- Script needs a change so that an object with zero properties that are not in a group, does not render table headings.


ghys:
Hi @AndyXMB,

> I thought I would try to 'do may part' and help with the component documentation. 

And I'm thrilled!

> I took the liberty of trying out a table layout for the automatically generated sections. I was hoping that this would result in a clearer layout, taking the inspiration from the way the Framework 7 documentation is laid out. If I'm honest, it's not as great as I hoped (due to the limitations of markdown tables).

I also considered a table at first but finally settled on simple lists because 1) the tables look a bit "unbalanced", you'll have the description column contents probably wrap several lines too often because of width constraints, 2) I tend to avoid big tables in general because on smaller mobile screens they will not fit and you'll have to constantly scroll horizontally.
It's also true that problems could maybe be solved with a bit of styling. We can certainly add CSS classes to the website for these particular cases.

> I think we could do better with HTML table and styles, but I'm not sure if this is allowed (maybe it would be okay as this is script generated)

We can indeed make it work. It's true also for the list item view (with `<span class="...">`). I kind of like the equivalent in Home Assistant for example https://www.home-assistant.io/lovelace/entity/#configuration-variables. I think it's the best compromise.

> I have added bookmark link generation in the script on each of the properties and property options.

Great idea!

> Created img folder

Can you make it `images`? This is more consistent with what's found in other places in the docs and I was ahead of you 😄:
https://github.com/openhab/website/blob/57955fee2bb4db17df661ef0640811e5257a2b4f/prepare-docs.rb#L396

A folder per component is fine and makes sense (and should work with the above).

> Used summary / details sections for the example code. Could see the page getting very long and cluttered otherwise...

Fine, I wouldn't have done so below a certain threshold, maybe anything under ~50-100ish lines is fine to display by default - one can wonder where is the code is for a second before you see it's collapsed. For more involved examples though it's a good idea.

> Think we should add a table of contents after the header image. 

`[[toc]]`: https://vuepress.vuejs.org/guide/markdown.html#table-of-contents
Note you also have other interesting things too, like
https://vuepress.vuejs.org/guide/markdown.html#custom-containers (including a collapsible one for the above)
and
https://vuepress.vuejs.org/guide/markdown.html#line-highlighting-in-code-blocks which I can see being _very_ useful to emphasize what the example tries to show.
There are a lot of things available with plugins too, some are available with by default like https://vuepress.vuejs.org/theme/default-theme-config.html#code-groups-and-code-blocks


The content is really what I had in mind, a few minor remarks:
- Links to Framework7 docs should be made to https://v5.framework7.io instead of the main website now (migrating to version 6 is no easy task since it involves upgrading to Vue 3 and changing a lot of stuff, so openHAB will stick to v5 for the foreseeable future). References in the components descriptions (in the code) should be updated.
- A link to https://openhab.org/docs/ui/building-pages.html#css-variables when mentioning styling would be nice
- Instead of "The standard f7-button capabilities are enhanced by the addition of specific actions to provide user interface functionality / work with openHAB objects etc." I would have added as a last section in Configuration, just after `<!-- GENERATED /props -->`, something like (perhaps do it as a rule, when applicable):
```
### Inherited Properties

The configuration is passed to the underlying `f7-button` component from Framework7.
Therefore, all compatible scalar [properties](https://v5.framework7.io/vue/button.html#button-properties) (except functions) not listed above are available for use.
```

Thanks a lot for this already! 👍

rkoshak:
I like what I'm seeing. I was nudged by Yannick to add to the discussion here consideration of examples. I've a couple of examples for oh-repeater posted to the forum at https://community.openhab.org/t/a-couple-of-simple-oh-repeater-examples/116818 and a few examples of several oh-list-items posted at https://community.openhab.org/t/example-custom-list-widgets/116265 which might be useful to include in these docs or to link to.

ghys:
@AndyXMB I think we should avoid using HTML to describe the props, I find it a little too" verbose".

I just noticed in the VuePress docs that we can define custom containers with a special syntax, they did it here: https://github.com/vuejs/vuepress/blob/3e1f5d844b16a3f829465a35e734a5f556c3dd22/packages/docs/docs/.vuepress/config.js#L83

So I can define a `props` container like this:

```
['container', {
      type: 'props',
      before: '<div class="component-props">',
      after: '</div>'
    }],
```

And add some styles like you defined directly in the website's CSS so you don't have to define them inline in the pages.

That way you can go back to the old layout with Markdown, just surround the options lists with:
```
::: props

:::
```
This will add the `<div class="component-props">` when building the site and everything below it will be handled with CSS.

Just define the first line including the name, type, label with HTML spans e.g.:

```
- <span class="name">{name}</span> <span class="type">{type}</span> <span class="label">{label}</span>
```
The rest (description, options table) can stay in Markdown like before :)

AndyXMB:
hi @ghys 

100% agree... This was just a 'quick and dirty' mock up of a style layout which might improve the scannability of the pages.  Basically is this a layout even worth pursuing? In essence this is a 'card' per property: 

image

- Header contains `property name` **Label** Description  
- 'Card' body contains data type and options list (flexbox used, but <ul>/<li> + <columns> looks good too)
- Options list uses breakpoints to change the number of 'columns' displayed. Ranges from 1 (mobile) to 5 (big tv). _Breakpoints may need a tweak if there are any very long option labels out there._ 
- Background colors allow you to easily scan for a likely property among the sea of (necessary) data. I'm no graphic designer and there seems to be few colors used in the website theme, so it's a lighter shade of OH orange and menu grey.  Component colors were tweaked where the contrast was insufficient.

To be honest I needed a change of scenery...  I've spent 2 days failing to get the oh-website VuePress build to work on Windows 10 (out of memory errors - 16GB installed).  I think it may be a bug, but I could have screwed something up...  I was hoping that I would not have to build out VM infrastructure to do this work :-( 

Andy

ghys:
This is very good!
I would still switch the data type and description - data type in a little badge in the header after the label, description in the body - since the description can be longer, and allows HTML with line breaks (not that much at the moment but we can plan for it). 

We can also consider definition lists (`dl/dt/dd` tags) for the options! (As long as it's still one line per option in the file.) See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl. I think they would be a good fit here.
Something like:
```
<dl>
  <dt>option1</dt><dd>Option label 1</dd>
  <dt>option2</dt><dd>Option label 2</dd>
  ...
</dl>
```
Then we can style them however we want, with columns like you did or anything else.
The site uses markdown-it to render Markdown so with [this plugin](https://github.com/markdown-it/markdown-it-deflist) we could use the syntax showcased in https://markdown-it.github.io for the definition lists:
```
option1
  ~ Option label 1
option2
  ~ Option label 2
...
```
which would be even more elegant.

Building the website locally can take a very long time if you include all the add-ons, if you don't need them. There's a "secret" trick to exclude them:
https://github.com/openhab/website/blob/57955fee2bb4db17df661ef0640811e5257a2b4f/.vuepress/config.js#L19
Set the `OH_NOADDONS` environment variable with `set OH_NOADDONS=1` before and they will be left out.

AndyXMB:
Hi @ghys 

Just done an update today to implement the custom container method you suggested.  This requires the following changes to config.js:

```
plugins: [
    'tabs',
    [
      '@vuepress/google-analytics',
      {
        'ga': 'UA-47717934-1'
      }
    ],
    ['container', {
      type: 'prop-head',
      before: '<div class="prop">',
      after: '</div>'
    }],
    ['container', {
      type: 'prop-opt',
      before: '<div class="options">',
      after: '</div>'
    }],
  ],
```
I think I've included all the changes you asked for.  I can't work out how styles are being implemented, so maybe someone could take the inline styles and get them into the CSS doc?

The only other substantive changes are:
- added a `General properties / Style` heading for properties not in a group (so these appear in the TOC).
- moved the `Slots` heading to under the `Inherited properties`, makes more sense I think.
- controversial this one... stripped the `<br>` tags from the description, replaced with a space.  I think this displays better all round.  If you want to leave that out, we can drop when we migrate the styles to the proper place.
- reduced the number of columns to max 3 with a bit more padding between.  Decided that columns were easier to implement than the flexbox approach. 

Are we getting closer?   Andy

AndyXMB:
@ghys 

Finally managed to find the solution to my VuePress issues.  Will post some workaround notes for Windows users to the contribute.md when I have a chance.

I've added the following to my local copy of VuePress `openhab-website\.vuepress\styles\index.styl`:

```
/* Component page styling */
.prop > ul 
    background-color rgba(230,72,25,0.3) 
    border-radius 5px 5px 0px 0px
    list-style-type none
    padding 10px
    margin: 10px 0px 0px 0px;
  
.prop > ul > li > a > code 
    background-color: rgba(255,255,255, 1)

.options 
    background-color #eaecef
    border-radius 0px 0px 5px 5px
    display flex
    flex-wrap wrap
    padding-left 5px
    margin-bottom 15px

.options > ul 
    list-style-type none
    padding 0px
    margin 5px
    columns 1
    column-gap 0px

.options > p 
    width 100%
    box-sizing border-box
    margin 5px

.options > p > br
    content ''

.options > p > br:after
    content ' '

.options > ul > li > a > code
    background-color white

/* 410, 620, 900, 1200 */
@media (min-width: 620px) 
    .options > ul 
        columns 1
        column-gap 0px

@media (min-width: 900px)
    .options > ul 
        columns 2
        column-gap 60px

@media (min-width: 1200px) 
    .options > ul
        columns 3
        column-gap 80px
```

I'd still like someone who really understands CSS to check this, as this is way outside my area of expertise.  

Any objection to creating a set of folders to host the code used in the examples?  My own OH instance is getting a bit cluttered with sample code widgets already and I'm only 2/3 through the oh-button control!  Keeping these in git so others can use / edit etc. seems like a good solution?  Was thinking same structure as we used for images with a subfolder for each control.

Long term perhaps we could then use https://v1.vuepress.vuejs.org/guide/markdown.html#import-code-snippets.  It's beta at the moment and I can't get it to work on my local instance. Though perhaps I just don't understand the filename syntax...

It's taking quite a lot of time to do this well, so would be good to know I'm on the right track.  Andy

AndyXMB:
@ghys 

Hi, I was wondering if you'd had a chance to take a look at what I've done so far. Been holding off doing much more until I had a 'green light'. Is this the type of content what you were looking for? Do we need any other changes to the layout etc.  If it's okay, I'll crack on with the other pages... Andy

ghys:
@AndyXMB sorry for falling behind these last few weeks...

I submitted a PR to your repository since I have no permission to push to your branch directly: https://github.com/AndyXMB/openhab-webui/pull/1
I changed my mind (again!) on the props generation - now using Vue components that the website is responsible for rendering, which are implemented in https://github.com/openhab/website/pull/276.
So it finally looks like:
```xml
<PropGroup name="actions" label="Action">
  Action to perform when the element is clicked
<PropBlock type="TEXT" name="action" label="Action">
  <PropDescription>
    Type of action to perform
  </PropDescription>
  <PropOptions>
    <PropOption value="navigate" label="Navigate to page" />
    <PropOption value="command" label="Send command" />
    <PropOption value="toggle" label="Toggle item" />
    <PropOption value="options" label="Command options" />
    <PropOption value="rule" label="Run rule" />
    <PropOption value="popup" label="Open popup" />
    <PropOption value="popover" label="Open popover" />
    <PropOption value="sheet" label="Open sheet" />
    <PropOption value="photos" label="Open photo browser" />
    <PropOption value="group" label="Group details" />
    <PropOption value="analyzer" label="Analyze item(s)" />
    <PropOption value="url" label="External URL" />
    <PropOption value="variable" label="Set Variable" />
  </PropOptions>
</PropBlock>
<PropBlock type="TEXT" name="actionUrl" label="Action URL" context="url">
  <PropDescription>
    URL to navigate to
  </PropDescription>
</PropBlock>
...
```
And looks like this (I altered the styling a little bit in the process to remove some colors and make it dark mode compatible):
image

But it doesn't define any styling, the components are responsible for the rendering, and it may even be changed afterwards. In the end I believe it's the right way to go. For instance required properties are displayed with a red asterisk, etc.

Sorry I misread you on the bookmarks, I do not think making links out of the prop IDs and option values to jump around the page to examples is the right way to go, it's not a bad idea per se, but should be implemented differently, for instance with the "prop block" reading frontmatter metadata on the page and generating appropriate links below the description/options.
So I removed them from the examples and we can figure it out in a follow-up PR.
On the other hand, props are now bookmarks themselves (see the # appearing when hovering over it above), it's important to have so we can deep link to a particular prop.

I have made minor changes to the 2 pages you wrote, which are great by the way, to better respect the structure that is now in the template - so the pages that haven't been edited yet have the ideal structure in comments that the first editor can read and eventually overwrite.  @openhab/documentation-maintainers have a rule to only put one sentence per line in the docs, so I applied it here as well. The updated versions of unaltered component pages is also included in https://github.com/AndyXMB/openhab-webui/pull/1.
Possibly your extensive examples of actions (in `oh-button`) could be in time moved to their dedicated page since they apply to multiple controls.

I think this can now be merged, especially since there is a dependency to https://github.com/openhab/website/pull/276 and maybe more involved to make it all work. I'll just need you to do two things:
1) merge https://github.com/AndyXMB/openhab-webui/pull/1
2) comment with a sign-off line like https://openhab.org/docs/developer/contributing.html#sign-your-work. Ignore the fact that it should've been in your commit messages, I'll include it in the final commit.

Once this is merged then you're more than welcome to make additional PRs with more content :)

Thanks for your hard work!

ghys:
> Long term perhaps we could then use https://v1.vuepress.vuejs.org/guide/markdown.html#import-code-snippets. It's beta at the moment and I can't get it to work on my local instance. Though perhaps I just don't understand the filename syntax...

We'll see, but as this is a VuePress feature let's not for the moment. It's hard to know if it'll work well when authoring the page in GitHub or VS Code or another generic Markdown editor, so for now let's embed the examples inline.

ghys:
@AndyXMB thanks for the merge, I resolved the conflicts, still waiting for a sign-off line then we can merge!

> 2. comment with a sign-off line like https://openhab.org/docs/developer/contributing.html#sign-your-work. Ignore the fact that it should've been in your commit messages, I'll include it in the final commit.

AndyXMB:
@ghys I have to be honest, I'm disappointed that we have lost everything but the use of columns.  Your changes add two additional concerns:

- The horizontal divider, introduced by adding the heading 4 to the property tag, actually acts to visually separate the related information.  This makes the page harder to read.  The rationale of adding the background was to unify the related data.  I would add this back, the horizontal separator will likely have less visual impact / act as intended to divide the properties from the description.
- The use of definition tags, doubles the vertical screen real estate used and interrupts the left to right reading flow (scan).  I'd revert this to a horizontal list unless there is some specific advantage to using definition tags that I don't understand.
- You should add styling to remove the `<br>` added to descriptions.  The instances where these exist not only consume more vertical space but also looks inconsistent when compared with the other descriptions that conventionally flow as sentences.  

I thought the bookmark feature worked well.  However I see that your addition of the <h4> property duplicated / broke the original property links. 

One of the key advantages of the bookmarks was that if an example to demonstrate this feature was not yet been done, then a click cause no real negative user experience i.e. nothing happened. You could also link multiple properties and/or options to the same example.  The last thing I would want to see is a load of `click here for example` hyperlinks - after all you just removed the same on the code blocks!  

Are you sure we are not over engineering this? Is this a simple as agreeing a standard for the bookmark labels to avoid the naming conflict?  I already used `#action-navigate` just in case there ever existed two properties with the same option name.  So we could simply add a prefix e.g. `#ex-action` for the property, `#ex-action-navigate` etc.?

Anyway I've signed off the code, I think.  I don't use Git very much, so please bear with me if I've got this wrong.  The instructions don't specify if the sign off needs to be in the summary or description, so I guessed description.

ghys:
@AndyXMB thanks for the honest feedback, just want to stress this is merely styling so it can be changed at will on the website code now, and the intention was not to undo anything you've done (the first iterations were closer to yours), nor discourage you from contributing more good examples!

You probably have a point about making things more compact to conserve vertical space, and the need for "scannability" (is that a word?).
The dl/dt tags for the options were introduced because they're probably the more semantically correct way of representing them.
According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl: 
> Common uses for this element are to implement a glossary or to display metadata (a list of key-value pairs).

Initially I had them on a single line like you did but then I felt there were also drawbacks of having them so - I felt it was cleaner to have the labels aligned, and when the label is too long (see `oh-category-axis` for example) it wraps to a second line, in a rather ugly way. So in a last minute decision I reverted to the default look. but I wouldn't mind reverting that to make it more compact.

Sorry but I also remain unconvinced by the "click on the property name to maybe or maybe not go to an example" feature. The intent to link to examples is clear to me and you because we discussed it, but maybe it would be a bit counter-intuitive to the casual reader - I haven't seen anything like this done anywhere else - I think it should be more explicit where these clicks link to. It could even be the other way around: in the example sections, explicitly mention: "related properties: property1, property2" with links to the reference above. - maybe even "related components: component1, component2" with links to their pages.

Anyways thank you, again just know nothing is set in stone wrt. styling, but the structure of the documents is clearer now.

AndyXMB:
@ghys Sorry, no time to work on anything this week.  PC power supply failed and only managed to strip it apart to confirm voltages today. Should be back to normal sometime next week.

PR #913, opened 2021-02-17
make trendline always show latest value (newest datapoint 60 minutes old otherwise)

tknaller:
make trendline always show latest value (newest datapoint 60 minutes old otherwise)


ghys (approved):
LGTM, thanks!

PR #917, opened 2021-02-19
Add missing link to f7 input docs

eikowagenknecht:
The link was missing. Not much more to say about this one.

Also aligned the style of those two links that are just next to each other to look the same.

This is what it looks like:

image

This is what it links to: https://framework7.io/docs/inputs.html#supported-inputs


ghys (approved):
Thanks!
Note that the framework7 docs links should be updated to https://v5.framework7.io but for the time being I'll allow it for consistency's sake, and because the API is AFAIK still the same.
There should/will probably be a PR updating all these links in bulk at some point.

PR #920, opened 2021-02-21
Show ThingUID instead of ThingTypeUID for discovery result

kaikreuzer:
See https://github.com/openhab/openhab-addons/pull/10154#issuecomment-779422785

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #921, opened 2021-02-22
Show ThingUID instead of ThingTypeUID for scan result

kaikreuzer:
Second (missing) part to https://github.com/openhab/openhab-webui/pull/920


ghys (approved):
Good catch!

Issue #919, opened 2021-02-19
Last item in list card can not be edited (CSS issue)


eikowagenknecht:
## The problem

With a list, the last item can not be edited:

This is not clickable:

![image](https://user-images.githubusercontent.com/1475672/108525778-c2174100-72d0-11eb-9ce7-bbdcf3166f67.png)

## Expected behavior

Should be clickable.

## Steps to reproduce

1. Create List
2. Make sure there is a card below
3. Try to edit the last entry

## Your environment
OH3.0.1

## Additional information

z-index CSS issue:
![image](https://user-images.githubusercontent.com/1475672/108526092-0acefa00-72d1-11eb-8f88-4ec6b7b32099.png)

The gray divider is obviously above the menu which makes it non-clickable.

Solution: Lower z-index of divider or raise z-index of menu. Not sure why this is 1000, there might be a reason for this.

PR #943, opened 2021-03-08
adjust configuration menu for list items, fixes #892

### LINKED ISSUE #892: oh-list-item not usable in widgets (visual editing not possible, empty output)
## The problem
**When using an oh-list-item as the main component in a custom widget, I cannot edit the widget properties anymore once the widget has been added to a list on a page.** When clicking "Configure Item", you can only edit the config of the underlying oh-list-item, but not the widget properties. The list item also often disappears from the page after saving and can only be found in YAML code. This is is a huge limitation for designing resusable list item (which are a main building block of UIs). 
More info can be found in the [Forum](https://community.openhab.org/t/example-custom-list-widgets/116265/18).

Widget Properties:
image

Properties when putting the widget on a page:
image

## Steps to reproduce

Create custom widget with oh-list-item as main component, such as [this](https://community.openhab.org/t/example-custom-list-widgets/116265) (or any other). Create a page, add row, column and list card and add the custom widget as list item. Try to edit the widget propoerties.

## Your environment

OH3

hubsif:
This is to fix the custom list item configuration of issue #892. 
To achieve this, I moved the level at which a list item menu is created to the parent component (oh-list), like it also is for all other components.

The bug report actually contains another issue, I think, about list items not being display in the widget editor. I'm going to open an extra github issue for that.


ghys (approved):
We lost the vertical alignment when the item's height is bigger than the button, but it's a small price to pay ;)

image

I added some last minute fixes to align the item buttons with the "+" button, and fix #919 in the process.
The reason these menus are behind other elements is because the `.list` CSS class defines `z-index: 1` - if you have a z-index anywhere in the tree it will behave like that. So I removed it but in edit mode only, in run mode it's probably best to keep things as they are as much as possible.

Build is green again, so let's merge.

Issue #678, opened 2020-12-20
Widget preview in metadata editor is not updated properly


XeS0r:
## The problem
If you have an equipment which is controllable by a slider widget you can add a default Standalone widget for this equipment. If you're in the editor you can see the selected slider preview on top of the settings. Unfortunately scale settings are not applied in the preview.

## Expected behavior
Changed settings should be visible in the preview.

## Steps to reproduce

1. Add Metadata -> Default Standalone Widet
2. Select Slider type
3. Choose display scale
4. Look at the preview
5. Nothing changes
6. Apply changes
7. Look at the equipment and now the scale is visible.

1. 

## Your environment

```yaml
Konfigurationsverzeichnis
/etc/openhab
Verzeichnis für Anwenderdaten
/var/lib/openhab
Verzeichnis für Protokolldateien
/var/log/openhab
Betriebssystem
Linux/5.4.0-1025-raspi (arm)
Java Runtime
Ubuntu
11.0.9.1 (undefined)
```

## Browser console

```txt

```

Issue #553, opened 2020-11-23
Deleting values when configuring grid for chart renders differently from initial setting


jpharvey:
## The problem
Changing a value when configuring the grid of a chart such as left or right affects the chart as expected. Then removing the value by pressing the x changes the way the chart is drawn to the same as 0 being entered instead of the initial default value.
The yaml in this case contains left: null etc. which causes the chart to be draw differently than if the value is missing.

## Expected behavior
THe chara

## Steps to reproduce

1. 

## Your environment

```yaml
component: oh-chart-grid
config:
  show: true
  width: 1400px
  height: 400px
  left: 480px
  right: null
slots: null

```

## Browser console

```txt

```

## Additional information
This is the second case of something similar i have reported. Would it be better for the x to remove the complete yaml entry instead of setting it to null and all rendering having to cope with null as well as no value?

ghys:
> This is the second case of something similar i have reported. Would it be better for the x to remove the complete yaml entry instead of setting it to null and all rendering having to cope with null as well as no value?

Definitely.
There's a reason why it's like this: https://github.com/openhab/openhab-webui/blob/0ebb8a4586686e0fe038fdcdcbe9c79eb21945b9/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue#L114 but obviously it has unfortunate side effects. Some API calls (e.g. `PUT /services/{serviceName}/config` will not delete an existing parameter if it's not present in the payload. More explanation here: https://github.com/openhab/openhab-webui/pull/339#discussion_r489632719
These cases should probably be identified and defaulting to null added as an opt-in option to the config sheet component for them.

PR #945, opened 2021-03-09
Rework itemMetadataWidget

hubsif:
Hi Yannick,

With this commit I addressed issue #678. I did quite a lot more than I had to, so here some notes:

I completely removed following code:
```js
        // set to '' when the default defines the option but the metadata doesn't (null would be better but the API then removes it)
        if (!this.metadata.config[key] && this.defaultComponent.config[key]) this.$set(this.metadata.config, key, '')
```

I tried real hard to find out what you added this for, there surely must have been a reason - but couldn't.
If by `!this.metadata.config[key]` you meant "if that key does not exist": that doesn't make sense when iterating through the keys. If you meant "if the config value is falsy and the default truthy": I can't see how it makes sense to set the value to an empty string (or from the comment the originally desired 'null'). Perhaps you remember what it's for, if it's not just some weird leftover.

With this removed, values like the slider scale, which defaults to true, can be set to false again. Also, values now can have a `0`, e.g. for the scaleSubsteps value.

<br />

The second major change is to make the preview widget dynamically update.

<br />

I then wanted to address another issue, which is that default values get reset to default when you empty its input field. Try e.g. to set another value for "Highlight Color" for cells. It defaults to `blue` and whenever you delete the last character (`b`), the value gets set to `blue` again.

At first I thought that comes from the many computed values. Later on I realized that this issue seems to lie in configSheet instead.
But while analyzing this, I discovered that with each keypress the many code for computed values is executed. As this seemed unnecesary, I reworked it - actually quite radically, I hope you like it.

Additionally, I streamlined the standard widget arrays and the ordered standard widget arrays into one. `personalWidgets` still gets computed, since like this it's updated automatically when loaded.

And I removed the `viewMode`, since it actually never changes. The `componentSelectKey` also seemed redundant.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-widget.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-widget.vue
@@ -2,44 +2,43 @@
   <div>
     <f7-block class="block-narrow widget-preview">
       <f7-col>
-        <generic-widget-component v-if="previewContext.component" :context="previewContext" />
+        <generic-widget-component v-if="previewContext.component" :context="previewContext" :key="previewWidgetKey" />
       </f7-col>
     </f7-block>
 
-    <f7-list v-if="viewMode === 'design' && defaultComponent.component">
-      <f7-list-item :key="componentSelectKey"
-                    :title="'Widget'" smart-select :smart-select-params="{ openIn: 'popup', searchbar: true, closeOnSelect: true, scrollToSelectedItem: true }" ref="widgets">
+    <f7-list v-if="defaultComponent.component">
+      <f7-list-item :title="'Widget'" smart-select :smart-select-params="{ openIn: 'popup', searchbar: true, closeOnSelect: true, scrollToSelectedItem: true }" ref="widgets">
         <select name="widgets" @change="updateComponent">
           <option value="">
             Default ({{ defaultComponent.component }})
           </option>
           <optgroup label="Standard Library (List)" v-if="namespace === 'listWidget'">
-            <option v-for="widget in orderedStandardListWidgets" :key="widget.name" :value="widget.name" :selected="metadata.value === widget.name">
+            <option v-for="widget in standardListWidgets" :key="widget.name" :value="widget.name" :selected="metadata.value === widget.name">
               {{ widget.label }}
             </option>
           </optgroup>
           <optgroup label="Standard Library (Cell)" v-else-if="namespace === 'cellWidget'">
-            <option v-for="widget in orderedStandardCellWidgets" :key="widget.name" :value="widget.name" :selected="metadata.value === widget.name">
+            <option v-for="widget in standardCellWidgets" :key="widget.name" :value="widget.name" :selected="metadata.value === widget.name">
               {{ widget.label }}
             </option>
           </optgroup>
           <optgroup label="Standard Library" v-else>
-            <option v-for="widget in orderedStandardWidgets" :key="widget.name" :value="widget.name" :selected="metadata.value === widget.name">
+            <option v-for="widget in standardWidgets" :key="widget.name" :value="widget.name" :selected="metadata.value === widget.name">
               {{ widget.label }}
             </option>
           </optgroup>
           <optgroup v-if="$store.getters.widgets.length" label="Personal Widgets">
-            <option v-for="widget in orderedPersonalWidgets" :value="'widget:' + widget.uid" :key="widget.uid" :selected="metadata.value.replace('widget:', '') === widget.uid">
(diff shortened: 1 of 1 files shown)


ghys:
> I tried real hard to find out what you added this for, there surely must have been a reason - but couldn't.
If by !this.metadata.config[key] you meant "if that key does not exist": that doesn't make sense when iterating through the keys. If you meant "if the config value is falsy and the default truthy": I can't see how it makes sense to set the value to an empty string (or from the comment the originally desired 'null'). Perhaps you remember what it's for, if it's not just some weird leftover.

https://github.com/openhab/openhab-webui/blame/main/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-widget.vue
It was added in #544 to fix #542 after #540.
This allows to keep the automatically-provided default widget (the metadata value is blank) and only override some of the preconfigured options. When setting it to null to clear it - like don't perform any action instead of the default one, like analyze - when posting to the REST API the server silently removed it, so it didn't work (when the config property is not set in metadata, use the automatically-provided option).

hubsif:
> don't perform any action instead of the default one

Ah ok, I actually thought it could be something like this, but seems I didn't stay on it enough.
I'll try to find a solution that fixes both, no-actions and intentionally falsy values.

hubsif:
Ok, so I examined this further and I now know why skipped the idea this might be to override an action default value to "none": It simply doesn't work for me! 😮 

When you select no action (empty option), whose resulting select option value is falsy, the config parameter gets fully deleted in configSheet, and thus in itemMetadataWidget's metadata.config. And with the parameter not being present in metadata.config, it can't be iterated over as key. Meaning: for me that line is actually never executed as intended. But this would mean that the fix never worked, which I can barely believe.

So, where's my fault? 🤔

ghys:
You're right it doesn't seem to work (anymore?).
That's probably since I added the `set-empty-config-as-null` prop to config-sheet in https://github.com/openhab/openhab-webui/blob/0a1848e9af6658b5f4034ee76597098fcecc4307/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue#L116...

Maybe it could be set in this case?
See https://github.com/openhab/openhab-webui/issues/553#issuecomment-732766927 and https://github.com/openhab/openhab-webui/pull/339#discussion_r489632719 for more explanation on why this parameter was sometimes needed, sometimes not.

ghys:
I struggled to try and finally have a jest test for this, like this one:
https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/test/jest/__tests__/config-parameter.test.js
(which miraculously still works, and I would really really like to improve in this area)

Turns out you have to mock the Vuex store, and the SSE connection API, but you still don't have a `Vue.prototype.$f7.utils` in the test Vue instance, so I can't "shallow mount" a Widget Metadata Editor for testing.
Not giving up yet but it looks like a challenge.

hubsif:
I think tests can be very helpful and are high level coding, but unfortunately setting them up and maintaining them can cost an awful lot of time 😃 

I was also looking for a fix for this and now I'm struggling with YAML:
somehow this
```js
{ config: { action: null } }
```

gets stringified to that:

```yaml
config:
  ? action
```

while the other way round (with correct yaml) it works just fine 🤨

hubsif:
ok, so enabling `setEmptyConfigAsNull` and checking for a string default value makes it all work (to set no action _and_ falsy values). Though this all seems more than a workaround than a perfectly clean solution for following reasons:
1. configSheet now sets the value to "null", just so that itemMetadataWidget can turn it into an empty string, just so that the API doesn't remove the config parameter when it shouldn't
2. the UI actually only displays the empty action option coincidentally, because parameterOptions can't match the empty string value to a select option

And of course it would be nicer to be able to display a "no action" or "none" for the empty string.

Best would be to have a consistent standard throughout the API and the UI for these things, so that all widgets could simply rely on the configuration components.

But then, ... it works 😃 . And maybe someday there'll be time to beautify things like these.

ghys:
> 1. configSheet now sets the value to "null", just so that itemMetadataWidget can turn it into an empty string, just so that the API doesn't remove the config parameter when it shouldn't

👍 The option is to modify the core to allow nulls in metadata config parameters, but I don't know if it makes sense/could lead to side effects.

> 2. the UI actually only displays the empty action option coincidentally, because parameterOptions can't match the empty string value to a select option

¯\\_(ツ)_/¯
That is acceptable because all non-required parameters have an empty option at the top, which would be selected if there's no matching option. I don't think it's necessary to have an explicit "no action" label.

Btw - I noticed a small bug when testing this, when you click on "Remove metadata", the dirty flag has to be cleared, because now you get the confirmation dialog... while the metadata is deleted in the background. But since it's not strictly related to this PR, it can be addressed separately.

ghys (approved):

PR #948, opened 2021-03-09
Refactor routes: Move admin pages away from entry point

ghys:
Load all settings pages, about, profile, and the HABot control asynchronously.
This also moves library dependencies to separate chunks, and thus reduces the
entry point by several hundred kb (~25%) if the user only accesses end-user pages,
until they go to these pages.


ghys:
### Before
image

### After
image

PR #950, opened 2021-03-10
 [MainUI] fix type "number" for list separator in Designer

### LINKED ISSUE #949: [MainUI - Sitemap Pagedesigner] Wrong type "number" for list separator in Designer
## The problem

Sitemap pagedesigner widget uses wrong type `number` for control `Separator`.  Therefore a correct separator for a list widget (like `,`) cannot be set via the designer. It is possible to set the separator via code (see screenshot below) but the set separator will not be displayed in the designer (and a following change+save of the separator with a string/text in the designer results in an unset of the seperator in the sitemap code... i guess because of validation?).
grafik

## Expected behavior

Control `separator` should have type `text` so it is possible to set the separator with the designer too.

## Steps to reproduce

1. make a sitemap (Settings -> Pages -> Create Sitemap)
2. add a list widget
3. try to set separator `,` in designer
4. save the sitemap (no error occures and separator is still visible in designer)
grafik
5. switch to tab `Code`
6. no separator prop was added in sitemap code
grafik
7. switch back to tab `Design`
8. separator `,` disappeared
grafik

## Your environment

```yaml
runtimeInfo:
  version: 3.1.0
  buildString: "Build #2211"
locale: de-AT
systemInfo:
  configFolder: c:\openhab\conf
  userdataFolder: c:\openhab\userdata
  logFolder: c:\openhab\userdata\logs
  javaVersion: 11.0.6
  javaVendor: Oracle Corporation
  javaVendorVersion: "18.9"
  osName: Windows 10
clientInfo:
  device:
    firefox: true
    os: windows
```

## Additional information

The problem seems to be in this line:

https://github.com/openhab/openhab-webui/blob/eb7596c42ddb9d73fd4f6e2dcd506fd38299474b/bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/widget-details.vue#L30

ListRenderer in BasicUI uses `getSeparator()` to split the string

https://github.com/openhab/openhab-webui/blob/eb7596c42ddb9d73fd4f6e2dcd506fd38299474b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ListRenderer.java#L64

and according to `openhab-core` separator is of type `string`:

https://github.com/openhab/openhab-core/blob/4a79e9bfe7dca9f6abd1e40d3eb398a9aa2c8988/bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/Sitemap.xtext#L104

I'm willing to submit a PR that fixes the wrong type in the `widget-details.vue` file.

DarkC35:
Fixes #949 

After changing the separator type to `text` it works as intended but there was one problem in following use case: when you need a space `' '` as separator character/string, the `updateParameter` method interprets it as a non-NaN value (since `isNaN(' ')` is `false`) and resultet in a `NaN` as input string. (This also occurs on other text inputs when you type a space, like url, refresh, frequency, ...)

I tought about two ways to solve this:
1. Only check for isNaN when `type="number"` (like in the second commit)
2. use `isNaN(parseFloat(value))` instead of `isNaN(value)`, since `parseFloat(' ') === NaN` so the check would work. This way had one problem when the number of space characters is important because `parseFloat('       1') === 1` and this will trim leading whitespaces of the input.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/widget-details.vue
+++ b/bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/widget-details.vue
@@ -27,7 +27,7 @@
           <f7-list-input v-if="supports('minValue')" label="Minimum" type="number" :value="widget.config.minValue" @input="updateParameter('minValue', $event)" clear-button />
           <f7-list-input v-if="supports('maxValue')" label="Maximum" type="number" :value="widget.config.maxValue" @input="updateParameter('maxValue', $event)" clear-button />
           <f7-list-input v-if="supports('step')" label="Step" type="number" :value="widget.config.step" @input="updateParameter('step', $event)" clear-button />
-          <f7-list-input v-if="supports('separator')" label="Separator" type="number" :value="widget.config.separator" @input="updateParameter('separator', $event)" clear-button />
+          <f7-list-input v-if="supports('separator')" label="Separator" type="text" :value="widget.config.separator" @input="updateParameter('separator', $event)" clear-button />
           <f7-list-item v-if="supports('switchEnabled')" title="Switch enabled">
             <f7-toggle slot="after" :checked="widget.config.switchEnabled" @toggle:change="widget.config.switchEnabled = $event" />
           </f7-list-item>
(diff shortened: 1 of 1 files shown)


ghys (approved):
LGTM, thanks

PR #952, opened 2021-03-10
[MainUI] fix lint npm scripts on windows

### LINKED ISSUE #951: [MainUI - Development] lint scripts do not work on Windows
## The problem

`npm run lint` and `npm run lint:fix` scripts do not work on Windows because double quotes are required for `glob` syntax.
> If you want to use node glob syntax, you have to quote your parameter (using double quotes if you need it to run in Windows)
see: https://eslint.org/docs/user-guide/command-line-interface

grafik

## Expected behavior

`npm run lint` and `npm run lint:fix` should work on Windows too.

## Additional information

There are currently some lint errors on `main` branch:
grafik

Those are not covered by the GH Action since it only covers `src`:

https://github.com/openhab/openhab-webui/blob/0a1848e9af6658b5f4034ee76597098fcecc4307/.github/workflows/mainui.yml#L42

Maybe the lint script should be unified and `npm run lint` used in the Action?

DarkC35:
Fixes #951 

Tested on Windows with powershell, cmd and wsl.
Not tested on Linux and macOS.

Up for discussion: Should `npm run lint` also be used in GH Action? (see #951)

Signed-off-by: Kevin Haunschmied <darkc.futarialone@gmail.com>


ghys:
Thank you, but I think in the end `npx eslint --ext js --ext vue src` is best, can you confirm and adjust please?

DarkC35:
> Thank you, but I think in the end npx eslint --ext js --ext vue src is best, can you confirm and adjust please?

They are both valid and seem to do the same, changed them to use extensions instead of glop pattern! But note that they are mutally exclusive:

> Note: --ext is only used when the arguments are directories. If you use glob patterns or file names, then --ext is ignored.

The only difference now:  `src` folder is used instead of root.

PR #955, opened 2021-03-12
Use commons-lang3 instead of commons-lang

wborn:
This allows for removing the old commons-lang dependency.


kaikreuzer:
Just for my understanding: Wasn't https://github.com/openhab/openhab-addons/issues/7722 about completely removing dependencies to Apache Commons Lang and not just about upgrading to version 3 of it?

wborn:
Yes it is, but we can easily get rid of commons-lang today by just changing the packages in the imports. commons-lang:2.6 is more than 10 years old. :-)

kaikreuzer:
Ok, so it's a first step on the way then...

kaikreuzer (approved):
Thanks!

PR #960, opened 2021-03-15
Wrap list items in list card in widget editor preview, fixes #923

### LINKED ISSUE #923: Preview in widget designer doesn't work for widgets based on oh-* components
## The problem

When you develop a widget and base it on some oh-* widget, e.g. oh-label-item, the preview does not show anything at all:

image

When you use an f7-* component instead it works, but looks broken (obviously because some parts from the oh-* component are needed):
image

This makes creating new widgets, especially list item widgets, quite hard because you need to actually place them on a page to see the result.

## Expected behavior
Preview should also work for widgets based on oh-* components.

## Steps to reproduce

1. Copy the YAML from one of [Rich Koshaks examples](https://community.openhab.org/t/example-custom-list-widgets/116265 ) and insert it into a new widget.

## Your environment
OH3.0.1

## Browser console
As soon as you type a valid oh-* name:
image

hubsif:
Result looks like this:

wrapped_list_item

It shows some additional information in the list card footer, so developers are made aware about what's happening behind the scenes. Can be omitted.

Roughly based on code in `item-metadata-widget`.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/developer/widgets/widget-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/widgets/widget-edit.vue
@@ -137,8 +137,7 @@ export default {
           ? {
             component: 'oh-list-card',
             config: {
-              mediaList: true,
-              footer: 'Detected list item as root component, wrapping in list card for preview'
+              mediaList: true
             },
             slots: {
               default: [this.widget]
--- a/bundles/org.openhab.ui/web/src/pages/developer/widgets/widget-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/widgets/widget-edit.vue
@@ -105,6 +105,8 @@ import { strOptions } from 'yaml/types'
 import ConfigSheet from '@/components/config/config-sheet.vue'
 import DirtyMixin from '@/pages/settings/dirty-mixin'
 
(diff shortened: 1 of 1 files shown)


ghys:
> It shows some additional information in the list card footer, so developers are made aware about what's happening behind the scenes. Can be omitted.

Not sure the footer is necessary to be honest.
Also a simple `/^oh-.+-item$/` match on the root component name (or `f7-list-item`) is perhaps enough to detect that it's based on list item widget?

hubsif:
> Not sure the footer is necessary to be honest.

That's up to you, I'm indecisive...

> Also a simple /^oh-.+-item$/ match on the root component name (or f7-list-item) is perhaps enough to detect that it's based on list item widget?

Nope, had the same thought, but that would also match e.g. 'oh-grid-item'. Ok, that's not to be used for widget development, but I then thought it's actually cleaner to check for "real" list-items than to "assume" from the name - who knows what else is coming? 😉

ghys:
> That's up to you, I'm indecisive...

Me too, but I finally decided against it, I know where you're coming from, but since it's the natural thing to do, we don't have to explicitly mention it, as it should be an appropriate reaction nobody will have issues with.
I removed the message. The rest is fine!

ghys (approved):

PR #965, opened 2021-03-20
Oh-Stepper component load default value when not initialized

### LINKED ISSUE #924: oh-stepper with variable always starts at 0
## The problem

If I create a oh-stepper component without a linked item and just a variable to store its state, the min values gets ignored on load / reload and it always starts at 0.

If I use the toggles then, to set the stepper +1 it sets the value to the min value +1 as expected (but ignoring the set min-value itself)

oh-stepper

## Expected behavior
Starting at the set min value on load/reload if a variable name is given.

## Steps to reproduce

Creating a oh-stepper component with a button as a trigger to reset the stepper.
```
- component: oh-stepper
   config:
     variable: addQuantity
     min: 1
     max: 10
     value: 1
     step: 1
- component: oh-button
   config:
      clearVariable:
         - addQuantity
```

## Your environment
openHAB 3.1.0
Build #2190

hoefnageltom:
When an oh-stepper component is created with a simple variable, the value is deleted within the mounted.  See issue #924 

This simple fix checks if the value is null and replaces it with the defaultValue.

I don't know if this is a proper solution, as I think the root of this issue is in the variable initialization but might be a good starting point for this issue.

closes #924 

Signed-off-by: Tom Hoefnagel <hoefnageltom@gmail.com>

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
@@ -1,5 +1,5 @@
 <template>
-  <f7-stepper ref="stepper" v-bind="config" :value="value" @stepper:change="onChange"
+  <f7-stepper ref="stepper" v-bind="config" :value="(value)? value : defaultValue" @stepper:change="onChange"
               :manual-input-mode="false" :format-value="formatValue" />
 </template>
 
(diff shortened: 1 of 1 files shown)


ghys:
I believe you if you found this happens to solve #924, but I also honestly don't understand how it's related.

Looking at the stepper code I don't even know why this `delete this.config.value` is necessary. IIRC it was because since there's a `v-bind="config"` and you happen to have a `value` parameter in your config, it could take precedence over the "proper" value binding, which is the computed property of the same name - so this code made sure it was discarded from the config if it existed.

The thing is, this shouldn't happen because `value` is not a valid config parameter, so you shouldn't ever set it; you either set `item` or `variable`.

(The other thing, unrelated, is that the variable-related parameters like `variable` and `clearVariable` are not even described in the definitions for `oh-stepper`, `oh-button` etc.)

hoefnageltom:
Hmm I see, I checked if I could come up with a more straightforward solution, but couldn't find one, I'll keep looking, but will close this PR for now as I don't think my solution is related to the origin of this issue.

PR #954, opened 2021-03-12
Run search filter on item list page reinit, fixes #861

hubsif:
Analyzing this I found out that the searchbar search gets overriden by virtual list's `replaceAllItems` and has to be rerun. So, the only required change here actually is:

https://github.com/hubsif/openhab-webui/blob/085bc59f5d99b875135c35c60cff04bf1bd61d9f/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue#L151-L154

While analyzing I stumbled across `initSearchbar` and alike, and this:

https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue#L168-L175

To me that looked like you were struggling with rendering and loading race conditions.

So, again I did more than required and tried to clean that up. I think the main issue was in the list to be conditionally rendered and thus not available for the searchbar on init:

https://github.com/openhab/openhab-webui/blob/f601eb242e3477f42888939eb79b41d7ee9492a0/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue#L54

I changed that to `v-show` instead and was then able to tidy up quite some.

Additionally, I discovered a "flicker" when going back from an item detail to the list: Due to the list only loading at "pageAfterIn", the transition would still show the filtered list and when finished switch to the loading components.
I wondered why you actually only load the page at "page:afterin" instead of e.g. "page:beforeIn"?
Anyways, I found it to be best situated in "page:init" and "page:reinit", I hope I didn't miss anything.

PR #973, opened 2021-03-23
Fix some misc minor issues

hubsif:
> Fix zero margin not applied in oh-grid-layout

Fixes zero margin issue reported in [comment in #835](https://github.com/openhab/openhab-webui/pull/835#issuecomment-803290075)

> Make slider label help more descriptive

I had to lookup the f7 docs to find out where this "label" actually sits. Perhaps this small addition also helps others like it would help me 😉.

> Fix load issue in item-list-vlist

Fixes an issue introduced with #954: I repeatedly hit the issue that when returning from metadata to item-details, the whole site would not react due to the item-list-vlist page being pre-loaded in background and not having $refs defined and thus hitting an exception - though not always 🤔. I didn't really track it down to the root, just fixed it by having it only load before the page is actually shown, which made it disappear.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/slider.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/slider.js
@@ -6,7 +6,7 @@ export default () => [
   pd('max', 'Max', 'Maximum value'),
   pd('step', 'Step', 'Minimum interval between values'),
   pd('vertical', 'Vertical', 'Display the slider vertically'),
-  pb('label', 'Display Label', 'Display a label above the slider knob'),
+  pb('label', 'Display Label', 'Display a label above the slider knob while sliding'),
   pb('scale', 'Display Scale', 'Display a scale on the slider'),
   pn('scaleSteps', 'Scale steps', 'Number of (major) scale markers'),
   pn('scaleSubSteps', 'Scale sub-steps', 'Number of scale minor markers between each major marker'),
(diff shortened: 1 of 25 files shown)


bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/slider.js
```diff
@@ -6,7 +6,7 @@ export default () => [
   pd('max', 'Max', 'Maximum value'),
   pd('step', 'Step', 'Minimum interval between values'),
   pd('vertical', 'Vertical', 'Display the slider vertically'),
-  pb('label', 'Display Label', 'Display a label above the slider knob'),
+  pb('label', 'Display Label', 'Display a label above the slider knob while sliding'),
```
ghys, line 9: If you change definitions would you mind running `node generate.js` from https://github.com/openhab/openhab-webui/tree/main/bundles/org.openhab.ui/doc/components/src - this should and will be part of a CI process but... as of now it's not...

hubsif: Thank you for pointing that out. I just did that and it resulted in much more files being changed 😳. Mostly docs on material icons have been added, but e.g. for `oh-input.md` the syntax has even changed (from `:` to `()`). My fault?

ghys: Most if not all of these changes come as a result of #917 - I forgot to add the same thing back! Thanks!

ghys (approved):
LGTM!

PR #975, opened 2021-03-25
recalculate oh-slider when rendered, fixes #967

### LINKED ISSUE #967: Wrong width of oh-slider component in oh-cell
## The problem

The slider component (and maybe other components?!) sets its width depending on the number of visible cell components in the row they resides in (asuming that each of the components will take up the same amount of the page-width - which would work in the row / column layout). 

So if e.g. 2 cells are in one row the slider will take up 50% of the avilable width of that page (even if only parts of it are filled with cells because of their fixed width) - Everything works as expected in the row/column layout.

oh-cell_slider
The upper row consits of 2 cells; the lower row uses the row / column layout (and works fine)

## Expected behavior
The used components inside the oh-cell should only use the available width of that cell.

## Your environment
OH 3.1.0 build 2269 on RPi 4

hubsif:


ghys (approved):
Seems to be pretty harmless, I haven't found that it caused a regression in performance or rendering, so LGTM!

Issue #626, opened 2020-12-12
Enable widgets to use Javascript


hubsif:
Hi!

### The problem

While trying to rebuild my "tablet screen" using OH3 Main UI with what I had done so far with HABPanel, I failed right away with my first widget in the upper corner which is a simple digital clock.
I couldn't find a clock standard widget, so I started diving into building custom widgets. I then discovered that it seems to not be possible to create a clock since it's not possible to use Javascript in custom widgets (please correct me if I'm mistaken, it would invalidate this whole issue).

### Your suggestion

I think a lot of upcoming custom widgets might want to make use of the possibilities and flexiblity that Javascript brings, and I think this is required to have the community build a whole lot of "cool" widgets. Having read that MainUI is meant to be able to replace HABPanel sooner or later, I'm of the opinion this feature might be somewhat crucial for its success.

So, I started experimenting again and came up with the result below (which took a lot more time than it seems 😄 ).

And again, since I'm not sure what's desired for OH3 and being a total newbie in vue and f7, I'm filing this as a suggestion to think about, not a ready to use pull request.

Actually, the only change is this so far:
```diff
diff --git a/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js b/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
index 148162a..645978c 100644
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
@@ -68,6 +68,19 @@ export default {
       return true
     }
   },
+  mounted () {
+    if (this.context.component.script) {
+      let script = document.createElement("script")
+      script.text = this.context.component.script;
+      let oldscript = this.$el.getElementsByTagName('script')[0]
+      if (oldscript)
+        this.$el.replaceChild(script, oldscript)
+      else
+        this.$el.appendChild(script);
+    } else if (this.$el.getElementsByTagName('script')[0]) {
+      this.$el.removeChild(script);
+    }
+  },
   methods: {
     evaluateExpression (key, value, context) {
       if (value === null) return null
```

which enables widgets to have an additional `script:` parameter defining some javascript. Example:

```yaml
uid: widget_ca6fb002bd
component: f7-card
config:
  content: clock
  class: myclass
script: '
  setInterval(function() {
    document.querySelector(".myclass .card-content").innerHTML = new Date().toLocaleTimeString();
  }, 1000);'
```

[...]

PR #979, opened 2021-03-26
Add stylesheet configuration to widgets

hubsif:
Hi @ghys,

I addressed something I think many powerusers have requested for: custom css stylesheets.

This first draft allows to define a stylesheet for a page, like this:

```yaml
config:
  layoutType: responsive
  label: stylesheet_test
  sidebar: true
  stylesheet: |
    .myoneclass {
      background-color: red;
    }
    .myotherclass {
      font-size: 40px;
    }
blocks:
...
```

The stylesheet gets added globally on pageIn and removed on pageOut (it's not scoped to the page, so it can generally also affect things like the panel, but I think powerusers can deal with that).

Please let me know if something like this is desired and if I should complete it (e.g. add to layouting). Please also share if a stylesheet per page is a good approach in your view.
Some users might also want to add a global stylesheet, so it affects all pages at once. I'm also planning on adding this (if ok by you), though the approach will definitely be different (requires a global config or a static include or similar).

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
@@ -9,6 +9,7 @@ import isoWeek from 'dayjs/plugin/isoWeek'
 import isToday from 'dayjs/plugin/isToday'
 import isYesterday from 'dayjs/plugin/isYesterday'
 import isTomorrow from 'dayjs/plugin/isTomorrow'
+import scope from 'scope-css'
 
 dayjs.extend(relativeTime)
 dayjs.extend(calendar)
@@ -46,7 +47,7 @@ export default {
       if (this.context.component.config) {
         if (typeof this.context.component.config !== 'object') return {}
         for (const key in this.context.component.config) {
-          if (key === 'visible' || key === 'visibleTo') continue
+          if (key === 'visible' || key === 'visibleTo' || key === 'stylesheet') continue
           this.$set(evalConfig, key, this.evaluateExpression(key, sourceConfig[key]))
         }
       }
@@ -66,6 +67,24 @@ export default {
         return false
       }
       return true
+    }
+  },
+  mounted () {
+    if (this.context.component.config && this.context.component.config.stylesheet) {
+      this.cssUid = 'scoped-' + this.$f7.utils.id()
+
+      this.$el.classList.add(this.cssUid)
+
+      let style = document.createElement('style')
+      style.id = this.cssUid
+      style.innerHTML = scope(this.context.component.config.stylesheet, '.' + this.cssUid)
+      document.head.appendChild(style)
+    }
+  },
+  beforeDestroy () {
+    if (this.cssUid) {
+      const scoped_stylesheet = document.getElementById(this.cssUid)
+      if (scoped_stylesheet) scoped_stylesheet.remove()
     }
   },
   methods: {
--- a/bundles/org.openhab.ui/web/package-lock.json
+++ b/bundles/org.openhab.ui/web/package-lock.json
@@ -28,6 +28,7 @@
         "nearley": "^2.19.6",
         "pkce-challenge": "^2.1.0",
         "qrcode": "^1.0.0",
+        "scope-css": "^1.2.1",
         "sprintf-js": "^1.1.2",
         "template7": "^1.4.2",
         "tern": "^0.23.0",
@@ -2901,13 +2902,21 @@
       },
       "engines": {
         "node": ">= 10.0"
+      },
+      "peerDependencies": {
+        "lodash": "^4.17.11"
       }
     },
     "node_modules/@bundle-stats/utils/node_modules/core-js": {
       "version": "3.9.1",
       "resolved": "https://registry.npmjs.org/core-js/-/core-js-3.9.1.tgz",
       "integrity": "sha512-gSjRvzkxQc1zjM/5paAmL4idJBFzuJoo+jDjF1tStYFMV2ERfD02HhahhCGXUyHxQRG4yFKVSdO6g62eoRMcDg==",
-      "dev": true
+      "dev": true,
+      "hasInstallScript": true,
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/core-js"
+      }
     },
     "node_modules/@cnakazawa/watch": {
       "version": "1.0.3",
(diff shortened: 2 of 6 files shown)


ghys:
I'm not opposed to it.

To be complete, normally I'm rather wary of giving so much latitude for security reasons and the fact that you could completely break the UI, but for stylesheets I'm not opposed to this approach as I believe it's less of a risk than allowing inline JavaScript (#626).
Still, it could be exploited to deceive and manipulate users, if abused. I believe it would still be wise to add a well-crafted Content Security Policy (like in [HABPanel](https://github.com/openhab/openhab-webui/blob/7cbdc3cc32a38d02480c9c345dcb77eea0c1d195/bundles/org.openhab.ui.habpanel/web/index.html#L13)) to at least prevent "`@import`ing" CSS from external files - i.e. all stylesheets should either be inline or loaded from the same origin i.e. server-relative URLs such as `/static/...` - that is including frameworks like Tailwind or Bootstrap, so they would have to be copied over locally and not loaded from a CDN. This gives the admin a chance to review what they're including and rest assured that the remote stylesheet will not change and silently include malicious code without notice.

I get the argument that adding the ability to override global styles is a powerful feature, but maybe it's also a bit dangerous. Scoping to the page would ensure that the sidebar remains unaffected and predictable so it's possible to bail out of a page with broken stylesheet overrides. I'm ready to be convinced that there are legitimate use cases where it would not be desired. On the contrary, it could be argued that it would be desirable to have the ability to add scoped stylesheets to every "root" UI component -not only pages but also widgets; so you could define classes that you use inside your widget that wouldn't "spill" to other widgets.

For adding global CSS we don't have a standard way of loading options for the entire UI, as I didn't find a need for them yet and it spares one additional request to the API before it can be rendered. I'm open to all ideas!

hubsif:
> On the contrary, it could be argued that it would be desirable to have the ability to add scoped stylesheets to every "root" UI component -not only pages but also widgets

I'm glad you said that, since the idea is actually as follows:
Extend general widget (widget-mixin maybe) and page configuration to support stylesheet definitions like above. These then get scoped (by some css scoping library) and are added when mounted and removed when unmounted.

This would allow for e.g. custom widgets to define their own stylesheet, which I think could drastically reduce repetitions and make them cleaner. And due to scoping they're limited to their component/widget. Being part of the widget YAML they're also included when shared.

As for security, I think a Content-Security-Policy is a simple and effective way to limit resources to the openHAB instance. I even don't think external imports are actually required and `@import` statements could simply be prohibitted.

Regarding global CSS my current suggestion would be to follow the simplest approach and have a predefined static style include added to the app, e.g. `/static/ui_custom.css`. If users want to add global css they can create that file and add their stylings.

hubsif:
So, originally I wanted to wait for your ok, but couldn't wait trying 😉.

I "moved" the stylesheet configuration to the widget mixin and added scoping. Surprisingly, this works smoothly for all I've tested so far, also in layout-edit and widget developer (latest at redraw). Though, it surely requires some more, thorough testing.
Also, I limited CSS to 'self' by extending the CSP.

Simple custom widget sample:

```yaml
uid: label_card_stylesheet
component: f7-card
config:
  stylesheet: |
    .testclass {
      color: red;
    }
slots:
  content:
    - component: Label
      config:
        class: testclass
        text: Test
```

When I first attended to adding this feature, I didn't think it was actually that simple to implement 😄.

ghys:
Works pretty well!
Tested with this example:
```yaml
component: f7-card
config:
  style:
    display: flex
    justify-content: center
  stylesheet: >
    .mylink {
      color: blue !important;
    } .mylink:hover {
      text-decoration: underline !important;
      transform: scale(1.5);
      transition: 0.3s all;
    }
slots:
  default:
    - component: oh-link
      config:
        text: Test
        class:
          - margin
          - mylink
```
A small bug, the stylesheet doesn't get initialized when you enter the widget editor; until you force a redraw or change the split orientation. On pages it works normally.
This seems to be fixed by setting ready to false (don't know why it is initialized to true) here:
https://github.com/openhab/openhab-webui/blob/e9e55060db1b8d31a019be98f089e2a96d68538f/bundles/org.openhab.ui/web/src/pages/developer/widgets/widget-edit.vue#L123

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
```diff
@@ -68,6 +69,24 @@ export default {
       return true
     }
   },
+  mounted () {
+    if (this.config && this.config.stylesheet) {
+      this.css_uid = 'scoped-' + Math.random().toString(36).substr(2, 9)
```
ghys, line 74: - `this.css_uid` should be initialized (to null) in `data ()` and also be camelCase instead of snake_case (i.e. `cssUid`).
- to generate a random ID `this.$f7.utils.id()` is used throughout the app, maybe use that here as well?

hubsif: - right, sorry, been sick last week (is that a valid excuse for all these "consistency faults"? 😆 )
- hm, I should have looked deeper into this. I thought that `id()` generates a uuid, which I thought is unnecessarily long in this case (well, eventually it wouldn't have mattered either). Checking the docs I now realize that by default it creates a 10 chars hex hash, which is about the same. So, will adjust, of course!

hubsif:
> This seems to be fixed by setting ready to false (don't know why it is initialized to true) here:

Hm, git blame: *13 months ago ... **ghys**!* 😜
Let me see if I can find a reason for this ...

I found another "blemish": At least when used in custom widgets, the stylesheet is applied as html attribute to the element. I don't think that breaks anything, as it should be ignored. Yet, it's an ugly spot, especially with longer stylesheets...

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
hubsif: ah, and regarding `data ()`:
I meanwhile read on multiple websites that it's not best practice to put variables in data() that don't need to be reactive, and that one should avoid simply initializing all "component variables" there. That's actually the reason I didn't put it there.
It's up to you how to handle this. Eventually, I think it doesn't hurt much, it just creates some getters and setters that are never called...

hubsif: > * `this.css_uid` should be initialized (to null) in `data ()` and also be camelCase instead of snake_case (i.e. `cssUid`).
>
> * to generate a random ID `this.$f7.utils.id()` is used throughout the app, maybe use that here as well?

fixed. what about `data ()` ?

hubsif:
> I found another "blemish": At least when used in custom widgets, the stylesheet is applied as html attribute to the element. I don't think that breaks anything, as it should be ignored. Yet, it's an ugly spot, especially with longer stylesheets...

fixed.

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
ghys: > what about data () ?

Your argument above seems fair enough to me.

ghys (approved):
Nice addition, now rather safe to add with the scoping and CSP, and which I suspect will boost the creativity of widget authors, can't wait to see the results!

I'd rather leave that out of the widget definitions (so it's not documented nor displayed on config GUIs), but neither are `visible`, `visibleTo`, `style`, `class` etc.
It'd be nice to make all of these at least appear in the YAML editors' autocompletion where relevant, eventually. This could be done in one go and directly in the hints file for components.

It would be nice to amend the documentation article to mention the new stylesheet parameter, though.

altaroca:
I just noticed that the `stylesheet` parameter doesn't work on the floor plan page. After some investigation I think there is an error on this line:
`this.$el.classList.add(this.cssUid)`
If `classList` is empty this will throw. It seems that vue optimizes components such that `v-if` results in an empty comment if not true. In the case of floor plans that leads to `this.$el` being just such a comment and `classList` being empty.
I do not have a good solution since it is difficult to find the widget's root element in this situation. Maybe scan the dom tree upwards to a known class? Thanks for your support.

PR #980, opened 2021-03-27
Show copy icon for thing and channel uids, fixes #396, fixes #475

### LINKED ISSUE #396: [MainUI] Allow the user to copy and paste the Thing ID and Channel IDs more easily
More discussion on this is located at https://community.openhab.org/t/openhab-3-0-my-getting-started-and-notes/106467.

One thing PaperUI got right was that nice little button to copy the Channel ID to the clipboard. As I migrate my Items and Things to OH 3 I find myself choosing better names for some of my Things and as a consequence the Channel IDs changed. I would paste the Items into the "add Item from text config" and edit the Channel there. But in another browser tab I'd have to navigate to the thing, then the Channel, then open the Channel config before I could get to a place where I could copy even a part of the ID. Even if the Thing ID were made selectable from the list of Things this workflow would be greatly improved. When I try to select the text though it won't let me (latest Firefox is the browser).

This is a usability related feature request.

### LINKED ISSUE #475: [MainUI] Missing quick way to copy the thing UID
The thing UID is something you can need when you want to move to a config file or even use it in a console command. This can be a long string and you don't want to type it but rather retrieve it from UI !
Until now, there is no way to copy it from the UI.
At the reverse, we have a way to copy the channel UID if you go in the channel detail/configuration page.

hubsif:
This is something I also have been missing and since I have not fully migrated yet, it still might come in handy 😉.

It adds small copy icons to quickly copy thing and channel uids. It required a little "creativity", e.g. for the Thing tab, since the disabled list-item-input wouldn't allow to place an icon next to it, and for the channel list as it seems to not be possible to prevent the accordion from toggling when clicked on a link inside the list-item's link (see [f7-forum](https://forum.framework7.io/t/framework7-vue-accordion-onclick-stoppropagation/3626)).

I think the result is acceptable and looks like this:

Screenshot 2021-03-27 at 17 55 52
Screenshot 2021-03-27 at 17 56 11
Screenshot 2021-03-27 at 17 56 51

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
+++ b/bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
@@ -8,7 +8,15 @@
               <f7-list-input label="Unique ID" v-if="createMode" type="text" placeholder="Required" :value="thing.ID"
                              @input="changeUID" info="Note: cannot be changed after the creation"
                              required validate pattern="[A-Za-z0-9_\-]+" error-message="Required. A-Z,a-z,0-9,_,- only" />
-              <f7-list-input label="Identifier" type="text" placeholder="Name" :value="thing.UID" disabled />
+              <f7-list-item>
+                <div slot="inner" class="item-label">
+                  Identifier
+                </div>
+                <div slot="inner" class="item-input-wrap">
+                  {{ thing.UID ? thing.UID : 'Name' }}
+                  <clipboard-icon v-if="thing.UID" slot="inner-end" :value="thing.UID" tooltip="Copy UID" />
+                </div>
+              </f7-list-item>
               <f7-list-input label="Label" type="text" :disabled="!ready || readOnly" placeholder="e.g. My Thing" :value="thing.label"
                              @input="thing.label = $event.target.value; $emit('updated')" required validate />
               <f7-list-input label="Location" type="text" :disabled="!ready || readOnly" placeholder="e.g. Kitchen" :value="thing.location"
@@ -36,11 +44,13 @@
 
 <script>
 import ThingPicker from '@/components/config/controls/thing-picker.vue'
+import ClipboardIcon from '@/components/util/clipboard-icon.vue'
 
 export default {
   props: ['thing', 'thingType', 'createMode', 'ready', 'readOnly'],
   components: {
-    ThingPicker
+    ThingPicker,
+    ClipboardIcon
   },
   methods: {
     computedThingUid () {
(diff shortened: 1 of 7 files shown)


hubsif:
Hm, don't know why webpack-stats failed, works for me locally...

ghys:
I think you forgot to add your `@/components/utils/clipboard-icon.vue` component.

hubsif:
right 🤦

bundles/org.openhab.ui/web/src/pages/settings/things/channel/channel-edit.vue
```diff
@@ -38,10 +42,12 @@
 
 <script>
 import ConfigSheet from '@/components/config/config-sheet.vue'
+import ClipboardIcon from '@/components/util/clipboard-icon'
```
ghys, line 45: Speaking of which - I've made an habit of specifying the extension when importing (except if it's .js and except in `@/assets/definitions/widgets` because the node loader is different from webpack and requires them to generate the docs).

```suggestion
import ClipboardIcon from '@/components/util/clipboard-icon.vue'
```

ghys (changes requested):

bundles/org.openhab.ui/web/src/components/thing/thing-general-settings.vue
```diff
@@ -36,11 +44,13 @@
 
 <script>
 import ThingPicker from '@/components/config/controls/thing-picker.vue'
+import ClipboardIcon from '@/components/util/clipboard-icon'
```
ghys, line 47: There as well.

bundles/org.openhab.ui/web/src/pages/settings/things/things-list.vue
```diff
@@ -121,11 +124,13 @@
 
 <script>
 import thingStatus from '@/components/thing/thing-status-mixin'
+import ClipboardIcon from '@/components/util/clipboard-icon'
```
ghys, line 127: And there.

hubsif:
Imports fixed. I usually try to comply to existing code as much as possible, somehow missed that today...

ghys:
No worries at all.

> I usually try to comply to existing code as much as possible

I'm glad, that's what I do too even if I don't agree with the choices made previously, it's just the right thing to do. Even if there are better ways to do things it's easier to migrate a consistent code base, posssibly with tooling, than deviate over time from said consistency and have a patchwork of what individual contributors thought was best, I think.
But I know in that case it was just an accident.

ghys (approved):

PR #993, opened 2021-04-07
Added trendSampling to doc

dominik2579:


ghys:
Thanks... I don't think your description is accurate though ;)
If you have for instance one data point every 5 minutes, the sampling is not the "amount of minutes between each point", rather the "amount of 5 minutes". I can't make my mind on how to express this correctly though.

Please consider the [sign-off requirements](https://www.openhab.org/docs/developer/contributing.html#sign-your-work) for contributions, if you can't add it to the commit messages, don't worry, include it as a comment on this PR and I will include it.

dominik2579:
Unfortunately I didn't find a fully accurate description that does not confuse unexperienced users. 
I thought the hint "Affected by persistence strategies different from 'every minute'" gives enough information to experienced users.
Another choise could be "Detail level of the trend line (default: 60). The lower, the more detailed the trend line.". That does not claim to explain the functionality in detail.

The sign-off requirements give exceptions to the following:
Your patch fixe
- Your patch fixes spelling or grammar errors.
- Your patch is a single line change to documentation.

I assume this to be last one :) ☝️ 
Anyway I updated the commit messages.

ghys:
Still not that thrilled that the description could be wrong, maybe something like:

_"Take only the nth data points retrieved from the persistence service (default: 60). Change to adjust the smoothness of the line."_

would be more correct.

But let's not split hairs... thanks.

ghys (approved):

Issue #906, opened 2021-02-11
Adding a Location with a name of an existing Location silently moves the original Location instead


eikowagenknecht:
## The problem
Adding a second Location with a name moves the original Location. See "Steps to reproduce" for details.

## Expected behavior
I'd expect an error message "Can't create another Location with that Name because it already exists." or at least a warning "Location already exists, continue and move it into this Location instead?".

## Steps to reproduce
1. Create Location "Root"
2. Create Location "TestFloor1" below "Root"
3. Create Location "TestFloor2" below "Root"
4. Create Location "Bathroom" below "TestFloor1"
5. Create Location "Bathroom" below "TestFloor2"

After step 4 it looks like this:

![image](https://user-images.githubusercontent.com/1475672/107620545-28f38500-6c55-11eb-9a28-b62ba1eb3356.png)

After step 5 it looks like this:

![image](https://user-images.githubusercontent.com/1475672/107620586-390b6480-6c55-11eb-809f-fce05a7f3170.png)

## Your environment
OH 3.0.1

## Browser console
Nothing to see here.

## Browser network traffic
Nothing to see here.

Issue #793, opened 2021-01-12
[Main UI] Caching for API responses


dlaplexurenet:
## The problem

Personally I use pretty often the "Add Items from Textual Definition". When you open this page, it's basically doing 3 XHR calls to load things, items and links. I have roughly 220 things and over 1400 items. After a restart of OpenHAB, this takes roughly 2s from things endpoint and 6s for items. After a couple of days, these waiting times go down hill (I opened 2109 in openhab-core for this).

Regardless from the root cause of this issue, I believe that the things, items and links endpoints are probably reloaded very often even if there's no need to load them from the server. Example: searching the items or display the badge with items count.

## Your suggestion

Would it be possible to load things, items and links and check the last modification of these objects on the server side? So before initiating a GET, you could do a HEAD, verify last modification or etag and compare with information in localStorage of the browser. As long as no changes have been made, all operations could be done on the data in the localStorage.

## Your environment

```yaml
runtimeInfo:
  version: 3.0.0
  buildString: Release Build
locale: en_LU
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 11.0.8
  javaVendor: Azul Systems, Inc.
  javaVendorVersion: Zulu11.41+75-CA
  osName: Linux
  osVersion: 5.4.72-v8+
  osArchitecture: arm
  availableProcessors: 4
  freeMemory: 77327728
  totalMemory: 259522560
bindings:
  - airquality
  - astro
  - exec
  - gardena
  - homeconnect
  - hpprinter
  - hue
  - icalendar
  - irobot
  - km200
  - knx
  - logreader
  - mqtt
  - mystrom
  - netatmo
  - ntp
  - samsungtv
  - sonos
  - telegram
  - unifi
  - unifiprotect
  - weathercompany
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: true
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: macos
    pixelRatio: 2
    prefersColorScheme: dark
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: N/A
[...]

PR #1000, opened 2021-04-15
Check if item exists when creating new, fixes #573

### LINKED ISSUE #573: Add link: make sure item doesn't exist if creating
## The problem

If I execute the function "Create a new Item" for a channel and I use an existing item name I get no hint that the item already exists. Instead the item will be linked to two channels.

## Expected behavior

If an existing item name is used in function "Create a new Item" either reject the creation of ask the user whether it is OK to create instead a link to the already existing item.

## Steps to reproduce

To demonstrate the problem, I have used the Astro binding:

Step 1: create a new item and link for a channel and name the item: ItemA
image

Step 2: select another channel, select "Create a new Item" and use the same item name as in step 1
image

The item "ItemA" now is linked to two channels:
image

## Your environment

- Build 2032
- Windows 10 with Zulu Java version 11
- Chrome Browser

## Browser console

```txt

```

hubsif:
This addresses #573. It adds an existence check to item-form.

Since the UI would load the items from the server multiple times and also each time the radio button is toggled between "new" and "existing", I modified `item-form` and `item-picker` to accept already loaded items as property. Items are now loaded in `link-add` (if required) and passed on to those components.

Originally I wanted to start migrating this to Vuex. But during planning I realized that this would become too big a change, since I think when doing this the whole REST API should be wrapped in a cache (e.g. [vuex-rest-api-cache](https://github.com/netsells/vuex-rest-api-cache)), so unchanged stuff doesn't always get loaded from the server. Since this would have had major impact in API communications and required a lot of time for testing, I dropped it.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -3,8 +3,7 @@
     <f7-list inline-labels no-hairlines-md>
       <f7-list-input v-if="!enableName" label="Name" type="text" placeholder="Name" :value="item.name" disabled />
       <f7-list-input v-else label="Name" type="text" placeholder="Name" :value="item.name"
-                     @input="item.name = $event.target.value" clear-button
-                     required validate pattern="[A-Za-z0-9_]+" error-message="Required. Alphanumeric &amp; underscores only" />
+                     @input="onNameInput" clear-button required :error-message="nameErrorMessage" :error-message-force="!!nameErrorMessage" />
       <f7-list-input label="Label" type="text" placeholder="Label" :value="item.label"
                      @input="item.label = $event.target.value" clear-button />
       <f7-list-item v-if="item.type && !hideType" title="Type" type="text" smart-select :smart-select-params="{searchbar: true, openIn: 'popup', closeOnSelect: true}">
@@ -46,7 +45,7 @@ import * as Types from '@/assets/item-types.js'
 import { Categories } from '@/assets/categories.js'
 
 export default {
-  props: ['item', 'enableName', 'hideCategory', 'hideType', 'hideSemantics', 'forceSemantics'],
+  props: ['item', 'items', 'enableName', 'hideCategory', 'hideType', 'hideSemantics', 'forceSemantics'],
   components: {
     SemanticsPicker
   },
@@ -54,7 +53,8 @@ export default {
     return {
       types: Types,
       categoryInputId: '',
-      categoryAutocomplete: null
+      categoryAutocomplete: null,
+      nameErrorMessage: ''
     }
   },
   methods: {
@@ -70,11 +70,30 @@ export default {
           }
         }
       })
+    },
+    onNameInput (event) {
+      this.item.name = event.target.value
+      this.validateName(this.item.name)
+    },
+    validateName (name) {
+      let oldError = this.nameErrorMessage
+      if (!/^[A-Za-z0-9_]+$/.test(name)) {
+        this.nameErrorMessage = 'Required. Alphanumeric & underscores only'
+      } else if (this.items.some(item => item.name === name)) {
+        this.nameErrorMessage = 'An item with this name already exists'
+      } else {
+        this.nameErrorMessage = ''
+      }
+      if (oldError !== this.nameErrorMessage) this.$emit('valid', !this.nameErrorMessage)
     }
   },
   mounted () {
     if (!this.item) return
     if (!this.item.category) this.$set(this.item, 'category', '')
+    if (this.enableName) {
+      if (!this.items) this.items = []
+      this.validateName(this.item.name)
+    }
     const categoryControl = this.$refs.category
     if (!categoryControl || !categoryControl.$el) return
     const inputElement = this.$$(categoryControl.$el).find('input')
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/add-from-thing.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/add-from-thing.vue
@@ -60,7 +60,7 @@
           <div>Loading...</div>
         </f7-block>
         <div v-else-if="selectedThing.UID && selectedThingType.UID">
-          <item-form v-if="createEquipment" :item="newEquipmentItem" :enable-name="true" :hide-type="true" :force-semantics="true" />
+          <item-form v-if="createEquipment" :item="newEquipmentItem" :items="items" :enable-name="true" :hide-type="true" :force-semantics="true" />
           <f7-block-title>Channels</f7-block-title>
           <f7-block-footer class="padding-left padding-right">
             Check the channels you wish to create as new Point items.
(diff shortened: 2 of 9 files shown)


hubsif:
Hooray, I caught the 1000! 😜

hubsif:
Just uploaded a slight adjustment that fixes the Vue warning to not alter props.

ghys:
Not a bad idea, but I think it would cause all items to be loaded again when the item form is displayed in the Model page.

model.vue => item-details.vue => item-form.vue

https://github.com/openhab/openhab-webui/blob/0bcbd8c8ad81011b37a923f82235b25b3c913ea9/bundles/org.openhab.ui/web/src/components/model/item-details.vue#L11-L16

The model component already has a flat list of items at hand so it could pass it to item-details as a prop (currently it has only a `model` prop which is the hierarchy as displayed in the tree view, not a flat list), which would in turn pass it to item-form.
 
Storing the items in Vuex is naturally the best way to go, but the challenge would become to detect/invalidate the cache when that list has changed - apart from keeping an EventSource open, which I don't think is reasonable, I have no good solution for this. That's why for instance the item picker fetches the list of items every time - they might have changed at any time, not only from the current UI tab. See also #793.

hubsif:
> Not a bad idea, but I think it would cause all items to be loaded again when the item form is displayed in the Model page.

I see. 

> Storing the items in Vuex is naturally the best way to go

So, does that mean passing the items as props would be ok (for now)? (since Vuex and caching would be more of a big, "global" change)

ghys:
> So, does that mean passing the items as props would be ok (for now)? (since Vuex and caching would be more of a big, "global" change)

Yes, sorry. I was talking "in the long run". Even if the info is rather volatile because it has to be checked regularly for changes, for instance the Model page could tell the Vuex store's items list to update and the children components would reuse it. There could even be some logic on how to update, that would avoid fetching the entire response if it's not necessary, like a `ETag`/`If-None-Match` or `Last-Modified`/`If-Modified-Since` headers combination - provided support is added in the backend (https://www.logicbig.com/tutorials/java-ee-tutorial/jax-rs/request-context.html).

hubsif:
> Not a bad idea, but I think it would cause all items to be loaded again when the item form is displayed in the Model page.
> model.vue => item-details.vue => item-form.vue

I just committed an update.

hubsif:
I just discovered that this actually also fixes #906.

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
```diff
@@ -19,31 +19,36 @@
...
 
-      <!-- Option to create new item (if not supplied by prop) -->
-      <f7-col v-if="!item">
-        <f7-block-title>Item</f7-block-title>
-        <f7-list media-list>
-          <f7-list-item radio :checked="createItem === false" value="false" @change="createItem = false" title="Use an existing Item" name="item-creation-choice" />
-          <f7-list-item radio :checked="createItem === true" value="true" @change="createItem = true" title="Create a new Item" name="item-creation-choice" />
-        </f7-list>
-      </f7-col>
+      <div v-if="!item && !items" class="text-align-center">
```
ghys, line 22: There's another preloader block, lines 77-80, conditionally displayed when `!ready` so you should either remove this one or the other (presumably the other) and alter its v-if or control `ready` to cover all readiness conditions.

Right now the two can be seen simulateously for a few milliseconds which is not a good look. If the item-form were displayed you would even have a third one.

bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -70,6 +78,36 @@ export default {
...
+      } else {
+        this.nameErrorMessage = ''
+      }
+      if (oldError !== this.nameErrorMessage) this.$emit('valid', !this.nameErrorMessage)
+    }
+  },
+  created () {
+    if (this.enableName) {
+      if (!this.items) {
+        this.$oh.api.get('/rest/items').then((items) => {
```
ghys, line 101: Is there a case where the items are not supplied by the parent component in the "items" prop and they have to be loaded by this component?
If the answer is no then maybe we can assume they will always be supplied, and remove this logic and the changes to the templates to include the preloader on lines 3-7?

ghys (changes requested):
Thanks, mostly LGTM, have a few final remarks.
Also it would be good to expand the cautious "sharing" of these lists of reusable items in the Vuex store as discussed, since passing such data as props down the component tree is not ideal and not a Vue best practice. But for now, it's fine!

bundles/org.openhab.ui/web/src/components/item/item-form.vue
hubsif: > Is there a case where the items are not supplied by the parent component in the "items" prop and they have to be loaded by this component?

There actually seem to be many: add-from-thing, item-edit, channel-list, ... They all don't load items to forward.

ghys: So say you want to "add a thing as equipment" which has 200 channels, click on Select All to check all these channels causing the item forms to render, it's going to call `/rest/items` 200 times and keep separate versions of the whole list in the individual item form instances? This doesn't seem right.

hubsif:
Ok, so I updated this and checked it more thoroughly 😅. Changes:
- `item-form` does not load items anymore. If `enableName` is true (which could be renamed to sth. like `editMode` 😉), it expects the items to be provided as prop
- all usages of `item-form` with `enableName` set to true now provide items
- for above `channel-list` now accepts items as a prop, required in `multipleLinksMode` for item-form
- `add-from-thing` now loads all items, to provide them to `item-form` and `channel-list` (so items get only loaded once)
- `item-edit` now loads all items, to provide them to `item-form` and `item-picker`

As discussed before, having the items handled and cached in a Vuex-store would be much better. I'm planning on doing some tests and maybe open a community forum thread for discussions regarding this.
Until then, I think the work provided here is ok for now.

bundles/org.openhab.ui/web/src/pages/settings/model/add-from-thing.vue
```diff
@@ -281,7 +282,11 @@ export default {
               groupNames: (this.parent) ? [this.parent.item.name] : []
             }
           }
-          this.ready = true
+
+          this.$oh.api.get('/rest/items').then((items) => {
```
ghys, line 286: No need to reload the items on every thing change IMO. We can assume they will not change while the user is on that screen (not strictly true, there could be external changes in the meantime, but it's a fair assumption). Only do it if `this.items.length === 0` otherwise simply set `this.ready` to true.

There are a few reports of instances where getting the items takes seconds (#793) - admittedly a problem in itself - so displaying the spinning loader too long after picking another thing should be avoided.

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
hubsif: the second preloader is now only shown if the first isn't

PR #1011, opened 2021-04-20
[BasicUI] create dynamic manifest, fixes #8

### LINKED ISSUE #8: [basicui] wrong start_url in manifest linked from sitemap pages
All sitemap pages include a link to a manifest that happens to be the same regardless of the actual sitemap selected, this manifest pointing itself to the root of basicui, i.e. /basicui/app.

this prevents a particular sitemap to be set onto the homescreen of smartphones as a web app (e.g. by means of the common "add to homescreen" functionality of smartphones). Instead the web app defined in that way will run correctly in full screen but will start with the list of all available sitemaps published at /basicui/app, instead of the one selected.

This problem was also reported [here](https://community.openhab.org/t/web-app-manifest-manifest-json-does-not-include-sitemap-parameter-for-desktop-shortcut-in-basic-ui/35891) together with a workaround via using a default sitemap. However this in turns prevents setting several distinct sitemaps in that way.

Another workaround that works for me (on an iPhone at least) is to set up a reverse proxy that will drop every manifest.json request, and that in turn will drive the smartphone into using the active url of the sitemap page (which indeed contains the sitemap parameter).

I wonder if that could be corrected. e.g. by having a distinct manifest.json returned per sitemap, that would point to the actual root page of the sitemap, i.e. with start_url=app?sitemap=xxxx

DarkC35:
Fixes #8 

This PR proposes a solution to set the sitemap query parameter to the `start_url` in `manifest.json` dynamically. 

It uses a new `ManifestServlet` that injects the query string to the `start_url` when the manifest is requested. It also considers the manifest as a snippet now and therefor overloads the `getSnippet` method with an optional `snippetExt` parameter to allow different extensions than `.html` as snippets.

Finally the `manifest.json` was added to the `main_static` snippet. (That's funny because the `/basicui/app` path uses this snippet that didn't use the manifest, but the snippet used for sitemaps pahts like `/basicui/app?sitemap=xxx` uses a manifest with `/basicui/app` as the `start_url`.)

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
@@ -114,20 +114,21 @@ public abstract class AbstractWidgetRenderer implements WidgetRenderer {
     }
 
     /**
-     * This method provides the html snippet for a given elementType of the sitemap model.
+     * This method provides the snippet for a given elementType and snippetExt of the sitemap model.
      *
      * @param elementType the name of the model type (e.g. "Group" or "Switch")
-     * @return the html snippet to be used in the UI (including placeholders for variables)
+     * @param snippetExt the extension of the snippet file (e.g. ".html" or ".json")
+     * @return the snippet to be used in the UI (including placeholders for variables)
      * @throws RenderException if snippet could not be read
      */
-    protected synchronized String getSnippet(String elementType) throws RenderException {
+    protected synchronized String getSnippet(String elementType, String snippetExt) throws RenderException {
         String lowerTypeElementType = elementType.toLowerCase();
         String snippet = SNIPPET_CACHE.get(lowerTypeElementType);
         if (snippet != null) {
             return snippet;
         }
 
-        String snippetLocation = SNIPPET_LOCATION + lowerTypeElementType + SNIPPET_EXT;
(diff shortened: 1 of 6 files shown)


lolodomo (approved):
LGTM, thank you and sorry for the very late review.

lolodomo:
@ghys : as this PR was old, I tried to recompile it and the build failed for several UIs due to unresolved import org.apache.commons.io. How can we proceed ?

ghys:
@lolodomo just discovered that finally you can rebase a PR from GitHub directly, it was introduced last February and I missed it then: https://github.blog/changelog/2022-02-03-more-ways-to-keep-your-pull-request-branch-up-to-date/

Note that the button didn't show at first because the option mentioned in the link above ("Always suggest updating pull request branches") was disabled in the repo settings. Then I just chose "Update with rebase" and clicked on the button as shown.

This is a neat feature, it could be enabled on the other repos as well!

As for this PR, unfortunately the build still fails because of a SAT error ("initial contribution" needed in the header). If the original author is not around anymore you might have to fix it yourself.

DarkC35:
Thanks for the approval! Unfortunately I think I have messed up my commit history since I used the GH WebUI to update the copyright year which produced a mismatch in the sign-off for the "DCO action". Any quick tips how to fix this mismatch? (Preferably with the exact commands I need since I rarely use `git rebase`...)

DarkC35:
@lolodomo thank you and sorry for any inconveniences!

lolodomo:
Thank you @DarkC35 for your contribution.

PR #1013, opened 2021-04-20
fix small bug in widget mixin

hubsif:
fix a small bug in widget-mixin, introduced by #979, ocurring when there's no `context` available (which happened regularly for me with `oh-icon`)


ghys (approved):
Thanks!

PR #1020, opened 2021-04-25
Rework metadata UI for new GA binding - Part 2

eikowagenknecht:
Sorry, I missed some things in the first PR you merged today.
- Added some more description text
- Moved some text from title to description where everything was crammed in the title
- Made URLs work as links
- Split the list in two option groups (like the Alexa metadata UI does, code is mainly from there):

image


ghys (approved):
Thanks! Code looks good, I don't use GA so cannot comment on the functionality but will trust you on that.

PR #1021, opened 2021-04-25
Fixes "Other Apps" menu not opening when the developer sidebar is active

eikowagenknecht:
Since there can't be two sidebars active at a time, it opens above the developer sidebar, but I think that's okay.

Before:
image

After:
image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -115,15 +115,16 @@
     </f7-panel>
 
     <!-- Right Panel -->
-    <f7-panel right reveal theme-dark v-if="ready">
+    <f7-panel v-if="ready" right theme-dark>
       <panel-right />
-    <!-- <f7-view url="/panel-right/"></f7-view> -->
     </f7-panel>
 
-    <f7-panel v-if="showDeveloperSidebar" right :visible-breakpoint="1280" resizable>
+    <!-- Developer Sidebar -->
+    <f7-panel v-if="showDeveloperSidebar" right resizable :visible-breakpoint="1280">
       <developer-sidebar />
     </f7-panel>
 
+    <!-- Main Content -->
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -115,15 +115,16 @@
     </f7-panel>
 
     <!-- Right Panel -->
-    <f7-panel right reveal theme-dark v-if="ready">
+    <f7-panel v-if="ready" right theme-dark>
```
ghys, line 118:
```suggestion
    <f7-panel v-if="ready" :reveal="!showDeveloperSidebar" right theme-dark>
```
What about condionally activating the reveal prop so the other apps sidebar pushes the content to the left as it used to, unless the developer sidebar is shown?

ghys (changes requested):

bundles/org.openhab.ui/web/src/components/app.vue
eikowagenknecht: I don't like the reveal effect too much anyways, so I just removed it. But if you want to keep it in situations where it's possible, we can of course do that.

Your suggested change doesn't work for me, though. It still always uses reveal when I try that. Perhaps your syntax is a little bit off? I'm just now learning vue and encountered this bug while trying to figure out how the Web UI works on the inside. So I'm not much help there (yet) ;-)

ghys: > I don't like the reveal effect too much anyways, so I just removed it.

Matter of taste, I guess. I prefer it for this pane as it appears more like a "drawer" below the home page to, well, "reveal" the other apps. But you're right that it doesn't work well when another right-hand pane is already displayed. It doesn't appear to be easy to control the effect after it's been initialized either.
I'd go as far as questioning whether we could hide the "other apps" button when the developer sidebar is displayed, since that page is presumably not used that much as it has limited functionality anyway, and if you need it, then you can temporarily hide the developer sidebar to show the button again.

PR #1027, opened 2021-04-28
Use vector version of openHAB logo

ecdye:
Signed-off-by: Ethan Dye <mrtops03@gmail.com>


hubsif:
Generally, a nice improvement 👍 
Though it doesn't work for me, e.g. it doesn't show the logo on the top left anymore. First guess is that the svg doesn't have a fixed size and thus doesn't get scaled properly. Did it work for you?

ecdye:
It does display for me but not fully the way it should, I'll work on playing with the settings for the css and stuff and get it fixed for all devices.

hubsif:
Hi @ecdye,
Looking good for the main site for me now!

I think it still needs some finetuning, though, e.g. the logo is huge on the HABot site.
Also, some sites are still using the non-vector "icons" (e.g. the HABot chat, about site, etc.). In my opinion we should use your good work as an opportunity to replace them all. Would it be possible for you to do that?

Note: Just making suggestions here, as I'm not in the position to make final decisions, that's up to @ghys.

ecdye:
Well, I actually don't use HABot so I think Ill actually save that for another day when I can properly test the UI and make sure it looks right. Why don't we just start with the mainUI for now.

bundles/org.openhab.ui/web/src/res/img/openhab-logo.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 170 37">
+	<style>
+		@media (prefers-color-scheme: dark) {
```
ghys, line 4: This isn't quite right because the user can override the preferred color scheme in the help & about page.
When the dark mode is active though, the root html element will have the `theme-dark` CSS class applied.

ghys (changes requested):
Nice initiative! I'll yest it next week but this should make the logo crisper on retina displays and the like 👍
Why remove the icons which are referenced in the web app manifest though? https://github.com/openhab/openhab-webui/pull/1027/files#diff-126354c0cac3eaa37186ee3abdc3db5810ca6cb4e5f71061f1f87cf77a15533b

bundles/org.openhab.ui/web/src/res/img/openhab-logo.svg
hubsif: Actually, I’d skip that and stick to the „coloured“ version. I think it looks much better than the white in any case, including the favicon.
E92A9593-0BA1-4AA4-904B-ECF9F54A059C

ecdye: I'm okay with either. Which would you prefer @ghys?

ghys: I made a white only image deliberately because IMO the contrast ratio of the gray circle/"HAB" against a dark the background is poor, especially compared to the orange parts which stand out as a result.
As for the favicon, in the png version the logo is inside a white circle.

I don't think it looks horrible either way but I'd stick to the white version in dark mode.

ecdye:
Support for vector icons os good on all except Safari according to my tests.

ecdye:
> Why remove the icons which are referenced in the web app manifest though? https://github.com/openhab/openhab-webui/pull/1027/files#diff-126354c0cac3eaa37186ee3abdc3db5810ca6cb4e5f71061f1f87cf77a15533b

I didn't remove them, I just minified and updated them to be consistent with the other icons.

bundles/org.openhab.ui/web/src/res/img/openhab-logo.svg
ecdye: I'm not super familiar with this web system, any tips on the right css for the override on the svgs icons?

ghys:
My bad, I was confused by the red sizes and thought they were gone 🤦 
However as stated above the white background circle was included on purpose to improve contrast, and you don't want a fully transparent logo because it doesn't stand out enough. So it needs to stay.

bundles/org.openhab.ui/web/src/res/img/openhab-logo.svg
hubsif: I don't think there's a straightforward solution that makes it possible to use inline CSS styling like the current `prefers-color-scheme: dark`.
I think for this to work the SVG has to be loaded inline, which would require using another webpack loader.
So, it's probably best to stick to two images.

ecdye:
@ghys Any chance for a review, I think I have addressed all your concerns.

ghys:
Thanks, almost all :)

The favicon should still have the white background as before for the same reasons - you won't see the gray circle (or barely) if the background is also a shade a gray:

Before/after:

image

The version on the left still looks better IMO.

And - more anecdotal - in the about page it's also fine to use a version of the logo with unaltered colors and a white background, instead of a monochrome version (and keep it pushed to the right to conserve vertical space):

Before
image

After
image

The logo on the top-left does look good and crisp though, nice! 👍

ecdye:
@ghys I think I got them all, does this look good?

ghys (approved):
Absolutely, thank you very much!

mstormi:
Thanks for this!
Ethan is there any chance for you to add another improvement and make the logo a *configurable* file ?
So that a user can put up a svg or png file of his own ?

ecdye:
Not very easily to my knowledge that would be a better project for the webui maintainers.

Issue #903, opened 2021-02-10
removing invalid link fails silently for links defined in .items file


eikowagenknecht:
## The problem

I am currently switching to defining my links in the .items file and encountered the following issue:

I have an invalid link (typo or whatever) and I see that link in the item details page. Clicking on it I'm asked if I want to remove it. Being curious what removing can to for an item that is defined in a file, I clicked on it and... 

![image](https://user-images.githubusercontent.com/1475672/107492463-c636b680-6b8c-11eb-8dfe-9a1d9e1edda3.png)

... nothing happens. At all. No error message telling me I can't do this, no log entry etc.

## Expected behavior

It should be displayed to the user why this isn't working (e.g. "You can't remove a link you defined in a file").

## Steps to reproduce

1.  Create broken link in file
2. Try to remove it with MainUI

## Your environment

OH 3.0.1
## Browser console

![image](https://user-images.githubusercontent.com/1475672/107492807-24639980-6b8d-11eb-9c6d-9e6ff8d70461.png)

So it *is* known to the frontend, just not showing it to the user. Should make the fix quite easy.

hubsif:
I started looking into this and thought it might be good to disable link editing at all for file-defined items, when I found that items defined in a file can have both, links created by file and links created via UI.
Unfortunately, in contrast to items, which return an attribute `editable` via REST, links don't share this information, so it cannot be determined if a link has actually been created by file or not 😕.

Actually, I think the best solution would be to add this to the core and then have non-editable links locked in the UI (indicated by a lock icon, just as with items).

Now, simply showing an error message in case the invalid link could not be deleted (just as with "unlink" in the link-edit screen) would be the quickest solution. Though, I think it's more like a workaround and changing the core would be much cleaner.

So, @ghys, in your experience, would it make sense to open a ticket to add this to the core or do you think it would be too complex to add this to the core (shortly) and we should go for the workaround instead (or do both 😉)?

ghys:
@hubsif this was already discussed in #693 and https://github.com/openhab/openhab-webui/issues/413#issuecomment-711411015. Adding the editable flag to the core shouldn't be too hard but nobody has tackled that problem yet.

Originally I thought it was acceptable to arbitrarily limit UI links to items also defined in the UI (which have an `editable` flag), but since it was not well accepted I reversed stance and allowed them back again. So I'm fine with the workaround only (display an error message when the link is not editable after attempting to delete it).

hubsif:
[...]

Issue #693, opened 2020-12-23
[MainUI] [OH3 Final] Items defined in .items file can be linked but not unlinked


eikowagenknecht:
## The problem

When you create your items with .items files and not the MainUI you can still link them in MainUI when you start in the "Model" view (see https://github.com/openhab/openhab-webui/issues/685). But you can not unlink them anymore in the items view (or anywhere else?).

Item created in MainUI (unlinking options are shown):
![image](https://user-images.githubusercontent.com/1475672/103031880-000d1880-455f-11eb-8d33-dc03f2b91379.png)

Item created in .items file (buttons are missing):
![image](https://user-images.githubusercontent.com/1475672/103031947-25018b80-455f-11eb-88a1-050cc803fd18.png)

## Expected behavior

Items that can be linked shouls also be unlinkable.

## Steps to reproduce

1. Create item in .items file
2. Go to Model view and Link it to a Point from there.
3. Try to unlink it

ghys:
This is a side effect of my decision to discourage item linking in the UI if the item is created in a file, discussed in https://github.com/openhab/openhab-webui/issues/413#issuecomment-719923775.
If you choose to use files for your items, you should link them there as well - `{channel="..."}`.
This is an arbitrary limitation, not sure it was a wise one after all, but it's either this or always show the buttons and let the user face an error if the link was made in the file, until we have a way of figuring out if it's actually editable by the UI.

eikowagenknecht:
I think this really is **not** a good limitation, because now items defined in .items files **can** be linked (see #685 where I describe how to do it) and never unlinked again (maybe it is possible in some other dialog as the linking is also only possible in some dialogs not others).

In #413 you say "I don't think there are many cases of links being created in the UI while the item is not."
I'm not sure that is true. I remember that when I started using openHAB some years ago that the guides I used suggested this practice and I sticked to it. Maybe today it's not encouraged as much, but as long as openHAB has this possibility I think the main user interface (mainUI) should also support it.

I understand that it is a bit annoying for users who  bind items in .items files when there is an "unlink" button in the mainUI. But I would prefer an error "Oops. Seems like you defined that link in a file. Please remove it there." over the current state.

The best solution would of course be to be able to determine from the UI side upfront where a link was defined and only provide edit / unlink functionality for the ones defined in mainUI. I guess this would need some API enhancement?

Issue #413, opened 2020-10-18
[MainUI] Channel links have to be locked when defined in conf files


lolodomo:
For the channel links created in conf files, the UI should not allow to remove them.
![image](https://user-images.githubusercontent.com/10186704/96367417-fece0280-114d-11eb-9151-7352bf3406c7.png)

A lock icon should be displayed somewhere and the 2 buttons "Unlink & Remove item" and "Unlink ionly" should be greyed and not clickable.

Additionally, for a link defined by the UI. if the item is defined in a conf file, the button "Unlink & Remove item" should be greyed and not clickable. Only the button "Unlink only" should be usable in this case.

ghys:
I think I had a look at the time, and links don't have an `editable` flag, so we can't know whether they come from textual configuration and therefore can't be removed or not. Maybe that would be relatively to figure out and implement in core.

> Additionally, for a link defined by the UI. if the item is defined in a conf file, the button "Unlink & Remove item" should be greyed and not clickable. Only the button "Unlink only" should be usable in this case.

Fair.

lolodomo:
@kaikreuzer  @wborn @cweitkamp : would it be possible to update the core to identify the channel links that are managed or not?

cweitkamp:
After a quick look I would say that it is possible. We have to Introduce an `EnrichedItemChannelLinkDTO` extending the `ItemChannelLinkDTO` by the requested `editable` property. If it is editable could be validated against the `ManagedItemChannelLinkProvider`. Straight forward. Similar to what we are doing for Things or Items.

ghys:
I removed the buttons when the item is not editable in #473, and filtered out non-editable items when reusing an existing item when creating a link from the UI. I think it's safe to assume that if an item is going to be defined with textual configuration, so is the link.

lolodomo:
You can have your items defined in config files but your things managed through the UI. In this case, it is important to be able to link to an item defined in a config file.

ghys:
Yes, but if you're defining your items in config files, chances are you're going to go ahead and define the link with
`{ channel="uid" }`
in the config file anyway. I don't think there are many cases of links being created in the UI while the item is not.

lolodomo:
Yes, that is the most probable but why adding this constraint ?

ghys:
Since the editable status of links cannot be determined yet, I figured it would be good to prevent these kinds of situations (links to non-editable items where we don't actually know if the link itself can be deleted) to happen in the first place?

PR #1029, opened 2021-04-29
Implement link editable flag

hubsif:
Related to #903, #693 and #413

Make use of the new `editable` flag for links:
- If a link is not editable, instead of "Save" a red lock icon is shown, the link's card footer is hidden (no "Unlink" buttons shown) and the Profile is read only.
- At "link-details" a lock is shown next to non-editable links, and the click action is disabled for non-editable, invalid links.


hubsif:
Ah, and the related server-side PR [openhab-core#2318](https://github.com/openhab/openhab-core/pull/2318) has been added to the milestone "3.1". Of course this change should not be included before then, and maybe also be added to the 3.1 milestone.

ghys (approved):
That's another A-OK from me!

PR #1031, opened 2021-05-01
Use SAT 0.11.1

wborn:
Signed-off-by: Wouter Born <github@maindrain.net>

### DIFF
--- a/bundles/org.openhab.ui.habot/src/main/resources/OH-INF/config/config.xml
+++ b/bundles/org.openhab.ui.habot/src/main/resources/OH-INF/config/config.xml
@@ -4,15 +4,6 @@
 	xmlns:config-description="https://openhab.org/schemas/config-description/v1.0.0"
 	xsi:schemaLocation="https://openhab.org/schemas/config-description/v1.0.0 https://openhab.org/schemas/config-description-1.0.0.xsd">
 
-	<config-description uri="ui:habot">
-		<parameter name="useGzipCompression" type="boolean">
-			<label>Use Compression</label>
-			<description>Serves the compressed web app static files with Content-Encoding: gzip. This decreases significantly the
-				download size of the app. This is NOT supported by openHAB Cloud/myopenhab.org!</description>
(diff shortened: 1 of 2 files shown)


wborn:
WDYT about these 2 changed lines @openhab/webui-maintainers?

bundles/org.openhab.ui.habot/src/main/resources/OH-INF/config/config.xml
```diff
@@ -6,7 +6,7 @@
 
 	<config-description uri="ui:habot">
 		<parameter name="useGzipCompression" type="boolean">
-			<label>Use compressed web app resources</label>
+			<label>Use Compression</label>
```
ghys, line 9: Why this change? It's obscure enough already and this makes it even more vague :)
For that matter this parameter was more of an experiment and never worked too well anyways (for instance AFAIK it still doesn't work with myopenhab.org/OH Cloud). Perhaps we should just remove the config description and simply keep it as an undocumented feature - that would remove the HABot entry from the configurable services.

wborn: It's because the new SAT version checks the label length and these 32 chars exceed the 25 chars limit (https://github.com/openhab/static-code-analysis/pull/360) so without fixing or supressing it the build would fail. Since this was the only change required, I reduced the length as it has a good description. I can also remove it if that makes you even more happy. :slightly_smiling_face:

ghys: Ah, that explains it! 👍  I really thought it was an accidental, unrelated change :)
If you don't maybe it's indeed time to remove ui:habot config-description. This parameter was confusing people and if it's ever needed it can be set via the console or other means.

wborn: OK it's gone now. :-)

ghys (approved):

PR #1036, opened 2021-05-03
Extend visible parameter evaluation, fixes #880

### LINKED ISSUE #880: edit metadata not transfering entries properly into the Code
## When you edit the metadata in the UI for visibility options, the code is not properly transferred into the code so the entry is not valid.

image

image

## Expected behavior

i believe it should be translated into 

image

## Steps to reproduce

1. 

## Your environment

```yaml

```

## Browser console

```txt

```

hubsif:
Interpret the visibility parameter set to the string "false" as boolean.


ghys (approved):
Thanks!

Issue #885, opened 2021-02-06
Consider item type for the selected profile when creating links


kaikreuzer:
When creating a link, the user can choose a profile to use.
Profiles have a set of supported item types. The "Timestamp on update" for example can only be linked to a `DateType` item.

<img width="742" alt="Screenshot 2021-02-07 at 00 54 50" src="https://user-images.githubusercontent.com/3244965/107132408-1be63700-68df-11eb-884b-f32219a08d74.png">

The UI does not consider this - when "Create a new Item" is selected, it will invariably create an item of the type that is defined for the channel, but not the type that is defined by the profile. This leads to an item being created with an invalid link:

<img width="736" alt="Screenshot 2021-02-07 at 00 57 04" src="https://user-images.githubusercontent.com/3244965/107132440-6cf62b00-68df-11eb-83ba-0a8199ef8731.png">

Likewise, when selecting an existing item ("Select item to link"), the list shows all available items, while it should only offer those of supported types.

ghys:
👀 

https://github.com/openhab/openhab-webui/blob/c3bafcf50b6cebdcb08b9713fa691ca30f610e39/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue#L259-L262

I had enough problems already determining the compatible item types (e.g. a Color channel can be linked to a Dimmer or Switch item etc.), started getting invalid 'The channel and item type are not compatible' messages, and then I noticed the profiles would make this even more complex so I kind of gave up 😅

kaikreuzer:
Maybe you find a quiet moment to revisit this - it'll imho be a huge simplification for the user if he only sees a filtered list, especially for large setups.

Rossko57:
As a sort-of related enhancement, the MainUI display of a channel and its linked Items seems to hide the _channel_ type.
It would be a useful enhancement to display this in channel details, as it is for Item types.
Example display in this post, and difficulty arising from missing info -
https://community.openhab.org/t/struggling-with-mqtt/118196

hubsif:
Hi!
I thought about working on this issue, starting by trying to understand the "base" behind this. Unfortunately, I didn't get further than general item linking:

> I had enough problems already determining the compatible item types (e.g. a Color channel can be linked to a Dimmer or Switch item etc.),

[...]

Issue #884, opened 2021-02-06
Profile list shows "Default" and "(no profile")


kaikreuzer:
When creating an item link, a selection like the following is shown:

<img width="743" alt="Screenshot 2021-02-07 at 00 38 41" src="https://user-images.githubusercontent.com/3244965/107132155-dc1e5000-68dc-11eb-9623-0e5a503f67f3.png">

Note that the [documentation](https://www.openhab.org/docs/configuration/items.html#profiles) says that the "Default" profile is used if no profile is defined - both these options are therefore the same for the user.

I'd suggest to remove the "(No Profile)" option and always show the "Default" instead at first place (and being selected if no profile is defined).
Additionally, since "Follow" is the counterpart that is always available, I'd fix its position to be the second entry in the list.

PR #1043, opened 2021-05-06
Link and Profile related fixes

hubsif:
This PR addresses #884 and #885. Changes:

- drop "no profile" and use the "Default" profile instead (link-add and link-edit)
- filter the item list for items compatible to the channel's type (link-add)
- only show profiles compatible to the current item (link-add, link-edit)

Note: This PR is based on the branch used for PR #1000, to avoid blocking after that has been merged.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
@@ -67,6 +67,7 @@
       <f7-col v-if="profileTypeConfiguration != null">
         <f7-block-title>Profile Configuration</f7-block-title>
         <config-sheet ref="profileConfiguration"
+                      :key="'profileTypeConfiguration-' + currentProfileType.uid"
                       :parameter-groups="profileTypeConfiguration.parameterGroups"
                       :parameters="profileTypeConfiguration.parameters"
                       :configuration="link.configuration"
@@ -125,7 +126,7 @@ export default {
       this.$oh.api.get('/rest/profile-types?channelTypeUID=' + this.channel.channelTypeUID + '&itemType=' + itemType).then((data) => {
         this.profileTypes = data
         this.profileTypes.unshift(data.splice(data.findIndex(p => p.uid === 'system:default'), 1)[0]) // move default to be first
-          .filter(p => !p.supportedItemTypes.length || p.supportedItemTypes.includes(this.item.type)) // only show compatible profile types
+        this.profileTypes = this.profileTypes.filter(p => !p.supportedItemTypes.length || p.supportedItemTypes.includes(this.item.type)) // only show compatible profile types
 
         this.$oh.api.get('/rest/links/' + itemName + '/' + channelUID).then((data2) => {
           this.link = data2
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
@@ -56,9 +56,8 @@
             <div>Loading...</div>
           </f7-block>
           <f7-list v-else>
-            <f7-list-item radio :checked="!currentProfileType" value="" @change="onProfileTypeChange()" title="(No Profile)" name="profile-type" :disabled="!link.editable" />
             <f7-list-item radio v-for="profileType in profileTypes"
-                          :checked="currentProfileType && profileType.uid === currentProfileType.uid"
+                          :checked="!currentProfileType && profileType.uid === 'system:default' || currentProfileType && profileType.uid === currentProfileType.uid"
                           :disabled="!link.editable"
                           @change="onProfileTypeChange(profileType.uid)"
                           :key="profileType.uid" :title="profileType.label" name="profile-type" />
@@ -123,9 +122,9 @@ export default {
       const itemName = this.item.name
       const itemType = this.item.type
       const channelUID = this.channel.uid.replace('#', '%23')
-      const getProfileTypes = this.$oh.api.get('/rest/profile-types?channelTypeUID=' + this.channel.channelTypeUID + '&itemType=' + itemType)
-      getProfileTypes.then((data) => {
+      this.$oh.api.get('/rest/profile-types?channelTypeUID=' + this.channel.channelTypeUID + '&itemType=' + itemType).then((data) => {
         this.profileTypes = data
+        this.profileTypes.unshift(data.splice(data.findIndex(p => p.uid === 'system:default'), 1)[0]) // move default to be first
 
         this.$oh.api.get('/rest/links/' + itemName + '/' + channelUID).then((data2) => {
           this.link = data2
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
@@ -125,6 +125,7 @@ export default {
       this.$oh.api.get('/rest/profile-types?channelTypeUID=' + this.channel.channelTypeUID + '&itemType=' + itemType).then((data) => {
         this.profileTypes = data
         this.profileTypes.unshift(data.splice(data.findIndex(p => p.uid === 'system:default'), 1)[0]) // move default to be first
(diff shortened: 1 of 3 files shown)


ghys:
Thanks! Nice additions at first sight, could you rebase now that most conflicting PRs have been merged?

hubsif:
> could you rebase now that most conflicting PRs have been merged?

done.

ghys:
> only show profiles compatible to the current item (**link-add**, link-edit)

Here's an issue I have with this change in one particular scenario: when you create a link and choose the "new item" option, the item type is set by default to the type of the channel and the "default" profile selected, which is good; however, the drawback is that it "hides" some useful profiles until the user changes the type of the item they're about to create.
For instance, they might not be aware of the existence of the "timestamp on update"/"timestamp on change" unless they change the item's type to DateTime.

So it's not a flaw in itself - just a matter of UX. You have to know that changing the item's type might unlock additional profiles to you.

The commented out message in red in the page that you removed:
```
      <!-- <f7-block v-if="!itemTypeCompatible()" class="text-color-red">
        The channel and the item type are not compatible.
      </f7-block> -->
```
was supposed to simply inform the user they're about to create a channel/item/combination that is not valid without having to discover what's possible by tinkering around, by leaving all the options visible.

So I'm wondering if the best course in this case only (creating both the link and the item) would be to reinstate this message, leave the list of profiles unfiltered, and maybe change the new item's type to the compatible one when another profile is chosen?

When the item already exists, of course, filtering the list makes sense.

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
```diff
@@ -123,9 +122,10 @@ export default {
       const itemName = this.item.name
       const itemType = this.item.type
       const channelUID = this.channel.uid.replace('#', '%23')
-      const getProfileTypes = this.$oh.api.get('/rest/profile-types?channelTypeUID=' + this.channel.channelTypeUID + '&itemType=' + itemType)
-      getProfileTypes.then((data) => {
+      this.$oh.api.get('/rest/profile-types?channelTypeUID=' + this.channel.channelTypeUID + '&itemType=' + itemType).then((data) => {
         this.profileTypes = data
+        this.profileTypes.unshift(data.splice(data.findIndex(p => p.uid === 'system:default'), 1)[0]) // move default to be first
+          .filter(p => !p.supportedItemTypes.length || p.supportedItemTypes.includes(this.item.type)) // only show compatible profile types
```
ghys, line 128: I get an error here:

_Uncaught (in promise) TypeError: this.profileTypes.unshift(...).filter is not a function
    at eval (link-edit.vue?c1b6:128)_

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift

> The unshift() method adds one or more elements to the beginning of an array and returns **the new length of the array.**

hubsif: oh wow! not sure how that made it into the commit, possibly happened during rebasing. Will correct.

hubsif:
> So I'm wondering if the best course in this case only (creating both the link and the item) would be to reinstate this message, leave the list of profiles unfiltered, and maybe change the new item's type to the compatible one when another profile is chosen?

For me the flow is: Choose an item, then select the profile (from those possible for that item). So, I'd be good with it. The user doesn't get any hints on which item types would be available for certain profiles anyway.

Suggestion: Simply add some text like: "Profiles available for this item type ({itemtype}):" before the profile list?

ghys:
> Suggestion: Simply add some text like: "Profiles available for this item type ({itemtype}):" before the profile list?

or disable/make read-only those in the list which are not compatible. The properties exist: https://v5.framework7.io/vue/radio.html

ghys:
```vue
          <f7-list-item radio v-for="profileType in profileTypes"
                        :checked="!currentProfileType && profileType.uid === 'system:default' || currentProfileType && profileType.uid === currentProfileType.uid"
                        :disabled="compatibleProfileTypes.indexOf(profileType) < 0"
                        :readonly="compatibleProfileTypes.indexOf(profileType) < 0"
                        :style="{ opacity: (compatibleProfileTypes.indexOf(profileType) < 0) ? 0.5 : 1 }"
```
Ugly code & styling - there isn't any styling for the disabled state amazingly...

image

image

hubsif:
> Ugly code & styling

By "styling", do you mean the "code for styling" or the displayed result? I think the displayed result is quite ok.

Right, the code doesn't look too nice, but if it's required 🤷?
I think it can be simplified a little, though:

```js
          <f7-list-item radio v-for="profileType in profileTypes"
                        :checked="!currentProfileType && profileType.uid === 'system:default' || currentProfileType && profileType.uid === currentProfileType.uid"
                        :disabled="compatibleProfileTypes.includes(profileType)"
                        :style="{ opacity: compatibleProfileTypes.includes(profileType) ? 0.5 : 1 }"
```

Like this it looks quite acceptable to me... (I dropped the "readonly" since I couldn't figure out what it's actually for and it didn't make any difference in the resulting html)

ghys (approved):
Thanks, I finished the PR while I had the branch checked out :)
Just fixed the error and added some keys to the config sheet in addition to the disabled profile list.

cweitkamp:
I am not able to use the `offset` (Versatz) Profile on quantitized Number Items anymore. Might this PR be related?

grafik

cweitkamp:
Additionally there is an JavaScript error in the developer console when trying to access an ItemChannelLink which uses an `offset` Profile.

grafik

hubsif:
Hi Christoph,
Seems a bug has been introduced with this change. Can you please share the server response for the "profile-types" call from the network tab for this?

cweitkamp:
Sure. Please let me know if you need more information.

```
[
  {
    "uid": "transform:MAP",
    "label": "MAP",
    "kind": "STATE",
    "supportedItemTypes": []
  },
  {
    "uid": "transform:SCALE",
    "label": "SCALE",
    "kind": "STATE",
    "supportedItemTypes": []
  },
  {
    "uid": "system:timestamp-update",
    "label": "Zeitstempel bei Aktualisierung",
    "kind": "STATE",
    "supportedItemTypes": [
      "DateTime"
    ]
  },
  {
    "uid": "system:range",
    "label": "Bereich",
    "kind": "STATE",
    "supportedItemTypes": [
      "Switch"
    ]
  },
  {
    "uid": "system:default",
    "label": "Standard",
    "kind": "STATE",
    "supportedItemTypes": []
  },
  {
    "uid": "system:hysteresis",
    "label": "Hysterese",
    "kind": "STATE",
    "supportedItemTypes": [
      "Switch"
    ]
  },
  {
    "uid": "system:timestamp-change",
    "label": "Zeitstempel bei Änderung",
    "kind": "STATE",
    "supportedItemTypes": [
      "DateTime"
    ]
  },
  {
    "uid": "system:follow",
    "label": "Folgen",
    "kind": "STATE",
    "supportedItemTypes": []
  },
  {
    "uid": "system:offset",
    "label": "Versatz",
    "kind": "STATE",
    "supportedItemTypes": [
      "Number"
    ]
  },
  {
    "uid": "basic-profiles:invert",
    "label": "Invertieren / Negieren",
    "kind": "STATE",
    "supportedItemTypes": [
      "Player",
      "Number",
      "Rollershutter",
      "Switch",
      "Dimmer",
      "Contact"
    ]
  },
  {
    "uid": "basic-profiles:battery-low",
    "label": "Niedriger Batteriestatus",
    "kind": "STATE",
    "supportedItemTypes": [
      "Switch"
    ]
  },
  {
    "uid": "basic-profiles:to-percent",
    "label": "In Prozent",
    "kind": "STATE",
    "supportedItemTypes": [
      "Rollershutter",
      "Dimmer",
      "Color"
    ]
  },
  {
    "uid": "basic-profiles:round",
    "label": "Runden",
    "kind": "STATE",
    "supportedItemTypes": [
      "Number"
    ]
  }
]
```

ghys:
It appears that the profile type defines a `supportedItemTypes` of "Number" which is perhaps checked a bit too strictly - it excludes the `Number:*` quantity types.

hubsif:
I thought I had taken that into account, but seems I only did for item filtering. Perhaps I should take a break...🙄
I just opened PR #1054 with a fix.

PR #1045, opened 2021-05-09
Use status instead of message for error responses to detect Unauthorized requests

### LINKED ISSUE #1018: Allow to disable Service Worker
## The problem

There is a bug in chrome and other browsers, related to a problem in the Web Fetch API in combination with Service Workers and Basic Authentication, see https://github.com/whatwg/fetch/issues/1132.
As a result it is not possible to use openHAB main ui behind a HTTPS reverse proxy with Basic Auth. On first page load it works. However one day later, the page is blank, because the Basic Auth prompt is not triggered by the Service Worker and therefore the access to the rest api is denied.

As you can see in the screenshot, the service worker caches the ui components, but the non-cached requests fail, because they don't trigger the Basic Auth prompt. 
image

## Your suggestion
This is a bug in Browser, not openHAB ui. However, while this bug is present as a workaround, openHAB should allow to disable the service worker.

## Your environment

```yaml
runtimeInfo:
  version: 3.0.2
  buildString: Release Build
locale: de-DE
systemInfo:
  configFolder: /etc/openhab
  logFolder: /var/log/openhab
  javaVersion: 11.0.9.1
  javaVendor: Raspbian
  osName: Linux
  osVersion: 5.10.17-v8+
  osArchitecture: arm
  availableProcessors: 4
  freeMemory: 59561608
  totalMemory: 144703488
bindings:
  - mqtt
  - openweathermap
  - tr064
  - upnpcontrol
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
    pixelRatio: 1
    prefersColorScheme: dark
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 16
    language: de-DE
    languages:
      - de-DE
      - de
      - en-US
      - en
    onLine: true
    platform: Win32
  screen:
    width: 2560
    height: 1440
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: dark
    filled: true
    bars: light
    homeNavbar: default
    homeBackground: default
  userAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML,
    like Gecko) Chrome/90.0.4430.85 Safari/537.36
timestamp: 2021-04-25T10:38:46.293Z

```

## Additional information

I use Apache Web Server as Reverse Proxy with TLS termination and Basic Auth.

Legion2:
With HTTP2 there is no status code  reason-phrase and even in HTTP1 it is optional. Therefore the response status codes must be used to detect unauthorized request errors.

fix #1018

@ghys based on your comment I assume the second argument is the integer status code, I did not test this code, but I think it should work.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -376,8 +376,8 @@ export default {
       const useCredentialsPromise = (useCredentials) ? this.setBasicCredentials() : Promise.resolve()
       return useCredentialsPromise
         .then(() => { return this.$oh.api.get('/rest/') })
-        .catch((err, status) => {
-          if (status === 401) {
+        .catch((err) => {
+          if (err === 'Unauthorized' || err === 401) {
             if (!useCredentials) {
               // try again with credentials
               this.loadData(true)
@@ -393,8 +393,8 @@ export default {
                   this.$oh.api.get('/rest/').then((rootResponse) => {
                     this.storeBasicCredentials()
                     this.loadData()
-                  }).catch((err, status) => {
-                    if (status === 401) {
+                  }).catch((err) => {
+                    if (err === 'Unauthorized' || err === 401) {
                       this.clearBasicCredentials()
                       this.loadData()
                       return Promise.reject()
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -376,8 +376,8 @@ export default {
       const useCredentialsPromise = (useCredentials) ? this.setBasicCredentials() : Promise.resolve()
       return useCredentialsPromise
         .then(() => { return this.$oh.api.get('/rest/') })
-        .catch((err) => {
-          if (err === 'Unauthorized') {
+        .catch((err, status) => {
+          if (status === 401) {
             if (!useCredentials) {
               // try again with credentials
               this.loadData(true)
@@ -393,8 +393,8 @@ export default {
                   this.$oh.api.get('/rest/').then((rootResponse) => {
                     this.storeBasicCredentials()
                     this.loadData()
-                  }).catch((err) => {
-                    if (err === 'Unauthorized') {
+                  }).catch((err, status) => {
+                    if (status === 401) {
                       this.clearBasicCredentials()
                       this.loadData()
                       return Promise.reject()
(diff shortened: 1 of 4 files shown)


ghys:
> @ghys based on your comment I assume the second argument is the integer status code, I did not test this code, but I think it should work.

I'm absolutely not 100% sure about this, it could be an object with the code inside. Will have to check.

ghys:
Well, actually...
[`Promise.reject()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) (and [`Promise.resolve()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) for that matter) can only pass one argument (the reason resp. the value) so this code:

https://github.com/openhab/openhab-webui/blob/19395dfacac21b3b8bec92c32c2444efa5c5434f/bundles/org.openhab.ui/web/src/js/openhab/api.js#L6-L9

is completely wrong 😠. In other terms `status` and `xhr` will always be undefined.
Simply rejecting with `err` isn't ideal because there are lots of instances in catch clauses where the reason is assumed to be the HTTP status message ('Not found', 'Unauthorized', 'Conflict' etc.) and puts it in a toast notification.

So I guess an ugly workaround would be to change the rejection above to `reject(err.message || err.status)` and then check on both the status and the message (`if (err === 'Unauthorized' || err === 401) {`) in the catch clauses, so that it fallbacks to status code when there's no message; that way there's no code change needed elsewhere but in your case where there's no HTTP status message `err` would still contain `401`.

Legion2:
@ghys I updated the code according to your comments.

ghys:
Thank you. Shame we have to deal with it like this but it'll do :)
One last thing, may I ask to add `|| err === 401` similarly in these two places too:

https://github.com/openhab/openhab-webui/blob/19395dfacac21b3b8bec92c32c2444efa5c5434f/bundles/org.openhab.ui/web/src/pages/home/homecards-mixin.js#L127

https://github.com/openhab/openhab-webui/blob/19395dfacac21b3b8bec92c32c2444efa5c5434f/bundles/org.openhab.ui/web/src/js/openhab/auth.js#L95

These are parts of the logic that will redirect to the authorize page when you enforce authorization for the non-admin ops (_Settings > API Security > Show advanced > Implicit user role for unauthenticated requests_ disabled).

ghys (approved):
Excellent, thanks!

PR #1046, opened 2021-05-10
Fix Z-Wave network map not rendering when a node has no zwave_neighbours property

TheNetStriker:
This pull requests fixes the error "Cannot read property 'split' of undefined" in the ZWave Network Map if some ZWave things do not have the "neighbours" property.

Signed-off-by: David Masshardt <david@masshardt.ch>


TheNetStriker:
Sorry, this is only my second pull request for OpenHab and I always forget to add the sign off line. I've added the sign off line to the first comment and also in this comment. 

Signed-off-by: David Masshardt <david@masshardt.ch>

PR #1052, opened 2021-05-17
Change unit from "lux" to "lx" on badges

cweitkamp:
- Change unit from "lux" to "lx" on badges

We are using unit signs for temperature and humidity. The correct unit for illumination is lx.

See also https://en.m.wikipedia.org/wiki/Lux


ghys (approved):
Makes sense, thanks!
The frontend doesn't have access to the extent of the UoM functionality available in the backend and relies on reasonable hardcoded defaults instead - but at least the unit symbols can be made right :)

PR #1054, opened 2021-05-18
Fix profile filtering

hubsif:
Fix bugs introduced with #1043


ghys (approved):
That should do the trick, thanks for the quick fix!

PR #1063, opened 2021-05-20
[Main UI] Added Badge for BatteryLow warning

### LINKED ISSUE #1060: [Main UI] Add system default Badge for BatteryLow
## The problem

I like to see Battery Low warnings in my location cards.

grafik

grafik

## Your suggestion

Add a new Badge to show Battery Low warnings on Location cards. Suggested Icon:

grafik

Suggested Matching Rule: Semantic Class (Point>BatteryLow), Switch Item: ON -> Icon is visible, OFF -> Icon is hidden

grafik

grafik

cweitkamp:
- Added Badge for BatteryLow warning

Closes #1060 

@ghys Untested new feature. I need your advice. Is it really that easy? Or did I miss something?

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
@@ -42,6 +42,7 @@ export default {
     return {
       badgeConfigs: {
         alarms: { icon: 'f7:exclamationmark_triangle_fill' },
+        battery: { icon: 'oh:lowbattery', state: 'on' },
         lights: { icon: 'oh:lightbulb' },
         windows: { icon: 'oh:window', state: 'open' },
         doors: { icon: 'oh:door', state: 'open' },
@@ -70,6 +71,10 @@ export default {
     query () {
       let direct, equipment, allPoints, points
       switch (this.type) {
+        case 'battery':
+          direct = findPoints(this.element.properties, 'Point_LowBattery', true)
+          if (direct.length) return direct
+          return findPoints(allEquipmentPoints(this.element.equipment), 'Point_LowBattery', true)
         case 'lights':
           direct = findPoints(this.element.properties, 'Point_Control', true, 'Property_Light')
           if (direct.length) return direct
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
@@ -42,7 +42,7 @@ export default {
     return {
       badgeConfigs: {
         alarms: { icon: 'f7:exclamationmark_triangle_fill' },
-        battery: { icon: 'oh:lowbattery', state: 'on' },
+        battery: { icon: 'f7:battery_25', state: 'on' },
         lights: { icon: 'oh:lightbulb' },
         windows: { icon: 'oh:window', state: 'open' },
         doors: { icon: 'oh:door', state: 'open' },
@@ -72,7 +72,7 @@ export default {
       let direct, equipment, allPoints, points
       switch (this.type) {
         case 'battery':
-          direct = findPoints(this.element.properties, 'Point_LowBattery', true)
+          direct = findPoints(this.element.properties, 'Point_Status_LowBattery', true)
           if (direct.length) return direct
           return findPoints(allEquipmentPoints(this.element.equipment), 'Point_LowBattery', true)
         case 'lights':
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -70,6 +71,10 @@ export default {
     query () {
       let direct, equipment, allPoints, points
       switch (this.type) {
+        case 'battery':
+          direct = findPoints(this.element.properties, 'Point_LowBattery', true)
```
cweitkamp, line 75: Should we check for property "Energy" too?

```suggestion
          direct = findPoints(this.element.properties, 'Point_LowBattery', true, 'Property_Energy')
```

ghys: > Should we check for property "Energy" too?

I think that's superfluous since the point class alone is self-describing.

Also it's `Point_Status_LowBattery`, not `Point_LowBattery`.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -42,6 +42,7 @@ export default {
     return {
       badgeConfigs: {
         alarms: { icon: 'f7:exclamationmark_triangle_fill' },
+        battery: { icon: 'oh:lowbattery', state: 'on' },
```
ghys, line 45: I prefer `f7:battery_25` to `oh:lowbattery` (monochrome and more representative I think:

image

image

ghys (changes requested):
Thanks, yes basically that's it - you got the class name wrong though and I'd suggest another icon.

The code of these status badges could definitely be improved to be more generic and extensible, so people could add their own badges (the code to map related items is rather similar for each types).

Perhaps a check on BatteryLevel points with a numerical state below 10 % could be also considered since this is what I usually see as channels, never "low battery" explicitely. Though probably LowBattery items could be linked to the actual level with a 0-10 Range profile...

cweitkamp:
Thanks for your feedback. I changed the class name and the icon. I finally got it running in my own dev environment.

It would be a nice enhancement if users can define their own Badges or change the existing logic. I already have another improvement in my mind. I would like to consider equipment of type "VoiceAssitant" and state "PLAY" for speakers.

I am no friend of fixed values. e.g. I would prefer 15% for Battery Low. Thus I think the Profiles Hysteresis and Range are the best solutions.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -70,6 +71,10 @@ export default {
     query () {
       let direct, equipment, allPoints, points
       switch (this.type) {
+        case 'battery':
+          direct = findPoints(this.element.properties, 'Point_Status_LowBattery', true)
+          if (direct.length) return direct
+          return findPoints(allEquipmentPoints(this.element.equipment), 'Point_LowBattery', true)
```
ghys, line 77: You missed this one!

```suggestion
          return findPoints(allEquipmentPoints(this.element.equipment), 'Point_Status_LowBattery', true)
```

ghys (changes requested):
Yes, the purpose of these badges wasn't originally designed with user extensibility in mind, rather provide sensible hardcoded defaults for common equipment so that stuff appears automatically, but of course it was a poor choice since the OH users like customizability, and since it proves quite popular and needs arise we can revisit this goal.

That's also why I mentioned the "battery level < 10%" as a suggestion as an additional criteria for bringing up the "battery low" icon, as many devices only have battery level channels that are likely to be mapped as-is with the default profile, and few users would bother to define dedicated "battery low" items with profiles for their equipment (or even know how to) if it's not offered as a channel.
But for now let's just agree on the BatteryLow point requirement 👍

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
cweitkamp: 🤦🏼 Ups ... changed it.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -70,6 +71,10 @@ export default {
     query () {
       let direct, equipment, allPoints, points
       switch (this.type) {
+        case 'battery':
+          direct = findPoints(this.element.properties, 'Point_Status_LowBattery', true)
```
ghys, line 75: I'm just realizing now that lines 75 & 76 here might even be unnecessary - these prioritize points directly under the Location over those under an Equipment... but having LowBattery points directly under a Location makes little sense I think, you would expect batteries to be under Equipment all the time?

ghys (approved):

ghys:
@cweitkamp I think an update to the docs (https://www.openhab.org/docs/tutorial/model.html#badges) is in order - I wish these would be in a reference doc rather than the tutorial...

cweitkamp:
Of course: https://github.com/openhab/openhab-docs/pull/1583.

PR #1066, opened 2021-05-25
item-picker adjustments

### LINKED ISSUE #1062: [Main UI] Managed Items cannot be assigned to Groups in Item Edit page
## The problem

Managed Items cannot be assigned to Groups in Item Edit page

grafik

grafik

## Expected behavior

Groups should be available.

## Steps to reproduce

1. Got to Settings -> Items page

2. Choose Item

3. Click Edit button in top right corner

## Your environment

```yaml

```

## Browser console

```txt

```

## Additional information

Workaround: Edit the Group and assign members.

hubsif:
_Edit:_ ~This fixes #1062.~ (fixed in #1073)

Additionally, I adjusted the following: with slow item loading I found that
- the item-picker element for groups looks exactly the same, but does not react to clicks unless the items are actually loaded. It now shows "Loading..." instead of "Parent Groups" and a spinner until items are ready. Let me know if the design is ok. (at first I wanted the popup to show a single loading item and then update to the items, but updates within a smart select seem to not be rendered. I the end I like the current solution even better.)
- the item picker always loads all items, even when only groups are required. In my test setup that makes a difference of about ~600 items or ~300KB. I adjusted the rest call to use the type filter prop if provided to the item-picker (and sincerely hope that despite my double checks to not have introduced new regressions again 😄.)

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
@@ -10,7 +10,7 @@
       <f7-button slot="media" icon-f7="list_bullet_indent" @click.native="pickFromModel" />
     </f7-list-item>
     <!-- for placeholder purposes before items are loaded -->
-    <f7-list-item link v-show="!ready" :title="title">
+    <f7-list-item link v-show="!ready" :title="title" disabled no-chevron>
       <f7-button slot="media" icon-f7="list_bullet_indent" @click.native="pickFromModel" />
     </f7-list-item>
   </ul>
@@ -50,7 +50,8 @@ export default {
     this.smartSelectParams.closeOnSelect = !(this.multiple)
     if (!this.items || !this.items.length) {
       // TODO use a Vuex store
-      this.$oh.api.get('/rest/items').then((items) => {
+      const filter = Array.isArray(this.filterType) ? this.filterType.join(',') : this.filterType
+      this.$oh.api.get('/rest/items' + (filter ? '?type=' + filter : '')).then((items) => {
         this.sortAndFilterItems(items)
       })
     } else {
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -79,7 +79,7 @@ export default {
       let oldError = this.nameErrorMessage
       if (!/^[A-Za-z0-9_]+$/.test(name)) {
         this.nameErrorMessage = 'Required. Alphanumeric & underscores only'
-      } else if (this.items.some(item => item.name === name)) {
+      } else if (Array.isArray(this.items) && this.items.some(item => item.name === name)) {
         this.nameErrorMessage = 'An item with this name already exists'
       } else {
         this.nameErrorMessage = ''
@@ -91,7 +91,6 @@ export default {
     if (!this.item) return
     if (!this.item.category) this.$set(this.item, 'category', '')
     if (this.enableName) {
-      if (!this.items) this.items = []
       this.validateName(this.item.name)
     }
     const categoryControl = this.$refs.category


bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
```diff
@@ -10,7 +10,8 @@
       <f7-button slot="media" icon-f7="list_bullet_indent" @click.native="pickFromModel" />
     </f7-list-item>
     <!-- for placeholder purposes before items are loaded -->
-    <f7-list-item link v-show="!ready" :title="title">
+    <f7-list-item link v-else title="Loading...">
```
ghys, line 13:
```suggestion
    <f7-list-item link v-show="!ready" :title="title" disabled no-chevron>
```
This is more subtle and doesn't break the item picker's "smart select" settings - it opens in a page instead of a popup and doesn't work at all when called from a popup, probably due to the VDOM patching.

bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
```diff
@@ -10,7 +10,8 @@
...
+      <f7-preloader />
```
ghys, line 14: Please remove this preloader. It changes the list item's height in some cases and really doesn't look good - it's too much "in your face" esp. when you have several of them in a config sheet. Applying the disabled style and removing the chevron conveys properly, and in a subtle way, that the control is not ready yet.

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
```diff
@@ -74,7 +74,7 @@ export default {
     return {
       ready: false,
       item: {},
-      items: [],
+      items: null,
```
ghys, line 77: While this probably solves #1062 indeed, it causes a new error in the console when creating a new item (Settings > Item > Add New Item):

```
[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "items"

found in

---> <ItemForm> at src/components/item/item-form.vue
       <F7Col>
         <F7Block>
           <F7PageContent>
             <F7Page>
               <ItemEdit> at src/pages/settings/items/item-edit.vue
```

ghys (changes requested):
Well thanks but this deserves some more work ;)

bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
hubsif: I assumed you wouldn't like the preloader too much 😄.
Though, I think your suggestion is a little too "subtle" - it doesn't indicate to the user that something's going to change, does it? You really don't even want to change the title to "Loading..."? IMHO there should be at least a tiny something telling the user that the page is not ready yet.

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
hubsif: Argh, missed that. Actually, that code line in question seems to have been able to cause these lines even before. Will adjust.

bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
hubsif: I'd like to learn, can you please explain why `v-show` here and not `v-else`?

bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
```diff
@@ -10,7 +10,8 @@
...
       <f7-button slot="media" icon-f7="list_bullet_indent" @click.native="pickFromModel" />
```
hubsif, line 15: while we're at it: without having read the code, I didn't realize the "list_bullet_indent"-icon actually was a button! 😮 
I even didn't really notice it, thought it's just some kind of menu icon. And I might not be the only one. What about giving the button a little of a "button styling", like a frame or some raised effect?

bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
ghys: Here's my take on it: the item picker is usually ready very quickly, maybe tenths of a second, and in many cases _before_ the user even has a chance to click on it, so they're not even aware there was something loading in the background in the first place. I never really had a case of an item picker being opened so quickly after being shown that it wasn't ready.

On the other hand, having something flickering before your eyes is almost always a bad look (unless it's an important part of the page), and here we'd be having every item picker briefly and conspicuously showing its "loading" state when shown, unnecessarily drawing the user's attention. I'd rather have the occasional user click twice because it was not ready the first time than the flickering.

So I believe in this case the subtler, the better. And of course it will be a moot point when we have a Vuex cache of items with a reasonable strategy to keep it up-to-date behind the scenes, after the cache is populated every item picker will be ready instantaneously.

bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
ghys: I tried different things and none really looked pleasing in all 3 themes and pixel densities, it was either too bulky or out of place - I wanted the item picker to remain essentially a list item without anything out of place. So I agree it's not ideal, but once to realize by hovering over the icon to discover it's a button, then you've discovered that feature. Happy to see some more designs but I really tried some months ago.

bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
ghys: Essentially because of https://vuejs.org/v2/guide/conditional.html#Controlling-Reusable-Elements-with-key.
Vue here would try to reuse the non-functional "loading" list item for the actual one and this leads to problems notably the smart select options not being considered. Another option would have been to add `key` attributes to both alternatives to prevent this re-using.

hubsif: ah, ok, thank you!

florian-h05:
Superseded by #2428.
The filter type improvement has not been picked up in #2428, for reason please read my comment in the new PR.

PR #1076, opened 2021-05-30
Corrected small typo

jossuar:


ghys (approved):
Thanks!

Signing requirement waived as per https://www.openhab.org/docs/developer/contributing.html#small-patch-exception

PR #1084, opened 2021-06-02
Update oh-repeater.md

jossuar:
Corrected display issue due to nested double quotes

Signed-off-by: Georgios Moutsos <georgios.moutsos@gmail.com>

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-repeater.md
+++ b/bundles/org.openhab.ui/doc/components/oh-repeater.md
@@ -38,12 +38,12 @@ Iterate over an array and repeat the children components in the default slot
     What to iterate on
   </PropDescription>
   <PropOptions>
-    <PropOption value="array" label="Array (default) in the "in" parameter" />
-    <PropOption value="range" label="Range of integers defined by "rangeStart", "rangeStop", "rangeStep"" />
-    <PropOption value="itemsInGroup" label="Member of the group defined in the "groupItem" parameter" />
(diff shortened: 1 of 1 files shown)


jossuar:
Please ignore the DCO check.
I have this repo probably not correctly setup.

ghys:
Thanks, but it needs to be changed at the source or it will be overwritten when the generate script runs again.
(or the script should take these into account and escape/replace the quotes properly).

jossuar:
Can you point me to the generation script? (and i will close this pull request)

ghys:
> Can you point me to the generation script? (and i will close this pull request)

Yes, sorry - the generation script is here: https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/doc/components/src/generate.js

This is the problematic line:
https://github.com/openhab/openhab-webui/blob/dd0472f0a509de61a750282603ff09cbd3bdc549/bundles/org.openhab.ui/doc/components/src/generate.js#L64

but there might be some others as well. Double quotes should be escaped in any case when there's concanated like this.

Another alternative is to simply get rid of the double quotes in the oh-repeater case at the source which is here:
https://github.com/openhab/openhab-webui/blob/dd0472f0a509de61a750282603ff09cbd3bdc549/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/repeater.js#L7-L14

If you match the docs Markdown file with these config parameters, even if you don't run the generator script, I believe this could be merged alright.

PR #1093, opened 2021-06-05
Set current item from selected name

### LINKED ISSUE #1079: Cannot select profile when linking to an existing item
## The problem

The list of profiles isn't displayed properly when creating a new link to an existing item.

## Expected behavior

The list should be provided, depending on the chosen item's type.

## Steps to reproduce

1. From a Things page in the Channels tab, open a channel and create a new link, leave the "Use an existing Item" option
2. Select an item in the "Item to Link" picker. A list of compatible profiles doesn't appear as it should (error below).

Workaround: the link can still be created with the default profile and then edited to configure the profile.

## Your environment

```yaml
runtimeInfo:
  version: 3.1.0
  buildString: "Build #2408"
```

## Browser console

```txt
vue.esm.js?a026:1897 TypeError: Cannot read property 'split' of undefined
    at eval (link-add.vue?46d5:190)
    at Array.filter (<anonymous>)
    at VueComponent.compatibleProfileTypes (link-add.vue?46d5:190)
    at Watcher.get (vue.esm.js?a026:4488)
    at Watcher.evaluate (vue.esm.js?a026:4593)
    at VueComponent.computedGetter [as compatibleProfileTypes] (vue.esm.js?a026:4845)
    at Object.get (vue.esm.js?a026:2104)
    at eval (link-add.vue?dc50:329)
    at Proxy.renderList (vue.esm.js?a026:2639)
    at Proxy.render (link-add.vue?dc50:324)
```

## Browser network traffic

N/A

## Additional information

In this code added by #1043:
https://github.com/openhab/openhab-webui/blob/9ac085ffae6194ca06b1e74523d0cc7415685106/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue#L185-L191
`currentItemType` can end up null, or equal to `this.selectedItemName` (from line 186) which is a string, not an item object, therefore doesn't have a type property.
This incidentally worked until #1054 added the `.split` without checking `currentItemType`.

hubsif:
Fixes #1079

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
@@ -158,7 +158,6 @@ export default {
         channelUID: null,
         configuration: {}
       },
-      selectedItem: null,
       selectedItemName: null,
       selectedThingId: '',
       selectedThing: {},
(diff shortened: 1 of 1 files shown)


ghys (approved):
Excellent, this seems to fix the profile selection indeed! Thanks

PR #1099, opened 2021-06-09
Upgrade echarts to 5.1.2

crnjan:
When displaying two values using a single gauge, on initial load gauge(s) are not rendered correctly, as per screen:

switching a tab or similar action fixes the problem

After upgrade to 5.1.1. this cannot be reproduced anymore - a similar discussion [here](https://community.openhab.org/t/pie-chart-problem/122604).

Signed-off-by: Boris Krivonog boris.krivonog@inova.si


NorbertHD:
Why not [v5.1.2](https://github.com/apache/echarts/releases/tag/5.1.2) with even more fixes?

crnjan:
At the time I created my branch 5.1.2 did not yet exist :) Will update my PR ...

ghys (approved):
Great, there shouldn't be any regressions with this upgrade (I hope!).
Thanks!

PR #1106, opened 2021-06-16
Enable linking of trigger channels

### LINKED ISSUE #886: Cannot link trigger channel to an item
When looking at a trigger channel of a thing, the UI does not allow creating an item link:

Screenshot 2021-02-07 at 01 01 23

The hint to use rules for using the channel is not really correct as their main purpose is to allow being linked to an item through a profile. Just like for state channels, I would expect the UI to show the available profiles (the ones that apply to trigger channels as per [the documentation](https://www.openhab.org/docs/configuration/items.html#profiles)).
It is probably not needed (and rather misleading) to offer the creation of a new item, but it should be feasible to create a link against an existing item that has the type that is supported by the selected profile.

hubsif:
Fixes #886

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
@@ -64,7 +64,7 @@ export default {
         const labelB = b.label || b.name
         return labelA.localeCompare(labelB)
       })
-      if (this.filterType) {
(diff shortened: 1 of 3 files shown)


ghys (approved):
This seems to be working nicely. Nice to have the trigger channels linkable. Thanks!

kaikreuzer:
Many thanks @hubsif!

PR #1109, opened 2021-06-20
Add Log Viewer (Frontail) to the Developer Tools

kubawolanin:
Welcome back! 👋🏻 

This PR draft introduces Log Viewer (Frontail) embedded directly in the UI. 
image

It also reacts to `filter` query string and passes it to Frontail, so we're able to quickly diagnose a problem:
image

Drawbacks:
* Frontail port is currently hardcoded to `9001` as I couldn't find a proper place to add it as a  configuration param. Basic UI offers some sort of settings (dark theme, png/svg icons etc), which main UI is currently missing.
* This will only work locally (as it tries to access `window.location.origin` with replaced port

Side note: it would be nice to introduce some sort of feature flag system in the main UI, so we're able to gradually turn on that kind of features to the broader audience (experiments?), make it's presence toggleable on the UI and ultimately - merge bigger changes (even not necessarily completed) more frequently, without them affecting the UX.

@ghys chapeau bas, awesome job on the UI!

Cheers,
Kuba

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -71,7 +71,12 @@
           </li>
 
           <f7-list-item link="/developer/" :title="$t('sidebar.developerTools')" panel-close
-                        :class="{ currentsection: currentUrl.indexOf('/developer/') >= 0 && currentUrl.indexOf('/developer/widgets') < 0 && currentUrl.indexOf('/developer/api-explorer') < 0 }">
+                        :class="{
+                          currentsection: currentUrl.indexOf('/developer/') >= 0 &&
+                            currentUrl.indexOf('/developer/widgets') < 0 &&
+                            currentUrl.indexOf('/developer/api-explorer') < 0 &&
+                            currentUrl.indexOf('/developer/frontail') < 0
+                        }">
             <f7-icon slot="media" ios="f7:exclamationmark_shield_fill" aurora="f7:exclamationmark_shield_fill" md="material:extension" color="gray" />
           </f7-list-item>
           <li v-if="showDeveloperSubmenu">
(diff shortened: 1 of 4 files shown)


ghys:
As you know Frontail is not part of the openHAB distribution _per se_, so we can't expect it to run on port 9001 by default, or at all, in which case this option shouldn't be displayed.
On the other hand, we can expect a sizable amount of users having opted to run openHABian on top of openHABian.
I think a sensible approach would be for openHABian to install a dedicated bundle whose purpose would be to add a "supervisor" REST API endpoint (`/rest/supervisor` or `/rest/openhabian`...), offering features beyond OH's scope:
- advertising which additional software is installed (Frontail, Grafana etc. and on which ports)
- retrieving system logs, event logs, application logs (and perhaps streaming them)
- package updates
etc.
That additional API endpoint would be detected by the UI when loading the `/rest/` root endpoint and would be used to unlock these additional options when relevant.

ghys:
> Side note: it would be nice to introduce some sort of feature flag system in the main UI, so we're able to gradually turn on that kind of features to the broader audience (experiments?), make it's presence toggleable on the UI and ultimately - merge bigger changes (even not necessarily completed) more frequently, without them affecting the UX.

For now I have avoided making a configurable service for the UI (to have a space for UI-related server-wide options - that's why the options in Help & About are local only), partly because it would be yet another request before anything is displayed, but mostly because now `/rest/services` is restricted to admins.
Ultimately a place to store UI-related options on the server could become necessary indeed.

Skinah:
One option that would be great to have is to turn off the watching of event.log so only the openhab.log is watched with a checkbox or some other control. I find the events annoying in frontail and prefer to use the `Event Monitor` that is in the Developer sidebar. Currently I turn the event.log to WARN level to achieve this.

hmerk:
> One option that would be great to have is to turn off the watching of event.log so only the openhab.log is watched with a checkbox or some other control. I find the events annoying in frontail and prefer to use the `Event Monitor` that is in the Developer sidebar. Currently I turn the event.log to WARN level to achieve this.

You can simply remove the events.log entry in the frontail.service file. That's how I did it.

Gifford47:
are there any news here? is this feature included in the next release? it would be very usefull for me.

splatch:
> are there any news here? is this feature included in the next release? it would be very usefull for me.

Sadly Kuba passed last summer. There will be no follow up from his account on this pull request.

PR #1110, opened 2021-06-20
HABot Polish language support

kubawolanin:
Still some strings remain untranslated, but I'm getting there.


florian-h05:
Closing as this went stale and according to a comment from @ghys in #142, Training data would also be needed.

PR #1115, opened 2021-06-23
Add Things filtering by binding

kubawolanin:
Demo:
filter1

Empty state:
image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/things-list.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/things-list.vue
@@ -14,6 +14,22 @@
           search-container=".contacts-list"
           search-in=".item-inner"
           :disable-button="!$theme.aurora" />
+        <f7-input
+          v-if="bindings.length"
+          class="bindings-filter"
+          type="select"
+          ref="bindings-filter"
+          @change="filterByBinding">
+          <option value="all">
+            All bindings
+          </option>
+          <option
+            v-for="(binding, index) in bindings"
+            :value="binding.id"
+            :key="index">
+            {{ binding.name }}
+          </option>
+        </f7-input>
       </f7-subnavbar>
     </f7-navbar>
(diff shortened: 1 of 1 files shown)


ghys:
Thanks for your work @kubawolanin.

Some objections I have about this PR:
- while it looks good enough on your GIFs on a desktop scenario, sharing the horizontal space with the search bar doesn't quite work on phone screens (see screenshots below). For instance on iOS I don't think there are examples where the search bar isn't full width (see also https://developer.apple.com/design/human-interface-guidelines/ios/bars/search-bars/). On Android it looks even more out of place.
I would be more than open to a common approach to filtering & sorting which would work across these list screens (things, items, pages, rule/scripts, widgets etc.): for example a button near the counter above the list that would open a popover/sheet with filtering & sorting controls, depending on the type of information that is listed - so that the items list could introduce other criteria like groups or tags. This would ideally come with a refactoring of these list screens, with all the common functionality moved in a common mixin or similar (a lot of code is duplicated in these, which obviously isn't ideal).
- the native dropdown isn't used anywhere else in the UI (afaik) and might look out of place depending on the OS
- there's an additional API call to get the binding labels (I added it early in the original development of this screen but then figured the ID alone could suffice).
- since there's an option to group by binding, having a filter with the same criteria seems redundant (I know it's not the same thing and there are legitimate cases where it could help, but still). Moreover on desktops at least the search box is focused by default and you can simply type "mqtt" or "zwave" to get a similar view.

image image

image image

kubawolanin:
Fair points @ghys, thanks for looking at this.

> Moreover on desktops at least the search box is focused by default and you can simply type "mqtt" or "zwave" to get a similar view.

Yeah, but typing sucks ;-) Especially in scenarios when you need to change some config property in several Things of a certain binding. 

Search filter persistence (when navigating through pages) would also be nice. It's a little missing thing that drove me nuts in Paper UI :)

Cheers

florian-h05:
Closing as this went stale.

PR #1132, opened 2021-07-14
[basicUI] Fix handling of buttons when only 1 button

### LINKED ISSUE #1131: [basicUI] Mappings with only one entry not correctly rendered

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

Mappings with only one entry are not correctly rendered.

My item:
```
Switch MuteAutoNoUpdate "mute (no autoupdate)" <mute> {channel="xxx"}
```
My  sitempa:
```
    Frame label="Mappings - autoupdate" {
        Switch item=MuteAutoNoUpdate
        Switch item=MuteAutoNoUpdate mappings=[ON="Muet"] 
        Switch item=MuteAutoNoUpdate mappings=[ON="Mute", OFF="Listen"] 
    }
```

The sound is muted so the item state is OFF.
Here is how it looks when I open the page:
image

The second widget (at right) is not as expected, it should be shown as "ppushed". It should be rendered similar as with the third widget.

## Expected behavior

It shoud like this when I open the page while the device is muted:
image

And like this when the device is unmuted:
image

The problem is clearly in `SwitchRenderer` where "mdl-button--accent" is not set when there is only one button !

lolodomo:
Fix #1131

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SwitchRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SwitchRenderer.java
@@ -126,13 +126,12 @@ public class SwitchRenderer extends AbstractWidgetRenderer {
                 if (commandDescription != null) {
                     for (CommandOption option : commandDescription.getCommandOptions()) {
                         // Truncate the button label to MAX_LABEL_SIZE characters
-                        buildButton(s, option.getLabel(), option.getCommand(), MAX_LABEL_SIZE, nbButtons > 1, item,
-                                state, buttons);
+                        buildButton(s, option.getLabel(), option.getCommand(), MAX_LABEL_SIZE, item, state, buttons);
                     }
                 }
             } else {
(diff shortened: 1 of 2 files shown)


lolodomo:
@kaikreuzer : could you please review this one ? It has no link with the autoupdate stuff.

kaikreuzer:
The code that you are changing suggests that the current behavior was a very deliberate decision and not an accidental bug.
Can you make any sense out of it, why it was done this way and what its intention might have been?

lolodomo:
> The code that you are changing suggests that the current behavior was a very deliberate decision and not an accidental bug.

I was expecting you will find that ;) I agree with you, this seems to be implemented deliberately like that.

> Can you make any sense out of it, why it was done this way and what its intention might have been?

I have strictly no idea why it was decided to not show the state of the button ! And no idea why to implement something different for 1 button versus n buttons.

kaikreuzer (approved):
Ok, thanks for your answer! Let's just hope that we don't overlook anything that might now break by this change...

PR #1138, opened 2021-08-08
[habot] Removing old ephemeral cards to debug

Skinah:
Believe this should not be info level as it is not useful to most users and my logs are getting it way too often.

To reproduce:

Type in "show temperature for 1 month".
The graph appears and the log now has an entry like:
Removing old ephemeral card d7xxx2-f1b4-4x5a-xxx9-x7f56xxxxxd

Signed-off-by: Matthew Skinner <matt@pcmus.com>


ghys (approved):
Makes perfect sense, thank you!

PR #1139, opened 2021-08-09
[basicUI] Fix color selector placement

### LINKED ISSUE #934: basicui: rgb color selector css improvements

- [ ] Basic UI

## The problem

RGB Color selector dot is not exactly placed, reason is the css styling

## Your suggestion

Change color picker handle css class ".colorpicker__handle" like following:
- remove current margin setting "`margin: -12px 0 -12px;`" which is also wrong (top and bottom are set here, not top and left)
- add "`transform: translate(-50%, -50%);`" what makes the px value irrelevant. works on all modern browsers. 

you instantly see a correct placement exactly in the middle for rgb 100% or on the outer cicle line for a single color 100%

SharedScreenshot

support for older browsers which is not realy relevant today, but possible:

```
-moz-transform: translate(-50%, -50%);
-ms-transform: translate(-50%, -50%);
-o-transform: translate(-50%, -50%);
-webkit-transform: translate(-50%, -50%);
transform: translate(-50%, -50%);
```

lolodomo:
Fix #934

Change proposed by @ako-3004 tested with Chrome/Edge/Firefox browsers on Windows 10 + Chrome on Android.

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -513,17 +513,15 @@ $colorpicker-mobile-size: 270px;
 	&__handle {
 		$handle-size: 24px;
 		position: absolute;
-		top: 50%;
-		left: 50%;
 		box-sizing: border-box;
 		-moz-box-sizing: border-box;
 		border: 4px solid white;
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
thanks!

PR #1148, opened 2021-08-25
Fix detection of protected user API operations

### LINKED ISSUE #1146: Deactivation of "Implicite user role" in API-settings blocks items state update
## The problem

When disabling the "Implicite user role" feature in API-Security settings, item states don*t get updated in MainUI anylonger.
Enabling the feature again leads to updated item states in MainUI. The issue is occuring both on RPi installation as well as on Windows installation. I*m using KNX binding to connect to my KNX installation. When switching on the light, MainUI does not show the updated state. In addition, when switching on the light using MainUI, the light goes on but I can*t switch off the light again. When refreshing the page, MainUI shows again the OFF state.
At the same time, BasicUI is working fine, i.e. item state get updated correctly.
When switching on "Implicite User Role" featue again, the item state gets updated correctly again.

## Expected behavior

MainUI should visualize the item state also with feature "Implicite User Role" deactivated.

## Steps to reproduce

I reproduced the problem using a fresh Openhab 3.1.0 installation on Windows machine.
1. Install KNX Binding (I don't know whether it really is linked to KNX)
2. Create a KNX Bridge and KNX Thing.
3. Create a switch item for a light.
4. Disable "Implicite User Role" feature in "API-Security" settings.
5. Switch on / off the light in MainUI / using physical switch.

## Your environment

N/A

## Browser console

N/A

## Browser network traffic

N/A

## Additional information

N/A

ghys:
The current implementation would set the access token before trying out calling `/rest/sitemaps` which would obviously always succeed, since the call would be made with the token. Therefore the "requireToken" flag would not be set properly and prevent the alternative SSE implementation (which allows headers) to be used.

Fixes #1146.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/auth.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/auth.js
@@ -84,17 +84,22 @@ export function storeBasicCredentials () {
 }
 
 export function setAccessToken (token, api) {
-  accessToken = token
-  if (!token || !api || requireToken !== undefined) return Promise.resolve()
-
-  // determine whether the token is required for user operations
-  return api.get('/rest/sitemaps').then((resp) => {
-    requireToken = false
+  if (!token || !api) return Promise.resolve()
+  if (requireToken === undefined) {
+    // determine whether the token is required for user operations
+    return api.get('/rest/sitemaps').then((resp) => {
+      accessToken = token
+      requireToken = false
+      return Promise.resolve()
+    }).catch((err) => {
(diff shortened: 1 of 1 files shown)


martinspinler:
Still I have related problem. In my configuration, the setAccessToken does a `GET /rest/sitemaps` with the `Authorization: Bearer` header. This request succeeds and sets `requireToken=false`.
But this causes that the newSSEConnection in sse.js doesn't sets  `X-OPENHAB-TOKEN` nor `Authorization: Bearer` headers. Then the `GET /rest/events/states` fails (401).
Did I miss something?

PR #1149, opened 2021-08-28
oh-icon component improvements, Iconify icon support

### LINKED ISSUE #1081: FR: Add materialdesignicons to MainUI
I was looking for some automative icons for the new carnet widget and found some at 
https://dev.materialdesignicons.com

Would it be possible to add those ?

Instructions for VueJS can be found here :
https://dev.materialdesignicons.com/getting-started/vuejs

ghys:
This allows the oh-icon to display icons from other sources than the
OH icon servlet; bundled Framework7/Material icons can also be displayed
by this component (use `f7:` or `material:` prefixes).
Specifying no prefix will assume an OH icon for backward compatibility.

This also introduces Iconify (https://icon-sets.iconify.design) as an
icon source, allowing access to 100,000+ icons from various icon sets,
to display them in widgets like cards or floorplan markers.

The Iconify option requires Internet connectivity on the client to access
the external API but the underlying library will cache the icons in
local/session storage to limit the requests and speed up the rendering.

Update/fix documentation and plan marker default icon registration.

Closes #1081 (as the `mdi` icon set is available though Iconify).


hmerk:
@ghys Great job Yannick !!

Issue #241, opened 2020-05-02
Modified files after build cause subsequent build to fail


wborn:
After running `mvn clean install` the following files are reported as modified by `git status`:

```shell
	modified:   bundles/org.openhab.ui.basic/npm-shrinkwrap.json
	modified:   bundles/org.openhab.ui.habot/web/package-lock.json
	modified:   bundles/org.openhab.ui/web/package-lock.json
```

When you then start another build with these modified files, the org.openhab.ui.basic bundle fails to build:

```
[INFO] --- frontend-maven-plugin:1.9.0:npm (gulp build) @ org.openhab.ui.basic ---
[INFO] Running 'npm run build' in /home/wouter/git/openhab/openhab-webui/bundles/org.openhab.ui.basic
[INFO] 
[INFO] > org.openhab.ui.basic@3.0.0 build /home/wouter/git/openhab/openhab-webui/bundles/org.openhab.ui.basic
[INFO] > gulp default
[INFO] 
[INFO] fs.js:35
[INFO] } = primordials;
[INFO]     ^
[INFO] 
[INFO] ReferenceError: primordials is not defined
[INFO]     at fs.js:35:5
[INFO]     at req_ (/home/wouter/git/openhab/openhab-webui/bundles/org.openhab.ui.basic/node_modules/natives/index.js:143:24)
[INFO]     at Object.req [as require] (/home/wouter/git/openhab/openhab-webui/bundles/org.openhab.ui.basic/node_modules/natives/index.js:55:10)
[INFO]     at Object.<anonymous> (/home/wouter/git/openhab/openhab-webui/bundles/org.openhab.ui.basic/node_modules/vinyl-fs/node_modules/graceful-fs/fs.js:1:37)
[INFO]     at Module._compile (internal/modules/cjs/loader.js:1158:30)
[INFO]     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
[INFO]     at Module.load (internal/modules/cjs/loader.js:1002:32)
[INFO]     at Function.Module._load (internal/modules/cjs/loader.js:901:14)
[INFO]     at Module.require (internal/modules/cjs/loader.js:1044:19)
[INFO]     at require (internal/modules/cjs/helpers.js:77:18)
[INFO] npm ERR! code ELIFECYCLE
[INFO] npm ERR! errno 1
[INFO] npm ERR! org.openhab.ui.basic@3.0.0 build: `gulp default`
[INFO] npm ERR! Exit status 1
[...]

PR #1152, opened 2021-08-31
Prevent modified files with Maven builds

wborn:
#1113 only partially solved #241.

It might be better to run `npm ci` by default for more reproducible builds and to prevent changes to `package-lock.json`.


kaikreuzer (approved):
Thanks!

PR #1173, opened 2021-10-01
Add rule template support

### LINKED ISSUE #1082: Add support for rule templates
Paper UI had support for rule templates, which is a feature still missing in the Main UI.

Rule templates are a feature to easily create rules that are not empty, but readily usable. Allowing users to choose a template should therefore best be added as an option to the "+" button for adding new rules.

Available rule templates can be retrieved through `/rest/templates`.
Besides a uid, a template mainly has a label and description which can be used to show them in a list.
Additionally, a rule template has a config description, i.e. a list of parameters that the user should be requested to fill in.

When this is done, all the UI needs to do is to POST to `rest/rules` and instead of providing lists of modules as normally, simply pass the `templateUID` and the `configuration` with it. This should be it - a new rule will be created!

ghys:
Closes #1082.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
@@ -82,41 +82,70 @@
           <f7-block-footer v-if="!isEditable" class="no-margin padding-left">
             <f7-icon f7="lock_fill" size="12" color="gray" />&nbsp;Note: this rule is not editable because it has been provisioned from a file.
           </f7-block-footer>
-          <f7-col v-if="isEditable" class="text-align-right justify-content-flex-end">
-            <div class="no-padding float-right">
+          <!-- <f7-col v-if="isEditable" class="text-align-right justify-content-flex-end">
+          </f7-col> -->
+          <f7-col v-if="createMode && templates.length > 0" class="new-rule-from-template">
+            <f7-block-title medium class="margin-bottom">
+              Create from Template
+            </f7-block-title>
+            <f7-list media-list>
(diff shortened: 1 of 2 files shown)


ghys:
@kaikreuzer can you please answer this as I'm not 100% positive I got it right:

- do the rule templates only assist in _creating_ the rule, and then the modules can be freely edited (and in that case, why save the templateUID and configuration as part of the rule?),

- or can the configuration be altered after the rule has been created, and the modules will be refreshed with the new values in placeholders (in that case, the UI should hide the modules and only display the parameters as described by the template when editing too?).

kaikreuzer:
> do the rule templates only assist in creating the rule, and then the modules can be freely edited

Yes, correct.

> why save the templateUID and configuration as part of the rule?
> can the configuration be altered after the rule has been created

The template UID is imho indeed not necessary.
The configuration is meant to enable configurations on rule level that are then automatically considered by the modules. It might well be that this is technically not yet fully implemented, so maybe better ignore these configurations in the UI for now.

If this is what you have implemented so far and you think it should be directly included in M3, please remove the WIP and I'll wait with launching the M3 build until the merge of this PR.

ghys:
Yes it's what was implemented. There's the little issue we experienced yesterday that the template parameters are not checked when the rule is created (like all required parameters have been filled in) and the template might not be applied if there are errors.
Other than that I believe it works well overall, we can probably add the parameter checks in a follow-up PR. It would be nice to open the marketplace to rule templates early to accept contributions and iron out the eventual bugs then.
If you agree then I'll merge this.

kaikreuzer (approved):
Alright, let's merge then - thank you!

PR #1175, opened 2021-10-06
Adds click action to oh-gauge-card

TheGreyDiamond:
This adds an action to the oh-gauge-card. This is not fully done, I still need to make the mouse a click pointer, so it is visible that you can click it.

- [ ] Make the mouse a click pointer
I will do this but I though to make this PR so people can discuss things I did wrong beforehand.


TheGreyDiamond:
Also if this gets eventually (hopefully) merged/accepted I would be grateful if someone could add the `**hacktoberfest-accepted**` tag so this can count towards it. (FAQ https://hacktoberfest.digitalocean.com/faq Answer 18 is about this)

bundles/org.openhab.ui/web/src/components/widgets/standard/oh-gauge-card.vue
```diff
@@ -3,8 +3,8 @@
     <f7-card-header v-if="config.title">
       <div>{{ config.title }}</div>
     </f7-card-header>
-    <f7-card-content class="oh-gauge-card display-flex justify-content-center">
-      <oh-gauge :context="childContext(context.component)" @command="onCommand" />
+    <f7-card-content class="oh-gauge-card display-flex justify-content-center clickPointer" @click.native="performAction" >
+      <oh-gauge :context="childContext(context.component)" @command="onCommand"  />
     </f7-card-content>
```
ghys, line 8: I would prefer the following:
```suggestion
    <f7-card-content class="oh-gauge-card display-flex justify-content-center">
      <f7-link v-if="config.action" class="oh-gauge-link" @click="performAction">
        <oh-gauge :context="childContext(context.component)" />
      </f7-link>
      <oh-gauge v-else :context="childContext(context.component)" />
    </f7-card-content>
```
That way, you get the hand pointer as well as a visual feedback when you hit the control - but only if an action is defined.

See also below for the definition of the `oh-gauge-link` CSS class.

bundles/org.openhab.ui/web/src/components/widgets/standard/oh-gauge-card.vue
```diff
@@ -13,15 +13,19 @@
 </template>
 
 <style lang="stylus">
+ .clickPointer {
+    cursor pointer
+  }
```
ghys, line 18:
```suggestion
.oh-gauge-link
  position absolute
  top 0
  left 0
  width 100%
  height 100%
```

ghys (changes requested):
Thank you for your first contribution!
I left some suggestions below.
Please also note that you have to add the action parameter group to the widget's definition so users can configure the action with the UI.
This is in `assets/definitions/widgets/standard/cards.js`.
Look for examples in the other widgets, you have to add something like:
```
  .paramGroup(actionGroup(null, 'Action to perform when the gauge is clicked'), actionParams())
```
at the end of the OhGaugeCardDefinition.
You'll have to regenerate the docs as well - this happens in `doc/components/src`, you have to run `node generate.js` and commit the modified `oh-gauge-card.md` (not the others if there's no need).

ghys (approved):
Thanks, all good now.

TheGreyDiamond:
Cool thanks!

PR #1180, opened 2021-10-17
Expose full item info to widgets

asmigala:
Allows to use the full item info in widget expressions, not just the state.

Most obvious use case is something like

```yaml
config:
  title: '=items[props.item].label 
```
but there are [other ideas](https://community.openhab.org/t/create-alternative-title-with-metadata/121056/3) [previously worked around](https://community.openhab.org/t/oh3-how-to-use-transformation-in-expressions-custom-widgets/109127/7) by using `oh-repeater`.

This version works by requesting the full item data from the api once,  when a new item is added to the tracking list. The state is then updated in the usual way.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/store/modules/states.js
+++ b/bundles/org.openhab.ui/web/src/js/store/modules/states.js
@@ -31,6 +31,7 @@ const handler = (context) => {
         }
 
         context.dispatch('updateTrackingList')
+        context.dispatch('retrieveItemInfo', itemName)
       }
       return context.state.itemStates[itemName]
     },
@@ -52,6 +53,12 @@ const getters = {
 }
 
 const actions = {
+  retrieveItemInfo (context, itemName) {
+    console.debug(`retrieving item info ${itemName}`)
+    this._vm.$oh.api.get(`/rest/items/${itemName}`).then((item) => {
+      context.commit('setItemInfo', { itemName: itemName, itemInfo: item })
+    })
+  },
   initializeTrackingStore (context) {
     if (stateTrackingProxy) return
     console.debug('Initializing state tracking store proxy')
@@ -140,8 +147,16 @@ const mutations = {
     state.keepConnectionOpen = value
   },
   setItemState (state, { itemName, itemState }) {
+    // get the existing item state (potentially with extra info)
+    // and overwrite with recieved state
+    itemState = { ...state.itemStates[itemName], ...itemState }
     Vue.set(state.itemStates, itemName.toString(), itemState)
     // state.itemStates[itemName] = itemState
+  },
+  setItemInfo (state, { itemName, itemInfo }) {
+    // get the existing item state and set the info (keeping the original state intact)
+    itemInfo = { ...itemInfo, ...state.itemStates[itemName] }
+    Vue.set(state.itemStates, itemName.toString(), itemInfo)
   },
   setPendingTrackingListUpdate (state, payload) {
     state.pendingTrackingListUpdate = payload


ghys:
Thanks, but I don't think it's reasonable to make a separate API request for every item that gets added to the state tracking list.

If you're browsing e.g. the home page or a complex page then you'll easily have dozens of items to track, that would mean dozens of additional requests - to get extra info that may or may not even be useful. This state tracking mechanism tries to be as efficient as possible to avoid performance issues, and the cost-benefit ratio seems too high here.

One possible way of doing this that would be acceptable is to take advantage of the currently empty response body of the `POST /rest/events/states/{connectionId}` call to get a filtered array of ItemDTOs based on the tracking list that was sent.

This needs changes to the core:
https://github.com/openhab/openhab-core/blob/9d9dc70d89407bdd55eb5a53e50b50bd56759fd5/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java#L199-L226

Then you can merge those results with the state updates coming from SSE.

asmigala:
> If you're browsing e.g. the home page or a complex page then you'll easily have dozens of items to track, that would mean dozens of additional requests

Right, I'm new to openahab, currently in the process of migrating from homeassistant, so I only have a handful of items set up. 

>  to get extra info that may or may not even be useful. This state tracking mechanism tries to be as efficient as possible to avoid performance issues

There should be no performance hit once the info is loaded. Also when the extra info is not used in the widgets, there should be no delay in showing the UI. Is there any performance test I could run to see the actual impact?

> 
> One possible way of doing this that would be acceptable is to take advantage of the currently empty response body of the `POST /rest/events/states/{connectionId}` call to get a filtered array of ItemDTOs based on the tracking list that was sent.
> 
> This needs changes to the core: https://github.com/openhab/openhab-core/blob/9d9dc70d89407bdd55eb5a53e50b50bd56759fd5/bundles/org.openhab.core.io.rest.sse/src/main/java/org/openhab/core/io/rest/sse/SseResource.java#L199-L226


Ok, I'll look into that, though I'm not sure about returning a list of items from a `/events/states/` endpoint. Maybe adding a new `filter` parameter to the `/items` [endpoint](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/item/ItemResource.java#L202-L208) and then requesting that from `updateTrackingList` (so only one extra API call) would be acceptable?

ghys:
> There should be no performance hit once the info is loaded. Also when the extra info is not used in the widgets, there should be no delay in showing the UI. Is there any performance test I could run to see the actual impact?

The tracking lists are reset often, when you switch pages etc. So in the end in a typical session you will end up making possibly hundreds of item info requests - many being made simulateously - causing network traffic, congestion, CPU time & battery usage to parse the responses etc.

To be honest, I have yet to see a compelling use case for the need to retrieve item information dynamically outside a oh-repeater, that couldn't be put directly on the page component; it's more visually compelling too since retrieving such info asynchronously causes a visual "flicker" until it's available.

That being said, there might be a solution - in the `setItemState` mutation instead of using the argument itself, augment it and construct/manage a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) based on it that would have a "smart" getter handler to make the request when a (well-known) property of an item other than `state` or `displayState` is requested. It could even return `state` when `displayState` is requested and there's none in the original object (another recent request https://community.openhab.org/t/expressions-items-not-accessible/125684/28).

florian-h05:
Closing this PR as it went stale. 

@asmigala In case you still want to get access to the full Item info, I would consider the solution using a Proxy proposed by @ghys in https://github.com/openhab/openhab-webui/pull/1180#issuecomment-946685771 should be fine.

PR #1182, opened 2021-10-18
Evaluate expressions in action analyzer items

asmigala:
Allows custom widgets to define the analyze action without hard coding the item, like 

```yaml
    config:
      action: analyzer
      actionAnalyzerItems:
        - '=props.item'
```

Ideally something like this should be added to all the other actions, let me know if there's an interest in doing that.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-actions.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-actions.js
@@ -211,7 +211,9 @@ export const actionsMixin = {
           break
         case 'analyze':
         case 'analyzer':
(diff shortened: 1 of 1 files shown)


ghys:
The `this.config` that is used as a basis in this function should already have the expressions evaluated.
You just can't mix the array YAML notation with expression in individual expressions, the whole parameter has to be an expression.
So instead of
```yaml
      actionAnalyzerItems:
        - '=props.item'
```
you should use:
```yaml
      actionAnalyzerItems: =[props.item]
```
> Ideally something like this should be added to all the other actions, let me know if there's an interest in doing that.

As said above AFAIK this should already be the case.

asmigala:
@ghys you're right, I was missing the array constructor in my previous attempts. Closing

PR #1187, opened 2021-10-25
Addition of graphical layer management for canvas layouts

tarag:
Addition of layers as an intermediate component between canvas layouts and canvas items.

Allows creation of pages with various set of widgets that can easily be toggled between one another. Layer visibility management in edition improves usability when lots of widgets are visible in one canvas.

This is not backwards compatible with previous canvas, so I think it should be merged before to many people start experimenting with canvases in the milestones.

Existing canvas layouts can be easily ported by adding the following YAML lines in the code editor, between canvas and the first canvas-item:
```yml
canvas: # existing line
  - component: oh-canvas-layer
    slots:
      default:
        - component: oh-canvas-item # existing line
```

Preview:
Screen Recording 2021-10-25 at 18 20 10


ghys:
> Having the visible property appear in the configuration panel for the layer would make sense. Is there a reason why it is not the case for widgets?

No reason, it's somewhat in the backlog but could possibly be added for all widgets that support it (as an advanced property group), as well as other common properties like `visibleTo`,  `class` or `style` (the latter giving you an editor tailored for CSS with autocompletion, perhaps).

tarag:
@ghys , do you thing we could get this merged before the imminent next milestone, so that people experimenting the canvas layout would not have to migrate their layout ?

ghys (approved):
@tarag I'd say it looks pretty good now.
I fixed a minor bug which broke the component docs generation script, which I run. It's a manual process for now, so please remember to give it a run when you add/change widget parameter definitions.

Also as https://github.com/openhab/openhab-docs/pull/1657 is still open, perhaps you could describe the layers feature there as well while you're at it?

Thanks!

tarag:
Thanks, I'll try to think about it for the parameter docs.

I'll update the docs for the layers feature soon.

PR #1201, opened 2021-11-08
Update icons F7 to v5.0.5 and Material to v6.1.1

woernsn:
As the bundled icons/fonts for F7 and Material are pretty out-of-date, I updated them to the latest released version.

Sources:
- ~~Framework7: https://github.com/centrual/Framework7/tree/v1.5.3/kitchen-sink-ios/fonts~~
- Framework7: https://github.com/framework7io/framework7-icons/tree/v5.0.5/package/fonts
- Material: https://github.com/jossef/material-design-icons-iconfont/tree/v6.1.1/dist/fonts


ghys:
Are you really sure of your source for the F7 icons? It looks like a fork and the last commit is 5 years old.

woernsn:
> Are you really sure of your source for the F7 icons? It looks like a fork and the last commit is 5 years old.

Whoosh. You are absolutely right. I did a quick search for the `.eot` file and this fork popped up first.

Nevertheless it seems as Framework7 dropped the `.eot` font with version 5.0.0. (https://github.com/framework7io/framework7-icons/blob/034cec9ec85ac28fe9b25c78be53d70971657e9b/CHANGELOG.md#features-1).

My main goal was to update the Material Font anyway. Does it make sense to ipdaye Framework7 to the latest `.eot`-supporting version - i.e. 4.0.2 (2021-02-19).
Or would updating Material Fonta only make more sense?

ghys:
> Nevertheless it seems as Framework7 dropped the .eot font with version 5.0.0.

Indeed - https://github.com/framework7io/framework7-icons/commit/fa7c51dbd8ef9ec5b85b7b2fbfa27c722903d8b9
So I'm wondering if we couldn't remove the .eot font support as well (I reckon it was mainly for IE which is not relevant anymore, and the UI isn't compatible with it anyway) - in `css/fonts.css`, so we could use the latest 5.0.5 version, and do the same for Material icons- only keep the CSS `@font-face` references to woff, woff2 & ttf formats.

woernsn:
Alright, it seems that
- `eot` is the only supported font face for IE8 and below
- `woff` is supported by IE11
- `woff2` is supported by everything else

I will update the PR accordingly.

ghys (approved):
Having confirmed that:

- the newest f7 icons are there and display properly
image

- the newest Material icons are there and display properly

...I think we're good there!

Thanks @woernsn!

PR #1202, opened 2021-11-08
[mainui] Fix addons details serviceId

J-N-K:
The serviceId for  details page in the addon store was not properly determined if the addonId did not start with `marketplace:`.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/addons/addon-info-table.vue
+++ b/bundles/org.openhab.ui/web/src/components/addons/addon-info-table.vue
@@ -42,7 +42,7 @@ export default {
     information () {
       let info = []
       if (!this.addon || !this.addon.id) return info
-      const source = this.addon.id.indexOf(':') > 0 ? this.addon.id.substring(0, this.addon.id.indexOf(':')) : "karaf"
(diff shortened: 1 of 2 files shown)


ghys:
You might want to address this problem too eventually:

https://github.com/openhab/openhab-webui/blob/9aeb301771be59ef9a117ce1d90dcde5943ca398/bundles/org.openhab.ui/web/src/components/addons/addon-info-table.vue#L50

i.e. every add-on not coming from the marketplace will be advertised as coming from the openHAB distribution.
This was to avoid a call to `/rest/addons/services` to get the label. Though it might be sensible to do that:
- serviceId=marketplace => Source: Community Marketplace
- serviceId=karaf => Source: openHAB Distribution
- otherwise: use the Id as the source, or call `/rest/addons/services` and find out the label asynchronously.

If you want to postpone that to another PR I can accept this one in the meantime.

J-N-K:
@ghys I tried to find a solution. WDYT?

ghys (approved):
LGTM - both the solution and the code.
Eventually the "3rd Party (serviceId)" could be improved with a request to get the real label but it's not that important right now.

PR #1212, opened 2021-11-23
Blockly enhancements 2 - scripts, ephemeris, persistence

stefan-hoehn:
Second part of the enhancements

- Add script support 
- Add ephemeris support
- Add persistence

... awaiting review

Note: I again struggled with the right way of the commits due to the bug fix included and the rebase with main...


stefan-hoehn:
Done with all implementations, now waiting for the review

ghys:
Okay first off, thanks!

I have one major hurdle with the date picker field: it is simply too big for what it brings to the table. It comes with another UI component library to render the calendar and as a result adds 5% to the entire size of the UI, or half a MB! The code wouldn't be loaded until the user brings up the Blockly editor but still. Also when the service worker is involved (if the UI is served using HTTPS) it would be preloaded anyway. The Actions tab has this report:

image

Since Framework7 has its own [date picker](https://v5.framework7.io/docs/calendar.html) so I think it would be best to reimplement our own date picker field using this component instead (similar to the item/thing pickers), shouldn't be too hard to do, I'll give it a try.

ghys (approved):
I made significant changes and I believe this is now ready for prime time.
Since we'll have a few weekly release candidates until the final 3.2 we'll have some time to test these!

Issue #1165, opened 2021-09-23
Build fails under MacOS on Apple M1 processors


J-N-K:
## The problem

Build fails because node.js 12.x is not available for ARM architecture.

## Expected behavior

Build succeeds.

## Steps to reproduce

1. Clone repo on a ARM machine running MacOS BigSur
2. `man clean install`

## Additional information

Solution: Upgrade to NodeJS 16.9.1 (Basic UI build fails with npm error on simple upgrade, Main UI succeeds)

ghys:
Fixed by #1353.

PR #1215, opened 2021-11-28
allow building of most projects on M1 processors

J-N-K:
This mainly upgrades NodeJS from 12.16.1 to 16.13.0. The only bundle that still can't be built is HaBOT because there are too many "old" dependencies that I am not able to update due to lack of knowledge.

Related to #1165 

Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ bin/
 features/*/src/main/history
 npm-debug.log
 node_modules/
+.idea
--- a/bundles/org.openhab.ui.basic/gulpfile.js
+++ b/bundles/org.openhab.ui.basic/gulpfile.js
@@ -3,7 +3,7 @@
 
 	var
 		gulp = require("gulp"),
-		sass = require("gulp-sass"),
+		sass = require('gulp-sass')(require('node-sass')),
 		uglify = require("gulp-uglify"),
 		eslint = require("gulp-eslint");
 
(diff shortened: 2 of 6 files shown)


ghys:
Thanks but it would be better to have a consistent version of node/npm across all bundles (HABPanel & HABot included) to avoid downloading and caching multiple versions. It was a mess before and https://github.com/openhab/openhab-webui/pull/202 made it a little better.

J-N-K:
I'm out then. I have no idea how to get that "quasar-cli" package updated. The last update is two years old and doesn't run with modern node as far as I found out (due to a dependency to fsevents/1.x which doesn't run on node 13+).

ghys:
@J-N-K since you figured out the issues w/ basic UI already can you reopen this one, as you were the first to on this? It turns out HABot does build on Linux (just not on macOS), which isn't critical imho.
Also while you're at it can you please set:
```xml
          <nodeVersion>v16.14.2</nodeVersion>
          <npmVersion>8.6.0</npmVersion>
```
If it builds and has no regressions then it's fine by me (and I'll close #1352).

J-N-K:
Looks like I'm not allowed to re-open a PR.

ghys:
Ok, me neither. There's a tooltip when I hover the button "the branch was force-pushed or recreated".
I'll incorporate your changes in #1352 and credit you with an `Also-by` sign-off then.

PR #1237, opened 2021-12-17
Restore JS highlighting for ECMAScript-2021 + Tern autocomplete

ghys:
Also adds separate "mostly" complete [Tern definitions](https://ternjs.net/doc/manual.html#typedef) for the JSScripting globals + standard library for JSScripting, assuming they are available as globals (the default).
The old Nashorn definitions or these new ones are loaded depending on the media type.

@digitaldan @jpg0 I would appeciate if you could think of these (in bundles/org.openhab.ui/web/src/assets/openhab-js-tern-defs.json) when you change the API of the standard library.
It doesn't have to be 100% exhaustive or anything (I didn't include the ZonedDateTime/JS-Joda types for instance) but whatever you can do helps!


jpg0:
@ghys is there no way these can be autogenerated? It would be possible to annotate the JS source to allow it. This would certainly make keeping the helper library current simpler.

(Note that the 'globals' are completely dynamic and defined by whatever plugins you have installed, so I suspect this would be much harder.)

digitaldan:
OMG , that's amazing @ghys !  I will take a look today , I will also start to think how to automate this in the future.  This really will be nice for users.

ghys:
> @ghys is there no way these can be autogenerated? It would be possible to annotate the JS source to allow it. This would certainly make keeping the helper library current simpler.

> (Note that the 'globals' are completely dynamic and defined by whatever plugins you have installed, so I suspect this would be much harder.)

Yes I think it'll be possible to at least partly generate it from JSDoc or similar.
It can also be made dynamic to some extent but as you said it can become tricky as we don't have direct access to the context from the client.

Let's merge this version then so it makes RC1 - there will be a window to make adjustments this weekend and prenty of time after the release to think about how to improve it ;)

digitaldan:
@Confectrician could our VS code plugin do something similar ?  I know autodetecting JS file that are rules would need some assistance, but we could support a header, commonly use by editors like vi or emacs for similar reasons

```javascript
/* openhab-rule:latest */
```

Confectrician:
This could be possible theoretically.

I have to admit, that the language and syntax parts didn't get any bigger updates after they have been introduced by kubawolanin at the very beginning of the extension.
(Because those language parts are really difficult for me to dig into.)

The extension will be active and available anyway when you are navigating in the openHAB conf folder.
So we have to check how we can tell vscode that the js files in a specific folder are openHAB related.

I will check where we would have to configure this.

ghys:
I'm not sure if Intellisense in VS Code is supposed to work out-of-the-box with `require()`, but I believe adding TypeScript typing modules does seem to be a good approach, e.g. see [day.js](https://day.js.org) for a simple example:

image

is defined in:
https://github.com/iamkun/dayjs/blob/4a7b7d07c885bb9338514c234dbb708e24e9863e/types/index.d.ts#L260-L269

which is referenced in `package.json`:
https://github.com/iamkun/dayjs/blob/4a7b7d07c885bb9338514c234dbb708e24e9863e/package.json#L6

For openhab-js, I tried starting a `index.d.ts`  (+ reference in `package.json` like this):
```ts
declare namespace openhab {
  declare interface Item {
    /** The name of the item */
    name: string;
    /** The type of the item */
    type: string;
    /** The tags of the item */
    tags: string[];
    /**
     * Members / children / direct descendents of the current group item (as returned by 'getMembers()'). Must be a group item.
     * @returns {Item[]} member items
     */
    members: Item[];
    /**
     * All descendents of the current group item (as returned by 'getAllMembers()'). Must be a group item.
     * @returns {Item[]} all descendent items
     */
    descendents: Item[];

    /**
     * Sends a command to the item
     * @param {String|HostState} value the value of the command to send, such as 'ON'
     * @see sendCommandIfDifferent
     * @see postUpdate
     */
    sendCommand(value: String|HostState);

    /**
     * Sends a command to the item, but only if the current state is not what is being sent.
     * Note
     * @param {String|HostState} value the value of the command to send, such as 'ON'
     * @returns {Boolean} true if the command was sent, false otherwise
     * @see sendCommand
     */
    sendCommandIfDifferent(value: String|HostState): boolean;
  }
  
  declare interface ItemsNamespace {
    /**
     * Gets an openHAB Item.
     * @memberOf items
     * @param {String} name the name of the item
     * @param {String} nullIfMissing whether to return null if the item cannot be found (default is to throw an exception)
     * @return {items.Item} the item
     */
    getItem(name: string, nullIfMissing?: string): Item;
  }
};

class openhab {
  static items: openhab.ItemsNamespace;
}

export = openhab;
```

Then it starts to work:

image

image

image

digitaldan:
Thats intersting, i knew TS allowed Intellisense auto completion, but  i had not thought about using it to just expose the libraries signature/interfaces , super interesting that day.js is doing just that and its not actually written in TS. 

@Confectrician if we could somehow infer then those imports on JS files in a openHAB directory,  that might be a good solution.  I like also that this would then not require changes to the VS Code plugin if we update the library (if it pulls latest)

Confectrician:
Would be a neat solution, but i am not sure how we could achieve this currently.
(As said, language parts are my waterloo category of the extension.)

Probably vscode does things on its own already, if you open a js file in the folder structure and a package.json with declared types file is available.

digitaldan:
this looks interesting as well, i may play with this on Sunday.

https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html

csowada:
ah, interesting to see a similar topic here. I use a similar module to ``openhab-js`` in TypeScript. But I want now to switch to the official libraries. For me it makes no sense to spend time on a special solution. Maybe I can help or my current project can show the TypeScript way. See here 

https://github.com/csowada/openhab-jsscripting-base

Confectrician:
@digitaldan @ghys 

Maybe a merged PR is the wrong place for this discussion.
May someone port this to a new open issue?

PR #1239, opened 2021-12-18
[blockly] add script param block

stefan-hoehn:
Signed-off-by: Stefan Höhn <stefan@andreaundstefanhoehn.de>

This finalizes the blockly "Run & Process" section by allowing provided parameters by a rule to be retrieved via a blockly written _script_  as this PR provides a simple block to retrieve the value via blockly

Use of the new parameter retrieval block

image

The rule that calls the script (for reference)
image

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-eventbus.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-eventbus.js
@@ -28,44 +28,4 @@ export default function (f7) {
     const value = Blockly.JavaScript.valueToCode(block, 'value', Blockly.JavaScript.ORDER_ATOMIC)
     return 'events.' + eventType + '(' + itemName + ', ' + value + ');\n'
   }
-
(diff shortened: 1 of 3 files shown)


ghys:
Thanks for the PR. Probably needs more work to consolidate the related concepts (the rule "context") and name things accurately:
- We already have a "contextual information" block in "Items & Things" (`oh_context_info`) and all things considered it should move to the "Run & Process" category (both the code in `blocks-eventbus.js` and the toolbox)
- This block should be renamed to `oh_context_attribute` and placed alongside `oh_context_info`
- `context.getAttribute` won't work in the new GraalVM engine; the attributes are also placed in `ctx["key"]` which will work so this should be used instead
- The label should read something like "context attribute {text shadow "key"}"
- The "run rule or script" label should be renamed to "run rule and script with context" (instead of "with parameters").

ghys (approved):
Let's merge this one as well as it was tested to work and adds an omission to the toolbox, so it makes the release!

Issue #1183, opened 2021-10-19
oh-icon - style is not applied to the underlying tags in all cases


crnjan:
## The problem

Playing with [heating widget](https://github.com/NRquadrat/oh3-widgetHeating/blob/2.0.0/oh3-widgetHeating.yaml#L748) and trying to use iconify icons - so replacing `f7-icon` with a `oh-icon` - the icon(s) were not displayed.

It turns out the problem is related to the `style` property and how it's applied.

## Expected behavior

One should be able to use custom `style` with `oh-icon`.

## Steps to reproduce

```
component: f7-card
config:
  title: TEST
slots:
  default:
    - component: oh-icon
      config:
        height: 28px
        icon: iconify:mdi:fan-speed-3
        style:
          position: absolute
          top: 5px
          left: 55px
```

will render

<img width="135" alt="Screenshot 2021-10-19 at 20 08 30" src="https://user-images.githubusercontent.com/3818211/137966621-b030a680-a2ff-40e5-98df-58afe341172a.png">

ignoring the content within `style`. As far as I can see from [code](https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/components/widgets/system/oh-icon.vue) `style` is only (fully) applied for `oh` icons, while not for others:

```
component: f7-card
config:
  title: TEST
slots:
  default:
    - component: oh-icon
      config:
        height: 28px
        icon: oh:light
        style:
          position: absolute
          top: 5px
          left: 55px
```

will render

<img width="107" alt="Screenshot 2021-10-19 at 20 17 36" src="https://user-images.githubusercontent.com/3818211/137967870-ded850bf-24af-4560-b183-88609f65edca.png">

since `style` is applied to underlying `img`, while not for other two (`f7-icon` and `iconify-icon`).

## Additional Info

Wrapping `oh-icon` in another block "fixes" the issue:

```
component: f7-card
config:
  title: TEST
slots:
  default:
  - component: f7-block
    config:
      style:
        position: absolute
        top: 5px
        left: 55px
    slots:
      default:
      - component: oh-icon
        config:
          height: 28px
          icon: iconify:mdi:fan-speed-3
```

renders:

<img width="114" alt="Screenshot 2021-10-19 at 20 28 16" src="https://user-images.githubusercontent.com/3818211/137969412-d4cc5547-eee7-40f1-94f8-fa14abbe83db.png">

[...]

PR #1243, opened 2021-12-18
Check config before allowing oh-icon style customization

### LINKED ISSUE #1242: Missing f7 / material icons in latest builds
I'm still trying to track down this on my end, but i just pulled down the latest build 3.2.0.202112181015 and none of my widgets that use `f7:...` or `material:...` are displaying.  Interesting enough, loading system local images with `oh:...` does work.  I also tested this on a version 3.2.0.202112122130 which i have running for development purposes, and the issue is not there, so either something is wrong with my main system, or this is a recent regression. 

theme

## Steps to reproduce

1. 

## Your environment

```yaml

```

## Browser console

```txt

```

ghys:
#1233 offers style customization properties for f7 and iconify icons but they end
up conflicting with those set by the `f7-icon` & `iconify-icon` components themselves.

Fixes #1242.


ghys:
Finally it appears checking that the `oh-icon` has a config set is enough to solve both #1242 and #1183.

ghys:
Merging as a critical bugfix, thanks @digitaldan for reporting it.

PR #1254, opened 2021-12-25
[blockly] fix item.getAttributes

stefan-hoehn:
Signed-off-by: Stefan Höhn <stefan@andreaundstefanhoehn.de>

Fix for what has been mentioned in https://community.openhab.org/t/blockly-reference/128785 (Documentation has already been updated related to this fix)

- [Get Item attribute](https://community.openhab.org/t/blockly-reference/128785#get-particular-attributes-of-an-item-14) has the wrong block type. As a result it cannot be connected to any other block and cannot be used. Fix: Return "String" instead of "any"
- [Get Item attribute returning categories](https://community.openhab.org/t/blockly-reference/128785#get-particular-attributes-of-an-item-14) returns an Array of Strings that cannot be used with further blocks. Fix: Return a Collection instead of String in case it is Set of attributes

**Work in Progress: Fix is coming**
- Get context attribute has the wrong block type. As a result it cannot be connected to any other block and cannot be used. Fix: Return "String" instead of "any"

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
@@ -87,25 +87,80 @@ export default function (f7) {
     return [code, 0]
   }
 
+  /*
+  * Provides all attributes from an item
+  * - name: String
+  * - label: String
+  * - state: State
+  * - category: String
+  * - tags: Array
+  * - groups: Array
+  * - type: String
+  * Blockly part
+  */
   Blockly.Blocks['oh_getitem_attribute'] = {
     init: function () {
+      let thisBlock = this
+      let dropdown = new Blockly.FieldDropdown(
+        [['name', 'Name'], ['label', 'Label'], ['state', 'State'], ['category', 'Category'], ['tags', 'Tags'], ['groups', 'GroupNames'], ['type', 'Type']],
+        function (newMode) {
+          thisBlock.updateType_(newMode)
+        })
       this.appendValueInput('item')
         .appendField('get ')
-        .appendField(new Blockly.FieldDropdown([['name', 'Name'], ['label', 'Label'], ['state', 'State'], ['category', 'Category'], ['tags', 'Tags'], ['groups', 'GroupNames'], ['type', 'Type']]), 'attributeName')
+        .appendField(dropdown, 'attributeName')
         .appendField('of item')
         .setCheck('oh_itemtype')
       this.setInputsInline(false)
-      this.setOutput(true, 'any')
+      this.setOutput(true, 'String')
       this.setColour(0)
-      this.setTooltip('Retrieve a specific attribute from the item')
+      this.setTooltip('Retrieve a specific attribute from the item. Note that groups and tags return a list and should be used with the loops-block \'for each item ... in list\'. ')
       this.setHelpUrl('https://www.openhab.org/docs/configuration/items.html')
+    },
+    /**
+    * Modify this block to have the correct output type based on the attribute.
+    */
+    updateType_: function (newAttributeName) {
+      let attributeName = this.getFieldValue('attributeName')
+      if (newAttributeName === 'Tags' || newAttributeName === 'GroupNames') {
+        this.outputConnection.setCheck('Array')
+      } else {
+        this.outputConnection.setCheck('String')
+      }
+    },
+    /**
+    * Create XML to represent the input and output types.
+    * @return {!Element} XML storage element.
+    * @this {Blockly.Block}
+    */
+    mutationToDom: function () {
+      let container = Blockly.utils.xml.createElement('mutation')
+      container.setAttribute('attributeName', this.getFieldValue('attributeName'))
+      return container
+    },
+    /**
+    * Parse XML to restore the input and output types.
+    * @param {!Element} xmlElement XML storage element.
+    * @this {Blockly.Block}
+    */
+    domToMutation: function (xmlElement) {
+      this.updateType_(xmlElement.getAttribute('attributeName'))
     }
   }
 
+  /*
+  * Provides all attributes from an item
+  * Code part
+  */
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
```diff
@@ -83,29 +83,84 @@ export default function (f7) {
...
+  /*
+  * Provides all attributes from an item
+  * Code part
+  */
   Blockly.JavaScript['oh_getitem_attribute'] = function (block) {
     const theItem = Blockly.JavaScript.valueToCode(block, 'item', Blockly.JavaScript.ORDER_ATOMIC)
     const attributeName = block.getFieldValue('attributeName')
-    let code = `${theItem}.get${attributeName}()`
-    return [code, 0]
+    let code = `(${theItem} instanceof org.openhab.core.items.GenericItem) ? ${theItem}.get${attributeName}() : itemRegistry.getItem(${theItem}).get${attributeName}()`
```
ghys, line 158: I'm really not sure we have to make the code more complex to handle this.

In the case of the _oh_getitem_state_ block, if we mistakenly supply an Item instance instead of a string, we get an exception that's actually confusing (the item exists):
image
```js
var i;


i = itemRegistry.getItem('Scene_General');
print(itemRegistry.getItem(i).getState());
```
> 13:53:26.751 [WARN ] [re.automation.internal.RuleEngineImpl] - Fail to execute action: script
java.lang.RuntimeException: org.openhab.core.items.ItemNotFoundException: Item 'Scene_General (Type=NumberItem, State=1, Label=Scene General, Category=house, Tags=[Status])' could not be found in the item registry
       at jdk.nashorn.internal.runtime.ScriptRuntime.apply(ScriptRuntime.java:531) ~[jdk.scripting.nashorn:?]
        at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:456) ~[jdk.scripting.nashorn:?]
...


however if we supply something of a wrong type to the _oh_getitem_attribute_ block we get an error that's much easier to understand:
image

```js
var n;


n = 'Scene_General';
print(n.getLabel());
```
> 13:57:31.201 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'blockly_new' failed: TypeError: n.getLabel is not a function in <eval> at line number 5

That's more of a general problem the user have to be aware of (some blocks accept inputs of certain types, as suggested by the shadow blocks), and if you trick your way into supplying wrong types using variables you accept the risk of running into errors. It's not linked to OH blocks, for example:

image

```js
var n, i;


n = 123;
print((n.join(',')));
```
> 14:05:07.345 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'blockly_new' failed: TypeError: n.join is not a function in <eval> at line number 5

So I would leave it as-is in this case.

ghys (commented):
Thanks, the dynamic type change looks nice. Some comments below:

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
rkoshak: > I'm really not sure we have to make the code more complex to handle this.

The case that came up that prompted this (at my suggestion) was when looping though the list of a Group's members one gets the Item Object. However, "get item state" expects an Item's name. 

I think it's somewhat unreasonable for the user to have to know and understand the difference in this one use case, especially when the same "get item state" block can be made to support both the Item's name and an actual Item Object transparently. 

It's the difference between:

image

which intuitively seems like it should work but will fail and forcing the users to use

image
which clearly works but is anything but intuitive. Why do I need to save the Item's state to a variable first? Why can't I just use "get state of item" directly? 

That is the specific problem being addressed here. I don't think, in this case at least, the users should have to know and deal with this type error so I'd say that merely changing the error so it's more meaningful is suboptimal. It would be more user friendly to just handle both cases (i.e. item name or item instance) transparently. 

As an aside, the number one problem I need to help new users with on the forum, regardless of the rules language chosen, are type problems. Anywhere we can avoid or sidestep a type errors for the users will be a big win. This is pretty firmly in the type error category.

I know we can't get away from all type errors, but this appears to be case where we can.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
```diff
@@ -83,29 +83,84 @@ export default function (f7) {
 
   Blockly.JavaScript['oh_getitem_state'] = function (block) {
     const itemName = Blockly.JavaScript.valueToCode(block, 'itemName', Blockly.JavaScript.ORDER_ATOMIC)
-    let code = `itemRegistry.getItem(${itemName}).getState()`
+    let code = `(${itemName} instanceof org.openhab.core.items.GenericItem) ? ${itemName}.state : itemRegistry.getItem(${itemName}).getState()`
```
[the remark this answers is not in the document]
stefan-hoehn: @ghys @rkoshak I would propose I rollback / take out the group handling fix here and rather wait for a final idea how to handle that, so we do not hold the other two fixes back.

stefan-hoehn:
@ghys @kaikreuzer Can someone help me to undo the last commit without a trace so we can go forward with this PR as these fixes are more important

I fear if I do that I will do something wrong with the PR. 

I then need to add another fix to this because I just found out that the following block has the wrong return type "any" and therefore cannot be added to any block and not used at all:

image

I only want to push this fix after we have removed the one for the group handling.

kaikreuzer:
@stefan-hoehn I have removed the second commit on your branch, now there's only the first one left. I hope this is what you asked for!

stefan-hoehn:
Thanks @kaikreuzer, yes I did and would you mind telling me how to do that myself in the future?

ghys (approved):
Thanks!

stefan-hoehn:
> Thanks!

Thanks to you, Yannick

Issue #753, opened 2021-01-03
Implement "PointGroup" semantic class


krocans:
## The problem
#
There is no way to group Controls/Points under Equipment at same time keeping semantic structure (tree) consistent. 
I guess this issue generalizes #749 

## Your suggestion
Imagine Equipment which has many controls/points like 20 or even more. Such equipment when correctly configured with semantic tags cerates enormously long list in Location and other cards and makes those hard to read/use.
Logical solution seems to be arrange those controls in some groups. "Group" in legacy sitemaps did that well but un OH3 UI this causes problem.
here is sample item setup with two points grouped.
```
Group TestRoom "Test room" <sofa> ["Room"]
Group TestEquipment "Test equipment" <receiver> (TestRoom) ["Receiver"]
Switch TestEquipment_Control1 "Test equipment control 1" <switch> (TestEquipment) ["Control"]
Group TestEquipment_ControlGroup "Test equipment control group" (TestEquipment)
Switch TestEquipment_Point1 "Test equipment point 1" <text> (TestEquipment_ControlGroup) ["Point"]
String TestEquipment_Control2 "Test equipment control 2" <text> (TestEquipment_ControlGroup) ["Control"]
```
In this setup non-semantic group TestEquipment_ControlGroup is not shown in cards at all and items are placed as standalone of semantic tree. (Probably for that setup it's expected behaviour)
![Grp_loc](https://user-images.githubusercontent.com/42302779/103485321-3029a680-4dfe-11eb-8e10-e09753b409c9.png)
![Grp_model](https://user-images.githubusercontent.com/42302779/103485322-328c0080-4dfe-11eb-9484-e1c0668e720c.png)

**When group is configured as "equipment in equipment"** 
```
Group TestEquipment_ControlGroup "Test equipment control group" (TestEquipment) ["Equipment"]
```
semantic model looks great
![Equi_model](https://user-images.githubusercontent.com/42302779/103485386-84cd2180-4dfe-11eb-8733-391b82efa427.png)
but items in group still are not shown in locafion/equpment cards
![Equi_loc](https://user-images.githubusercontent.com/42302779/103485397-a1695980-4dfe-11eb-8663-9ef32248c837.png)
And bogus Miscellaneous Equipment is created in equipment tab 
![Equi_equi](https://user-images.githubusercontent.com/42302779/103485420-d1b0f800-4dfe-11eb-9133-2219d6c817f4.png)
[...]

PR #1264, opened 2022-01-06
Labelling of items in Model tab pages (Equipment & Properties)

tarag:
In order to have something understandable in Equipment/Properties model tab pages when having many Equipment and/or Properties of the same type (potentially having the same user-facing label), it is necessary to label them with something that allows to distinguish them from one another, which is usually the location of the equipment/point item.

This labelling is redundant with the labelling of the Location items to which these Equipment/Properties items may belong and leads to additional configuration work. In addition, it will clutter the pages in the Locations tab because all the items displayed there would have one common part which is the location, providing no user value in this context but more visual load (particularly problematic on small mobile screen where the end of the label, which may contain the most meaningful part will be trimmed).

This pull request provides some changes providing model context for the Equipment and Properties model tab pages. This  should allow having simple labels definition for items belonging to the model:

- In Equipment tab page cards: display the Location / Upper-level-equipment path in the divider title before items belonging to some equipment, trimming to preserve the end of this label if needed
- In Properties tab page cards: display the Location / Upper-level-equipment path in the subtitle of the item, instead of the item name (which I believe is not meant to be displayed in first-level user-facing non admin-level task pages)

Here is an illustration of the issue with the current version (3.2.0) with either short or complete labels used, and the resulting effect of the patch using the short labels.

Current status with model-redundant labels
current-long labels

Current status with short labels
current-short labels

Pull request version with short labels
patched-short labels

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/cards/card-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/card-mixin.js
@@ -65,6 +65,12 @@ export default {
     }
   },
   methods: {
+    itemPathLabel (item) {
+      if (!item.modelPath) return '(?) > ' + item.name
+      return item.modelPath.map((parent) => {
+        return parent.label || parent.name
+      }).join(' > ')
+    },
     cardOpening () {
       this.cardId = this.title + '-' + this.$f7.utils.id()
       history.pushState({ cardId: this.cardId }, null, window.location.href.split('#card=')[0] + '#' + this.$f7.utils.serializeObject({ card: this.element.key }))
--- a/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
@@ -35,13 +35,13 @@ export default {
   },
   computed: {
     listContext () {
-      const standaloneEquipment = this.element.equipment.filter((i) => i.points.length === 0).map((i) => itemDefaultListComponent(i.item, true))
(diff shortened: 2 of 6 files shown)


ghys (approved):
Really nice addition!
I just tested it on my prod and it works really well, the separator title trimming is a nice touch ;) Improving these home page cards is definitely something I want to push for the next release and I was actually planning something like that but never got around to it, so you beat me to it 🙂 Just know that it all might be revisited (see #753 for a related discussion). But I'm happy to have this improvement in the meantime while we discuss.

I just had a minor comment on some Stylus syntax that I ended up fixing myself since I had the branch checked out to test.

Thank you very much!

tarag:
I'm glad you like it, thanks for the merge. I am really looking forward to have the default model tab pages providing a user-friendly and useful result without too much tweaking and I really think this patch helps in this matter. 

There is still one improvement relevant to sorting that I'll try to perform as you can see that the `WidgetOrder` is not relevant in the Equipment and Properties pages. I'll try to find a way to pull Location related order and have it applied in these pages. WDYT?

My goal is to be able to rely only on the auto-generated model pages - and some canvas pages of course 😉 

I had not seen the discussion #753 while searching for related issues. I agree with the ideas discussed there, and I think it could further improve the usability when having many Points. I solve this now with a "maintenance" role where these advanced points are visible but this implies logging-in as another user which is not the best solution. Yet I believe my PR is  still relevant even if some of the changes discussed there would be performed as it only deals with labelling.

PR #1266, opened 2022-01-08
Update license headers to 2022

wborn:
I have updated the headers because it is now 2022. :slightly_smiling_face:


kaikreuzer (approved):
Thanks for taking care of it!

PR #1274, opened 2022-01-11
[habot] Add german tests and training data

### LINKED ISSUE #1271: [HABot] DeactivateObjectSkill does not properly recognize locations in German

- [ ] Basic UI
- [ ] HABPanel
- [x] HABot
- [ ] CometVisu

## The problem
As there is too less training data, the skill to deactivate an object in German does not recognize locations.

In these examples, only a single entity "licht" (=light) is found, but not the location ("wohnzimmer"=living room, "küche"=kitchen)
* mache das Licht im Wohnzimmer aus
* schalte in der Küche das Licht aus

I will provide more training data in a fork and additional unit tests with similar utterances.

JanMattner:
Especially for the activate and deactivate object skills
and various declension of locations. Other skills needed more training
data to prevent bias by sole training data amount.
Fixes #1271

Signed-off-by: Jan Mattner <jan.mattner@gmail.com>


ghys:
Hi, thank you very for contributing to the HABot training, it is always appreciated 🙂  - the more the better!

The DCO failure is a technicality (_Commit sha: 8519ea4, Author: Jan Mattner, Committer: Jan Mattner; Expected "Jan Mattner Jan.Mattner@aitgmbh.de", but got "Jan Mattner jan.mattner@gmail.com"._)... which one would you like in the final commit?
No need to do anything on your part.

However it would be great if you could solve the build error: check https://ci.openhab.org/job/PR-openHAB-WebUI/1088/console at the end, there are some format violations.

JanMattner:
Thanks @ghys.
About the DCO error: the mail in the commit message is correct (jan.mattner@gmail.com). The other one is just set in my local git config.

I have fixed the whitespace error, but somehow, the whole file changed, at least that's what shown in the history in GitHub, but locally in my history I just deleted the whitespaces in line 87. The spotless:check goal renders no errors locally.

Now I'm lost. I haven't found a way to solve this and I cannot locally reproduce the style error from the server build. Do you have an idea what could be the problem? Some kind of different encoding or line ending issue?

ghys:
Looks like you're on Windows and you've changed the encoding to CRLF, those need to be set back to Unix-style (LF).
In Eclipse that would be in File > Convert Line Delimiters to > Unix (IIRC).

JanMattner:
Thanks @ghys for the hint, that seemed to be the problem.
I am running VS Code in Windows and as suggested in the spotless error message, I used `mvn sportless:apply`. Seems like in Windows, this changes all line endings by default to CRLF (I will discuss this in a separate issue).
PS: in VS Code this is easy to see and change, see https://stackoverflow.com/a/48694365

PR #1283, opened 2022-01-18
Disable user-select on user facing pages (run mode) to prevent text selection issues

tarag:
Disable user-select on user facing pages (run mode) to prevent text selection issues

On iOS particularly, there are some issues when performing involuntary long presses while browsing and manipulation of main UI page, text in the displayed page or sometimes in background invisible pages can be highlighted which disrupts user manipulation. This patch prevents user selection on runtime pages to avoid these issues.

I had no access to Android device to test if the issue occurs but the patch has been made to also cover this platform.

Example of issue before PR applied:
user-select-orig

Result with PR applied:
user-select-patched

### DIFF
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -274,12 +274,8 @@ html
 
 // Disable user select causing distracting text selection upon long press on phones/tablets
 .disable-user-select
-  -webkit-user-select none
-  -moz-user-select none
-  -ms-user-select none
   user-select none
 
 // Disable user dragging of background images in canvas layouts
 .disable-user-drag
   user-drag none
-  -webkit-user-drag none
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -270,3 +270,16 @@ html
 
 // .card-backdrop-out
 //   animation card-backdrop-fade-out 400ms forwards
+
+
+// Disable user select causing distracting text selection upon long press on phones/tablets
+.disable-user-select
+  -webkit-user-select none
+  -moz-user-select none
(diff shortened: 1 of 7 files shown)


ghys:
Ok, what about the widgets allowing user input though? (`oh-input`, `oh-rollershutter` depending on config)

bundles/org.openhab.ui/web/src/css/app.styl
```diff
@@ -270,3 +270,16 @@ html
...
 // .card-backdrop-out
 //   animation card-backdrop-fade-out 400ms forwards
+
+
+// Disable user select causing distracting text selection upon long press on phones/tablets
+.disable-user-select
+  -webkit-user-select none
+  -moz-user-select none
+  -ms-user-select none
+  user-select none
```
ghys, line 280: Normally the vendor prefixed versions are added for you by Stylus so you shouldn't need to set them explicitely.
See for instance `oh-cell.vue`.

tarag: Fixed.

tarag:
> Ok, what about the widgets allowing user input though? (`oh-input`, `oh-rollershutter` depending on config)

I just tested with an input text card. Even with the page preventing user select upon long press, long press works inside of the text field so selection, copying and pasting is possible. I assume this directive is overridden by the browser for user input html fields.
I don't understand what kind of input there would be in roller shutter cards. It looks only like buttons to me. Can you explain ?

ghys:
> I don't understand what kind of input there would be in roller shutter cards. It looks only like buttons to me. Can you explain ?

Sorry I meant the oh-stepper and derivatives - you can configure them to have an input box to type the desired number (it's even the default IIRC).

ghys (approved):
Okay, let's try it out - if it causes problems we can always adjust later. Thanks!

Issue #1296, opened 2022-02-04
Single tap on floor plan on iOS does not work


georg138:
When I try to single tap on a marker on a floor plan, it actually issues a double tap. I'm trying to use a floor plan to toggle a light. This works fine on chrome on desktop but not on my mobile running iOS (at least chrome and Safari). I can see the icon is toggled for a very short time and then toggled again. As a result, the light usually stays off (or on).  Rarely it works. The following example demonstrates this issue:
```yaml
config:
  label: Sandbox
  sidebar: true
markers:
  - component: oh-plan-marker
    config:
      name: New Marker
      action: command
      actionItem: test
      actionCommand: '=(items.test.state > 0) ? "0" : "100"'
      actionCommandAlt: "100"
      icon: "=(items.test.state > 0) ? 'f7:circle_fill' : 'f7:circle'"
    slots:
      default: []

```
I found the following issue with leaflet which I think causes this:
https://github.com/Leaflet/Leaflet/issues/7255#issuecomment-849638476
Disabling tap as proposed there fixes the issue for me.

Issue #389, opened 2020-10-13
[Main UI] Toggling a marker on floorplan does not work properly on iOS


Trickx:
openHAB version: 3.0.0
I created a simple floorplan with a marker to toggle some lights.
While the marker properly toggles my lights accessing the page via Firefox or Chrome under Window 7 or Macosx,
it does not under iOS (iPhone SE 1. Gen or iPad Mini 3, both running iOS 14.0.1).
It looks like the action, toggling in my case, is executed twice per touch event in most cases.
The lights start to flicker a bit, but remain/return in their original state.
Sometimes it works though.

ghys:
Is this issue still current? Can anyone confirm this? I couldn't reproduce it on an iPhone recently.

Ronson1909:
Yes, I can confirm this issue with current openHAB 3.0.1 Release Build in the iOS Webinterface (but have not yet tested it in different browsers). Would be great to get it fixed.

PR #1297, opened 2022-02-04
Fix single tap on floor plan on iOS

georg138:
Fixes https://github.com/openhab/openhab-webui/issues/1296


ghys (approved):
Many thanks for researching this @georg138!
Looks like #389 is the same issue, so I'll add it to the final commit message.
Reading through https://github.com/Leaflet/Leaflet/issues/7255#issuecomment-849638476 I believe setting `tap` to false doesn't seem to have adverse effects.
It also looks like it was solved by https://github.com/Leaflet/Leaflet/pull/7026 which was merged in November 2021 so updating Leaflet might also work.

Issue #1304, opened 2022-02-08
Page to select bridge does not honor multi-line descriptions


FSeidinger:
## Adding a bridge with multi line descriptions

As described in the documentation of labels and descriptions of things:
https://www.openhab.org/docs/developer/bindings/thing-xml.html#formatting-labels-and-descriptions
Multi-line descriptions can have multiple lines with a subset of HTML markup.

The first stage of the bridge selection process in the `choose-thing-type.vue` does not honor these kind of descriptions and renders them as plain text including the HTML markup.

![image](https://user-images.githubusercontent.com/4890418/153079453-c39774eb-3181-446b-b3f2-3b3b931f409c.png)

## Expected behavior

The page should honor the multi-line descriptions and render them accordingly like the `thing-add-vue` does.

## Steps to reproduce

1. Open the page `Settings/Things`
2. Hit the `+` button to add a new Thing
3. Select the binding from the selection by clicking on it
4. See the formatting of the description

## Your environment

```yaml
runtimeInfo:
  version: 3.2.0
  buildString: Release Build
locale: default
systemInfo:
  configFolder: C:\Users\frank\PROGRA~1\Java\OPENHA~1.0\conf
  userdataFolder: C:\Users\frank\PROGRA~1\Java\OPENHA~1.0\userdata
  logFolder: C:\Users\frank\PROGRA~1\Java\OPENHA~1.0\userdata\logs
  javaVersion: 11.0.13
  javaVendor: Eclipse Adoptium
  javaVendorVersion: Temurin-11.0.13+8
  osName: Windows 10
  osVersion: "10.0"
  osArchitecture: amd64
  availableProcessors: 16
  freeMemory: 234811608
  totalMemory: 322961408
bindings:
  - knxng
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
    pixelRatio: 1
    prefersColorScheme: light
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: 8
    hardwareConcurrency: 16
    language: de-DE
    languages:
      - de-DE
      - de
      - en-US
      - en
      - it-IT
    onLine: true
[...]

PR #1306, opened 2022-02-09
Extract first line from multi-line thing description for the ThingTyp…

FSeidinger:
Fixes issue #1304 by extracting the first line of a multi-lined thing description.


ghys:
Thanks - although IMHO it would be enough to just split the string on `<br` and take the first part (see my remark @ https://github.com/openhab/openhab-webui/issues/1304#issuecomment-1034053065), building elements seems a little overkill, wdyt?

FSeidinger:
> Thanks - although IMHO it would be enough to just split the string on `<br` and take the first part (see my remark @ [#1304 (comment)](https://github.com/openhab/openhab-webui/issues/1304#issuecomment-1034053065)), building DOM elements seems a little overkill, wdyt?

As mentioned in the issue, you will find that as a standard solution for handling with inner HTML. I do not have any worries on computational power or memory consumption for just temporarily creating a single `<div>` tag and parse some lines of HTML. Bonus is that the script is easy to read and comprehend and the call is very short and also much more comprehend-able than the very long line with ternary operator and two or three parenthesis stacks.

But hey, I just tried to help. At last this is not my code and I have no preferences, beside that the bug gets fixed :-)

In the end, you decide.

ghys (approved):
Ok, fair enough :)

PR #1307, opened 2022-02-09
New option to display equipment as accordion in Locations cards

tarag:
This PR provides a new option to display equipment as accordion in Locations cards, also giving the possibility to promote the main item of the equipment (widgetOrder == 0) as the accordion opening item (or equipment with a single point), while keeping the functionality of the control in the list item:

Screen Recording 2022-02-09 at 14 11 21
 

This is an improvement when having very long list of equipment in one location, as it give a fast access to each point, while not having to scroll a lot and be overwhelmed by the quantity of items. A good selection of the main item for each equipment and selection of the 'promote' option further increases speed of use.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
@@ -24,20 +24,20 @@ export const OhHomePageDefinition = () => new WidgetDefinition('oh-home-page', '
 const EquipmentListParameterGroup = () => pg('equipmentList', 'Equipment List', 'General settings for equipment lists in this tab')
 
 const EquipmentListParameters = () => [
-  pt('eqptNesting', 'Display sub-equipment levels')
+  pt('equipmentNesting', 'Display sub-equipment levels')
     .o([
       { value: 'nested', label: 'Nested pages for sub-equipment (default)' },
       { value: 'accordion', label: 'Equipment items grouped as accordion cards' }
     ]),
-  pb('eqptPromoteSingle', 'Promote single item equipments', 'Display groups with a single Point as a single item'),
-  pb('eqptPromoteMain', 'Promote the main item of an equipment', 'Promote the main item of an equipment (widgetOrder equal to 0) as the equipment representation'),
-  pt('eqptPromotedLabel', 'Label promoted elements', 'Choose what elements to display in label of promoted items')
+  pb('equipmentPromoteSingle', 'Promote single item equipments', 'Display groups with a single Point as a single item'),
+  pb('equipmentPromoteMain', 'Promote the main item of an equipment', 'Promote the main item of an equipment (widgetOrder equal to 0) as the equipment representation'),
+  pt('equipmentPromotedLabel', 'Label promoted elements', 'Choose what elements to display in label of promoted items')
     .o([
       { value: 'equipment', label: 'Label of the equipment' },
       { value: 'separator', label: 'Separator character (>)' },
       { value: 'item', label: 'Label of the item within the equipment' }
     ]).m().v((value, configuration, configDescription, parameters) => {
-      return configuration.eqptPromoteSingle === true || configuration.eqptPromoteMain === true
+      return configuration.equipmentPromoteSingle === true || configuration.equipmentPromoteMain === true
     })
 ]
 
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
@@ -29,15 +29,17 @@ const EquipmentListParameters = () => [
       { value: 'nested', label: 'Nested pages for sub-equipment (default)' },
       { value: 'accordion', label: 'Equipment items grouped as accordion cards' }
     ]),
-  pb('equipmentPromoteSingle', 'Promote single item equipments', 'Display groups with a single Point as a single item'),
-  pb('equipmentPromoteMain', 'Promote the main item of an equipment', 'Promote the main item of an equipment (widgetOrder equal to 0) as the equipment representation'),
-  pt('equipmentPromotedLabel', 'Label promoted elements', 'Choose what elements to display in label of promoted items')
+  pb('equipmentPromoteSingle', 'Promote single Points', 'Flatten equipment which only have a single Point (only display the Point as the Equipment\'s representation')
+    .v((value, configuration, configDescription, parameters) => { return configuration.equipmentNesting === 'accordion' }),
+  pb('equipmentPromoteMain', 'Promote the main Point of an Equipment', 'Promote the main Point item of an Equipment (widgetOrder equal to 0) as the Equipment\'s representation')
+    .v((value, configuration, configDescription, parameters) => { return configuration.equipmentNesting === 'accordion' }),
+  pt('equipmentPromotedLabel', 'Label for promoted elements', 'Choose what to include in the labels of promoted items')
     .o([
       { value: 'equipment', label: 'Label of the equipment' },
       { value: 'separator', label: 'Separator character (>)' },
       { value: 'item', label: 'Label of the item within the equipment' }
     ]).m().v((value, configuration, configDescription, parameters) => {
-      return configuration.equipmentPromoteSingle === true || configuration.equipmentPromoteMain === true
+      return (configuration.equipmentPromoteSingle === true || configuration.equipmentPromoteMain === true) && configuration.equipmentNesting === 'accordion'
     })
 ]
 
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
@@ -21,6 +21,32 @@ export const OhHomePageDefinition = () => new WidgetDefinition('oh-home-page', '
       ]).m()
   ])
 
+const EquipmentListParameterGroup = () => pg('equipmentList', 'Equipment List', 'General settings for equipment lists in this tab')
+
+const EquipmentListParameters = () => [
+  pt('eqptNesting', 'Display sub-equipment levels')
+    .o([
+      { value: 'nested', label: 'Nested pages for sub-equipment (default)' },
+      { value: 'accordion', label: 'Equipment items grouped as accordion cards' }
+    ]),
+  pb('eqptPromoteSingle', 'Promote single item equipments', 'Display groups with a single Point as a single item'),
+  pb('eqptPromoteMain', 'Promote the main item of an equipment', 'Promote the main item of an equipment (widgetOrder equal to 0) as the equipment representation'),
+  pt('eqptPromotedLabel', 'Label promoted elements', 'Choose what elements to display in label of promoted items')
+    .o([
+      { value: 'equipment', label: 'Label of the equipment' },
+      { value: 'separator', label: 'Separator character (>)' },
+      { value: 'item', label: 'Label of the item within the equipment' }
+    ]).m().v((value, configuration, configDescription, parameters) => {
+      return configuration.eqptPromoteSingle === true || configuration.eqptPromoteMain === true
+    })
+]
+
+export const OhLocationsTabParameters = () => new WidgetDefinition('oh-locations-tab', 'Locations Tab', 'The tab showing all locations of the installation')
+  .paramGroup(EquipmentListParameterGroup(), EquipmentListParameters())
+
+export const OhEquipmentTabParameters = () => new WidgetDefinition('oh-equipment-tab', 'Equipment Tab', 'The tab showing all equipment of the installation, by category')
+  .paramGroup(EquipmentListParameterGroup(), EquipmentListParameters())
+
 const ModelCardParameterGroup = () => pg('card', 'Model Card', 'General settings for this card')
 
 const ModelCardParameters = () => [
--- a/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
@@ -30,6 +30,7 @@ import ModelCard from './model-card.vue'
 
 export default {
   mixins: [mixin, CardMixin],
+  props: ['tabContext'],
   components: {
     ModelCard
   },
--- a/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
@@ -24,7 +24,7 @@
 
 <script>
 import mixin from '@/components/widgets/widget-mixin'
-import itemDefaultListComponent from '@/components/widgets/standard/list/default-list-item'
+import itemDefaultListComponent, { equipmentListComponent } from '@/components/widgets/standard/list/default-list-item'
 import CardMixin from './card-mixin'
 import ModelCard from './model-card.vue'
 
@@ -36,31 +36,9 @@ export default {
   },
   computed: {
     listContext () {
-      const standaloneEquipment = this.element.equipment.filter((i) => i.points.length === 0).map((i) => itemDefaultListComponent(i.item, this.itemPathLabel(i.item)))
-      const equipmentWithPoints = this.element.equipment.filter((i) => i.points.length !== 0).map((i) => {
-        return [
-          {
-            component: 'oh-list-item',
-            config: {
-              title: [this.itemPathLabel(i.item), i.item.label || i.item.name].filter((label) => label && label.length > 0).join(' > '),
-              divider: true
-            }
-          },
-          ...i.points.map((p) => itemDefaultListComponent(p))
-        ]
-      })
-
       return {
         store: this.$store.getters.trackedItems,
-        component: {
-          component: 'oh-list',
-          config: {
-            mediaList: true
-          },
-          slots: {
-            default: [...standaloneEquipment, ...equipmentWithPoints].flat()
-          }
-        }
+        component: equipmentListComponent(this.element.equipment, this.tabContext, false)
       }
     }
   }
--- a/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
@@ -24,7 +24,7 @@
 
 <script>
 import mixin from '@/components/widgets/widget-mixin'
-import itemDefaultListComponent, { equipmentListComponent } from '@/components/widgets/standard/list/default-list-item'
+import { equipmentListComponent } from '@/components/widgets/standard/list/default-list-item'
 import CardMixin from './card-mixin'
 import ModelCard from './model-card.vue'
 
--- a/bundles/org.openhab.ui/web/src/components/widgets/standard/list/default-list-item.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/standard/list/default-list-item.js
@@ -105,6 +105,8 @@ export default function itemDefaultListComponent (item, itemNameAsFooterOrLocati
   return component
 }
 
+/* The functions below deal with specifically with equipment representation in the home page cards */
+
 export function itemPathLabel (item) {
   if (!item.modelPath) return '(?) > ' + item.name
   return item.modelPath.map((parent) => {
--- a/bundles/org.openhab.ui/web/src/components/widgets/standard/list/default-list-item.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/standard/list/default-list-item.js
@@ -104,3 +104,54 @@ export default function itemDefaultListComponent (item, itemNameAsFooterOrLocati
 
   return component
 }
+
+function promotedEquipmentContext (item, config) {
+  let c = itemDefaultListComponent(item)
+  const parts = (config.eqptPromotedLabel && config.eqptPromotedLabel.length > 0) ? config.eqptPromotedLabel : false
+  c.config.title = [
+    !parts || parts.includes('equipment') ? (item.parent.label || item.parent.name) : null, // Default setting: display parent name
+    parts && parts.includes('separator') ? '>' : null,
+    parts && parts.includes('item') ? (item.label || item.name) : null
+  ].flat().join(' ')
+  return c
+}
+
+export function itemAccordionEquipmentComponent (item, config) {
+  console.log('Accordion equipment context for ' + item.name)
+  if (item.equipmentOrPoints.length === 0) {
+    // Item is a point or equipment without points or sub-equipment
+    return itemDefaultListComponent(item)
+  }
+
+  if (item.equipmentOrPoints.length === 1 && config.eqptPromoteSingle) {
+    // TODO: take into account visibility for promoting single elements (do not count siblings not visible)
+    return promotedEquipmentContext(item.equipmentOrPoints[0], config)
+  }
+
+  // Try to promote main item based on widgetOrder metadata
+  let promoted = config.eqptPromoteMain ? item.points.find((p) => {
+    return p.metadata && p.metadata.widgetOrder && p.metadata.widgetOrder && p.metadata.widgetOrder.value && (+p.metadata.widgetOrder.value) === 0
+  }) : null
+
+  if (promoted) console.log(`Will promote ${promoted.name} for ${item.name}`)
+
+  let c = promoted ? promotedEquipmentContext(promoted, config) : itemDefaultListComponent(item)
+  c.config.action = undefined
+  c.config.after = ''
+  c.slots = {
+    accordion: [
+      {
+        component: 'oh-list',
+        config: {
+          mediaList: true,
+          accordionEquipment: true
+        },
+        slots: {
+          default: item.equipmentOrPoints.filter((i) => { return i !== promoted }).map((i) => itemAccordionEquipmentComponent(i, config))
+        }
+      }
+    ]
+  }
+
+  return c
+}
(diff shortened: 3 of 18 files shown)


ghys:
Thank you very much for this @tarag because it is pretty much what I had in mind for quite some time - I even made some experiments a year ago or so but didn't pursue them further because I had some problems with accordions as you found out yourself.

I'm giving it a quick try and it works pretty well - with a few rendering issues but nothing major, we can address them as we go along - there is something that I can't figure out though (I only looked at the code in passing yet):

The accordions in the location cards seem to work when enabled but not the equipment cards:

image

image

So I'm not sure if i'm missing something.

Also maybe it would be nice to repeat the options on the cards to override those set at the tab level
(not the biggest fan of the "eqpt" prefix in `eqptNesting: accordion` prefix to be honest, `equipmentNesting` looks better IMHO 🙂).

Great work overall though, I think we can push this forward.

> I would like to propose to remove the Equipment | Property top level distinction in Location/Equipment cards. I really don’t see what is the purpose of this or how to explain a user what he will find behind each category. I don’t use Properties at to level in Locations but looking at the openhab demo I really wonder what is the purpose of this distinction in Locations card (I don’t question the purpose of properties in the model of course). @ghys , do you think this distinction is really useful in this part of the UI?

It was implemented only because the model allow this (property -isPartOf-> location) and it's useful in some cases, like for NLP (HABot & co.), when you ask "what's the temperature in the living room?" the Measurement>Temperature property is supposed to take precedence over any such property part of equipment in that room.
Mostly the badges in the location card header respect this as well, but you can't do advanced things on badges like open the analyzer. Are you proposing to remove the properties from the location card or just the segmented button to have them separate?

tarag:
> Thank you very much for this @tarag because it is pretty much what I had in mind for quite some time

Glad you like it, it did not really experiment it other than testing but I really think it’s an improvement. 

> I even made some experiments a year ago or so but didn't pursue them further because I had some problems with accordions as you found out yourself.

The PR doesn’t eventually have lots of code lines for this mechanism but it took quite a few hours of trial and error, head scratching and deciphering f7 code to reach this state. F7 does lots of behind the scene work with events that would prevent reaching the expected result. It does rely on what appears to be some internal API but I don’t think it will change. Maybe we could ask for an evolution of f7 to support this  without fighting against it….

> The accordions in the location cards seem to work when enabled but not the equipment cards

This is not yet implemented. It will rely on the same mixin but there are a few labelling changes to make in addition.  I wanted to show the principle first before continuing. 

> Are you proposing to remove the properties from the location card or just the segmented button to have them separate?

The latter. I perfectly understand the reason to have `Location`-level `Property` ( or `Point` for that matter). What I don’t really understand is why they are separate (at least in the demo example it doesn’t really make sense to me). I’d propose to group them with the `Point` and `Equipment`, or maybe display the non-`Property` `Point`s along.  In that case the designer of the model could use an enclosing equipment to decide where he wants to display the final item. Maybe this separate category makes sense for Location settings that would affect automation linked to the room or multiple equipment  (eg: activate presence simulation, activate automatic light switch on, pair speakers with other rooms). I’m just trying to understand…

> Btw another idea that I had, off-topic for this PR but still, is to have an "advanced" flag for items defined in metadata

I am also thinking a bit about how to handle this. The current solution I use with other account/roles and use of `visibleTo` is not good. 

With this PR I it’s less useful though because these items would not be displayed when the accordion is closed. I also thought defining a sub-equipment labelled « Advanced »  that would gather all these items but it is somehow an abuse of the model and a dedicated metadata flag and dedicated UI control would definitely be better.

tarag:
> Also maybe it would be nice to repeat the options on the cards to override those set at the tab level

I was also thinking about this capability, maybe not necessarily providing the UI for it (just allowing it if defined in YAML). 


> (not the biggest fan of the "eqpt" prefix in eqptNesting: accordion prefix to be honest, equipmentNesting looks better IMHO 🙂).

No problem for the options names, I’ll use the full « equipment »  word.

tarag:
@ghys, did you have a look at the latest commit ? Could you tell me what would be needed to go forward (you said there were some cosmetic issues) ?

Related to our discussion, Points belonging to the model directly under a Location are now displayed under the 'Properties' sub-part of location cards (they would not show-up anywhere before...). So even if I don't necessarily plan to use them I thinks that's an improvement.

ghys:
>>  @ghys, did you have a look at the latest commit ? Could you tell me what would be needed to go forward (you said there were some cosmetic issues) ?

>> The accordions in the location cards seem to work when enabled but not the equipment cards

> This is not yet implemented. It will rely on the same mixin but there are a few labelling changes to make in addition. I wanted to show the principle first before continuing.

I admit I didn't, but I thought it was still a work in progress?
I'll have a fresh look in the next few days!

tarag:
> I admit I didn't, but I thought it was still a work in progress?

There is always room for improvement but I believe it’s quite consistent right now. 

In addition, merging would allow me to tackle the configuration of the labelling that was raised in the community discussions.

tarag:
This should work for equipment (it works for me). Are you sure you fetched the latest commit?

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
```diff
@@ -163,6 +163,7 @@ export default {
           this.$set(this.widgetVars, varKey, this.context.vars[varKey])
         }
       }
+      Object.assign(widget.slots, this.context.component.slots)
```
tarag, line 166: This breaks custom widget loading elsewhere (I have the issue on canvas layers). I did this to allow custom widgets showing in the right part of the accordion trigger. @ghys do you have an idea on how to achieve this without breaking stuff? I'm clearly not at ease with this mechanism of slots, widgets...

tarag:
@ghys, sorry for harassing you 😉 Could you tell me what should be improved before a merge is possible ?

ghys:
@tarag thanks for the update, I just had a look (I have low bandwidth for OH these days) and had to set up my dev environment again.
Unfortunately the location cards don't display at all for me with the latest changes:

image

So it's probably linked to the changes wrt. the card model building in homecards-mixin.js although I have to diagnose why (there's nothing in the console).

In any case since there's clearly potential for regressions it needs to be tested extensively (since there are still no unit/e2e tests although the infrastructure is there).
I'll find what the issue is and get back to you soon.

bundles/org.openhab.ui/web/src/pages/home/homecards-mixin.js
```diff
@@ -64,74 +64,102 @@ export default {
...
+      delete item.semanticLoopDetector
+      item.parent = parent
+      item.children = []
+      item.locations = []
+      item.points = []
+      item.properties = []
+      item.equipment = []
+      item.equipmentOrPoints = []
+
+      if (parent) {
```
ghys, line 82: here you add the item to the `filteredItems` only if it has a parent, which is surely wrong - there can be items at the root level that are part of the card model.

tarag: Thanks for noticing this. I am actively testing this PR (by actually using it) on two installations with hundreds of items and so far I did not have any issue. But in my case there are no top level equipment, points or properties (everything is below the top-level location) so I did let this issue slip through.

I added some test items/groups and it appears to work fine now.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
```diff
@@ -21,6 +21,32 @@ export const OhHomePageDefinition = () => new WidgetDefinition('oh-home-page', '
...
+
+const EquipmentListParameters = () => [
+  pt('equipmentNesting', 'Display sub-equipment levels')
+    .o([
+      { value: 'nested', label: 'Nested pages for sub-equipment (default)' },
+      { value: 'accordion', label: 'Equipment items grouped as accordion cards' }
+    ]),
+  pb('equipmentPromoteSingle', 'Promote single item equipments', 'Display groups with a single Point as a single item'),
+  pb('equipmentPromoteMain', 'Promote the main item of an equipment', 'Promote the main item of an equipment (widgetOrder equal to 0) as the equipment representation'),
+  pt('equipmentPromotedLabel', 'Label promoted elements', 'Choose what elements to display in label of promoted items')
```
ghys, line 34:
```suggestion
  pt('equipmentPromotedLabel', 'Label for promoted elements', 'Choose what to include in the labels of promoted items')
```

bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
```diff
@@ -21,6 +21,32 @@ export const OhHomePageDefinition = () => new WidgetDefinition('oh-home-page', '
...
 
+const EquipmentListParameterGroup = () => pg('equipmentList', 'Equipment List', 'General settings for equipment lists in this tab')
+
+const EquipmentListParameters = () => [
+  pt('equipmentNesting', 'Display sub-equipment levels')
+    .o([
+      { value: 'nested', label: 'Nested pages for sub-equipment (default)' },
+      { value: 'accordion', label: 'Equipment items grouped as accordion cards' }
+    ]),
+  pb('equipmentPromoteSingle', 'Promote single item equipments', 'Display groups with a single Point as a single item'),
```
ghys, line 32:
```suggestion
  pb('equipmentPromoteSingle', 'Promote single Points', 'Flatten equipment which only have a single Point (only display the Point as the Equipment's representation'),
```

bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
```diff
@@ -21,6 +21,32 @@ export const OhHomePageDefinition = () => new WidgetDefinition('oh-home-page', '
...
+  pb('equipmentPromoteMain', 'Promote the main item of an equipment', 'Promote the main item of an equipment (widgetOrder equal to 0) as the equipment representation'),
```
ghys, line 33:
```suggestion
  pb('equipmentPromoteMain', 'Promote the main Point of an equipment', 'Promote the main Point item of an equipment (widgetOrder equal to 0) as the Equipment's representation'),
```

bundles/org.openhab.ui/web/src/assets/definitions/widgets/home/index.js
```diff
@@ -21,6 +21,32 @@ export const OhHomePageDefinition = () => new WidgetDefinition('oh-home-page', '
       ]).m()
   ])
 
+const EquipmentListParameterGroup = () => pg('equipmentList', 'Equipment List', 'General settings for equipment lists in this tab')
+
+const EquipmentListParameters = () => [
+  pt('equipmentNesting', 'Display sub-equipment levels')
+    .o([
+      { value: 'nested', label: 'Nested pages for sub-equipment (default)' },
+      { value: 'accordion', label: 'Equipment items grouped as accordion cards' }
```
ghys, line 30: Wouldn't it be better to display the 3 parameters below only when `equipmentNesting === 'accordion'` witb `v()` (if I understand correctly they are only applicable in that case)?

bundles/org.openhab.ui/web/src/components/cards/equipment-card.vue
```diff
@@ -24,42 +24,21 @@
 
 <script>
 import mixin from '@/components/widgets/widget-mixin'
-import itemDefaultListComponent from '@/components/widgets/standard/list/default-list-item'
+import itemDefaultListComponent, { equipmentListComponent } from '@/components/widgets/standard/list/default-list-item'
```
ghys, line 27: It seems `itemDefaultListComponent` is not used anymore so you can remove it from this import.

bundles/org.openhab.ui/web/src/components/widgets/standard/list/default-list-item.js
```diff
@@ -104,3 +104,92 @@ export default function itemDefaultListComponent (item, itemNameAsFooterOrLocati
 
   return component
 }
+
```
ghys, line 107: This file is about default list item representation but everything below this point is about equipment representation in the home page cards.
IMHO it could use another dedicated file but I'm fine to leave it here if you add a comment about what these functions do.
```suggestion

/* The functions below deal with specifically with equipment representation in the home page cards */
```

bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
```diff
@@ -2,7 +2,7 @@
   <f7-input v-if="!config.item || !config.sendButton" class="input-field" ref="input" v-bind="config" :value="(config.type.indexOf('date') === 0) ? valueForDatepicker : value" @input="$evt => updated($evt.target.value)" :change="updated" @calendar:change="updated" @texteditor:change="updated" @colorpicker:change="updated" />
   <f7-row no-gap v-else class="oh-input-with-send-button">
     <f7-input class="input-field col-90" ref="input" v-bind="config" :value="(config.type.indexOf('date') === 0) ? valueForDatepicker : value" @input="$evt => updated($evt.target.value)" :change="updated" @calendar:change="updated" @texteditor:change="updated" @colorpicker:change="updated" />
-    <f7-button class="send-button col-10" v-if="this.config.sendButton" @click="sendButtonClicked" v-bind="config.sendButtonConfig || { iconMaterial: 'done', iconColor: 'gray' }" />
+    <f7-button class="send-button col-10" v-if="this.config.sendButton" @click.stop="sendButtonClicked" v-bind="config.sendButtonConfig || { iconMaterial: 'done', iconColor: 'gray' }" />
```
ghys, line 5: Could you elaborate why you added "stopPropagation" modifiers on all these widgets? (not that it doesn't make sense I just wanted to know why you made these changes, as it could have potential impacts unrelated to this PR)

bundles/org.openhab.ui/web/src/pages/settings/pages/home/home-edit.vue
```diff
@@ -46,41 +46,59 @@
...
+                :configuration="page.slots.locations[0].config"
+                @updated="dirty = true" />
+            </div>
+
+            <div v-if="currentModelTab === 'equipment'">
+              <config-sheet
+                :parameterGroups="equipmentTabParameters.props.parameterGroups || []"
+                :parameters="equipmentTabParameters.props.parameters || []"
+                :configuration="page.slots.equipment[0].config"
+                @updated="dirty = true" />
```
ghys, line 67: I would put these config sheets below the card list actually (so there is some consistency when switching tabs from Locations & Equipment to Properties).

ghys (changes requested):
Thanks a lot, it seems to work better at least in my case!
I left some comments below:

bundles/org.openhab.ui/web/src/components/widgets/standard/list/oh-list-item.vue
```diff
@@ -15,16 +17,93 @@
...
+    display: block
+    width: 100%
+    height: 100%
+    position: absolute
+    top: 50%
+    color: var(--f7-list-chevron-icon-color)
+
+.media-list .accordion-item-opened  > .item-content > .item-inner > .item-title-row:before
+    content: var(--f7-accordion-chevron-icon-up)
+
```
ghys, line 106: I think these styles should be cleaned up:
- Stylus allows the syntax `property value` instead of `property: value` and it's widely adopted elsewhere so please change accordingly `margin-top: 10px` ➡️ `margin-top 10px`
- `oh-accordion-item` should probably be renamed to something more specific like `oh-equipment-accordion`
- Everything related to equipment accordions should be put beneath a `.oh-equipment-accordion` selector. Generic selectors like `.list ul ul` have the potential of breaking other things since the styles are not scoped to this component.

ghys (changes requested):
One more thing wrt. additional styles in `oh-list-item`

bundles/org.openhab.ui/web/src/components/widgets/standard/list/default-list-item.js
tarag: OK, I'll add this comment. I think the functions belong in this file because there could be updates to other part of the web UI to allow display of an Equipment as accordion in other list contexts.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
tarag: Without these modifiers, the user action performed on the control (toggle, button, stepper...) will at the same time toggle the accordion visibility. So they are definitely necessary, I hope I have not forgotten any.

bundles/org.openhab.ui/web/src/pages/settings/pages/home/home-edit.vue
tarag: I made the change but I think it could lead to a little bit less discoverability. One can think that the page ends with the dynamic list of locations, equipment...

Screenshot 2022-04-23 at 16 59 45

tarag:
Not sure what this not successful check is about. I think everything should be all good now.

bundles/org.openhab.ui/web/src/components/widgets/standard/list/default-list-item.js
ghys: Fair enough, although it would need proper support (namely a "model context") which would then have to be carved out the code dedicated to home page cards.

ghys (approved):
Right, I think we can at last ship this change. The code is clear enough and I believe there aren't any obvious regressions esp. in the behavior of system widgets, but if we missed something it can be dealt with.

To avoid another back-and-forth I finished cleaning up the styles in oh-list-item (putting everything under the dedicated class), and tweaked the loading indicator in the home page designer to avoid flickering, a minor change.

The build failure is not related, the version of HABot's code when you forked the main branch doesn't build with the latest core, it's been fixed since.

Many thanks for all your work and your patience in the last few months!

SR-G:
Upgrading from 3.3.0-M3 to 3.3.0-M4, i have an issue : the "location" cards are now nearly empty and a lot of items are not displayed there anymore (whereas still attached to location groups).

Same with 3.3.0-SNAPSHOT (but one-week ago SNAPSHOT was OK).
If i downgrade to 3.3.0-M3, everything is fine.
So : 
- with 3.3.0-M3, i have all my location cards + all the attached devices (8 cards)
- with 3.3.0-M4, i have only 3 cards (5 are missing)

For example this card is shown without any issues in 3.3.0-M3 : 
openhab-card

Whereas it's not even displayed at all in 3.3.0-M4.

My semantic model is still OK / still the same in both situations (and everything is coming from .items files).

openhab-model

Is there anything to change ? I tried the various new options, without any change on the display of theses location cards.

tarag:
Could you send me your .items file ?

tarag:
It would be helpful to check if anything is displayed in your browser's javascript console.

SR-G:
I just restarted with .M4 : 
```
docker run \
        --name openhab3 \
 (...)
        openhab/openhab:3.3.0.M4-debian    
```
to check chrome console's : there is nothing at all displayed in console.

I don't have any errors in openhab.log, by the way (outside a few WARN like some (for now) empty .things or .rule files) or a few weird things (like an error "cannot instale bindings" (whereas they are already there ...), but not related (and same ones with .M3).
Maybe something is wrong in my .items files (i'm just migrating from OH2 to OH3 and in the end starting everything from scratch) but anyway, they have been built in such a way that they were working with .M3, and they are not working anymore with .M4 as far as i can tell (and i was upgrading to .M4 as there are other issues that i were encountering and that are now solved).

tarag:
If the location tiles where displayed in M3 and not in M4, it could very well be linked to this PR. What would really be helpful is the result of the REST API call of items, including the metadata. You can send it via email, I don't think there is anything too much sensitive (it's the current state of your installation).

SR-G:
Yes, sure.
By the way, indeed, i have 0 evidence that the problem is coming from this PR, but by checking the content of .M4, it seemed to be the most possible one (but maybe i'm wrong).

tarag:
After a quick look to the files you sent me, it seems that the missing location cards between M3 and M4 have no equipment/points, and are therefore not displayed.  Looking at your items, it appears that some equipment have 2 parents: one container Equipment, and one Location. This is for example the case of the following Equipment:

```
    {
        "members": [],
        "link": "http://192.168.8.4:8080/rest/items/gDevice_HueCapteurPresenceEntree",
        "state": "NULL",
        "metadata": {
            "semantics": {
                "value": "Equipment_Sensor",
                "config": {
                    "hasPoint": "HueCapteurEntreeTemperatureBatteryLow",
                    "hasLocation": "gEntree",
                    "isPartOf": "gDevices_Hue"
                }
            }
        },
        "editable": false,
        "type": "Group",
        "name": "gDevice_HueCapteurPresenceEntree",
        "label": "Hue / Entrée / Capteur de présence",
        "category": "sensor",
        "tags": [
            "Sensor"
        ],
        "groupNames": [
            "gEntree",
            "gDevices_Hue"
        ]
    }
```

Therefore it is displayed in the Equipment tab but not in the Location one, because somehow Equipment belonging has precedence over Location belonging.

This PR makes the assumptions that the model to be consistent with the definitions given in the [docs](https://www.openhab.org/docs/tutorial/model.html), and more precisely with these sentences: 
_An Equipment can only be the direct member of zero or one Location, or the direct member of zero or one Equipment. Put another way, an Equipment can only be the direct member of one Group that has a semantic tag._

_A Point can only be the direct member of zero or one Location, or the direct member of zero or one Equipment. Put another way, a Point can only be the direct member of one Group that has a Location or Equipment tag._

@ghys : Are these descriptions no longer accurate ? Not having a single parent will lead to all sorts of difficulties one all my recent PRs... For example, what parent to display in the path of the item, the Equipment or the Location ?

ghys:
> (...) Put another way, an Equipment can only be the direct member of one Group that has a semantic tag.

This is false! You can only have at most one parent **for a given relation type** (_isPartOf_, _isPointOf_, or _hasLocation_, not considering _hasPoint_ because it's not reliable) but it is perfectly valid to have multiple relations when they're not incompatible.
So an Equipment can be part of another superequipment and also have a location (for instance, a lightbulb both part of some central lighting system and also located somewhere), and a Point can be both a point of an equipment and directly related to a location (for instance, a temperature setpoint both part of a thermostat equipment and defining the target temperature of the room itself).

> For example, what parent to display in the path of the item, the Equipment or the Location

In the Location card, I think the equipment path (as you already know the location), and in the Equipment card, the location path.

tarag:
> In the Location card, I think the equipment path (as you already know the location), and in the Equipment card, the location path.

This change, and the others that might be link are not trivial so at least I made a quick fix that restores the way Equipment and Points were previously selected for Location cards. At least it will prevent users from seeing disappearing items in their Location cards, while a more in-depth solution is found.

A new PR has been made: #1385

matchews:
I'm running 3.3.0.M7.  Noticed some issues a couple releases ago, but not sure of the exact version it started.

With the new accordion options disabled, something appears to have disrupted my existing tree structure using the group details action.
image

image

This image shows the members of gHayward are all displayed rather than being hidden until I select it.  This was accessed from the location card.
image

From the overview layout, using a shortcut to the gHayward group, it still mostly works.  It just isn't showing the item to display (water temp mem).
image

Any input is appreciated!

Thanks

tarag:
I did not check again as I'm running everywhere on the milestone 3.3, but my understanding from version 3.2 code is the following. When displayed in location cards, an Equipment would be displayed as a header (divider) and all the points of that equipment would be displayed below, but sub-Equipments would not be displayed at all. To get sub-equipment displayed then the parent equipment should not have had any point children, and in that case the Equipment would be displayed as a list item leading to the group contents (not only the model contents). This is probably what you were used-to from the Hayward OmniLogix equipment since it does not have Point children.

I thought this behaviour was a bug, or at least something only partially implemented so in this PR, I made sure that all the model children of an equipment are displayed in the location cards, wether they are points or sub-equipment. I agree that it changes what you were used to but is this really an issue @matchews? No you get the contents of the OmniLogix equipment with one less click. On the other hand, if you'd prefer not to have everything displayed at the first level then maybe switching to accordion is a good solution for you ? Then we can see how to get the temperature displayed where you'd like.

What do you think about this unexpected change @ghys? Was 3.2 design about sub-equipment on purpose. I don't recall that from our discussion.

ghys:
>  To get sub-equipment displayed then the parent equipment should not have had any point children, and in that case the Equipment would be displayed as a list item leading to the group contents (not only the model contents). This is probably what you were used-to from the Hayward OmniLogix equipment since it does not have Point children.

I think it works even with Points.

I have 2 cases of a location>equipment>sub-equipment>points tree they're still working as expected both with the accordion feature disabled and enabled.
When it's disabled only the top-level equipment are displayed in a flat list with dividers, and the sub-equipment groups are rendered with the default list item widget, which in case of Groups is to open a "group details" popup.
When it's enabled these list items will act as an accordion instead of opening the popup and you get a tree of your model instead.

> image

@matchews if I'm not mistaken, your temperature point above is a member of 2 equipment groups at the same time ("Body of Water" and "Heater") and that is not supported by the semantic model - perhaps that's what causing the problems.
What's supported though is to make it (and others, like the setpoint) a member of the Location group (gPool) and they will then normally appear as badges in the card's header.

matchews:
I've removed the duplicate group memberships on any items with more than one parent with no change in behavior.  The new accordion feature is nice, but still does not show the "Item to Display" as setup in the default list item meta data.  With the new Equipment List/Display sub-equipment levels feature set to either option selected, or no option selected, I no longer see the "Item to display".  Thanks again for the help.

ghys:
Ah OK, I just got what you meant by "item to display", you redefined the 'item' property of the group to display the temperature.
Then your culprit is #1398. For group items like yours it redefine the `after` property to clear the value in an attempt to avoid displaying unnecessary "NULL" states, and it takes precedence over the redefned item's state too.
There's no equivalent workaround I'm afraid at the moment but you might want to set the `badge` property to `=items.YourTempItem.state` for a similar look.

tarag:
@ghys, maybe we can revert #1398 and do the simple and evident correction you suggested in the comments there instead. There might be other people using itemToDisplay in groups, I had not thought of this setting possibility.

@matchews, if you use accordion presentation and also 'promote' the temperature item by setting the widgetOrder to 0 and selecting the right option, you will get the temperature displayed as the header of your heater group.

tarag:
I just issued the corresponding PR #1435.

PR #1313, opened 2022-02-16
Taphold action on oh-label-card and oh-button

tarag:
This PR provides another configurable action upon taphold (or calling contextual menu on desktop) in oh-label-cards and oh-buttons.

This is quite interesting for wall mounted interfaces when using canvas layout and a floor plan. A button controlling a light might do the basic action upon touch (toggle the light), and display a popup with additional actions upon hold (like dimming, color...).


ghys:
Thanks! For completeness sake, could you elaborate how to chose to add this to oh-label-card & oh-button only and not components which perhaps could benefit from this as well?

tarag:
It's just a start to see how things goes (from a usage standpoint). I think these are the most generic components on which actions are used. In addition, there are others like toggle or roller shutter on which I don't think it would be a great choice to have this kind of action.

ghys (approved):
Alright, thanks!

PR #1317, opened 2022-02-18
[MainUI] new HomeKit group mapping editor

yfre:
Complex HomeKit accessories like Thermostats or Fan with Lights are modelled as groups in openHAB and the users have to assign HomeKit characteristics to corresponding group members - openHAB items. 

this can be done already today at item level, i.e. adding HomeKit metadata to items, one after another. 
but an overview at group level was missing. 

this PR adds editor at group level for HomeKit metadata so that user can map HomeKit characteristics to openHAB item on one screen. 

it looks like this. The example shows Fan with Light. 
It lists all characteristics supported by HomeKit for these accessory types. Mandatory are marked with *. User can select the openHAB group item for each characteristics. And press "Update group members" button on the top to add required HomeKit metadata to the openHAB items. 

Group can be save also without this mapping and mapping can be done later at item level as well. 

Overview
Screenshot 2022-02-18 at 11 14 29

Select OH item
Screenshot 2022-02-18 at 11 14 41

please review the code, UI design and wording especially thoroughly as im not a frontend or javascript developer. 

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/metadata/homekit.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/metadata/homekit.js
@@ -1,209 +1,210 @@
-const accessories = {
+export const accessories = {
   'AirQualitySensor': [
-    'AirQuality',
-    'OzoneDensity',
-    'NitrogenDioxideDensity',
-    'SulphurDioxideDensity',
-    'PM25Density',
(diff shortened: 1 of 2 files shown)


ghys (approved):
Sorry for the delay, glad to have some HomeKit love in the UI - I have nothing to say except "great work"!

Issue #1273, opened 2022-01-11
OH 3.2 - 404 error for fonts and icon/number


Daniel-at-git:
## The problem
I get some 404 errors in Chrome Console in basicUI
![image](https://user-images.githubusercontent.com/63794766/149004355-87ea5cba-d60f-4f0c-84e5-e837a1de3788.png)

(basicUI load fine I just saw this errors in console)

## Expected behavior
No 404 errors in console ;-)

## Steps to reproduce
Start Developer Tool (F12) and reload your basicUI

## Your environment

```
root@openhab:~# openhab-cli info

Version:     3.2.0 (Build)

User:        openhab (Active Process 5900)
User Groups: openhab tty dialout audio

Directories: Folder Name      | Path                        | User:Group
             -----------      | ----                        | ----------
             OPENHAB_HOME     | /usr/share/openhab          | openhab:openhab
             OPENHAB_RUNTIME  | /usr/share/openhab/runtime  | openhab:openhab
             OPENHAB_USERDATA | /var/lib/openhab            | openhab:openhab
             OPENHAB_CONF     | /etc/openhab                | openhab:openhab
             OPENHAB_LOGDIR   | /var/log/openhab            | openhab:openhab
```

I'm not an expert but is this the data for webui?
```
jar tf /var/lib/openhab/tmp/kar/openhab-addons-3.2.0/org/openhab/ui/bundles/org.openhab.ui.habpanel/3.2.0/org.openhab.ui.habpanel-3.2.0.jar |grep Roboto
web/fonts/roboto/Roboto-Regular.woff2
web/fonts/roboto/Roboto-Regular.eot
web/fonts/roboto/Roboto-Regular.ttf
web/fonts/roboto/Roboto-Regular.svg
web/fonts/roboto/Roboto-Regular.woff
```
Then we have to add "/roboto" to roboto.css.
But I don't have an idea for 404 on numbers.

Daniel-at-git:
@maggu2810 Is it broken by
https://github.com/openhab/openhab-webui/commit/1a36cd49150eb262c2d05f1b41b94f13af34e161

Should we modify in
bundles/org.openhab.ui.basic/src/main/resources/web/roboto.css
all "fonts/" to "/habpanel/fonts/roboto/" or should Robot-Regular be bundled in basicui?

maggu2810:
Hi @Daniel-at-git,
sorry I am not part of the "game" anymore.
You need to discuss with active members.

[...]

PR #1322, opened 2022-02-19
Fixed browser console error for Roboto-Regular (#1273)

Daniel-at-git:
Modified Roboto-Regular to habpanel
Removed Roboto-Medium. It's not available
Signed-off-by: Daniel Schröter <d.schroeter@gmx.net>


ghys:
Sorry, this introduces implicit dependencies between UIs which is wrong. You can't link resources from one UI to another because you can't expect HABPanel to be installed along with Basic UI. Every UI should be self-sufficient.

PR #1330, opened 2022-03-03
[mainui] Add ability to link Group items

J-N-K:
Related to https://github.com/openhab/openhab-core/issues/257

Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #1335, opened 2022-03-04
Add aggregation function for play/pause

J-N-K:
Follow-Up to #1330


ghys (approved):
LGTM, wasn't aware of those, very valuable, thanks!

PR #1340, opened 2022-03-12
Remove openhab-iconset-classic feature

wborn:
The bundle of this feature is now part of the openhab-runtime-ui feature.

See openhab/openhab-distro#1371


kaikreuzer (approved):
Thanks!

PR #1345, opened 2022-03-22
Remove HABot service configuration

wborn:
There is now a "HABot" service configuration entry without any configuration parameters because those got removed in #1031.

habot


kaikreuzer (approved):
Thanks!

PR #1350, opened 2022-04-08
Use displayState in oh-input

mherwege:

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
@@ -12,6 +12,7 @@ export default () => [
   pb('validate', 'Validate', 'When enabled, input value will be validated based on input type'),
   pb('validate-on-blur', 'Validate on blur', 'Only validate when focus moves away from input field'),
   pi('item', 'Item', 'Link the input value to the state of this item'),
+  pb('displayState', 'Display State', 'Display formatted state as defined in state description'),
   pt('defaultValue', 'Default value', 'Default value when not found in item state or variable'),
   pt('variable', 'Variable', 'Name of the variable to set when the input changes')
 ]
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
@@ -12,7 +12,7 @@ export default () => [
   pb('validate', 'Validate', 'When enabled, input value will be validated based on input type'),
   pb('validate-on-blur', 'Validate on blur', 'Only validate when focus moves away from input field'),
   pi('item', 'Item', 'Link the input value to the state of this item'),
-  pb('displayState', 'Display State', 'Display formatted state as defined in state description'),
+  pb('useDisplayState', 'Use Display State', 'Use the formatted state as the value for the input control'),
   pt('defaultValue', 'Default value', 'Default value when not found in item state or variable'),
   pt('variable', 'Variable', 'Name of the variable to set when the input changes')
 ]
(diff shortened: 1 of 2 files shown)


ghys:
Thanks, but there's no context provided so not sure what to make of this. What's the rationale here? It could be perfectly reasonable to think an input widget would rather be "fed" the actual state and not its displayState even if it exists (you could have another widget to show the displayState alongside the input).
At best this should be made an option - and kept as-is as the default to avoid breaking changes.

mherwege:
@ghys Thank you for your feedback.

I mostly use oh-input for entering things like meter readings. An example: I have an official, non smart, meter for my solar production. I need to manually read the value and communicate the reading to the authorities to receive subsidies for my solar production. Every 1000kWh produced will give me a certificate.
I built some rules that calculate an estimated meter reading continuously to inform me when I approach the 1000kWh value. The calculation is a bit off from the real meter, so I need to adjust it regularly with the real value.
Now the oh-input will show me the calculated value as a float with a lot of digits. This just looks ugly and is not relevant. Using displayState would solve that. It makes using the default widget easy as well. I have strong doubts the full float value is relevant in any use case.
It also allows me to show it in the most appropriate UOM. Input in other UOM remains possible.

See screenshot:
image

You are right this is a breaking change. But I still have to find a use case where displayState would not give me a better value than the raw state. Using a formatting on a free input beyond simple number formatting always will be tricky. E.g. I also use it to do text input to be able to name playlists. Such text does not have any formatting. A simple map transformation on a free form input value is also not likely, as it would be difficult to interpret all possible input values.

I can investigate making this a parameter, but in my view, this adds complexity and the current proposed solution just makes things consistent across the UI. I expected to get the displayState when someting is shown, even in an input field.

mherwege:
@ghys I made displayState optional. While this would break backward compatibility, I would argue the inverse parameter 'forceRawState' would actually make more sense and be more consistent overall. But I am OK with the this solution as well.

ghys:
> You are right this is a breaking change. But I still have to find a use case where displayState would not give me a better value than the raw state.

You can always find some. Imagine a scene where you have a Number item for a scene and a state description mapping like 1=Morning, 2=Evening, 3=Night etc. In such a case for a state of "1" the displayState would be "Morning".
Now someone might choose to have an input to select the scene from 1, 2 or 3. With this change they would have to input (as in type in) "Morning", which is probably not what they want.

mherwege:
> Number item for a scene and a state description mapping like 1=Morning, 2=Evening, 3=Night etc.

I know and understand, but wouldn’t a list item be a much more appropriate input widget in this case? I personally only use an input type widget where there is no better way possible. After all, it is not the most practical widget type on a touch display.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
```diff
@@ -12,6 +12,7 @@ export default () => [
   pb('validate', 'Validate', 'When enabled, input value will be validated based on input type'),
   pb('validate-on-blur', 'Validate on blur', 'Only validate when focus moves away from input field'),
   pi('item', 'Item', 'Link the input value to the state of this item'),
+  pb('displayState', 'Display State', 'Display formatted state as defined in state description'),
```
ghys, line 15:
```suggestion
  pb('useDisplayState', 'Use Display State', 'Use the formatted state as the value for the input control'),
```

bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
```diff
@@ -37,7 +37,11 @@ export default {
       } else if (this.config.sendButton && this.pendingUpdate !== null) {
         return this.pendingUpdate
       } else if (this.config.item && this.context.store[this.config.item].state !== 'NULL' && this.context.store[this.config.item].state !== 'UNDEF') {
-        return this.context.store[this.config.item].state
+        if (this.config.displayState) {
```
ghys, line 40:
```suggestion
        if (this.config.useDisplayState) {
```

ghys (changes requested):
Here's a better example:

image

(regular DateTime item with a state description pattern: `%1$tA, %1$td.%1$tm.%1$tY`)
Default standalone widget:
```yaml
value: oh-input-card
config:
  type: date
  footer: =items.TestDate
  sendButton: true
```
In this case the `displayState` would probably not be parsed if set as the `<input type="date">` value (depends on the pattern), whereas the `state` always will.

Below are some wording suggestions for the new option.

mherwege:
@ghys Thank you for the example and the feedback. I have made the suggested adjustments.

PR #1352, opened 2022-04-13
Upgrade node to v16, npm to v8

ghys:
(DRAFT)

As Node.js v12 is becoming EOL at the end of the month (https://nodejs.org/en/about/releases/), upgrade the version used for the Maven builds (as well as npm) keeping it consistent for all UIs: main, Basic, HABot.

Closes #1165.

PR #1353, opened 2022-04-13
Allow building most bundles on Apple Silicon processors and upgrade tooling.

J-N-K:
Follow-Up to #1215 (can't be re-opened for unknown reasons).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ bin/
 features/*/src/main/history
 npm-debug.log
 node_modules/
+.idea
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ You find the following repository structure:
 
 In most cases, you **don't** need a Java development environment for UI development.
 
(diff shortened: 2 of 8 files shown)


J-N-K:
Trying to figure out if it's HaBOT related

J-N-K:
@ghys How so we proceed here? Shall I remove the upgrade for habot? As I said before, unfortunately I don‘t have enough knowledge to upgrade the dependencies that block the beide upgrade.

ghys:
Right, let's abandon the harmonization of versions and let's keep the HABot build on node v12 then, at least for now.

ghys:
Thanks! LGTM.
One last thing (if you don't mind!), @wborn made 2 pertinent remarks https://github.com/openhab/openhab-webui/pull/1352#pullrequestreview-941885004 & https://github.com/openhab/openhab-webui/pull/1352#pullrequestreview-941923637. Could you address them?

J-N-K:
Sure.

PR #1354, opened 2022-04-18
Include instructions on how to access HABPanel

murfett-au:
Added instructions on how to access HAB Panel as it was not covered elsewhere in the document, and is information I found very useful as a first time user of HABPanel.


murfett-au:
Will re-try with signed commit. Closing.

Issue #551, opened 2020-11-23
[MainUI] Custom logo and CSS


tarag:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=94563140)](https://www.bountysource.com/issues/94563140-mainui-custom-logo-and-css?utm_source=94563140&utm_medium=shield&utm_campaign=ISSUE_BADGE)

## The problem

It would be great to be able to customise the MainUI look using a custom image instead of the openhab logo in the upper-left corner and also a different color scheme.

## Your suggestion

Provide an administrator configuration option to select a file that would be served in conf/html directory for logo and CSS, or use a predefined name.

## Your environment

OH3 M2

hmerk:
What is the purpose of your request.
As this is openHABs main UI, it should show the logo.

kaikreuzer:
I guess the use case is white-labelling - we also had the option for other UIs in the past to add a fragment for customization (own css, assets, etc.). I wouldn't necessarily make it customizable through the UI, but rather statically (like the fragment concept) as the use case is typically for people using it in a commercial context.

Imho not a top prio, but a valid feature request.

tarag:
That's right, but there is also the case where pure  _users_ of openHAB in a house might not be familiar at all about what it is. That's not very meaningful to have the link to the overview page behind this logo. A picture/drawing of the house or some other picture might be more appropriate.

I am not talking about the 'help and about' page where the logo makes more sense of course with the software version.

Concerning the way to do it, of course since it is mostly a one-time modification for a setup, there is no need to have a UI for this but it would be great to be able to do it without rebuilding/patching the distro.

I just thought it would be an easy mod and way to reward through bounty. By the way  @kaikreuzer, there does not seem a way to add the 'bounty' label to issues in this repo. I have other ideas for more substantial ones...

kaikreuzer:
> That's not very meaningful to have the link to the overview page behind this logo.

Well, it is the common practice for websites that clicking on the logo brings you back to the main page.
And showing the openHAB logo here makes it clear that it is openHAB that the user interacts with.
But I agree that it is currently pretty huge and not that helpful, especially when in the side menu on a phone:

<img width="241" alt="Screenshot 2020-11-27 at 09 10 49" src="https://user-images.githubusercontent.com/3244965/100425979-74ec5000-3090-11eb-870e-072f9be4a4c4.png">

[...]

PR #1359, opened 2022-04-25
Allow resource overrides for main UI.

splatch:
This commit solves #551 where user requested a way to customize UI colors or logo.
Now, with OSGi bundle fragment, just like in OH 2.5 it is possible to put new/additional resources.
Just create an OSGi fragment with patch/app/index.html, patch/app/css/xyz.css.
After reload of system, if fragment is matched properly, patched index will take a rule over default one.

### DIFF
--- a/bundles/org.openhab.ui/bnd.bnd
+++ b/bundles/org.openhab.ui/bnd.bnd
@@ -1,2 +1,3 @@
 Bundle-SymbolicName: ${project.artifactId}
 Service-Component: OSGI-INF/*.xml
+Bundle-ClassPath: /patch,.


ghys (approved):
LGTM

digitaldan:
Hey @splatch, i'm not familiar with bundle fragments, do you have an example somewhere that targets either the main ui or paper ui?  Just curious how you target a fragment project in eclipse (i assume thats where you start) to work with a  specific OH bundle.

splatch:
@digitaldan quickest tips I can give you - make sure that module you build have a `Fragment-Host` header pointing to openhab main ui bundle (`org.openhab.ui`). Inside IDE/maven module create a `src/main/resources/patch/app/index.html`, it will then override a main entry point used for browser. The whole trick in this PR is manipulation of bundle classpath so it first looks at `patch/` and then own resources/classes. Since compiled OH main ui is packaged inside app/ you need to make all overrides through src/main/resources/patch/app.

I know there are million of things webpack/browser is doing so customization capabilities are fairly limited. You should be starting from copying of index.html used by UI version you have. With this you can start injecting an extra css and maybe other stuff (haven't tested it to be fair) just like you would do with standard HTML stuff ie. ```<link href="css/extra.css" rel="stylesheet">```. Just remember to place it in ``src/main/resources/patch/app/css/`.

Issue #1349, opened 2022-04-05
[HabPanel] Feature request - Show transformed text of value in Dummy widget


BasvanH:
## The problem
We would like to display the transformed text of a value.

## Your suggestion
Add an checkbox option to display the transformed text of a value.

Issue #1294, opened 2022-02-01
[HabPanel] Feature request - Add color selection to Clock widget


BasvanH:
## The problem
We would like to customize the color of the Clock widget without using CSS.

## Your suggestion
Add color customization just like the Label widget has.

Issue #1293, opened 2022-02-01
[HabPanel] Feature request - Add color selection to Dummy widget


BasvanH:
## The problem
We would like to customize the color of the Dummy widget without using CSS.

## Your suggestion
Add color customization just like the Label widget has.

PR #1380, opened 2022-04-29
Dummy and Clock widget improvements

BasvanH:
Signed-off-by: Bastiaan an Haastrecht b.vanhaastrecht@gmail.com
Adds color customization to Clock and Dummy widget.
Adds option to show state description in Dummy widget.

Closes: #1349
Closes: #1293
Closes: #1294

### DIFF
--- a/bundles/org.openhab.ui.habpanel/doc/habpanel.md
+++ b/bundles/org.openhab.ui.habpanel/doc/habpanel.md
@@ -168,7 +168,7 @@ The following built-in widgets are available:
 
 ![Dummy widget](images/habpanel_widget-dummy.png)
 
-The so-called dummy widget (whose name is explained by historical reasons - it evolved from the first developed widget) displays the current state of an item without any interactivity, along with a label and an optional icon. It has a few options to customize it's appeareance.
+The so-called dummy widget (whose name is explained by historical reasons - it evolved from the first developed widget) displays the current state of an item without any interactivity, along with a label and an optional icon.
 
 #### Switch (switch)
 
@@ -230,7 +230,7 @@ The frame widget displays an external web page in a HTML `<iframe>`.
 
 ![Clock widget](images/habpanel_widget-clock.png)
 
-The clock widget displays an analog or digital clock with color customization. It can also be used to display the current date.
+The clock widget displays an analog or digital clock. It can also be used to display the current date.
 
 #### Chart (chart)
 
--- a/bundles/org.openhab.ui.habpanel/doc/habpanel.md
+++ b/bundles/org.openhab.ui.habpanel/doc/habpanel.md
@@ -168,7 +168,7 @@ The following built-in widgets are available:
 
 ![Dummy widget](images/habpanel_widget-dummy.png)
 
-The so-called dummy widget (whose name is explained by historical reasons - it evolved from the first developed widget) displays the current state of an item without any interactivity, along with a label and an optional icon.
+The so-called dummy widget (whose name is explained by historical reasons - it evolved from the first developed widget) displays the current state of an item without any interactivity, along with a label and an optional icon. It has a few options to customize it's appeareance.
 
 #### Switch (switch)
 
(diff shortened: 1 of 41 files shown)


bundles/org.openhab.ui.habpanel/doc/habpanel.md
```diff
@@ -168,7 +168,7 @@ The following built-in widgets are available:
 
 ![Dummy widget](images/habpanel_widget-dummy.png)
 
-The so-called dummy widget (whose name is explained by historical reasons - it evolved from the first developed widget) displays the current state of an item without any interactivity, along with a label and an optional icon.
+The so-called dummy widget (whose name is explained by historical reasons - it evolved from the first developed widget) displays the current state of an item without any interactivity, along with a label and an optional icon. It has a few options to customize it's appeareance.
```
ghys, line 171:
```suggestion
The so-called dummy widget (whose name is explained by historical reasons - it evolved from the first developed widget) displays the current state of an item without any interactivity, along with a label and an optional icon. It has a few options to customize its appeareance.
```

ghys (commented):
Many thanks, this looks quite good (especially the i18n stuff), a few minor remarks on the docs, but LGTM.

bundles/org.openhab.ui.habpanel/doc/habpanel.md
```diff
@@ -230,7 +230,7 @@ The frame widget displays an external web page in a HTML `<iframe>`.
 
 ![Clock widget](images/habpanel_widget-clock.png)
 
-The clock widget displays an analog or digital clock. It can also be used to display the current date.
+The clock widget displays an analog or digital clock with color customization. It can also be used to display the current date.
```
ghys, line 233: Not sure this addition was necessary to be emphasized here tbh :)

ghys (approved):
Merging this small change as non disruptive, besides it was positively reviewed a while ago and I was just waiting for feedback on my doc changes which I ended up reverting yesterday.

jsiemon:
The changes made between M7 and RC1 seem to have broken the Dummy Widget when implemented  with HabPanel and used on an iPad 3 or earlier.  IPad 4and later seem unaffected.  I believe many, not all, HabPanel users use older hardware, otherwise they would likely use MainUI instead of HabPanel.  I view this as a regression for a feature implemented after the features were frozen.

kaikreuzer:
@ghys Can you confirm that those changes might have increased the minimal browser requirements?
I leave it to your judgement on whether this is something that we can live with or if there's any need to act.

jsiemon:
Whatever changes were made seems only to have negatively affected the Dummy Widget.  The Clock Widget displays fine on the iPad 3.

bundles/org.openhab.ui.habpanel/web/app/widgets/dummy/dummy.widget.js
```diff
@@ -47,7 +47,21 @@
                 vm.value = "N/A";
                 return;
             }
-            var value = item.transformedState || item.state;
+
+            if (vm.widget.usedescription) {
+                var options = item.stateDescription.options;
+                if (Array.isArray(options)) {
+                    var option = options.find((element) => element.value === item.state);
```
ghys, line 54: This is probably the line that is causing problems (arrow functions are iOS Safari 10+, `find` is iOS Safari 8+).
The iPad 3 apparently runs on iOS 9.

Reference
https://caniuse.com/arrow-functions
https://caniuse.com/array-find
https://en.wikipedia.org/wiki/IPad_(3rd_generation)#Software

It could probably be chenged to `options.find((function (element) { return element.value === item.state; })` to make it compatible with iOS 8+.

jsiemon: That would be great if it can be fixed as proposed.  I know this is old hardware, but it works great with HabPanel otherwise and HabPanel is a great legacy  alternative to the new MainUI that does require more modern hardware to run.  Your efforts are greatly appreciated.

ghys: #1433 reimplements the code with techniques found elsewhere (and it seems to work in iOS 6).

image

Thank you for reporting that late minute bug.

PR #1382, opened 2022-05-01
add more datetime support, type-independent log block , CRLF block

stefan-hoehn:
1) Added multiple ZonedDateTime-Blocks based on ideas and implementations of deibich - see https://community.openhab.org/t/date-and-time/131329

- added oh_zdt_now
- added oh_zdt_create
- added many more date patterns that can be detected by the getZonedDateTime()-Method used by oh_zdt_fromText
- added more flexible oh_zdt_amend (flexible way of assembling various combinations of amending a ZDT via oh_zdt_temporal_unit) instead of three specific "copy of" of deibichs idea
- added oh_zdt_temporal_unit and oh_zdt_temporal_unit_field for oh_zdt_amend
- added oh_zdt_fromItem
- added oh_zdt_compare
- added oh_zdt_between
- added oh_get_zdt_part
- added oh_get_time_between
- supports using variables
- rearranged order of the blocks 
- moved persistence specific date blocks to persistence section

Note: set item from deibichs implementation was not ported as this can be done with a normal "send command"
2) OH Blocks are expanded by default
3) Logging-Block allows any input type

Also-by: deibich
Big thanks goes to @tweini within the blockly community.


ghys:
Thank you Stefan, a few remarks as I was testing this PR:

image

I'm not sure "ZDT" as a technical acronym would be understood by novices; there ought to be a more user-friendlier identifiers for them

image

Not sure what this does, I have some idea with my technical background but regular users will IMO surely be confused.

stefan-hoehn:
> Thank you Stefan, a few remarks as I was testing this PR:
> 
> image
> 
> I'm not sure "ZDT" as a technical acronym would be understood by novices; there ought to be a more user-friendlier identifier for them

I agree on the ZDT-acronym. The reason I used it that I took over from deibich's implementation which you liked pretty well. How about replacing it with datetime or even zonedDatetime. 

> image
> 
> Not sure what this does, I have some idea with my technical background but regular users will IMO surely be confused.

I actually showed this several people including @Confectrician in a live demo before completely implementing it. As soon as you we have merged this code I will document it and you will see it is straight forward and easy to use.

How to use it 
- add any ZDT
- open the cog and add as many temporal unit in the mutator
- add one of the two temporal unit blocks (either the flat field or the input, which allows more flexibility, to your block)
- change the type of the temporal unit block with the -> of the block.

https://user-images.githubusercontent.com/5937600/167250758-a4d41020-4e79-4592-85c8-09c9f50619cb.mov

Once you get the hang of it it is really straight forward and pretty easy. I intend to add an animated gif to the docu showing how to use it

Confectrician:
Hi all,

Jumping in here through the mention.
Stefan and I had a video call during the implementation of this and he showed me his thoughts behind it.

Yes. There might be kind of a learning curve.

But when you are a bit familiar with it, you can use one single block in a highly flexible way and just have to combine the same parts in a different way.

ghys:
I agree this is great stuff but I'm really still unconvinced about "ZDT" - it really doesn't mean anything for the casual user (and you can't expect casual users to go check out the openHAB docs to figure that out, I believe especially in Blockly things have to be approachable and tailored for non-technical users) - however if it was spelled out completely as "ZonedDateTime" or even "zoned date time" then they could at least google it and learn about what they're dealing with. Just my opinion ;)

Confectrician:
Hm Stefan agreed in renaming already and is probably only waiting for feedback on his suggestions.

>I agree on the ZDT-acronym. The reason I used it that I took over from deibich's implementation which you liked pretty well. *How about replacing it with datetime or even zonedDatetime.*

My 2 Cents:
I would go for DateTime to keep it simple and probably make some note in the docs/help that it technically is a zoned datetime.

stefan-hoehn:
> I agree this is great stuff but I'm really still unconvinced about "ZDT" - it really doesn't mean anything for the casual user (and you can't expect casual users to go check out the openHAB docs to figure that out, I believe especially in Blockly things have to be approachable and naturally tailored for non-technical users) - however if it was spelled out completely as "ZonedDateTime" or even "zoned date time" then they could at least google it and learn about what they're dealing with. Just my opinion ;)
> 
> Edit: turns out searching for "zoned date time" gives at least relevant results on:
> 
> * Google: https://www.google.com/search?q=zoned+date+time
> * DuckDuckGo: https://duckduckgo.com/?q=zoned+date+time
> 
> So this would have my preference.

As mentioned above  am fine with both date time and zoned date time. As date time is shorter I will change it that asap.

ghys:
> I agree on the ZDT-acronym. The reason I used it that I took over from deibich's implementation which you liked pretty well. How about replacing it with datetime or even zonedDatetime.

Sorry I completely missed that part. I just was focused on the "add temporal units" block video ;)
The difference - I think - between adding a block library that has "ZDT" blocks and having "ZDT" blocks in the built-in library is that in the former case you _choose_ to add that block library so you have a chance to read what it does and what "ZDT" means. In the latter case you don't, you're just presented with "ZDT" and have to figure it out.

stefan-hoehn:
:-)
 Very well so, I will change it to show "datetime" in the blocks then. I like that also much more than ZDT.

stefan-hoehn:
@ghys As the next milestone is around the corner would it be possible for you (or anyone else) to merge my Pull Request?

stefan-hoehn:
Thanks for being so rigid on the review (I looked at it) - I appreciate it.

One question regarding

> :expanded="$f7.device.desktop"

How does that work?

ghys:
I restored the original package.json / package-lock.json as I haven't been able to immediately fix the peer dependencies issues. I'll address that another time in another PR.

As answered on the forum please refrain from updating unrelated dependencies and committing/submitting the result as a PR (or more generally anything unrelated to the PR itself), it avoids some headaches! :)

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
```diff
@@ -13,31 +13,128 @@ export function addOSGiService (serviceName, serviceClass) {
...
+      `    case 19: return ${zdt}.parse(datetime + '+00:00', dtf.ofPattern('yyyy-MM-dd HH:mm:ssz'));`,
+      `    case 25: return ${zdt}.parse(datetime, dtf.ofPattern('yyyy-MM-dd HH:mm:ssz'));`,
+      `    case 23: return ${zdt}.parse(datetime + ' +00:00', dtf.ofPattern('yyyy-MM-dd HH:mm:ss.SSS z'));`,
+      `    case 26: return ${zdt}.parse(datetime + ' +00:00', dtf.ofPattern('yyyy-MM-dd HH:mm:ss.SSSSSS z'));`,
+      `    case 29: return ${zdt}.parse(datetime, dtf.ofPattern('yyyy-MM-dd HH:mm:ss.SSSSz'));`,
+      `    case 32: return ${zdt}.parse(datetime, dtf.ofPattern('yyyy-MM-dd HH:mm:ss.SSSSSSSz'));`,
+      `    case 28: return ${zdt}.parse(datetime.slice(0,26) + ':' + datetime.slice(26,28), dtf.ofPattern('yyyy-MM-dd HH:mm:ss.SSSSz'));`,
+      `    default: return ${zdt}.parse(datetime);`,
+      '  }',
+      '}'
```
ghys, line 62: Honestly (and I missed it the previous time), I'm not the biggest fan of this implementation, especially as you have a block that allows the user to submit arbitrary strings to this function. Moreover it's kind of opaque when you see this big switch/case block added to your script. Surely extracting values with a all-encompassing regular expression then using [`ZonedDateTime.of`](https://howtodoinjava.com/java/date-time/zoneddatetime-class/#22-create-zoneddatetime-with-values) would have been more elegant.
I'll accept still as-is, considering that eventually Nashorn support will be dropped and then we'll be able to use functions from openhab-js (or add support directly there) which would make these kinds of utility functions obsolete.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-dateoffsets.js
```diff
@@ -173,14 +618,193 @@ export default function (f7) {
...
+    switch (temporalPart) {
+      case 'getMilli':
+        temporalPart = `getLong(${chrono}.MILLI_OF_SECOND)`
+        break
+      case 'getMicro':
+        temporalPart = `getLong(${chrono}.MICRO_OF_SECOND) % 1000`
+        break
+      case 'getNano':
+        temporalPart = `getLong(${chrono}.NANO_OF_SECOND) % 1000`
+        break
```
ghys, line 763: I didn't quite understand why the `% 1000` for micro & nano. For nano I seem to always get 0.
In fact I wonder if anything below a millisecond is really necessary, who would ever need them?

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
```diff
@@ -13,31 +13,128 @@ export function addOSGiService (serviceName, serviceClass) {
...
+    'createZonedDateTime', [
+      'function ' + Blockly.JavaScript.FUNCTION_NAME_PLACEHOLDER_ + '(year, month, day, hour, minute, second, nano, offsetString, timezoneString) {',
+      '  stringToParse = \'\' + year;',
+      '  stringToParse += \'-\' + (\'0\' + month).slice(-2);',
+      '  stringToParse += \'-\' + (\'0\' + day).slice(-2);',
+      '  stringToParse += \'T\' + (\'0\' + hour).slice(-2);',
+      '  stringToParse += \':\' + (\'0\' + minute).slice(-2);',
+      '  stringToParse += \':\' + (\'0\' + second).slice(-2);',
+      '  stringToParse += \'.\' + nano + offsetString + \'[\' + timezoneString + \']\';',
+      `  return ${zdt}.parse(stringToParse, dtf.ISO_ZONED_DATE_TIME);`,
```
ghys, line 75: This seems to be an implementation of [`ZonedDateTime.of`](https://howtodoinjava.com/java/date-time/zoneddatetime-class/#22-create-zoneddatetime-with-values) so why not use it directly?

Since these functions are injected when you inject date support whether you need to call them of not, and are quite long, they tend to clutter users' scripts very quickly, so the fewer, the better. Besides we should IMHO use the best possible reference implementation when injecting things into user scripts.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
```diff
@@ -13,31 +13,128 @@ export function addOSGiService (serviceName, serviceClass) {
...
+  let chrono = Blockly.JavaScript.provideFunction_(
+    'chrono',
+    ['var ' + Blockly.JavaScript.FUNCTION_NAME_PLACEHOLDER_ + ' = Java.type("java.time.temporal.ChronoField");'])
+  return chrono
+}
+
+export function addDateComparisonSupport () {
+  let zdtCompare = Blockly.JavaScript.provideFunction_(
+    'zdtCompare', [
+      'function ' + Blockly.JavaScript.FUNCTION_NAME_PLACEHOLDER_ + '(zdt1, zdt2, compareOp, precision, compDate) {',
```
ghys, line 103: Similarly, instead of injecting this huge utility function, since those compareOp/precision/compDate are fields, you could generate the code dynamically, using [temporary variables](https://developers.google.com/blockly/guides/create-custom-blocks/caching-arguments#temporary_variables) to store copies of  `zdt1` & `zdt2` with the "precision" and "compDate" logic, then generate the code that you need dynamically.
In other words this big switchs/cases utility function should have been implemented in the code generation for the blocks instead.

bundles/org.openhab.ui/web/src/pages/settings/rules/script/blockly-editor.vue
```diff
@@ -318,7 +319,7 @@
         </block>
       </category>
 
-      <category name="openHAB" colour="0">
+      <category name="openHAB" colour="0" :expanded="$f7.device.desktop">
```
ghys, line 322: This should only be expanded on desktops so using Framework7 [Device](https://v5.framework7.io/docs/device) property available under the `this.$f7` in Vue components.

ghys (approved):
I'm leaving some notes on the implementation of those big utility functions, and other remarks, but I won't ask you to address those at least not in this PR.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
stefan-hoehn: Not to defend myself but to explain why I did it: The original block
image
was only able to take a this simple date format and I got complaints in the forum that this block was always resetting the time. I then started to think about what would be the most common patterns that someone would use as a string (including of what is being produced by OH itself), so I added all this cases (which was a pretty tedious job and not really fun) like these (and others)

image

I totally agree that the code itself with the switch based on the length of the string isn't really the nicest way of doing it but I wasn't able to do it better myself. I even started doing it via regex but actually failed to do it correctly .

So, yes, I agree the code could probably be done better and hopefully it can be approved eventually.
Thanks for pointing out ZonedDateTime.of . I hadn't known that one. I never stop learning on dates and times...

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
stefan-hoehn: Good hint.

> so why not use it directly?

for two reasons:

1) I took that over from deibichs implementation directly
2) I wasn't aware of that zdt.of

I will see if I can optimize that eventually.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
stefan-hoehn: I will add that to my todo list and see what I can do here.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-dateoffsets.js
stefan-hoehn: First of all I didn't want to deviate from deibich's implementation as he had implemented nanos (and yes, you could of course question whether nanos are really necessary). I then went further because millis and nanos where there but micros was missing, so for consistency reasons I have added them. 

Regarding your note I checked the behaviour with

image

which results into 

>2022-05-30T11:34:26.855407+02:00[Europe/Berlin]
>2022-05-30T11:34:26.855407123+02:00[Europe/Berlin]
>123

So, you are right that now does not include Nanos itself, so we could have omitted them. Sure, you can set and add and subtract them but the question really is how much sense that makes.

Would you vote for removing nanos everywhere? I would then remove them with another PR.

PR #1384, opened 2022-05-02
Add ManagedProvider components for storing configurations

J-N-K:
Depends on https://github.com/openhab/openhab-core/pull/2947

Due to the changes in https://github.com/openhab/openhab-core/pull/2617 storing configurations on server-side now requires a `ManagedProvider`.  

Also-By: Łukasz Dywicki <luke@code-house.org>


ghys (approved):
thanks. there's also a provider needed for the system:sitemap namespace but putting that one in core would be a better fit.

J-N-K:
I'm currently working on a solution that does not require separate service classes for each namespace.

PR #1385, opened 2022-05-03
Fixes the parsing of the model for Location cards following PR #1307.

tarag:
Changes the parsing of the model so that Equipment or Points that are respectively sub-Equipment or Part Of an Equipment while having a Location parent can be displayed in the relevant Location tab card.

This fixes the issue described there: PR https://github.com/openhab/openhab-webui/pull/1307#issuecomment-1115416483

No changes are performed on the labelling.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/home/homecards-mixin.js
+++ b/bundles/org.openhab.ui/web/src/pages/home/homecards-mixin.js
@@ -58,12 +58,12 @@ export default {
       if (!item.metadata || !item.metadata.semantics) return
       if (item.modelPath) return item.modelPath
       let parent = null
-      if (item.metadata.semantics.config && item.metadata.semantics.config.isPointOf) {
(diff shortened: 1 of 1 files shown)


ghys (approved):
Thanks. This is acceptable to fix the immediate problem but ultimately I'd like to have a more thorough discussion on how to determine the "path" of an equipment to display in a card, as the semantic model is an oriented graph but not necessarily a tree as was assumed in #1307 (sorry for not picking up on that when reviewing). So there might be multiple "paths" for an item, but in the end it's not immediately obvious nor trivial how to choose the "best" one to display to the user.

Most users consider the semantic model as a _Location{0..n} > Equipment{0..n} > Point{0..n}_ tree, surely because the Model settings page does display it as such a tree but items may be duplicated there (as pointed out - by me even!):
- https://community.openhab.org/t/item-equipment-or-point/122184/5
- https://community.openhab.org/t/getting-started-with-oh3-rewriting-the-tutorial-6-model-your-home-with-items-link-channels/101008/13

PR #1386, opened 2022-05-06
Adds WebRTC support to video widget

digitaldan:
This adds WebRTC support to the video widget for 3rd party systems that support posting a SDP offer and returning a SDP answer.  There are a number of system this may work with including some nest cameras and RTSPtoWeb (https://github.com/deepch/RTSPtoWeb)  

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-video-webrtc.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-video-webrtc.vue
@@ -4,7 +4,7 @@
     :autoplay="this.startManually ? false : true"
     :controls="this.hideControls ? false : true"
     playsinline
-    style="max-width: 100%; max-height: 100%;">
+    style="max-width: 100%">
     Sorry, your browser doesn't support embedded videos.
   </video>
 </template>
@@ -15,14 +15,12 @@ export default {
   props: {
     src: { type: String },
     stunServer: { type: String },
-    enableTrickleIce: { type: Boolean },
     startManually: { type: Boolean },
     hideControls: { type: Boolean }
   },
   data () {
     return {
-      webrtc: null,
-      isClosed: false // if we closed the webrtc stream
+      webrtc: null
     }
   },
   watch: {
@@ -40,19 +38,20 @@ export default {
   },
   methods: {
     closeConnection () {
-      this.isClosed = true
       if (this.webrtc) {
+        this.webrtc.isClosed = true
         this.webrtc.close()
+        // see https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/close
+        this.webrtc = null
       }
     },
-    startPlay () {
+    startPlay (withTrickleIce = true) {
       this.closeConnection()
-      this.isClosed = false
       if (!this.src) {
         return
       }
       const self = this
-      self.webrtc = new RTCPeerConnection({
+      const webrtc = new RTCPeerConnection({
         iceServers: [
           {
             urls: [self.stunServer || 'stun:stun.l.google.com:19302']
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/components/widgets/system/oh-video-webrtc.vue
```diff
@@ -0,0 +1,112 @@
...
+      this.closeConnection()
+      this.isClosed = false
+      const self = this
+      if (!self.src) {
+        return
+      }
+      self.webrtc = new RTCPeerConnection({
+        iceServers: [
+          {
+            urls: ['stun:stun.l.google.com:19302']
```
ghys, line 57: Thanks.
I'm not familiar with WebRTC so I'm just getting learned on that!

Some good articles:
https://www.html5rocks.com/en/tutorials/webrtc/infrastructure/
https://medium.com/tenable-techblog/using-webrtc-ice-servers-for-port-scanning-in-chrome-ce17b19dd474

This is arguably a good hardcoded default, but would you say it's right not be made overridable?

digitaldan: >I'm not familiar with WebRTC so I'm just getting learned on that!
Its a neat technology but complicated technology ! I had to do a lot with it in a former life. 

Yeah, so 99% of the time everyone uses the google stun servers, although i have run my own coturn server (https://github.com/coturn/coturn) in a commercial setting years ago when needing to support private TURN connections (which is rare these days).  I figured there was little chance to need to change this , but can add this as a "advanced option" for sure if we think it's important!

digitaldan:
Hey @ghys just a quick question, i am noticing that `beforeDestroy` does not get called when i navigate away from the page , i actually have to navigate to 2 pages,  before `beforeDestroy` gets called, for example,  video_page -> page 1 -> page 2.    And whats also odd is that if i click the `< back` navigation on that third page back to the second one,  my video page gets loaded again, even though it is not  visible (page 1 is ) .   This is probably a bug that needs to be opened separately, but any  thoughts there?  My vue skills are close to zero :-)

digitaldan:
@ghys i'm pretty close to done on this one, just doing some testing across devices and networks (cellular, vpn, etc..).

digitaldan:
Just one more comment on the vue/f7 lifecycle methods not getting called as i expected.  As i dig deeper into this i'm realizing this is more complicated due to how vue tries to reuse components?  I came up with a "creative" way to clean up my view when a route change happens, but i'm hoping there is a better way 🙏 .  This involves storing the current route when a component is mounted, and then listening for router events and cleaning up if our route is moved away from.  Feels hacky, but it works. 

```javascript
mounted () {
    console.debug("mounted: ", this.$f7.views.main.router.currentRoute.url)
    this.routeName = this.$f7.views.main.router.currentRoute.url
    this.$f7ready(() => {
      this.$f7.on('routeChange', this.onRouteChange)
    })
  }

...

 onRouteChange (newRoute, previousRoute) {
      console.log("onRouteChange", newRoute, previousRoute);
      if (previousRoute.url === this.routeName) {
        this.closeConnection()
      }
    }
```

FYI we also have the image widget (when using the refresh interval)  as well as the current videojs widget that would need similar logic.  Right now they will keep running even though they are not visible.

digitaldan:
See #1390 for my lifecycle issues

ghys:
Right, sorry for the delay, let's take these lifecycles issues to #1390 then, I responded there.

> Hey @ghys just a quick question, i am noticing that beforeDestroy does not get called when i navigate away from the page

As discussed in #1390 it's not that simple - you have the previous top-level (aka. main view) page still in the DOM for the eventual back transition, and then you have modals (like popups), you can show a sequence of them (this is the demo's model explorer but you can do the same with your pages using Popup actions):

chrome_2022-05-15_20-14-46

I believe all these modals stay in the DOM while you're not done with them.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-video-webrtc.vue
ghys: > Its a neat and complicated technology ! I had to do a lot with it in a former life.

Sure looks like it - I had my share of complicated technologies in former lives but not this one 😄 

> I figured there was little chance to need to change this , but can add this as a "advanced option" for sure if we think it's important!

You're the subject matter expert so if you don't believe it's necessary, then fine! I only wanted to challenge you with it because I suspected users who are adamant about a wholly locally-hosted solution might not appreciate the hardcoded Google server.

digitaldan:
Ready !

ghys (approved):
Thanks!
Merging late to have it in 3.3RC1 because a new option for the video widget is not a fundamental change and has little potential for disrupting anything else. I trust that you tested it yourself and it'll work nicely for others as well!

Issue #1391, opened 2022-05-17
Support for click actions on chart elements


splatch:
## The problem

OH charts are fairly powerful thanks to use of ECharts. Since ECharts do support million of various options, including ones related to events related to mouse or click triggers. In this regard OH does not provide a equal configuration options ie. for map and chart widget.
Currently there is no way to ie. define a popup or detail page for `oh-time-series` or other graph elements.

## Your suggestion
By quick look into code I think echarts/vue-echarts are perfectly fine with multiple events, however chart mixins and components used there doesn't seem to be compatible with `actionsMixin` used in all other places which may react to user actions.

## Your environment
Example yaml which could be utilized as an idea

```yaml
    - component: oh-aggregate-series
      config:
        gridIndex: 0
        xAxisIndex: 0
        yAxisIndex: 0
        type: bar
        dimension1: date
        stack: shared
        name: "Electricity consumption"
        item: G_Total_Electricity
        action: navigate
        actionPage: page:electricity_consumption_details
```

An example of working proof of concept code:
![oh_chart_action](https://user-images.githubusercontent.com/199488/168791407-3c6f44c5-5801-4f56-bfc4-64923e311150.gif)

(obviously difficulty here is that echart is not under a vuejs control)

PR #1392, opened 2022-05-17
Add support for navigate action after click on series data.

splatch:
An attempt to solve #1391 as there are some difficulties in getting a full support for options provided by action handlers in vue components.


ghys:
Thanks, but honestly I would prefer having the entire set of actions available... It's not obvious for the user that only the "navigate" action is working for the chart series, yet the parameters allow them to choose.
Maybe sending a f7 event from the series and then catching them in the oh-chart-component (where `performAction` would be called) could work, but it would require some changes in `performAction` (in the `widget-actions` mixin) to allow passing a config object from another component.

splatch:
Superseded by #1417.

PR #1394, opened 2022-05-19
Addition of page action configuration option to override vars in target page

tarag:
Provides ability to transmit local context upon page switching. The target page is able to react to the calling page via the contents of the variable.

For example: 
I have a canvas page with different layers for the ground floor (doors, hvac, lights) which visibility is managed through a var and buttons. It also has a button to navigate to the page for the second floor which has similar layers. Using this evolution I can select the same kind of layer when navigating across these 2 pages by overriding the var defining layer visibility.


ghys (approved):
Good enough, could probably use a doc update but then again, I don't believe `defineVars` at the page level is documented so it goes for both.
Thanks.

PR #1395, opened 2022-05-20
Configuration of context labels in Equipment and Properties card of homepage tabs.

tarag:
Addition of configuration options available are the path in model (joining parents label), the parent label, the item name or nothing. The possibility to trim beginning or end of path is also configurable.
If nothing is selected, will display the footer defined in metadata (if any).

This should hopefully satisfy users who complained about the lack of configurability when the improvements of PR #1264 were introduced.


ghys (approved):
Thanks, LGTM.
I don't remember if I already told you this, but I think it could be great to have these settings on a per-card basis too, because deviating from the tab's default might make sense in some cases (like no context in some cards if the label is enough).

PR #1396, opened 2022-05-20
Fixes item value hidden for equipment accordion button

tarag:
Value was hard coded to avoid the infamous NULL on groups (PR #1285) without having to configure items, but this causes issues for legitimate uses where you'd want to see the value displayed.


ghys (approved):
thx!

PR #1398, opened 2022-05-23
Prevent some undesired 'NULL' labels in lists for un-typed Groups.

tarag:
When non-typed Group items are rendered automatically in lists as oh-label-item, or following listWidget metadata application, the label displayed is 'NULL'. On these un-typed Group items this can be corrected without hiding information side-effect since there will never be a state with any meaning. It will always be NULL.

This PR forces the label not to be displayed.

Combined with last PR #1396, it allows automatic generation of Model UI that makes sense and does not require manual metadata adjustment when equipment are configured as accordion. This also fixes automatically generated hierarchical Group browsing list pages. For typed groups, manual setting of metadata (PR #1285) is still required if needed.


ghys (approved):
Agreed on the principle, should be fine as the group should (hopefully) have a meaningful state only when it has a groupType.
Another option could simply be checking whether `item.state === 'NULL'`.

PR #1401, opened 2022-05-30
[blockly] add help button to OH docs, provide blockly source maps for debugging

stefan-hoehn:
1) Add a help button to the category flyouts that directly opens the respective help page on openhab-docs, e.g. like so

image

This is available for all openhab categories as well as in the list and color category
It will most likely lead the blockly users much quicker to the help pages as the experience in the forum shows that the help link of the context menu of an individual block is easily overlooked.

2) Allow better debugging for blockly code by resolving the source maps for blockly.

It allows much easier development of further blockly functionality as the debugger will now resolve and show the non-minified javascript code when stepping into it.

To be able to use that, source-map-loader (at least 1.1.3) should be installed via NPM.  

Again big thanks goes to @tweini within the blockly community.

### DIFF
--- a/bundles/org.openhab.ui/web/build/webpack.config.js
+++ b/bundles/org.openhab.ui/web/build/webpack.config.js
@@ -86,7 +86,11 @@ module.exports = {
           resolvePath('node_modules/ssr-window')
         ]
       },
-
+      {
+        test: /(blockly\/.*\.js)$/,
+        enforce: "pre",
+        use: ["source-map-loader"],
+      },
       {
         test: /\.vue$/,
         use: 'vue-loader'
--- a/bundles/org.openhab.ui/web/build/webpack.config.js
+++ b/bundles/org.openhab.ui/web/build/webpack.config.js
@@ -89,7 +89,7 @@ module.exports = {
       {
         test: /(blockly\/.*\.js)$/,
         enforce: "pre",
-        use: ["source-map-loader"],
+        use: (env === 'development' || buildSourceMaps) ? ["source-map-loader"] : [],
       },
       {
         test: /\.vue$/,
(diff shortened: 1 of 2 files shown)


ghys:
Wrt. help buttons - fair enough, on a phone it would look like this though:

image

Wouldn't it be better to emphasize the blocks first and move the help button to the end?
Also I think simply "Help" is sufficient as you clearly know which category you're on.

bundles/org.openhab.ui/web/build/webpack.config.js
```diff
@@ -86,7 +86,11 @@ module.exports = {
           resolvePath('node_modules/ssr-window')
         ]
       },
-
+      {
+        test: /(blockly\/.*\.js)$/,
+        enforce: "pre",
+        use: ["source-map-loader"],
+      },
```
ghys, line 93:
```suggestion
      {
        test: /(blockly\/.*\.js)$/,
        enforce: "pre",
        use: (env === 'development' || buildSourceMaps) ? ["source-map-loader"] : [],
      },
```
I have to confirm this suggestion above works but I'd not include source maps in production builds in any case - unless the environment variable SOURCE_MAPS is set, then it would build the source maps for the entire UI, see lines 20 & 44 in these files. Source maps can be heavy in size and easily add megabytes to the openHAB distribution archive.
If `env === 'development'` (i.e. when you use `npm start` to start the UI with the development server) then it's fine to include any source maps.

stefan-hoehn: I totally agree not to have in production.

stefan-hoehn:
I actually wouldn't like to have the button on the end because I am sure it would be overlooked 90% of the time. That was actually my intention that people see it right away. So I'd rather leave it there. I wouldn't mind though, to change them to "Help" only (just wanted to be a bit nicer here ;-) ) - that also makes it shorter then on mobile. I will change them all to "Help" then.

ghys (approved):
Alright - looks good, thanks for all your valuable (and I believe important!) work on Blockly!

Issue #1390, opened 2022-05-14
Vue lifecycle issues 


digitaldan:
While building out an enhanced  video widget i noticed that every time a widget is navigated to the Vue `mounted` lifecycle hook is called.  My understanding is that this  means Vue is creating a new instance of this component into the DOM.  If you navigate 10 times to the same widget, you will see the `mounted` function called 10 times, so there no re-use of the widget.   

Unfortunately `unmounted` is never being called when a widget is routed away from, either in a full page transition or a popup.   This leads to issues:

1. Leak of Vue components in the Dom that are never reused or cleaned up
2. Components that have background services like timers or video streams run indefinitely 
3. Because of 1 and 2, my guess is the UI starts to progressively slow down the longer it is up and used without a browser reload.  

I have a hacky workaround where i have a timer that updates a value for Vue to render in my widget, then another timer that fires if the component  stops rendering , which happens when its no longer visible, then i can at least cleanup, but the widget remains somewhere hidden until a browser reload.   I can't seem to find any other event to trigger a cleanup on.

digitaldan:
Just a revision to this.  This is not happening in all cases, it depends on how you navigate.  `beforeDestroy` does get called on component that were navigated to from within a page and then back out of using the `back` nav button , including closing pop-ups on the page.  

For example, on my main homepage, if i click on a button on the main page that navigates to another page, then click back, `beforeDestroy` gets called, likewise if i trigger a popup to a page, then close it, `beforeDestroy` gets called.  But, if i use the side menu to navigate between pages, then the currently showed component/page does not get notified, and `beforeDestroy` does not get called.  The mixed behavior was confusing me, as it seemed to work and then not. 

So i don't think this is nearly as bad as i originally thought :-) But still would like to know how to fix it .

hubsif:
Hi @digitaldan,
Framework7 keeps one page in the background, i.e. when you navigate to another page the current page is just shifted to background for quicker back navigation. That's why Vue's `beforeDestroy` does not get called before you move two pages forward. Is this maybe what you're experiencing?

[...]

PR #1404, opened 2022-06-04
[WIP] Foreground detection mixin service for widgets

digitaldan:


digitaldan:
Hi @ghys, this is your solution as discussed in #1390 .  Since this is needed for more then just the video widget (image widget as well), i was thinking making it an optional mixin for those types of widgets might be a good option?  I'm importing this now in the WebRTC widget which works nicely.

hubsif:
Good idea to provide this in a mixin 👍. If it works as expected from what I read in #1390, I'm going to use it for my sip widget as well.

digitaldan:
Thanks @hubsif , to implement you do something like this in your widget

```javascript
import foregroundService from '../widget-foreground-service'

export default {
  mixins: [foregroundService],
  name: 'OhVideoWebRTC',
  methods:  {
      startForegroundActivity () {
        this.startPlay()
      },
      stopForegroundActivity () {
        this.stopPlay()
      }
  }
}
```

there is also a variable `isForeground` which is true if it's in the foreground, i use this when a property changes (my src value)  and i need to know if i should start the video or not.

ghys:
Yes, thank you! I had an idea of adding this to the widget mixin at first but subscribing to events unnecessarily was not the right way to go; making it separate is a good option.

A minor remark is that `startForegroundActivity` or `stopForegroundActivity` could be called needlessly, it's not a problem as long as their implementations take this into account - so for instance `startForegroundActivity` should detect if the activity is already started and do nothing in that case.

bundles/org.openhab.ui/web/src/components/widgets/widget-foreground-service.js
```diff
@@ -0,0 +1,47 @@
...
+  data () {
+    return {
+      pageEl: null,
+      inForeground: false
+    }
+  },
+  mounted () {
+    // determine the current page - support: https://caniuse.com/?search=closest
+    this.pageEl = this.$el.closest('.page')
+    const isPopup = this.$el.closest('.popup') !== null
```
ghys, line 11: Could you elaborate on this? I tested the original code with popups and found out that it seemed to work as expected?

digitaldan: Yeah, it was working  for me as well when i had the logic directly in my widget , but when using as a mixin, it was not finding the `page-current` class.

I actually can't test this right now to show some examples, the build seems to be broken?  I rebased to `main` and now I am getting the error when running `npm start`:

```
Module not found: Error: Can't resolve 'source-map-loader' in '/Users/daniel/openhab-main/git/openhab-webui/bundles/org.openhab.ui/web'
``` 
Along with a whole lot of other source-map-loader errors.  I tried blowing away my node modules and reinstalling, and when that did not work, i cloned a clean copy of the repo, did a `npm install` then `npm start` and am seeing the same error.

digitaldan: So it looks like this line https://github.com/openhab/openhab-webui/blob/f1851e9be2c995a35b59fe6171b4b12c8859515c/bundles/org.openhab.ui/web/build/webpack.config.js#L92 which was introduced in commit https://github.com/openhab/openhab-webui/commit/f1851e9be2c995a35b59fe6171b4b12c8859515c  is causing my local build to fail, simply removing this test works FYI.

digitaldan: so back to the topic at hand, if i dump out the `this.$el.closest('.page')` to the console during a popup, you can see it does not contain the `page-current` class.  I'm not sure why thats different now, maybe a loading order issue?

image

digitaldan: >  and npm i source-map-loader should also solve the problem for the time being 

Yeah, that was the first thing i tried, but it led to some package conflict which i thought could happen when your npm cache is out of sync (it was not the issue in this case), so when trying from a fresh clone failed,  I kinda gave up debugging the issue :-)

>That would be another overlook - I'm assuming you want the widget to work when it's in a page presented as a popup.

i have only tried this PR running the widget in popups and in full pages,  but not a page in a popup (or other modals)

>So this "page-current class means the page is already is the foreground" logic is wrong

So does that mean this technique is not viable, or needs to be tweaked in your opinion?

bundles/org.openhab.ui/web/src/components/widgets/widget-foreground-service.js
```diff
@@ -0,0 +1,47 @@
...
+    // start the foreground activity immediately if the page
+    // is already in the foreground when the widget is mounted
+    // or is being shown within a popup
+    if (this.pageEl.classList.contains('page-current') || isPopup) {
+      this.inForeground = true
+      this.startForegroundActivity()
+    }
+    this.$f7.on('pageAfterIn', this.onPageAfterIn)
+    this.$f7.on('pageBeforeOut', this.onPageBeforeOut)
```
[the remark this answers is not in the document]
ghys: > Yeah, that was the first thing i tried, but it led to some package conflict which i thought could happen when your npm cache is out of sync (it was not the issue in this case), so when trying from a fresh clone failed, I kinda gave up debugging the issue :-)

This should probably be addressed asap as it's only useful to @stefan-hoehn at the moment and breaks the build for others.

> So does that mean this technique is not viable, or needs to be tweaked in your opinion?

It's just a matter of making it work in as many combinations of pages & modals as possible. It would appear at mounting time modals are in the DOM under a background div with the `framework7-modals` class (then they're moved elsewhere). There's no need to subscribe to events when in a modal, since they are temporary in nature and destroyed when closed. A caveat is that a modal may be covered by another modal (or navigate to a page while still in the background) but in this case the foreground activity won't stop.
The suggestion above seems to work in most cases:

chrome_2022-06-10_19-46-00

ghys (commented):
Continuing the conversation above with a suggestion:

ghys:
@digitaldan as the feature freeze is imminent would be able to address the minor change above and use the mixin in #1386 in time?

ghys (approved):
@digitaldan as you're busy with the iOS app as discussed I finished this PR to include my earlier remark about modals. This PR can be safely merged because it is not currently used and won't even be compiled in before #1386 uses it.

PR #1406, opened 2022-06-10
Add source-map-loader dependency

ghys:
(the last version that supports webpack 4)
Also enforce the use of the SOURCE_MAPS env variable for the Blockly rule.


ghys:
@stefan-hoehn this means you would have to set the environment variable `SOURCE_MAPS` before even in development, running `npm start`, to get the Blockly source maps. Sorry for the inconvenience but in the end I believe scoping development aids as much as possible to avoid overhead and general hassle is the best way to go :)

PR #1407, opened 2022-06-13
Equipment as accordion click through fixes

tarag:
Fixes for click through custom widgets containing either oh-button or oh-knob used in equipment accordion list. Click through leads to both the accordion opening and the action on the widget upon user action.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-button.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-button.vue
@@ -1,5 +1,5 @@
 <template>
-  <f7-button v-bind="config" @click="clicked" @taphold.native="onTaphold($event)" @contextmenu.native="onContextMenu($event)">
+  <f7-button v-bind="config" @click.stop="clicked" @taphold.native="onTaphold($event)" @contextmenu.native="onContextMenu($event)">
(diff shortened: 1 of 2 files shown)


ghys (approved):
LGTM, thanks!

@tarag bonus unrelated question: normally top-level accordion items should have a hand cursor and some highlighting (in desktop theme) when hovered, like https://v5.framework7.io/vue/accordion#examples or the Things' channel lists, but that's not the case for equipment accordions, is this intentional? It's kind of counterintuitive.

chrome_2022-06-14_20-10-00

tarag:
> LGTM, thanks!

Thanks for the merge!

> @tarag bonus unrelated question: normally top-level accordion items should have a hand cursor and some highlighting (in desktop theme) when hovered, like https://v5.framework7.io/vue/accordion#examples or the Things' channel lists, but that's not the case for equipment accordions, is this intentional? It's kind of counterintuitive.
> 

This is definitely not intentional and must have slipped through the copy of style from the original F7 widget. I'll try to have a look at it. I agree with you that some additional cueing of the fact that there are some items hidden would be good, also for mobile. But the pointer is a start.

By the way, I see that you use colour lights. I still couldn't find a way to avoid the click trough effect while clicking on the colour widget. I've not figured out how the events are handled in this one.

PR #1410, opened 2022-06-17
Blockly datesupport fix

stefan-hoehn:
fix: when providing an item state to getZonedDateTime() it is not detected as string by ecma-script and replace() fails -> convert explicitly to String first

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/script/blockly-editor.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/script/blockly-editor.vue
@@ -369,12 +369,12 @@
           </block>
           <block type="oh_item" />
           <sep gap="48" />
-          <block type="oh_thing" />
           <block type="oh_getthing_state">
             <value name="thingUid">
               <shadow type="oh_thing" />
             </value>
           </block>
+          <block type="oh_thing" />
         </category>
 
         <category name="Timers &amp; Delays">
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/pages/settings/rules/script/blockly-editor.vue
```diff
@@ -368,6 +367,7 @@
               <shadow type="oh_item" />
             </value>
           </block>
+          <block type="oh_item" />
```
ghys, line 370: I would keep this green block where it was because it sort of serves as "header" of the group of block types related to items, the same way the green "thing picker" is the header of the block (or put these at the end of each group).

stefan-hoehn: I hear what you saying and I was thinking about it but I noticed that many people use these and think it is the state but it actually is "not more than a string". I would then go for putting both to the end which is probably a good compromise.

stefan-hoehn:
It seems during the rebase I didn't signoff. How can I amend that later without making things worse?

bundles/org.openhab.ui/web/src/pages/settings/rules/script/blockly-editor.vue
ghys: Thinking about this further I would also put the "send command/update state" block immediately after "get state" as they're somewhat related and probably the most used.

ghys (approved):
No worries about the DCO, I will include it in the squashed commit to main.

PR #1412, opened 2022-06-17
oh-icon: Verify icon property is a string before performing string operations.

### LINKED ISSUE #1411: oh-icon breaks when icon property does not evaluate to a string
## The problem

If the icon property of the oh-icon component evaluates to something other than a string, an error will be generated on the console. Subsequent updates to the data will not cause the value to be re-evaluated.  Thus, if the evaluation error is transient, such as when item state data is slow to be received from the server, the icon will not appear.

The exception logged is one along the lines of:

TypeError: indexOf is not a function. [....]

## Expected behavior

Icon should appear when the item state is updated and the icon property can be properly evaluated. Additionally, an exception shouldn't be logged to the console, as it's not helpful for the end user (that is, the error relates to the UI assuming the icon property is a string, not to the evaluation itself failing).

## Your environment

OH 3.2.x and 3.3.0-SNAPSHOT

## Additional information

oh-icon.vue assumes that the value of the icon property is always a string and performs indexOf() operations without checking first. This assumption can fail when the icon property is set using an expression and the expression results in a value that isn't a string (such as a TypeError due to either an incorrect expression, or when the item state is slow to arrive from the server.

I have a proposed fix and will submit a PR referencing this issue.

hww3:
When the value of the icon property in the oh-icon component does not evaluate to a string, an exception is thrown. Subsequent updates to the underlying data are not evaluated, leaving the icon absent. This change checks the type of the icon property before performing substring operations.

Tested on OH 3.3.0-SNAPSHOT on MacOS and illumos. 

Fixes #1411.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-icon.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-icon.vue
@@ -48,12 +48,14 @@ export default {
     iconType () {
       const icon = (this.context) ? this.config.icon : this.icon
       if (!icon) return 'oh'
+      if (!(typeof icon === 'string' || icon instanceof String)) return 'oh'
       if (icon.indexOf('f7') === 0 || icon.indexOf('material') === 0) return 'f7'
       if (icon.indexOf('if') === 0 || icon.indexOf('iconify') === 0) return 'iconify'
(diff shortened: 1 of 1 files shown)


ghys (approved):
Thank you for the fix!

PR #1416, opened 2022-06-19
[homekit] add support for parameter groups

yfre:
this PR

- adds missing parameter for homekit accessory of type battery
- adds support for parameters groups at complex accessories so that it is clear to which accessory parameter belongs


ghys (approved):
Thanks!

PR #1430, opened 2022-06-24
Optimize SVG images

dilyanpalauzov:
with SVGO - https://github.com/svg/svgo

As result less data will be transferred to the browser.


ghys:
Thanks for this initiative (and #1431), but since images were mostly sourced from different places which probably curated them well enough, applying a blanket optimization pass could have unexpected side effects or other implications. So those would have to be assessed, analyzed and processed which takes time.

florian-h05:
Closing this PR as it went stale.

PR #1435, opened 2022-06-27
Hide 'NULL' group labels in oh-label-item, second take.

tarag:
Hides 'NULL' labels in oh-label-item while not preventing display of labels for Group where the item to display option has been set.

Reverts undesired effect of PR #1398.


ghys (approved):
Thanks, the build failure is unrelated and has been fixed since.

PR #1408, opened 2022-06-15
Provide link styling to list item used as equipment accordion item.

tarag:
Provide link styling for equipment accordion item (pointer and hover colour), probably only for desktop.

PR #1436, opened 2022-06-27
Reverts equipment accordion link styling.

tarag:
Previous change introduced in PR #1408 with link class for equipment as accordion items caused issue in the events handling, preventing promoted widgets from working (oh-toggle).

@ghys, since the initial PR was only cosmetic and only affects desktop platforms, I propose to revert it while I can find a better solution. It would be a pity not to have promoted items working correctly for accordion equipments in 3.3.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/standard/list/oh-list-item.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/standard/list/oh-list-item.vue
@@ -32,6 +32,12 @@
   .item-link .item-inner:after
     transition-duration 300ms
 
+  > .item-content
+    cursor pointer
+
+    &:active
+      background var(--f7-list-link-pressed-bg-color)
+
   .list,
   .block
     margin-top 0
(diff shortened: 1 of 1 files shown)


tarag:
I assume f7 framework does behind the scene event handling when the element is set to 'link' class.

ghys:
> @ghys, since the initial PR was only cosmetic and only affects desktop platforms, I propose to revert it while I can find a better solution. It would be a pity not to have promoted items working correctly for accordion equipments in 3.3.

Afraid you're too late, 3.3 was built last night and is just released. 🤷‍♂️ Feel free to update this PR with a solution since there's no rush.

tarag:
Hi @ghys , could you review please ?

This is not just some styling as this PR fixes interaction on the right widgets of accordion items (otherwise broken due to F7 event handling of link items).

ghys (approved):
Thank you @tarag and apologies for the delay. I tested it and it seems to solve the problem indeed. 👍

PR #1452, opened 2022-07-17
Allow changing log-levels for add-ons

### LINKED ISSUE #1351: Allow changing add-on log-levels
## The problem

Add-on logger levels can't be changed from the UI. See openhab/openhab-webui#1376. Core support was added in https://github.com/openhab/openhab-core/pull/2772.

## Your suggestion

Bildschirmfoto 2022-04-10 um 13 30 56

## Additional information

This was done by changing `addon-details.vue`

```
              <div class="addon-header-actions">
                <f7-preloader v-if="isPending(addon)" color="blue" />
                <f7-button v-else-if="addon.installed" class="install-button" text="Remove" color="red" round small fill @click="openAddonPopup" />
                <f7-button v-else class="install-button" :text="installableAddon(addon) ? 'Install' : 'Add'" color="blue" round small fill @click="openAddonPopup" />
                <f7-link v-if="!isPending(addon) && bindingInfo && bindingInfo.configDescriptionURI" icon-f7="gears" tooltip="Configure Binding" color="blue" :href="'/settings/addons/' + bindingInfo.id + '/config'" round small />
                <select v-if="addon.installed" class="button button-round button-fill color-lime">
                  <option value="TRACE">
                    Log: TRACE
                  </option>
                  <option value="DEBUG">
                    Log: DEBUG
                  </option>
                  <option value="INFO">
                    Log: INFO
                  </option>
                  <option value="WARN">
                    Log: WARN
                  </option>
                  <option value="ERROR">
                    Log: ERROR
                  </option>
                </select>
              </div>
```

I'm pretty sure this is not the correct way to do it, but I have no idea what is the correct way. Also functionality is missing, that's why I didn't make a PR.

J-N-K:
Closes #1351 

Depends on https://github.com/openhab/openhab-core/pull/3045

As discussed the gear wheel is now shown on each add-on page if either a config-description URI or a set of logger-packages is found.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/addons/addon-config.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/addons/addon-config.vue
@@ -116,12 +116,11 @@ export default {
 
     this.$oh.api.get(requestUri).then(data => {
       this.addon = data
-      let configDescriptionURI = this.addon.configDescriptionURI
+      let configDescriptionURI = this.addon.configDescriptionURI || ''
       if (configDescriptionURI) {
         this.$oh.api.get('/rest/config-descriptions/' + configDescriptionURI).then(data2 => {
           this.configDescription = data2
-
-          this.bindingId = this.strippedAddonId
+          this.bindingId = configDescriptionURI.substring(configDescriptionURI.indexOf(':') + 1)
           this.$oh.api.get('/rest/addons/' + this.bindingId + '/config').then(data3 => {
             this.config = data3
           })


J-N-K:
Bildschirmfoto 2022-07-17 um 13 43 52

ghys:
Tried it and I don't see the configuration for the Kodi binding anymore:

image

image

The reason is that the configDescriptionURI is empty (3.3.0 Release) in `/rest/addons/binding-kodi`:

image

Since it appears to work in your screenshot above, I would suspect that the configDescriptionURI _can_ eventually appear in the `/rest/addons` response, but in this case it does not? Otherwise it might be safer to still try to retrieve it from `/rest/bindings` if is identified as a binding, like before.

J-N-K:
Did you try with a snapshot after https://github.com/openhab/openhab-core/pull/3045 was merged?

```
{
  "id": "binding-kodi",
  "label": "Kodi Binding",
  "version": "3.4.0.SNAPSHOT",
  "compatible": true,
  "contentType": "application/vnd.openhab.feature;type=karaf",
  "link": "https://www.openhab.org/addons/bindings/kodi/",
  "author": "openHAB",
  "verifiedAuthor": true,
  "installed": true,
  "type": "binding",
  "configDescriptionURI": "binding:kodi",
  "keywords": "",
  "countries": "",
  "connection": "",
  "properties": {},
  "loggerPackages": [
    "org.openhab.binding.kodi"
  ]
}
```

ghys:
Ah, my bad, I thought the 3.3.0 core was sufficient, I missed you mentioned the dependency to https://github.com/openhab/openhab-core/pull/3045. Will do.

ghys:
Also in the `/settings/things/add` page where you select the binding:

image

The gear icon link to configure the binding (if it has a configDescription) needs to be updated to the new route.
It's important because it's AFAIK still the only way to configure a binding that wasn't added from the distribution.
Also it could make sense to display this icon regardless of whether a configuration exists now since the screen also includes the log level section.

J-N-K:
Currently the "old route" will still work. This will change with [openhab-core#3050](https://github.com/openhab/openhab-core/pull/3050/files) but that's more at the beginning of next year. There I'm also working on reporting manually installed add-ons (i.e. in addons folder) in the REST API, to we can drop the configuration at that place.

ghys:
Yes the route works but it fails to find the add-on it should pull the config from:

When clicking that gear icon besides the Hue entry it navigates to `/settings/addons/hue/config` which fails to render:

image

This is due to a `GET /rest/addons/hue` which results in a 404 error (the former call was `/rest/bindings/hue`).
Now it could be tricky to translate the bindingId (`hue`) we get from the binding list page into an add-on id. It could be `binding-hue` or `marketplace:{something}` or yet something else entirely depending on the underlying add-on service.
I have no good solution for this, and I can only offer to keep the old `binding-config.vue` along with a separate route in `routes.js` for this page. It will not feature the log level sections of the new add-on configuration page, but these can still be accessed from the add-on store.

bundles/org.openhab.ui/web/src/pages/settings/addons/addon-config.vue
```diff
@@ -0,0 +1,142 @@
...
+        <config-sheet
+          :parameter-groups="configDescription.parameterGroups"
+          :parameters="configDescription.parameters"
+          :configuration="config" />
+      </f7-col>
+    </f7-block>
+    <f7-block form v-if="loggerPackages.length > 0" class="service-config block-narrow">
+      <f7-block-title>Add-on log settings</f7-block-title>
+      <f7-col>
+        <f7-list class="col wide">
```
ghys, line 23:
```suggestion
      <f7-col>
        <f7-block-title medium>Add-on log settings</f7-block-title>
        <f7-list class="col wide">
```

bundles/org.openhab.ui/web/src/pages/settings/addons/addon-config.vue
```diff
@@ -0,0 +1,142 @@
...
+        <f7-link @click="save()" v-if="$theme.md" icon-md="material:save" icon-only />
+        <f7-link @click="save()" v-if="!$theme.md">
+          Save
+        </f7-link>
+      </f7-nav-right>
+    </f7-navbar>
+    <f7-block form v-if="configDescription && config" class="service-config block-narrow">
+      <f7-block-title>Add-on configuration</f7-block-title>
+      <f7-col>
+        <config-sheet
```
ghys, line 14:
```suggestion
      <f7-col>
        <f7-block-title medium>Add-on configuration</f7-block-title>
        <config-sheet
```

ghys (changes requested):
Alright, great. Quite unfortunate to have to have 2 separate pages to configure the bindings, but it's only a little overhead and it solves other problems until all add-ons can be configured, so "oh well".

Last cosmetic suggestions below on the section titles in the add-on config page to make them look better. They are also bigger because there might be config description groups in the configuration, which would render group titles on the same level.

image

image

ghys (approved):
OK, LGTM!
Since it's still technically a draft, please mark it ready when you are.

J-N-K:
I would prefer to keep user-facing changes for beginning of next year (i.e. after release), otherwise we'll have the same discussion as in https://github.com/orgs/openhab/teams/architecture-council/discussions/15 again: Not enough visible changes to justify a change in Java version.

bundles/org.openhab.ui/web/src/pages/settings/addons/addon-details.vue
```diff
@@ -204,6 +204,10 @@ export default {
     }
   },
   computed: {
+    showConfig () {
+      console.error('configUri' + this.addon.configDescriptionURI)
```
ghys, line 208: I had another check: can you remove this debugging leftover?

ghys (changes requested):

ghys (approved):

ghys:
This now needs rebasing and fixing the conflicts since #1468 was merged.

J-N-K:
Done. However, to work a regression core needs to be fixed: https://github.com/openhab/openhab-core/pull/3322

hmerk:
As https://github.com/openhab/openhab-core/pull/3322 was merged 3 days ago, could this one be merged as well ?
Would love to try this feature with one of the next SNAPSHOTs ....

hmerk:
@ghys Thanks Yannick !

hmerk:
@J-N-K Great job, working as expected 👍

PR #1458, opened 2022-07-22
fix identification of ruby scripts in editor

ccutrer:
I confused the branches of the if-else, and wrote the condition as if
it was the positive branch. now it's "if there is no module, or it's
not a DSL script, or it's not a ruby script, then mark it as error"

I apologize - I've definitely compiled and tested this locally now, since the ruby side is also submitted now and I can test the full stack.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
@@ -137,7 +137,6 @@ export default {
       currentModuleConfig: {},
       scriptModuleType: null,
       languages: null,
-      moduleError: null,
       script: '',
       mode: '',
       eventSource: null,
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
```diff
@@ -261,9 +261,9 @@ export default {
         }
 
         if (!this.currentModule ||
-            this.currentModule.type.indexOf('script') !== 0 ||
-            (this.currentModule.type === 'jsr223.ScriptedAction' &&
-             this.currentModule.configuration.type === 'application/x-ruby')) {
+            !(this.currentModule.type.indexOf('script') === 0 ||
+             (this.currentModule.type === 'jsr223.ScriptedAction' &&
+              this.currentModule.configuration.type === 'application/x-ruby'))) {
```
ghys, line 266: Since the module types to look for are in the end:
1. `script.ScriptAction`
2. `script.ScriptCondition`
3. `jsr223.ScriptedAction`

I suppose we can simplify those by trying to find `.Script`:

`if (!this.currentModule || this.currentModule.type.indexOf('.Script') < 0)`
(or with a regex of matching this in an array of hardcoded known types)

It makes little sense to scope `jsr223.ScriptedAction` to just Ruby as this module type can be used with other languages too.

(Also please mind the DCO check)

ccutrer: I just removed this check completely. seems like if a module has a script in its configuration, we should just try and show it. no need for additional checks on the module type or script type

ghys: Yes, agreed, no other module type has an unrelated `script` configuration parameter and it would make potential (1st or 3rd party) new module types that accept such a parameter to work instantly.

ghys (approved):

PR #1459, opened 2022-07-30
Fix foreground service with modal pop-ups

digitaldan:

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-foreground-service.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-foreground-service.js
@@ -6,7 +6,7 @@ export default {
     }
   },
   mounted () {
-    const isInModal = this.$el.closest('.framework7-modals')
+    const isInModal = this.$el.closest('.framework7-modals') || this.$el.closest('.modal-in')
     if (isInModal) {
       this.inForeground = true
       this.startForegroundActivity()


digitaldan:
@ghys , so i ran into a situation where when in a pop-up , the foreground service would not work the first time the widget would appear, but it would work subsequent times.  This was super odd.  First thing i tried to debug this was loading the mounted logic in the next `tick` cycle thinking maybe this was a race condition when the widget is first rendered before caching, but that did not work.

The next thing i tried was traversing the page element's parent class list to see what was being used when mounted, so something like

```javascript
let parent = this.$el
while (parent) {
    console.log(parent.className)
    parent = parent.parentNode
} 
```
There was probably a better way of getting these, but 🤷 

When its loads the first time, this is the list of classes up to the parent element
```
09:12:19.655 widget-foreground-service.js?daf4:13 player
09:12:19.655 widget-foreground-service.js?daf4:13 card
09:12:19.655 widget-foreground-service.js?daf4:13 page-content
09:12:19.655 widget-foreground-service.js?daf4:13 page
09:12:19.655 widget-foreground-service.js?daf4:13 popup modal-in
09:12:19.655 widget-foreground-service.js?daf4:13 framework7-root theme-dark
09:12:19.655 widget-foreground-service.js?daf4:13 theme-dark aurora device-pixel-ratio-2 device-desktop device-macos with-modal-popup
```

When it loads the 2nd time and more this is the list of classes
```
09:13:07.655 widget-foreground-service.js?daf4:13 player
09:13:07.655 widget-foreground-service.js?daf4:13 card
09:13:07.655 widget-foreground-service.js?daf4:13 page-content
09:13:07.655 widget-foreground-service.js?daf4:13 page
09:13:07.655 widget-foreground-service.js?daf4:13 popup
09:13:07.655 widget-foreground-service.js?daf4:13 framework7-modals
09:13:07.655 widget-foreground-service.js?daf4:13 framework7-root theme-dark
09:13:07.655 widget-foreground-service.js?daf4:13 
09:13:07.655 widget-foreground-service.js?daf4:13 theme-dark aurora device-pixel-ratio-2 device-desktop device-macos
```

The different is that in the first case, it has this "modal-in" class on it,  but not the "framework7-modals".  And in the second case it's the reverse of that.   Not sure if this is the right fix, but it's working for me. I did try this out in pop ups, pop over, and sheet and it also appears to still work as desired.

ghys:
Alright, if it works for you, then it works for me!
I was pretty sure about all modals being placed under `framework7-modals` (when trying to find a common class that could handle all modals incl. popups, popovers & sheets) but I guess it's more complicated than that after all.

> There was probably a better way of getting these, but 🤷

To put breakpoints in Vue components when debugging, it's a little weird, I hit <kbd>Ctrl-P</kbd> in the Chrome developer tools and type the name of the file, there would usually be several results, and one of them (usually the last one) will be displaying the "real" source.  It's the one with the source mapped from `webpack-internal:///node_modules/babel-loader/...` in the status bar below the code. Then you can go ahead and place your breakpoint and it should be hit. Another technique is to temporarily add `debugger` statements in the code and reload.

ghys (approved):

PR #1461, opened 2022-07-31
Fix alexa power state default attribute

jsetton:


ghys (approved):
Perfect as always, and I trust you on the functional side!

PR #1462, opened 2022-08-06
Fixes the stepper from making phantom commands when a value is outside of the min/max

digitaldan:

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
@@ -16,7 +16,9 @@ export default {
   computed: {
     value () {
       if (this.config.variable) return this.context.vars[this.config.variable]
-      const value = this.toStepFixed(parseFloat(this.context.store[this.config.item].state))
+      let value = this.toStepFixed(parseFloat(this.context.store[this.config.item].state))
+      if (this.config.min !== undefined) value = Math.max(value, this.config.min)
+      if (this.config.max !== undefined) value = Math.min(value, this.config.max)
       return value
     }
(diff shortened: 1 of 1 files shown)


digitaldan:
So this was a hard one to track down.  I have a variable speed pump for my pool that was randomly being sent commands to set it to '0', was driving me a bit crazy as i could not figure out what was sending those.  Turns out it was a stepper widget on my pool page.  

Long story short, i have a min and max value (0 and 3450) set on the stepper widget.  It looks like we pass this config directly to the F7 stepper widget and ignore it when processing updates.  The problem is that:

1. my pump would send "-1" occasionally when its initializing, 
2. We pass -1 to the stepper, who then immediately changes that to 0 to conform to min/max, 
3. this causes a change event on the widget (as if a user did the change) 
4. and so we fire off a command to the item to set it to 0. 

So every time the stepper received a -1, it would fire off,  on its own,  a command to change the item to 0

ghys (approved):
Yeah, that's a common problem with several widgets like these which led to a number of similar reports.

The only event that can be reacted to is usually an _onChange_ event, but there's no indication (AFAIK) if the change was due to the user interacting directly with the control, or because its value was updated due to an external change (the item state being updated in another way). It is fired in both cases.
In the latter case, if the new externally updated state value isn't valid for the control it could adjust to the closest valid value, and that could in turn lead to a new command being sent (some widget code _onChange_ implementations may decide to send the command only when the current item state is different to the new value, but that's the case here).

...it's clearly not ideal. 
But your fix seems to be sound in the stepper's case at least for your use case so LGTM. If it leads to regressions (like new rogue commands being sent due to this change), then we can revisit and decide what to do.

PR #1470, opened 2022-08-21
Change lazy loading config

wborn:
The add-on image downloading is a lot faster when using non-sequential lazy loading.

The images on GitHub are downloaded using HTTP/2 which allows for a lot of concurrency. :slightly_smiling_face: 

See: https://framework7.io/docs/lazy-load#lazy-load-app-parameters

Did you ever test such a configuration @ghys?

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -336,6 +336,11 @@ export default {
         },
         touch: {
           tapHold: true
+        },
+        // Lazy loading settings
+        lazy: {
+          threshold: 50,
+          sequential: false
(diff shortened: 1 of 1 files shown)


ghys (commented):
Wow, I meant to eventually address this slow loading issue and it is really all that was needed? 😅

Last time I checked _raw.githubusercontent.com_ was still using HTTP/1.1 so I planned to bring them back to _[next.]openhab.org_ (there were caching issues with CloudFlare).

I didn't know about this option and it makes sense to allow concurrent loading, of course. I'm not sure why it's disabled by default.

Thanks for the investigation!

ghys (approved):

Issue #1466, opened 2022-08-15
expandable f7-card not getting focus when used in an oh-masonry frameowrk


jsjames:
## See https://community.openhab.org/t/expandable-card-is-blurry-when-expanded-in-oh-masonry/138125/8

It appears that the surrounding <div> around the oh-masonry should include a z-index value in order to place higher on the z stack.  The oh-block operates correctly and has a higher z-index than the background

## Expected behavior

When f7-card is expanded, it should be the focus.

## Your environment

Running Openhab 3.3.0 on raspberry pi. 

```yaml
config:
  label: Scratch
  sidebar: true
blocks:
  - component: oh-block
    config: {}
    slots:
      default:
        - component: oh-grid-cells
          config: {}
          slots:
            default:
              - component: widget:ExpandableCardTest
                config: {}
masonry:
  - component: oh-masonry
    slots:
      default:
        - component: widget:ExpandableCardTest
          config: {}
grid: []
canvas: []
```

```
uid: ExpandableCardTest
tags: []
props:
  parameters: []
  parameterGroups: []
timestamp: Aug 15, 2022, 7:10:06 AM
component: f7-card
config:
  hideNavbarOnOpen: false
  class: lazy no-margin
  expandable: true
  style:
    border-radius: 6px
    height: 9.6em
    width: 17em
slots:
  default:
    - component: f7-card-content
      config:
        class: lazy card-opened-fade-in
        style:
          color: white
          background-image: url("https://source.unsplash.com/category/nature")
          background-size: 100% auto
          background-position: center
          background-repeat: no-repeat
      slots:
        default:
          - component: oh-link
            config:
              class: card-prevent-open
              style:
                margin: 0em 0.25em 0 -0.7em
                color: var(--f7-card-text-color)
                opacity: 0.4
                z-index: 99
              visible: true
              iconF7: power
              iconSize: 22
              round: true
              action: toggle
              actionItem: =props.switchItem
              actionCommand: ON
              actionCommandAlt: OFF
          - component: oh-link
            config:
              visible: true
              class: card-prevent-open
              style:
                margin: 0 0.25em
                color: var(--f7-card-text-color)
                opacity: 0.4
                z-index: 99
              iconF7: gear_alt
              iconSize: 22
              round: true
              action: group
    - component: f7-card-content
      config:
        class: lazy card-opened-fade-out
        style:
          color: white
          background-image: url("https://source.unsplash.com/category/interiors")
          background-size: 17rem 10rem
[...]

PR #1472, opened 2022-08-23
Fix z-index of oh-masonry component (#1466)

jsjames:
This fix sets the z-index for the parent of the oh-masonry component in the oh-layout-page.  This was necessary to ensure that when an f7-card was expanded, the focus was on the expanded card.

Signed-off-by: Jeff James <jeff@james-online.com>

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/layout/oh-layout-page.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/layout/oh-layout-page.vue
@@ -15,7 +15,7 @@
       </f7-block>
 
       <hr v-if="context.editmode">
-      <f7-block v-if="context.component.slots.masonry && context.component.slots.masonry.length">
+      <f7-block v-if="context.component.slots.masonry && context.component.slots.masonry.length" style="z-index: auto !important">
         <oh-masonry
           :context="childContext(context.component.slots.masonry[0])"
           v-on="$listeners" />


bundles/org.openhab.ui/web/src/components/widgets/layout/oh-layout-page.vue
```diff
@@ -15,7 +15,7 @@
       </f7-block>
 
       <hr v-if="context.editmode">
-      <f7-block v-if="context.component.slots.masonry && context.component.slots.masonry.length">
+      <f7-block v-if="context.component.slots.masonry && context.component.slots.masonry.length" :style="{ 'z-index': 'auto !important' }">
```
ghys, line 18: I have to check for side effects, but using the regular `style` property would work just as well:
```suggestion
      <f7-block v-if="context.component.slots.masonry && context.component.slots.masonry.length" style="z-index: auto !important">
```
No need to involve Vue.JS if there's no need for it.

jsjames: I will remove this as I left this in by mistake.  I was trying to model this similar to how the z-index is set in the oh-block.vue ( style="{ 'z-index': (context.editmode) ? 100 - context.parent.component.slots.default.indexOf(context.component) : 'auto !important' }" ), but that broke things in edit mode, so I just hardcoded.

ghys (approved):
Sorry for the delay, LGTM...!

Issue #324, opened 2020-09-07
[MainUI] New sitemaps do not understand "visibility" & "valuecolor"


digitaldan:
When trying to import my legacy sitemap using the New UI's sitemap editor, the lexer is choking on parsing the "valuecolor" attribute

```
Unexpected identifier token: "valuecolor". Instead, I was expecting to see one of the following:
```

An example sitemap entry using value color:
```
Text item=F2_Bedroom valuecolor=[==OFF="GRAY",!=OFF="green"]
Text label="Security [%s]" item=AlarmMode valuecolor=[==0="green", !=0="red"] icon="dd_security"
```

ghys:
Visibility & color rules are not supported yet in "managed" sitemaps :( So they were left out of the grammar as well.

https://github.com/openhab/openhab-core/blob/ab8aa9ed3af42eeccfd091c4a7085b57c4ac9945/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/components/UIComponentSitemapProvider.java#L270

stritti:
I have the same issue transfere my sitemap from 2.5 to 3.0: 
https://github.com/smart-swimmingpool/openhab-config/blob/master/sitemaps/pool.sitemap

sirupflex:
The lack of support for visibility & valuecolor is quite painful and, in my opinion, contradicts a basic rule with OpenHAB. The basic rule is "either do everything in the UI or in text files", otherwise things get very messy and unpredictable with OpenHAB. With the lack of support for these important tags, you force people to maintain existing sitemaps in text files while pages are created in the UI, with no direct visibility of each other.

Please consider the implementation with high priority. In addition, I highly recommend the "visibility" feature to be implemented also for pages in general. If you want to have a user-friendly page that only displays relevant info based on a current status, you do need this feature. Today, you need to list a lot of items on a page but many of them might e.g. be displayed as "UNDEF" and therefore not adding any value to the page, confusing non-geek family members. Please think user-focused here, not technology-focused! OpenHAB 3 has fantastic potential, let's not forget the most important part, i.e. the interface to "general users", i.e. "the rest of the family", otherwise OpenHAB will remain an "IT geek tool". Thank you very much!

ghys:
> In addition, I highly recommend the "visibility" feature to be implemented also for pages in general.

https://www.openhab.org/docs/ui/building-pages.html#widgets-definition-usage

[...]

PR #1487, opened 2022-09-01
Add missing sitemap attributes

mherwege:

This PR adds the possibility to configure sitemap widget parameters so far only configurable through sitemap files. It should allow fully copying sitemap files into the UI and correctly interpreting it, with the possibility to modify in the UI.

This PR depends on [PR#3075](https://github.com/openhab/openhab-core/pull/3075)


mherwege:
[PR#3075](https://github.com/openhab/openhab-core/pull/3075) has been merged, so this PR should now work in a recent snapshot.

ghys:
Many thanks @mherwege for - finally! - implementing this much requested and desired feature.
This looks good at first sight! 👍
But I'll make a thorough test & review in the following days, so thanks for your patience.

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/attribute-details.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <f7-card v-if="widget">
+    <f7-card-content v-if="attributes.length">
+      <f7-list inline-labels sortable @sortable:sort="onSort">
```
ghys, line 4: Nice idea to allow sorting, to enable sortable mode by default you would need:
```suggestion
      <f7-list inline-labels sortable sortable-opposite class="sortable-enable" @sortable:sort="onSort">
```
The handlers display but sorting doesn't quite work well (the indexes should be recomputed after the sorting and the new order is changed afterwards...)

ghys (changes requested):
Thanks @mherwege - here are some remarks and findings after testing:

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/dslUtil.js
```diff
@@ -4,18 +4,30 @@ function writeWidget (widget, indent) {
...
+        if (key === 'item' || Number.isFinite(widget.config[key])) {
+          dsl += widget.config[key]
+        } else if (['mappings', 'visibility', 'valuecolor', 'labelcolor', 'iconcolor'].includes(key)) {
+          dsl += '['
+          const arrayDsl = widget.config[key].map((v) => {
+            // Anything after the first comparator that is a string should be in quotes.
+            // Also quote string if no comparator (i.e. fixed labelcolor or valuecolor).
+            let value = v.substring(0, v.search(/[=<>]/))
+            value += v.substring(v.search(/[=<>]/)).replace(/[A-Za-z][A-Za-z0-9 _-]*/, function (x) { return '"' + x + '"' })
+            return `${value}`
```
ghys, line 24: When I try this:

image

it works properly:

```
Text icon="frontdoor" item=HS1DSZDoorSensor_DoorSensor label="Front Door" valuecolor=[=="CLOSED"="blue",=="OPEN"="red"]
```

However when changed to:

image

which should be valid - https://www.openhab.org/docs/ui/sitemaps.html#label-and-value-colors - _"If an operator is not specified, the operator will default to =="_, the generated code becomes:

```
Text icon="frontdoor" item=HS1DSZDoorSensor_DoorSensor label="Front Door" valuecolor=[CLOSED=""blue"",OPEN=""red""]
```
and parsing chokes on the repeated double quotes around the colors.

The same seems to happen when an item name is specified in the designer e.g. `HS1DSZDoorSensor_LastChange=="Uninitialized"` - extra double quotes are added:
```
        Text icon="frontdoor" item=HS1DSZDoorSensor_DoorSensor label="Front Door" visibility=[HS1DSZDoorSensor_LastChange==""Uninitialized""]
```

In fact the examples in the sitemaps docs should probably all be tested & working.

ghys (changes requested):
One additional comment that was left out:

bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
```diff
@@ -77,23 +84,49 @@ Widgets -> Widget
...
+WidgetVisibilityAttrValue -> %lbracket _ Visibilities _ %rbracket                 {% (d) => d[2] %}
+WidgetColorAttrName -> %widgetcolorattr
+WidgetColorAttrValue -> %lbracket _ Colors _ %rbracket                            {% (d) => d[2] %}
 
 Mappings -> Mapping                                                               {% (d) => [d[0]] %}
   | Mappings _ %comma _ Mapping                                                   {% (d) => d[0].concat([d[4]]) %}
-Mapping -> MappingCommand %equals MappingLabel                                    {% (d) => d[0][0].value.toString() + '=' + d[2][0].value.toString() %}
-MappingCommand -> %number | %identifier | %string
+Mapping -> MappingCommand _ %equals _ MappingLabel                                {% (d) => d[0][0].value.toString() + '=' + d[4][0].value.toString() %}
+MappingCommand -> %identifier | %string
```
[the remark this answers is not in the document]
mherwege: This shouldn't have been removed

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/attribute-details.vue
mherwege: I actually didn't touch this code, but it indeed didn't work. It cost me some effort, but I think I figured out how to make it work. The index as a key doesn't do it, I needed to construct a key that changes when moving an item in the list.

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/dslUtil.js
mherwege: I did one correction, and for me it now works on all examples that are in the documentation on the website. I did have to submit a small PR to correct a syntax error in one of the examples. Can you check again?

mherwege:
@ghys I think I am done now. I wasn't happy on not being able to easily remove an added empty attribute. I now take care of it when converting to DSL and saving the sitemap, removing empty attribute lines.
All other remarks in the review should have been covered as well.

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/dslUtil.js
ghys: It seems there are still some cases where the quotes aren't right (the color names don't have their quotes anymore) :)
(in fact I'd like to ask, why not simply copy what the user is providing verbatim?)

image

outputs:

```
visibility=[Day_Time=="Morning",Temperature>19]
labelcolor=[Temperature==22="green",Last_Update=="Uninitialized"=gray,LastUpdate=="OPEN"=red,22="green","gray"]
```

ghys:
>  I wasn't happy on not being able to easily remove an added empty attribute (the clear button is not shown with an empty input). I now take care of it when converting to DSL and saving the sitemap, removing empty attribute lines.

Yeah that was a problem indeed. It would be better I think to try removing them altogether if the arrays are empty, not just before the save, because when you remove all items for an attribute and switch to the Code tab you get this error that you can't get rid of:
```
Error: Syntax error at line 9 col 92:

     Text icon="window" item=FGDW002WindowSensor_Open label="Window" visibility=[]
                                                                                 ^
Unexpected rbracket token: "]". Instead, I was expecting to see one of the following:
```

bundles/org.openhab.ui/web/src/pages/settings/pages/sitemap/sitemap-edit.vue
```diff
@@ -176,15 +204,16 @@
 <script>
 import SitemapCode from '@/components/pagedesigner/sitemap/sitemap-code.vue'
 import WidgetDetails from '@/components/pagedesigner/sitemap/widget-details.vue'
-import MappingDetails from '@/components/pagedesigner/sitemap/mapping-details.vue'
+import AttributeDetails from '@/components/pagedesigner/sitemap/attribute-details.vue'
 import DirtyMixin from '../../dirty-mixin'
+import { isArray } from 'zrender/lib/core/util'
```
ghys, line 209: This isn't right - zrender is a graphic library used in ECharts, importing it like that would screw up the webpack chunks and introduce an unwanted dependency.
You can just use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray.

ghys (changes requested):

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/attribute-details.vue
ghys: It works great 👍

bundles/org.openhab.ui/web/src/pages/settings/pages/sitemap/sitemap-edit.vue
mherwege: Oops, indeed, that is what I meant to use. Sorry about that.

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/dslUtil.js
mherwege: This was the original code, only for mapping:
```
      } else if (key === 'mappings') {
        dsl += '['
        const mappingsDsl = widget.config.mappings.map((m) =>
          `${m.split('=')[0]}="${m.substring(m.indexOf('=') + 1)}"`
        )
        dsl += mappingsDsl.join(',')
```
Notice it also adds quotes as they come without quotes from the widget.config. I tried generalizing for all, but that makes it more complex. Not sure if it works without these gymnastics.
Anyway, I thing the missing element is to do the quoting not just once, but for all instances of strings, so just missing a 'g' in the search.

mherwege:
> Yeah that was a problem indeed. It would be better I think to try removing them altogether if the arrays are empty, not just before the save, because when you remove all items for an attribute and switch to the Code tab you get this error that you can't get rid of:
> 
> ```
> Error: Syntax error at line 9 col 92:
> 
>      Text icon="window" item=FGDW002WindowSensor_Open label="Window" visibility=[]
>                                                                                  ^
> Unexpected rbracket token: "]". Instead, I was expecting to see one of the following:
> ```
> 
> (to be clear it's not your fault, that was an already existing bug, but maybe it'd be nice to have it solved.)

I now don't create attributes for empty attribute arrays in DSL, but leave the empty lines until saving. That still allows the user to enter something before save. On save, the remaining empty ones will get removed.

ghys (approved):
Alright LGTM now!
Closes #324 - Many thanks again for implementing this after 2 years! 🥳

jimmys01:
Great Work Thank to all that made this possible!
I have some issues.
-Commends in the file make it complain.
-At sliders the  `sendFrequency` attribute is not recognized.
 example
`Slider item=RGB_middle_Speed  sendFrequency=1000 switchSupport minValue=1 maxValue=40 step=2 labelcolor=[RGB_middle_Unreach==CLOSED="red"]`

crcowan:
I find that when the text of the sitemap has a quoted string:
`Text icon="text" item=Picture_Information_CaptionAbstract label="Caption/Abstract" visibility=[Picture_Information_CaptionAbstract!="Not Available"]
              `

That quoted string looses the quotation in the YAML:
`{
                          "component": "Text",
                          "config": {
                            "icon": "text",
                            "item": "Picture_Information_CaptionAbstract",
                            "label": "Caption/Abstract",
                            "visibility": [
                              "Picture_Information_CaptionAbstract!=Not Available"
                            ]
                          }`

This causes the error:
`2023-08-10 15:03:20.358 [WARN ] [omponents.UIComponentSitemapProvider] - Syntax error in visibility rule 'Picture_Information_CaptionAbstract!=Not Available' for widget Text`

mherwege:
@crcowan This should be fixed with the proposed core PR.

PR #1493, opened 2022-09-17
Update oh-icon.md

EngiNearEngiFar:
wrong link for f7 icons -> fixed

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-icon.md
+++ b/bundles/org.openhab.ui/doc/components/oh-icon.md
@@ -30,7 +30,7 @@ Display an openHAB icon
 <PropGroup label="General">
 <PropBlock type="BOOLEAN" name="icon" label="Icon">
   <PropDescription>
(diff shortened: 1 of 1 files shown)


ghys:
Ah yes, thanks for noticing, however this part of the docs are generated from the actual config description which would be here:
https://github.com/openhab/openhab-webui/blob/fe0e2db56f01261673bbaf6f53c2c228d8e66ea3/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/icon.js#L4
Then you have a script to generate the actual docs from these files, I can lead you though it.

EngiNearEngiFar:
This is the first time I make changes. I changed the link in the config and created a new pull request. 
https://github.com/openhab/openhab-webui/pull/1497
After that the script can be executed?

ghys:
> After that the script can be executed?

Normally yes but since the change is small the two PRs you created is all that's needed (a single one would have been best), I'll merge them both with the "small patch exception" so there's no DCO requirement. Please mind the sign-off rules for PRs if you make significant changes in the future.

PR #1495, opened 2022-09-18
[Blockly] Fix date comparison

Confectrician:
The `isEqual` method is not known and therefore produces Errors, when the correspondign block is usec.
According to the zdt docs the `equals` method should be feasable for comparison.

Signed-off-by: Jerome Luckenbach <github@luckenba.ch>


stefan-hoehn:
This would have run into an exception as "zdt.isEqual", which does not exist, was called instead of zdt.equals.

image

Tested by me and works now.

ghys (approved):
All good then - Thanks to you both!

PR #1498, opened 2022-09-20
Thing "actions" config param handling made generic

### LINKED ISSUE #1491: [thing-details] Extend special handling of "actions" parameter group to all bindings
## The problem
Currently, the Thing Details view handles Z-Wave binding special, and presents a list of <b>actions</b> (`<parameter-group name="actions">`) as clickable buttons, with confirmation prompt. 

This behavior is nice and desirable, but **specific to Z-Wave binding** only:

- https://github.com/openhab/openhab-webui/blob/60fd5f3c4636764c1b5908b542f601a95fc6af48/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue#L374-L378

- https://github.com/openhab/openhab-webui/blob/60fd5f3c4636764c1b5908b542f601a95fc6af48/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue#L110-L119

## Your suggestion
Make the abovementioned behavior more **generic** (remove hard-code for Z-Wave, allow **other bindings** to use it).

### Exemplary proposal
Use the "action" behavior IFF the following heuristic match is true:
`groupName == 'actions' && groupLabel === 'Actions' && groupDescription === 'Actions' && action.type === 'BOOLEAN'`
^--- The values have been picked to be backwards-compatible with [existing Zwave binding's implementation](https://github.com/openhab/org.openhab.binding.zwave/blob/6fe69d4b2267cf8657e4c437e505e2eb4151a18b/src/main/java/org/openhab/binding/zwave/internal/ZWaveConfigProvider.java#L195). Assumption is these values would never be i18n'ed and we want to avoid extra conditions. 

_Note:_ A bigger (more semantically correct) change may include changing the Configuration Description DSL to include a new parameter group and/or type (say: `action`) and remove the heuristic matching by group/param name. This would be a OH-core change, though and given this Z-Wave hardcode has been around for ~3y now, seems to be an overkill to introduce now IMHO.

### Rationale
It is customary for some devices to include a "_reset to factory default_" functionality. Z-Wave binding's paradigm of showing such one-off "commands" as actions seems appropriate and would create consistent UX.

## Your environment

```yaml
runtimeInfo:
  version: 3.4.0
```

## Additional information
_Disclaimer:_ I'm currently developing a binding which would benefit from this functionality.

mbronk:
For Main UI's Thing-Details view.
Fixes #1491

Signed-off-by: Mateusz Bronk <bronk.m+gh@gmail.com>

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -118,7 +118,7 @@
                 <div v-html="actionGroup.group.description" />
               </f7-block-footer>
               <f7-list>
-                <f7-list-button v-for="action in actionGroup.actions" :color="(action.verify) ? 'yellow' : 'blue'" :key="action.name" :title="action.label" @click="action.execute()" :tooltip="action.description" />
+                <f7-list-button v-for="action in actionGroup.actions" :color="(action.verify) ? 'yellow' : 'blue'" :key="action.name" :title="action.label" @click="action.execute()" />
               </f7-list>
             </f7-col>
           </f7-block>
@@ -385,6 +385,7 @@ export default {
             this.configDirty = false
             this.thingDirty = false
 
+            // gather actions (rendered as buttons at the bottom)
             let bindingActionsGrouped = this.getBindingActions(this.configDescriptions)
             let bindingActionsNames = bindingActionsGrouped.flatMap(g => g.actions).flatMap(a => a.name)
             this.configDescriptions.parameters = this.configDescriptions.parameters.filter(p => !bindingActionsNames.includes(p.name)) // params except actions
@@ -393,11 +394,13 @@ export default {
             let allActions = bindingActionsGrouped
             this.getUiActions().forEach(uiAction => {
               let existingGroup = allActions.find(g => g.group.name === uiAction.group.name)
-              if (existingGroup) { // existing (binding-side) group found, *prepending* UI actions
+              if (existingGroup) {
+                // existing (binding-side) group found, *prepending* UI actions
                 existingGroup.actions = uiAction.actions.concat(existingGroup.actions)
-                if (uiAction.group.label !== undefined) { existingGroup.group.label = uiAction.group.label }
-                if (uiAction.group.description !== undefined) { existingGroup.group.description = uiAction.group.description }
-              } else { // no action group from binding, adding the UI actions into their own group (appending at the very end)
+                if (uiAction.group.label !== undefined) existingGroup.group.label = uiAction.group.label
+                if (uiAction.group.description !== undefined) existingGroup.group.description = uiAction.group.description
+              } else {
+                // no action group from binding, adding the UI actions into their own group (appending at the very end)
                 allActions = allActions.concat([uiAction])
               }
             })
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -107,16 +107,23 @@
           </f7-col>
         </f7-block>
 
-        <f7-block class="block-narrow" v-if="ready && !error && thingType && thingType.UID.indexOf('zwave') === 0">
-          <f7-col>
-            <f7-block-title>Z-Wave</f7-block-title>
-            <f7-list>
-              <f7-list-button color="blue" title="View Network Map" @click="openZWaveNetworkPopup" />
-              <f7-list-button color="blue" v-for="action in zwaveActions" :key="action.name" :title="action.label" @click="doZWaveAction(action)" />
-            </f7-list>
-          </f7-col>
-          <z-wave-network-popup :opened="zwaveNetworkPopupOpened" @closed="zwaveNetworkPopupOpened = false" />
-        </f7-block>
+        <!-- Actions -->
+        <div v-if="ready && !error">
+          <f7-block class="block-narrow" v-for="actionGroup in configActionsByGroup" :key="actionGroup.group.name">
+            <f7-col>
+              <f7-block-title class="parameter-group-title">
+                {{ actionGroup.group.label }}
+              </f7-block-title>
+              <f7-block-footer class="param-description" v-if="actionGroup.group.description">
+                <div v-html="actionGroup.group.description" />
+              </f7-block-footer>
+              <f7-list>
+                <f7-list-button v-for="action in actionGroup.actions" :color="(action.verify) ? 'yellow' : 'blue'" :key="action.name" :title="action.label" @click="action.execute()" :tooltip="action.description" />
+              </f7-list>
+            </f7-col>
+          </f7-block>
+        </div>
+        <z-wave-network-popup :opened="zwaveNetworkPopupOpened" @closed="zwaveNetworkPopupOpened = false" v-if="ready && !error && thingType && (thingType.UID.indexOf('zwave') === 0) " />
 
         <f7-block class="block-narrow" v-if="ready && thing.editable">
           <f7-col>
@@ -192,6 +199,11 @@ textarea.textual-definition
 .ios .code-popup
   margin-bottom 44px !important
 
+p.action-description
+  margin-top: 0
+  font-size: 80%
+  color: var(--f7-block-footer-text-color)
+
 .code-popup
   width 100%
   position fixed
@@ -234,6 +246,8 @@ textarea.textual-definition
 
 <script>
 import YAML from 'yaml'
+import groupBy from 'lodash/groupBy'
+import map from 'lodash/map'
 
 import ConfigSheet from '@/components/config/config-sheet.vue'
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
```diff
@@ -107,12 +107,12 @@
...
-        <f7-block class="block-narrow" v-if="ready && !error && thingType && thingType.UID.indexOf('zwave') === 0">
+        <f7-block class="block-narrow" v-if="ready && !error && thingType && (thingType.UID.indexOf('zwave') === 0 || configActions.length > 0)">
           <f7-col>
-            <f7-block-title>Z-Wave</f7-block-title>
+            <f7-block-title>Actions</f7-block-title>
             <f7-list>
-              <f7-list-button color="blue" title="View Network Map" @click="openZWaveNetworkPopup" />
-              <f7-list-button color="blue" v-for="action in zwaveActions" :key="action.name" :title="action.label" @click="doZWaveAction(action)" />
+              <f7-list-button color="blue" title="View Network Map" @click="openZWaveNetworkPopup" v-if="thingType.UID.indexOf('zwave') === 0" />
+              <f7-list-button :color="action.verify? 'yellow' : 'blue'" v-for="action in configActions" :key="action.name" :title="action.label" @click="doConfigAction(action)" :tooltip="action.description" />
```
ghys, line 115:
```suggestion
              <f7-list-button :color="action.verify ? 'yellow' : 'blue'" v-for="action in configActions" :key="action.name" :title="action.label" @click="doConfigAction(action)" :tooltip="action.description" />
```

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
```diff
@@ -371,10 +371,12 @@ export default {
             this.configDirty = false
             this.thingDirty = false
 
-            // special treatment for Z-Wave actions
-            if (this.thingType.UID.indexOf('zwave') === 0) {
-              this.zwaveActions = this.configDescriptions.parameters.filter((p) => p.groupName === 'actions')
-              this.configDescriptions.parameters = this.configDescriptions.parameters.filter((p) => p.groupName !== 'actions')
+            // special treatment for actions (heuristic match by group name/label)
+            if (this.configDescriptions.parameterGroups.filter((pg) => pg.name === 'actions').every((pg) => pg.label === 'Actions' && pg.description === 'Actions')) {
```
ghys, line 375: This heuristic is good enough but since parameter groups have a "context" attribute:

https://next.openhab.org/javadoc/latest/org/openhab/core/config/core/dto/configdescriptionparametergroupdto

I'm wondering whether this could perhaps be the way to go - look for e.g. an "actions" context in the parameter group definition instead of checking for labels & descriptions - it certainly would be more elegant IMO.

We could still make an exception for the Z-Wave things unless @cdjackson agrees to add the "actions" context for the actions parameter group for them.

I'm not sure how easy that is to add such contexts to parameter groups, whether in declarative or imperative form.

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
```diff
@@ -443,12 +445,12 @@ export default {
...
       })
     },
-    doZWaveAction (action) {
+    doConfigAction (action) {
       let thing = this.thing
       let save = this.save
       if (action.type !== 'BOOLEAN') return
       this.$f7.dialog.confirm(
-        `${action.label}?`,
+        `${action.description ?? action.label}?${action.verify ? '<p><small><strong class="text-color-yellow">WARNING:</strong>&nbsp;This action may be dangerous!</small></p>' : ''}`,
```
ghys, line 453:
```suggestion
        `${action.description ?? action.label} ? ${action.verify ? '<p><small><strong class="text-color-yellow">WARNING:</strong>&nbsp;This action may be dangerous!</small></p>' : ''}`,
```
I'm not sure I understand this...

ghys (changes requested):
Thanks for the implementation, some remarks (and further discussion) below!

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
mbronk: commited as-is (will squash later) - or feel free to squash if/when merging

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
mbronk: Indeed, there seems to be a "param group context" defined, though not documented too widely. Per [this](context) it would seem it may be a freeform string (the per-param context seems to be a pre-defined list).
Let me try to see how easy it is to set/use and report back (may take me about a week to do so, though)...

I 100% agree it would be more elegant than this heuristic.
That said, I think the ROI is there when ZWave binding also adopts it, otherwise we're back into heuristic-land :). Not sure what'd be the timeline, but how about a 2-step approach? I.e. this heuristic *for now* (possibly with an "or" for a by-context match) and follow through with a proper update if/once ZWave binding changes?

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
mbronk: Oh. and the abovementioned code was not a ternary IF statement! It is just a good old questionmark (to be rendered as-is in the output :) ).

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
ghys: > I.e. this heuristic for now (possibly with an "or" for a by-context match) and follow through with a proper update if/once ZWave binding changes?

Works for me!

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
ghys: > The "actions" behaviour of Main UI does ask for a confirmation for all the actions, of course... so I figured to bring back at least some of the "verify" semantics, and add a notion of "hey, the binding developer considered this dangerous".

Wow I completely forgot this "verify" attribute existed... 🤯

Fully agree with your assessment - this particular line of code I considered quite "convoluted" though with the muliple ternary expressions. If I understand it right it does display the description or label, or when there's none, displays a warning message *if* the verify flag is true? Is that your intent?

mbronk: > or when there's none....

As per my 2nd comment... the `?` here is not a ternary expresssion, actually, rather the same questionmark that existed in the original. The intent is to:
  1.  Display description (and fallback to label if none)
  1. (orthogonal to the above): If `verify==true` --> add an extra warning.

I now see how poorly it reads code-wise (it was much better in IDE, with coloring, etc.). Unfortunately, I have literally never written anything in vue, so while I assume using some templated text/conditional controls would make it more readable, it would take me some extra learning to find the syntax out. Do you happen to have any tips in particular?

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
mbronk: The "by-context" method does work indeed, so I added it as primary.
Any param group constructed with `.withContext("actions")` will now be picked as action container.

As per our discussion, I've left the heuristic match by `name` and `label` for compat with today's Z-wave binding. 
Note I had to pull the "by description" criterion, since Z-Wave doesn't use it uniformly and only sets it on non-controller devices...

--
While at it, I attempted to make it a bit more generic (hopefully the code explains the intent well enough). Do ask if it raises any concerns.

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
mbronk: Extracted to separate variables - hope it'll be more readable now.

ghys:
@mbronk Please review the ESLint errors: https://github.com/openhab/openhab-webui/actions/runs/3165087894/jobs/5209020387
(you can also see _some_ in the Files tab but there's a limit on how many can be displayed per file, so some are missing towards the end of the file in that view).

If you can, use an editor that can highlight ESLint errors, for instance VS Code with the ESLint plugin - it's easier!

mbronk:
@ghys Yeah, sorry about that! 
I don't have any git hooks set up for this repo, and completely forgot to run `npm run lint` manually before pushing... Since the CI only runs after manually approved for 1st time committers... I then have forgotten about it :/ 

Anyways, eslint should be happy now. Could you please approve another CI build?

ghys (approved):
Thanks - LGTM!
I just removed the tooltip with the description on the action buttons because it didn't look too good on mobile IMO, especially with potentially long descriptions. You have that description on the confirmation dialog anyway so it's probably enough...?
Nice new feature!

(build failures were temporary, the build actually succeeded: checks/webpack-stats are green)

PR #1499, opened 2022-09-23
[WIP] Monitoring SSE Connection Health

digitaldan:
Supports monitoring the "alive" message type in the upcoming PR https://github.com/openhab/openhab-core/pull/3086

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -385,6 +385,19 @@ export default {
       return (this.serverUrl || window.location.origin)
     }
   },
+  watch: {
+    '$store.state.states.sseConnected': {
+      handler: function (connected) {
+        console.debug('sseConnected', connected)
+        if (window.OHApp && typeof window.OHApp.sseConnected === 'function') {
+          try {
+            window.OHApp.sseConnected(connected)
+          } catch {}
+        }
+      },
+      immediate: true // provides initial (not changed yet) state
+    }
+  },
   methods: {
     loadData (useCredentials) {
       const useCredentialsPromise = (useCredentials) ? this.setBasicCredentials() : Promise.resolve()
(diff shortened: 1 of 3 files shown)


digitaldan:
Hey @ghys this PR is allows the monitoring of the SSE connection health to wrapping applications, but i did not add anything yet to the app itself to try and reconnect since the browser does a pretty good job of trying to establish the connection on its own.  Wrapping applications may have more knowledge then the browser based on changes to Wifi, LTE or VPN which may require a change in host name / IP and a reload of the app. 

This is the first time i have messed with Vue state management, not sure if thats the right path or not to track connected state.

digitaldan:
Hey @ghys, I did look into reconnection options in the client, i'm not 100% sure how that would work.

Couple questions/options

1. Is that a path we want to continue with? 
2. Or, do we want to simply raise a banner notifying the user of a connection error with the option to reload? 
3. Or do we want to review/merge this with the current logic for now so mobile can take advantage of it, and continue to research a client alert/solution?

Thanks!

ghys (approved):
I think option 3. is fine for now - at least it will print those "SSE timed out" warnings on the console when the connection seems to have dropped, so we could better analyze/debug in which circumstances it happens without further impact.
Eventually we could first display an invitation to refresh as you propose, before implementing a full transparent reconnection mechanism.

Another optimization idea I had in mind for a long time is to leverage the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) (see also https://whatwebcando.today/foreground-detection.html) to, for example:
1) stop processing events (and perhaps even close the connections) on `visibilitychange` to hidden, no need to do that especially on mobile and battery-powered devices when e.g. the user switches to another app
2) make sure the connections are properly working and/or refreshed when the app is visible again - it might make sense to store the time of the last received ping (or connection opened?) and act on `visibilitychange` to visible when it seems to be older than 12 seconds or if `sseConnected` is already false.

Not sure if that API works properly in webviews (like in the iOS or Android apps).

For the time being let's just give you the mobile hook you need :) Code LGTM.

PR #1510, opened 2022-10-07
[MainUI] Access page-content dimensions

florian-h05:
## Improvements

- ~~Update the docs for the `actionPhotos` property to explain how to set up the config Array in YAML syntax.~~ (See #1519)
- ~~Add support for evaluating expressions in config Arrays of custom widgets.~~ (See #1519)
- Access the current dimensions (height; width) available for a page's content (useful for building widgets with dynamic height ...).

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
@@ -122,7 +122,7 @@ export default {
             theme: this.$theme,
             themeOptions: this.$f7.data.themeOptions,
             device: this.$device,
-            screen: this.getPageContentRect(),
+            screen: this.getScreenInfo(),
             JSON: JSON,
             dayjs: dayjs
           })
@@ -145,14 +145,20 @@ export default {
         return value
       }
     },
-    getPageContentRect () {
+    getScreenInfo () {
       const pageCurrent = document.getElementsByClassName('page-current').item(0)
       const pageContent = pageCurrent.getElementsByClassName('page-content').item(0)
       const pageContentStyle = window.getComputedStyle(pageContent)
 
       return {
-        width: pageContent.clientWidth - parseFloat(pageContentStyle.paddingLeft) - parseFloat(pageContentStyle.paddingRight),
-        height: pageContent.clientHeight - parseFloat(pageContentStyle.paddingTop) - parseFloat(pageContentStyle.paddingBottom)
+        width: window.screen.width,
+        height: window.screen.height,
+        availWidth: window.screen.availWidth,
+        availHeight: window.screen.availHeight,
+        colorDepth: window.screen.colorDepth,
+        pixelDepth: window.screen.pixelDepth,
+        viewAreaWidth: pageContent.clientWidth - parseFloat(pageContentStyle.paddingLeft) - parseFloat(pageContentStyle.paddingRight),
+        viewAreaHeight: pageContent.clientHeight - parseFloat(pageContentStyle.paddingTop) - parseFloat(pageContentStyle.paddingBottom)
       }
     },
     childContext (component) {
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
@@ -122,6 +122,7 @@ export default {
             theme: this.$theme,
             themeOptions: this.$f7.data.themeOptions,
             device: this.$device,
+            screen: this.getPageContentRect(),
             JSON: JSON,
             dayjs: dayjs
           })
@@ -142,6 +143,16 @@ export default {
         return evalArr
       } else {
         return value
+      }
+    },
+    getPageContentRect () {
+      const pageCurrent = document.getElementsByClassName('page-current').item(0)
+      const pageContent = pageCurrent.getElementsByClassName('page-content').item(0)
+      const pageContentStyle = window.getComputedStyle(pageContent)
+
+      return {
+        width: pageContent.clientWidth - parseFloat(pageContentStyle.paddingLeft) - parseFloat(pageContentStyle.paddingRight),
+        height: pageContent.clientHeight - parseFloat(pageContentStyle.paddingTop) - parseFloat(pageContentStyle.paddingBottom)
       }
     },
     childContext (component) {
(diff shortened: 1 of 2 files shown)


florian-h05:
@ghys 
This one should be fast to review as there are no large changes to the codebase.

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
```diff
@@ -122,6 +122,7 @@ export default {
             theme: this.$theme,
             themeOptions: this.$f7.data.themeOptions,
             device: this.$device,
+            screen: this.getPageContentRect(),
```
ghys, line 125: New additions to the context should also be added to _hint-components.js_ for the autocompletion dropdown, as well as the docs:
https://www.openhab.org/docs/ui/building-pages.html#dynamically-configuring-components-with-expressions
(just realizing Number is missing there).

bundles/org.openhab.ui/web/src/pages/page/page-view.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <f7-page @page:afterin="onPageAfterIn" @page:beforeout="onPageBeforeOut" hide-bars-on-scroll :style="pageStyle" class="disable-user-select">
-    <f7-navbar v-if="!page || !page.config.hideNavbar" :back-link="(showBackButton) ? $t('page.navbar.back') : undefined" class="disable-user-select">
+  <f7-page @page:afterin="onPageAfterIn" @page:beforeout="onPageBeforeOut" hide-bars-on-scroll :style="pageStyle" class="disable-user-select" id="page">
+    <f7-navbar v-if="!page || !page.config.hideNavbar" :back-link="(showBackButton) ? $t('page.navbar.back') : undefined" class="disable-user-select" id="page-navbar">
```
ghys, line 3: There could be 2 top-level pages in the DOM at the same time (one hidden in the background for the back transition) so ids would not be unique

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
```diff
@@ -144,6 +145,16 @@ export default {
         return value
       }
     },
+    getPageContentRect () {
+      const page = document.getElementById('page') !== null ? document.getElementById('page').getBoundingClientRect() : null
+      const navbar = document.getElementById('page-navbar') !== null ? document.getElementById('page-navbar').getBoundingClientRect() : null
+      const toolbar = document.getElementById('page-toolbar') !== null ? document.getElementById('page-toolbar').getBoundingClientRect() : null
```
ghys, line 151: - See remark below about uniqueness of pages. The home page also has a navbar (potentially with a dynamic height) & bottom toolbar which is not taken into account here. Look for the element with class `view` or `page-current` instead. Then the page area excluding navbar & toolbar is the descendent of `page-current` with the `page-content` class.
- When the page is displayed in a popup/modal those values don't reflect the popup's size, not sure if that's what you want? Pages in modals also have `page-content` which would be the area of the modal
- These values don't update when the window is resized or the screen orientation changed, if it's not an issue then fine, otherwise a `calc(100vh - var(--f7-navbar-height))` or similar in CSS would be more elegant and better practice than getting pixel values for positioning

ghys (changes requested):
The docs update are appreciated and the expressions in arrays are good.

Almost missed the last-minute feature though, would have preferred a separate PR since there are some problems with it (you can revert and cherry-pick).

florian-h05:
@ghys 

I‘ve opened #1519 which only includes the docs updates and the support for evaluation of expressions in Arrays.

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
florian-h05: > Look for the element with class view or page-current instead. Then the page area excluding navbar & toolbar is the descendent of page-current with the page-content class.

Thanks for the tip, I've implemented this :+1: .

I am now selecting/getting the elements by their respective class name and I am also subtracting the padding from width and height.

> These values don't update when the window is resized or the screen orientation changed,

I am aware that these values are not dynamic, but as long as I find no pure CSS way to do comparisons like the following, I need those value inside the JS expressions:

```yaml
    - component: oh-image
      config:
        url: image-url
        refreshInterval: 250
        lazy: false
        style:
          margin-left: "=screen.width <= (screen.height * (4/3)) ? '0' : 'calc(calc(100% - calc(96vh - var(--f7-navbar-height) - var(--f7-toolbar-height)) * (4/3)) / 2)'"
          border-radius: var(--f7-card-expandable-border-radius)
          width: "=screen.width <= (screen.height * (4/3)) ? '100%' : 'calc(calc(96vh - var(--f7-navbar-height) - var(--f7-toolbar-height)) * (4/3))'"
```

> otherwise a calc(100vh - var(--f7-navbar-height)) or similar in CSS would be more elegant and better practice than getting pixel values for positioning

I am using CSS calc for the exact positioning, getting the pixel values is only important for making the decision which CSS statement should be passed.

> When the page is displayed in a popup/modal those values don't reflect the popup's size, not sure if that's what you want? Pages in modals also have page-content which would be the area of the modal

That's the problem I have no solution for, but personally I am not using popups in combination with getting the pixel values of width and height.

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
florian-h05: I added it to `hint-components.js`.

Docs updates are on the way (see https://github.com/florian-h05/openhab-docs/commit/4e611c49c0a679c254f9e8eb6da5366ca2ff5e6c).

ghys (approved):
Okay then, so since you've named the object `screen` I added some info from
https://developer.mozilla.org/en-US/docs/Web/API/Screen
(not everything because we don't want people to rely on e.g. `orientation` which is not available on Safari, but only those with widespread support - https://www.w3schools.com/jsref/obj_screen.asp)

The former `screen.width` and `screen.height` are moved to resp. `screen.viewAreaWidth` and `screen.viewAreaHeight`.

Test (entire page):

```yaml
config:
  label: Fullscreen Test
  sidebar: true
blocks:
  - component: oh-block
    config: {}
    slots:
      default:
        - component: oh-grid-row
          config: {}
          slots:
            default:
              - component: oh-grid-col
                config: {}
                slots:
                  default:
                    - component: Label
                      config:
                        text: =screen
                        style:
                          width: ='calc(' + screen.viewAreaWidth + 'px -
                            2*var(--f7-block-padding-horizontal))'
                          height: ='calc(' + screen.viewAreaHeight + 'px -
                            2*var(--f7-block-padding-vertical))'
                          background: red
masonry: null
grid: []
canvas: []

```

florian-h05:
@ghys
Thanks for the help!
I‘ll update the docs accordingly ASAP.

florian-h05:
@ghys 
Docs PR is now open: https://github.com/openhab/openhab-docs/pull/1894.

PR #1522, opened 2022-10-09
Update widget expression capabilities

### LINKED ISSUE #1506: Change expression parsers
As the community gets more involved in advanced custom widget creation, more and more users are running up against the limitations of the expression-eval library.  In particular, I've seen many issues with users wanting regex expressions or js methods that require functions as parameters.  Expression-eval does not handle either of these two cases.

There's a recent fork of expression-eval, [jse-eval](https://github.com/6utt3rfly/jse-eval/), which adopts the jsep plug-in system which, in turn would allow the addition of regex and arrow plug-ins.  Is it feasible/acceptable to move MainUI from expression-eval to jse-eval?

JustinGeorgi:
This resolves #1506 and adds several other smaller enhancements to address common requests about widget capability seen in the forums.

Signed-off-by: Justin Georgi justin.georgi@gmail.com


JustinGeorgi:
Hi @ghys , I've included a couple of changes to the widget expression system here:
1) Move to jse-eval: The breaking changes noted in the jsep docs due to the updated version don't appear to apply to anything here, so the jse-eval does seem to drop in cleaning for expression-eval.  With a couple days of testing, I have found no impacts on existing widget expressions. 

2) Add the arrow function and regex plugins to jse-eval

3) Evaluate the component type:  Now components can be dynamically selected rather than needing to define multiple components with complex `visible` values if something dynamic about the component (e.g., item type) requires different component types.

4) Add a `Content` component: This is analogous to the `Label` but does not have a container.  Some of the `f7` components require actual text inside the created tags that does not have additional html tags (e.g., the options of a select input), the `Content` component allows this.  In my novice understanding, this does not introduce any injection issues, so I think it's safe.

5) Added the `@` unary operator based on [discussion in the forums](https://community.openhab.org/t/any-documentation-around-for-valid-javascript-in-expressions-in-widgets/139658): The operator definition appears be evaluated separate from other evaluate calls but I cannot find a way to include a context in the operator definition.  So, the `items` variable cannot be included in the operator as far as I can tell.  That means the format is restricted to `@items['itemname']` which is still an improvement.

6) Added `user` to the widget expression context: `user` just contains a string of the user name if a user is logged in.  Currently whole widgets can be shown or hidden based on the user, but widget content cannot be made response to the user.  This allows expressions to check the user for responsive content.

7) Added 'screen' to the widget expression context: `screen` replicates the screen information shown in the technical details to facilitate widget responsiveness based directly on screen size for some advanced widgets.

JustinGeorgi:
Do'h I see that @florian-h05 beat me to the screen definition by a couple of hours.  I'll pull that part out.

JustinGeorgi:
@ghys not sure what's going on with this.  I missed the sign-off on the initial commit so I followed the rebase steps outlines to fix it and it's not clear to me exactly what the result is.  Should I just start a new PR (and not mess up the commits this time)?

ghys:
> Move to jse-eval: The breaking changes noted in the jsep docs due to the updated version don't appear to apply to anything here, so the jse-eval does seem to drop in cleanly for expression-eval. With a couple days of testing, I have found no impacts on existing widget expressions.
> Add the arrow function and regex plugins to jse-eval

Excellent!

> Evaluate the component type: Now components can be dynamically selected rather than needing to define multiple components with complex visible values if something dynamic about the component (e.g., item type) requires different component types.

Not sure how it works/if this PR includes the change?

> Add a Content component: This is analogous to the Label but does not have a container. Some of the f7 components require actual text inside the created tags that does not have additional html tags (e.g., the options of a select input), the Content component allows this. In my novice understanding, this does not introduce any injection issues, so I think it's safe.

Good idea - maybe we could also allow slots (can be combined with 3. above) so we can build an arbitrary DOM hierarchy (of HTML or even SVG tags) using components & slots instead of using the `f7-row` technique with the `tag` parameter, would be nice to have something focused on that.

> Added the @ unary operator based on [discussion in the forums](https://community.openhab.org/t/any-documentation-around-for-valid-javascript-in-expressions-in-widgets/139658): The operator definition appears be evaluated separate from other evaluate calls but I cannot find a way to include a context in the operator definition. So, the items variable cannot be included in the operator as far as I can tell. That means the format is restricted to @items['itemname'] which is still an improvement.

Works with this code 😉 (note the explicit comparisons with `undefined` to avoid falsy displayStates like `'0'` or `'false'` to fall back to the raw state:
```js
import store from '@/js/store'
...
expr.addUnaryOp('@', (itemName) => {
  const itemState = store.getters.trackedItems[itemName]
  if (itemState === undefined) return '-'
  if (itemState.displayState === undefined) return itemState.state
  return itemState.displayState
})
```
image

> Added user to the widget expression context: user just contains a string of the user name if a user is logged in. Currently whole widgets can be shown or hidden based on the user, but widget content cannot be made response to the user. This allows expressions to check the user for responsive content.

Any reason not to expose the whole `this.$store.getters.user`? Would expose the role with `user.roles[0]` which unlocks interesting options (username moved to `user.name`).

JustinGeorgi:
> Not sure how it works/if this PR includes the change?

You may be thinking bigger at the moment than this change is.  I just changed:
```javascript
    componentType () {
      return this.context.component.component
    },
```
to
```javascript
    componentType () {
      return this.evaluateExpression('type', this.context.component.component)
    },
```
It doesn't permit arbitrary tags, it just allows the use of expressions for the value of `component`.  Here's an example.  My old Dynamic Favorites Widget has this in the middle of it:
```yaml
- component: oh-toggle-card
  config:
    item: =loop.favItem.name
    title: "=loop.favItem.metadata ? loop.favItem.metadata.FavLabel.value : loop.favItem.label"
    visible: =loop.favItem.type=='Switch'
- component: oh-player-card
  config:
    item: =loop.favItem.name
    title: "=loop.favItem.metadata ? loop.favItem.metadata.FavLabel.value : loop.favItem.label"
    visible: =loop.favItem.type=='Player'
- component: oh-slider-card
  config:
    item: =loop.favItem.name
    scale: true
    scaleSteps: 2
    title: "=loop.favItem.metadata ? loop.favItem.metadata.FavLabel.value : loop.favItem.label"
    visible: =loop.favItem.type=='Dimmer'
- component: oh-rollershutter-card
  config:
    item: =loop.favItem.name
    title: "=loop.favItem.metadata ? loop.favItem.metadata.FavLabel.value : loop.favItem.label"
    visible: =loop.favItem.type=='Rollershutter'
``` 
This is long, ungainly, and in the page editor renders all version of every widget for each item which kinda gets in the way. 
 If component uses expression then it reduces to this:
```yaml
- component: =(loop.favItem.type=='Switch')?('oh-toggle-card'):(loop.favItem.type=='Player')?('oh-player-card'):(loop.favItem.type=='Dimmer')?('oh-slider-card'):(loop.favItem.type=='Rollershutter')?('oh-rollershutter-card'):(false)
  config:
    item: =loop.favItem.name
    title: "=loop.favItem.metadata ? loop.favItem.metadata.FavLabel.value : loop.favItem.label"
    scale: true
    scaleSteps: 2
```
and renders correctly in the page editor.

> Good idea - maybe we could also allow slots (can be combined with 3. above) so we can build an arbitrary DOM hierarchy (of HTML or even SVG tags) using components & slots instead of using the f7-row technique with the tag parameter, would be nice to have something focused on that.

If you don't think this opens up too many potential issues.  I'm game.  I can see having `ht` be a prefix for arbitrary component names that render directly as html compliant tags.  Something like this:
```yaml
- component: ht-table
  config:
    class: fancy-table
  slots:
    default:
      component: ht-tr
...
```
```html
<table class="fancy-table>
  <tr>
...
  </tr>
</table>
```
I think that this might be a separate PR, however.  While I'm confident I can work most of that out, there are some pieces I'll have to learn to do it. 

> Any reason not to expose the whole `this.$store.getters.user`? Would expose the role with `user.roles[0]` which unlocks interesting options (username moved to `user.name`).

Mostly caution on my part.  But I think you're right, having the role available does have some worthwhile uses.  I'm happy to make that change.

> Works with this code :wink: (note the explicit comparisons with `undefined` to avoid falsy displayStates like `'0'` or `'false'` to fall back to the raw state) - also adding a '@@' operator to force the raw state:

Oh!  I got close with one of my versions...it didn't occur to me to import store directly though.  Live and learn.  Thanks.

I like the `@@` too.  Nice touch.  I'll factor that in too.

JustinGeorgi:
> I think that this might be a separate PR, however. While I'm confident I can work most of that out, there are some pieces I'll have to learn to do it.

...or not.  Looks like it was much easier than I expected.  With the `ht-` commit, the following widget:
```yaml
component: ht-div
config: {}
slots:
  default:
    - component: ht-blockquote
      config:
        class: new-quote-tag
      slots:
        default:
          - component: Content
            config:
              text: ="html tags " + (props.tagVar || "work")
    - component: ht-table
      config:
        style:
          background: green
      slots:
        default:
          - component: ht-tr
            config:
              style:
                vertical-align: top
            slots:
              default:
                - component: ht-td
                  slots:
                    default:
                      - component: Content
                        config:
                          text: Cell 1
                - component: ht-td
                  config:
                    style:
                      background: yellow
                  slots:
                    default:
                      - component: Content
                        config:
                          text: "Cell "
                      - component: ht-span
                        config:
                          style:
                            font-size: 32px
                        slots:
                          default:
                            - component: Content
                              config:
                                text: 2
                - component: ht-td
                  slots:
                    default:
                      - component: Content
                        config:
                          text: Cell 3
```
is rendered as:
```html
<div>
  <blockquote class="new-quote-tag">
    <!--fragment#3093688ff0#head-->
    html tags work
    <!--fragment#3093688ff0#tail-->
  </blockquote>
  <table style="background: green;">
    <tr style="vertical-align: top;">
      <td>
        <!--fragment#c1887fca28#head-->
        Cell 1
        <!--fragment#c1887fca28#tail-->
      </td>
      <td style="background: yellow;">
        <!--fragment#4b24a9591c#head-->
        Cell 
        <!--fragment#4b24a9591c#tail-->
        <span style="font-size: 32px;">
          <!--fragment#17d250b2021#head-->
            2
          <!--fragment#17d250b2021#tail-->
        </span>
      </td>
      <td>
        <!--fragment#bca24b37a1#head-->
          Cell 3
        <!--fragment#bca24b37a1#tail-->
      </td>
    </tr>
  </table>
</div>
```
image

It does run into trouble rendering svgs, but only because it appears that the `<path>` tag means something else in vue.  All the other basic svg pieces appear to work correctly as far as I can tell.

ghys:
Nice!

> I can see having ht be a prefix for arbitrary component names that render directly as html compliant tags.

I think `html-` would be nicer, or even no prefix at all, just put it last with `v-else-if="visible'"`, we shall see if it poses problems. After all Vue components are all prefixed.
Also I would still offer the ability to put some content, before the slots, so you don't actually need Content subcomponents in most cases (and probably improves performance too) - your example above would be just:
```yaml
component: div
config: {}
slots:
  default:
    - component: blockquote
      config:
        class: new-quote-tag
        text: ="html tags " + (props.tagVar || "work")
    - component: table
      config:
        style:
          background: green
      slots:
        default:
          - component: tr
            config:
              style:
                vertical-align: top
            slots:
              default:
                - component: td
                  config:
                    text: Cell 1
                - component: td
                  config:
                    text: "Cell "
                    style:
                      background: yellow
                  slots:
                    default:
                      - component: span
                        config:
                          text: 2
                          style:
                            font-size: 32px
                - component: td
                  config:
                    text: Cell 1
```

And adding the "template" & "objects" jse-eval plugins would probably be nice too, I've missed them sometimes.

Example adapted from the placeholder you get when you add a new widget:
```yaml
uid: widget_fb590daf3c
props:
  parameterGroups: []
  parameters:
    - name: item
      label: Item
      type: TEXT
      context: item
      description: An item to display
tags: []
component: f7-card
config:
  title: =props.item?`State of ${props.item}`:'Set props to test!'
  footer: =`Raw = ${@@props.item}`
  content: =@props.item
```
Last thing, when you're typing the dot after <kbd>items.</kbd> you're offered a list of items to autocomplete (it completes in `<ItemName>.state`). It would be nice to have after typing <kbd>@</kbd> (completing in `'<ItemName>' i.e. adding quotes)

https://github.com/openhab/openhab-webui/blob/6e9cc922fe1bf5285ea018238faa05ad37268bb4/bundles/org.openhab.ui/web/src/components/config/editor/hint-components.js#L117-L122

https://github.com/openhab/openhab-webui/blob/6e9cc922fe1bf5285ea018238faa05ad37268bb4/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue#L241-L244

JustinGeorgi:
> I think html- would be nicer, or even no prefix at all, just put it last with v-else-if="visible'", we shall see if it poses problems. After all Vue components are all prefixed.

That works for me.  I was going for some consistency with the `ht-` prefix, but no prefix is just that much more readable.  Un-prefixed html and svg tags now work (except for `<path>` still).

> Also I would still offer the ability to put some content, before the slots, so you don't actually need Content subcomponents in most cases (and probably improves performance too)

To do this up right, I had to not bind the entire config, but just `class` and `style` keys (binding the whole config, of course, just put a `text` or `content` attribute in the tag, which only sort of works).  That does mean that these components cannot have events bound to them.  I assume you find this to be a feature and not a bug :wink:, so I have left it at that instead of trying to figure out how to pass generalized events through.  There are a few places where I could see events being really handy for some truly advanced UI (e.g., a `<canvas>` could be used to great effect now for background weather animations or sun rise/set events), but those few cases probably aren't worth the headache.

In the end I moved the key for this from `text` to `content`.  I felt that makes it a little more clear what really going on with these tags and parallels better the `content` component name when those need to be worked in as well.

> And adding the "template" & "objects" jse-eval plugins would probably be nice too, I've missed them sometimes.

Done.

> Last thing, when you're typing the dot after items. you're offered a list of items to autocomplete (it completes in <ItemName>.state). It would be nice to have the same after typing @ (completing in '<ItemName>' i.e. adding quotes)

Done.  Thanks for pointing this out.  I wouldn't have thought of it myself, and it's a great addition (and was a fun dive into some pieces I hadn't examined at all yet).

bundles/org.openhab.ui/web/src/components/widgets/generic-widget-component.vue
```diff
@@ -13,10 +13,21 @@
   <div v-else-if="componentType && componentType === 'Label' && visible" :class="config.class" :style="config.style">
     {{ config.text }}
   </div>
+  <fragment v-else-if="componentType && componentType === 'Content'">
+    {{ config.text }}
+  </fragment>
   <pre v-else-if="componentType && componentType === 'Error' && visible" class="text-color-red" style="white-space: pre-wrap">{{ config.error }}</pre>
+  <component v-else-if="visible" :is="componentType" :class="config.class" :style="config.style" @command="onCommand">
```
ghys, line 20:
```suggestion
  <component v-else-if="visible" :is="componentType" v-bind="config">
```

> To do this up right, I had to not bind the entire config

Why not?
Having all attributes available is a must for SVG and yes, even the events work - not that I'm particularly happy about that indeed, but well... We could bind an object with all keys beginning with `on` removed but let's leave it for now.

> In the end I moved the key for this from text to content

Good call - it's true that it will add a `content` attribute to the element in addition to putting it before the slots but it's not that much of a problem, `content` as an attribute is only used in `<meta>` - https://www.w3schools.com/tags/att_content.asp

Example:
```yaml
uid: new_expressions
tags: []
props:
  parameters:
    - description: Disable the button
      label: Disabled
      name: disabled
      required: false
      type: BOOLEAN
  parameterGroups: []
timestamp: Oct 11, 2022, 2:05:10 PM
component: button
config:
  class:
   - padding
   - margin
  style:
    width: 200px
  type: button
  content: Button is
  disabled: =props.disabled
  onclick: alert('hmm not sure this is good')
slots:
  default:
    - component: span
      config:
        content: =props.disabled?'Disabled':'Enabled'
        style:
          font-size: 20px
          color: =props.disabled?'red':'green'
```

ghys (commented):
I think we're getting close...!

> Done. Thanks for pointing this out. I wouldn't have thought of it myself, and it's a great addition (and was a fun dive into some pieces I hadn't examined at all yet).

I realize I pointed you to the wrong part of script-editor.vue above, but you figured it out :)

bundles/org.openhab.ui/web/src/components/widgets/generic-widget-component.vue
JustinGeorgi: Ok, sounds good.  If you're not concerned about the content attribute, then this is the better way to go.

> I realize I pointed you to the wrong part of script-editor.vue above, but you figured it out :)

The trickiest part was actually realizing I also had to modify `filterPartialCompletions` in hint-utils.js.  I got everything else working quickly and then spent an hour trying to figure out why my hint list kept coming back empty...

ghys (approved):
I think we're good to go.

Would you please update the docs to mention the `user` context object and the new features, similar to https://github.com/openhab/openhab-docs/pull/1894?

Btw I was trying to build data tables, nice to see that it's finally easily possible 😃:
```yaml
uid: data_table
tags: []
props:
  parameters: []
  parameterGroups: []
timestamp: Oct 11, 2022, 5:16:01 PM
component: div
config:
  class:
    - card
    - data-table
    - data-table-collapsible # Remove if the table should not be collapsible
    - data-table-init # Only useful for collapsible tables
slots:
  default:
    - component: div
      config:
        class:
          - card-header
      slots:
        default:
          - component: div
            config:
              class:
                - data-table-title
              content: Nutrition
    - component: table
      config: {}
      slots:
        default:
          - component: thead
            slots:
              default:
                - component: tr
                  slots:
                    default:
                      - component: th
                        config:
                          class:
                            - label-cell
                          content: Dessert (100g serving)
                      - component: th
                        config:
                          class:
                            - numeric-cell
                          content: Calories
                      - component: th
                        config:
                          class:
                            - numeric-cell
                          content: Fat (g)
                      - component: th
                        config:
                          class:
                            - numeric-cell
                          content: Carbs
                      - component: th
                        config:
                          class:
                            - numeric-cell
                          content: Protein (g)
          - component: tbody
            slots:
              default:
                - component: oh-repeater
                  config:
                    fragment: true
                    for: food
                    in:
                      - name: Frozen Yoghurt
                        calories: 159
                        fat: "6.0"
                        carbs: =@'ElectricityMeter_Power'
                        protein: "4.0"
                      - name: Ice Cream Sandwich
                        calories: 237
                        fat: "9.0"
                        carbs: 37
                        protein: "4.4"
                      - name: Eclair
                        calories: 262
                        fat: "16.0"
                        carbs: 24
                        protein: "6.0"
                      - name: Cupcake
                        calories: 305
                        fat: "3.7"
                        carbs: 67
                        protein: "4.3"
                  slots:
                    default:
                      - component: tr
                        slots:
                          default:
                            - component: td
                              config:
                                class:
                                  - label-cell
                                content: =loop.food.name
                                data-collapsible-title: Dessert
                            - component: td
                              config:
                                class:
                                  - numeric-cell
                                content: =loop.food.calories
                                data-collapsible-title: Calories
                            - component: td
                              config:
                                class:
                                  - numeric-cell
                                content: =loop.food.fat
                                data-collapsible-title: Fat (g)
                            - component: td
                              config:
                                class:
                                  - numeric-cell
                                content: =loop.food.carbs
                                data-collapsible-title: Carbs
                            - component: td
                              config:
                                class:
                                  - numeric-cell
                                content: =loop.food.protein
                                data-collapsible-title: Protein (g)
```
image

JustinGeorgi:
> I am eagerly waiting for this update. Thanks for this improvement!
> 
> I have something I don't understand. I am trying to use `@jsep-plugin/object` through the following a widget expression: `"='Test' + {a:1}['a']"`. It does not work in 3.4M5 ( -> displays undefined), whereas it works as expected on my local build.
> 
> Any idea ?

I haven't had the time recently to update to M5 yet, but your expression works just fine in M4.

Have you tried enclosing the object in an extra set of parens? `"='Test' + ({a:1})['a']"`?

PR #1524, opened 2022-10-10
put boxes arround addons cards for better seperation

niwla23:
This is the current addon store. It can be quite hard to tell to which image the text belongs as there is no visual seperation.
image

This PRs adds better grouping by putting a simple gray box around the addon. The color matches the one used for secondary backgrounds at other parts of the main UI.:

image

EDIT: added different device views and titles are white again

There is one thing that is kind of hacky about this, I had to change `width calc(100% - 5rem)` to `width 210px` because otherwise addons with overflowing names would overflow the box. However, I have found that 210px works reasonably well with all screen sizes.
If someone has a cleaner solution to this, please tell me.


niwla23:
BTW, in case this gets merged I would be glad if a maintainer could add the hacktoberfest-accepted label so it counts towards hacktoberfest.com :)

niwla23:
I just noticed the titles somehow have turned orange, I will change them back to white soon

niwla23:
image
Made the titles white again

niwla23:
image
I added &nbsp for cards that have no headline set so the cards don't have different heights

bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
```diff
@@ -48,21 +59,20 @@
...
     line-height 1.75
     .addon-card-label
       text-overflow ellipsis
-      overflow hidden
+      overflow clip
       white-space nowrap
-      width calc(100% - 5rem)
+      // width calc(100% - 5rem)
+      width 210px
+      color #fff
```
ghys, line 75:
```suggestion
      color var(--f7-list-item-title-text-color)
```

bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
```diff
@@ -1,45 +1,56 @@
...
-  padding 5px 0px 0px
+  padding 5px
+  width: 100%
   position relative
-  margin-bottom 1rem
+
+  .addon-card-inner
+    width 100%
+    height 100%
+    background-color #1c1c1d
```
ghys, line 46:
```suggestion
    background-color var(--f7-list-link-hover-bg-color)
```
And I would actually prefer it on `:hover` so that the look is the same as the list items (the color would change on the desktop theme only, no change for iOS or android)

bundles/org.openhab.ui/web/src/components/addons/addons-section.vue
```diff
@@ -75,6 +75,7 @@
     align-content flex-start
     align-items flex-end
     flex-wrap wrap
+    // gap 0.5rem
     padding-left calc(var(--f7-safe-area-left) + var(--f7-list-item-padding-horizontal))
```
ghys, line 79:
```suggestion
    padding-left calc(var(--f7-safe-area-left)
```
This was actually there before, but I don't know why there was this left padding... It looks better without.

ghys (changes requested):
Reminder to always test design changes in the 6 theme+dark mode combinations, and never hardcode colors (except in very specific circumstances), as it doesn't look good at all in the light theme:

image

niwla23:
Can someone please review?

bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
```diff
@@ -1,45 +1,57 @@
...
-    <div class="addon-card-title">
-      <div class="addon-card-title-after">
-        <f7-preloader v-if="addon.pending" color="blue" />
-        <f7-button v-else-if="addon.installed" class="install-button prevent-active-state-propagation" text="Remove" color="red" round small @click="buttonClicked" />
-        <f7-button v-else class="install-button prevent-active-state-propagation" :text="installActionText || 'Install'" color="blue" round small @click="buttonClicked" />
-      </div>
-      <div class="addon-card-label">
-        {{ addon.label }}
+  <f7-link v-if="addon" class="addon-card padding-right-half" :href="addon.id">
+    <div class="addon-card-inner card-expandable">
```
ghys, line 3: Those cards will expand and going back will bug unless you add the `card-prevent-open` class as well.

Like I said IMO the other/better alternative is add `background-color var(--f7-list-link-hover-bg-color)` it on `.addon-card:hover`, without any border, so that the look of these "cards" is the same as the list items - it solves your initial issue (highlighting the image + info text) at least on desktops and remains consistent across designs.

ghys (changes requested):
Still doesn't work so good, sorry!

(btw - I'm in Asia without a laptop until November so I have very few opportunities for review).

bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
niwla23: I will take a look at it when I am back home (a few days). The reason i used the card-expand class is that it fits the design of other cards. For example, there seems to be some box shadow on the card class

niwla23: I changed card-expandable to just card now, I don't know anymore why I used exandable cards in the first place. 

> Like I said IMO the other/better alternative is add background-color var(--f7-list-link-hover-bg-color) it on .addon-card:hover

You mean instead of cards at all so only the currently hovered addon is highlighted by background color? As you already wrote, this would only fix it on PC and I feel like it wouldn't be the best UX if you can only tell distinguish the current card from the others but not those apart

> so that the look of these "cards" is the same as the list items 

image

This is how it looks combined with list items. I think it looks okay and it looks like the cards are meant to promote some featured addons when combined with list items, so is it that bad when they look a bit different?

ghys (approved):
Alright, fair enough - I added the highlight-on-hover anyway and fixed a subtitle (transformations section) in the process, which was wrong and probably so for quite some time ;)

PR #1530, opened 2022-10-17
[MainUI] Improve sitemap validation

mherwege:
Building upon the more complete support of the sitemap DSL in the UI, this PR adds more validation logic:

- Validate all required parameters are provided when interpreting DSL.
- Validate all required parameters are privded when saving sitemap from sitemap builder (but still save if this is not the case).
- Validate syntax of parameter values when saving in sitemap builder.
- Improved parameter types in sitemap builder, dropdown for some parameters.

@ghys I use the f7.dialog popup to list detected errors in the sitemap, but it becomes too small when there are many. I am not sure this is the best way of doing this. To widen the dialog, I would need to provide a CSS class, but I don't know much about that. Do you have suggestions?

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
+++ b/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
@@ -35,8 +35,9 @@
     gt:               '>',
     equals:           '=',
     comma:            ',',
-    NL:               { match: /\n/, lineBreaks: true },
+    NL:               { match: /\n/, lineBreaks: true }
   })
+  const requiresItem = ['Group', 'Chart', 'Switch', 'Mapview', 'Slider', 'Selection', 'List', 'Setpoint', 'Colorpicker', 'Default']
 
   function getSitemap(d) {
     return {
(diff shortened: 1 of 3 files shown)


ghys (approved):
Thanks @mherwege - this looks useful.

> @ghys I use the f7.dialog popup to list detected errors in the sitemap, but it becomes too small when there are many. I am not sure this is the best way of doing this. To widen the dialog, I would need to provide a CSS class, but I don't know much about that. Do you have suggestions?

I had a look and made some improvements in https://github.com/openhab/openhab-webui/pull/1530/commits/81195d29df0e89fccb2304280414fdd0c8f85598.
You can also cancel the saving:

image


While testing with one of my sitemaps I found another bug with the mapping serialization - if you have an accented character it doesn't work, it puts quotes around the accented character:

image

image

I figure this is a bug with your previous PR, not this one. Would you like to address it here or in a follow-up PR? This one LGTM.

mherwege:
> Would you like to address it here or in a follow-up PR? This one LGTM.

I prefer looking at it in a follow-up PR.

PR #1519, opened 2022-10-08
[MainUI] Evaluate expressions in configuration Arrays & Update `actionPhotos` docs

florian-h05:
@ghys 

This PR brings the docs updates and support for expressions in configuration arrays from PR #1510.

## Improvements

- Update the docs for the `actionPhotos` property to explain how to set up the config Array in YAML syntax.
- Add support for evaluating expressions in config Arrays of custom widgets.

PR #1531, opened 2022-10-20
Fixes evaluating arrays - cached expressions can be overwritten

crnjan:
Seems as there is an issue in https://github.com/openhab/openhab-webui/pull/1519 that can cause cached expressions to override each-other (some of my existing widgets are broken in latest snapshot).

Also removed what seems obsolete since arrays are now parsed as part of `evaluateExpression` itself.


ghys (approved):
LGTM, thanks for the fix!

PR #1540, opened 2022-11-12
Filter rules/scripts by id and rule status

stefan-hoehn:
Signed-off-by: Stefan Höhn <mail@stefanhoehn.com>

I always missed the possibility to filter rules by 

- rule status (enabled, disabled, idle, running) -> item-text
- rule id -> item-after

as they are shown in the list but not taken into account during the filtering.

hence, I have added these (item-text and item-after) to the searchbar field list.


stefan-hoehn:
Sorry, I forget the signoff.

I added it manually in the PR description. Can someone retrigger the build?

PR #1545, opened 2022-11-17
New sub-component for oh-cards footers to allow several text layouts

tarag:
Allows nice layout of several text fragments in the footers with the layout built-in f7 framework. The use of this components needs the user to define an array in the YAML view.


ghys (approved):
This is good. Thanks!
I would like to have a discussion on having component's slots match slots in the original components' slots though, it's a challenge code-wise but we can achieve this. The goal would be to allow slots that you can learn and rely on from the Framework7 documentation to be reflected in the OH widgets/components, in the YAML definition.

tarag:
Thanks.

Concerning the use of slots, that exactly what I thought when I made this PR. It would really be a great to be able to use all the slots of the original f7 components. That would provide so much more flexibility.

PR #1548, opened 2022-11-19
[BasicUI] Fix the color picker

### LINKED ISSUE #1140: [basicUI] Wrong behaviour of the colorpicker
I see a big bug with the colorpicker in BasicUI. When you use it to select a color near the limit of the circle for a hue color light and then wait for the state update, you see the selector moving near the center of the circlie which maps the real color of the light (not the expected color). So the scolorpicker allows to select a color which is not correctly handled by the hue system.

Even with a color item not linked to any device, I can see that if I select a color at the limit of the circle and then close/reopen the popup, the selector can be outside the circle ! This probably explains the strange behaviour with the hue light.

The popup after moving the selector to the limit of the circle:
image

Then after I close and reopen the popup:
image

Here are associated logs:
```
12:27:45.918 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'MagicColor' received command 305,19,100
12:27:47.447 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'MagicColor' received command 305,19,100
12:27:47.449 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'MagicColor' received command 305,19,100
12:27:47.449 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'MagicColor' received command 312,32,100
12:27:47.449 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'MagicColor' received command 312,32,100
12:27:47.449 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'MagicColor' received command 312,32,100
12:27:47.449 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'MagicColor' changed from 305,19,100 to 312,32,100
12:27:47.734 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'MagicColor' received command 312,32,100
```

Sometimes, the selector is moved near the center of the circle after reopening the popup.

I have the strange feeling that the selector allows to select invalid color ?

lolodomo:
Fix #1140

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1092,7 +1092,8 @@
 				maxRadius = _t.image.clientWidth / 2,
 				offsetX = pos.x - maxRadius,
 				offsetY = pos.y - maxRadius,
-				radius = Math.sqrt((offsetY * offsetY) + (offsetX * offsetX));
+				radius = Math.sqrt((offsetY * offsetY) + (offsetX * offsetX)),
+				angle = Math.atan2(offsetX, offsetY) / Math.PI / 2;
 
 			if (radius > maxRadius) {
 				var
@@ -1107,11 +1108,8 @@
 			_t.handle.style.top = (pos.y / _t.image.clientWidth) * 100 + "%";
 
 			var
-				angle = offsetX >= 0 ?
-						(Math.PI * 2 - Math.atan(offsetY / offsetX) + Math.PI / 2) / (Math.PI * 2) :
-						(Math.PI * 2 - Math.atan(offsetY / offsetX) - Math.PI / 2) / (Math.PI * 2),
 				hsv = {
-					h: isNaN(angle) ? 0 : angle,
+					h: angle >= 0 ? angle : 1 + angle,
 					s: radius / maxRadius,
 					v: 1
 				},
@@ -1402,7 +1400,7 @@
 			_t.modalControl = new Colorpicker(_t.modal.container, _t.value, function(color) {
 				_t.value = Colorpicker.hsv2rgb(color);
 				emitEvent(
-					Math.round((color.h * 360) % 360) + "," +
+					(Math.round(color.h * 360) % 360) + "," +
 					Math.round(color.s * 100) + "," +
 					Math.round(color.v * 100)
 				);
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1089,17 +1089,18 @@
 				};
 			}
 			var
-				maxR = _t.image.clientWidth / 2,
-				offsetX = pos.x - maxR,
-				offsetY = pos.y - maxR,
-				r = (offsetY * offsetY) + (offsetX * offsetX);
+				maxRadius = _t.image.clientWidth / 2,
+				offsetX = pos.x - maxRadius,
+				offsetY = pos.y - maxRadius,
+				radius = Math.sqrt((offsetY * offsetY) + (offsetX * offsetX));
 
-			if (r > (maxR * maxR)) {
+			if (radius > maxRadius) {
 				var
-					ratio = 1 - Math.abs(maxR / Math.sqrt(r));
+					ratio = 1 - Math.abs(maxRadius / radius);
 
 				pos.x -= (offsetX * ratio);
 				pos.y -= (offsetY * ratio);
+				radius = maxRadius;
 			}
 
 			_t.handle.style.left = (pos.x / _t.image.clientWidth) * 100 + "%";
@@ -1111,7 +1112,7 @@
 						(Math.PI * 2 - Math.atan(offsetY / offsetX) - Math.PI / 2) / (Math.PI * 2),
 				hsv = {
 					h: isNaN(angle) ? 0 : angle,
-					s: Math.sqrt(r) / maxR,
+					s: radius / maxRadius,
 					v: 1
 				},
 				hsl = hsv2hsl(hsv);
@@ -1402,7 +1403,7 @@
 				_t.value = Colorpicker.hsv2rgb(color);
 				emitEvent(
 					Math.round((color.h * 360) % 360) + "," +
-					Math.round((color.s * 100) % 100) + "," +
+					Math.round(color.s * 100) + "," +
 					Math.round(color.v * 100)
 				);
 			});


andrewfg (changes requested):
As mentioned in the issue, I think you should use the now officially recommended 'Math.atan2()' function..
```
angle = Math.atan2(offsetX, offsetY) / Math.PI / 2;
```

andrewfg:
> I was talking about saturation, not hue. Saturation is between 0 and 100.

Yes I certainly agree that a saturation value of 1 should NOT be rounded to 0. !!

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1111,7 +1112,7 @@
 						(Math.PI * 2 - Math.atan(offsetY / offsetX) - Math.PI / 2) / (Math.PI * 2),
 				hsv = {
 					h: isNaN(angle) ? 0 : angle,
```
andrewfg, line 1114: And if you use 'Math.atan2()' it also eliminates the need for the NaN() check below, since the new code always produces a valid angle number.

lolodomo:
Is it `Math.atan2(offsetX, offsetY)` or `Math.atan2(offsetY, offsetX)` ?

andrewfg:
> Is it `Math.atan2(offsetX, offsetY)` or `Math.atan2(offsetY, offsetX)` ?

It is definitely `Math.atan2(offsetX, offsetY)` !!

I think your question is because the JavaDoc refers to `Math.atan2(offsetY, offsetX)`? But the JavaDoc is referring to a calculations in a cartesian coordinate system where XY Origin is bottom left, and angles are calculated anti-clockwise. However we have an inverted cartesian coordinate system where XY Origin is top left, but the angles are still calculated anti-clockwise. So we have to flip the calculation compared to the JavaDoc..

andrewfg:
Below is my JUnit code..

```
    private static final DecimalType RED = HSBType.RED.getHue();
    private static final DecimalType YELLOW = HSBType.fromRGB(255, 255, 0).getHue();
    private static final DecimalType GREEN = HSBType.GREEN.getHue();
    private static final DecimalType CYAN = HSBType.fromRGB(0, 255, 255).getHue();
    private static final DecimalType BLUE = HSBType.BLUE.getHue();
    private static final DecimalType MAGENTA = HSBType.fromRGB(255, 0, 255).getHue();
    private static final DecimalType WHITE = HSBType.WHITE.getHue();

    @Test
    void testHMouseHit() {
        System.out.print(RED + " => ");
        mouseDown("r", 150, 300);

        System.out.print(YELLOW + " => ");
        mouseDown("y", 300, 235);

        System.out.print(GREEN + " => ");
        mouseDown("g", 300, 60);

        System.out.print(CYAN + " => ");
        mouseDown("c", 150, 0);

        System.out.print(BLUE + " => ");
        mouseDown("b", 0, 65);

        System.out.print(MAGENTA + " => ");
        mouseDown("m", 0, 240);

        System.out.print(WHITE + " => ");
        mouseDown("w", 150, 150);

        System.out.println("sqrt: " + Math.sqrt(0));
        System.out.println("atan: " + Math.atan2(0, 0));

        for (int x = -10; x <= 10; x++) {
            for (int y = -10; y <= 10; y++) {
                mouseDown("", x, y);
            }
        }
    }

    protected static class HSV {
        double h;
        double s;
        double v;

        protected HSV(double h, double s, double v) {
            this.h = h;
            this.s = s;
            this.v = v;
        }
    }

    protected static class XY {
        int x;
        int y;

        protected XY(int x, int y) {
            this.x = x;
            this.y = y;
        }
    }

    private void mouseDown(String label, int X, int Y) {
        XY pos = new XY(X, Y);

        final double clientWidth = 300;

        double maxRadius = clientWidth / 2;
        double offsetX = pos.x - maxRadius;
        double offsetY = pos.y - maxRadius;
        double radius = Math.sqrt((offsetY * offsetY) + (offsetX * offsetX));

        double angle = Math.atan2(offsetX, offsetY) / Math.PI / 2;

        HSV hsv = new HSV( //
                Math.min(1.0, Math.max(0.0, (offsetX >= 0 ? angle : 1 + angle))), //
                Math.min(1.0, radius / maxRadius), //
                1);

        assertTrue(hsv.h >= 0.0);
        assertTrue(hsv.h <= 1.0);

        assertTrue(hsv.s >= 0.0);
        assertTrue(hsv.s <= 1.0);

        if (!label.isEmpty()) {
            System.out.println(String.format("%s: h:%d, s:%d, v:%d", label, (int) (hsv.h * 360.0),
                    (int) (hsv.s * 100.0), (int) (hsv.v * 100.0)));
        }
    }
```

bundles/org.openhab.ui.basic/web-src/smarthome.js
andrewfg: And, I forgot to mention: the old angle hue code requires the modulo 1.0 correction, whereas my new Math.atan2 code produces a result naturally in the range 0.0 .. 1.0 ..

andrewfg:
@lolodomo as I did not hear from you, I was wondering if you still have doubts about the atan2() code?

The geometry is in the picture below which shows the relationship between atan2(y, x) and atan2(x, y) .. they are simply the opposite angles. And as the widget screen coordinates have an inverted y axis this means that atan2(x, y) is the correct one for our case.

Furthermore atan produces results in the range -PI .. +PI so we need to
- convert (0 .. PI) => (0.0 .. 0.5) 
- convert (-PI .. -0) => (0.5 .. 1.0)

```
temp = Math.atan2(x, y) / Math.PI / 2;
result = offsetX >= 0 ? temp : 1 + temp;
```

The benefit of this is that it always produces naturally a value in the range (0.0 .. 1.0) so A) the NaN check becomes redundant, and B) the additional (0.0 .. 1.0) range normalisation function also becomes redundant.



IMHO we should try to get this into M5 if at all possible..

lolodomo:
> @lolodomo as I did not hear from you, I was wondering if you still have doubts about the atan2() code? 

I just did not find/take time yet to test your proposed code. But this is not ignored or forgotten.

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: Done and tested

lolodomo:
@kaikreuzer : please review and merge before building 3.4M5. This is a fix for a major and annoying bug in Basic UI.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1089,29 +1089,28 @@
...
 			_t.handle.style.top = (pos.y / _t.image.clientWidth) * 100 + "%";
 
 			var
-				angle = offsetX >= 0 ?
-						(Math.PI * 2 - Math.atan(offsetY / offsetX) + Math.PI / 2) / (Math.PI * 2) :
-						(Math.PI * 2 - Math.atan(offsetY / offsetX) - Math.PI / 2) / (Math.PI * 2),
 				hsv = {
-					h: isNaN(angle) ? 0 : angle,
-					s: Math.sqrt(r) / maxR,
+					h: Math.min(1.0, Math.max(0.0, offsetX >= 0 ? angle : 1 + angle)),
```
andrewfg, line 1112: @lolodomo I think that strictly speaking the Math.min and Math.max are no longer needed, since (as far as I can tell), the angle code now always produces a value in the range (0.00 .. 1.00). However the min/max check certainly doesn't do any harm.

```
offsetX >= 0 ? angle : 1 + angle
```

andrewfg (approved):
LGTM

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: Are you sure that "1 + angle" can never be greater than 1 ?

andrewfg: Pretty much ;) ..

reason is that angle is only > 0 if offsetX is > 0 (so in other words 1 + angle is < 1)

alternate (perhaps even better) code could be ..

```
// old
offsetX >= 0 ? angle : 1 + angle
..
// better
angle >= 0 ? angle : 1 + angle
```

lolodomo: Done

lolodomo:
@kaikreuzer : if you could include this PR in 3.4M5, that would be cool.

andrewfg:
@kaikreuzer : gentle reminder that if you could include this PR in 3.4M6, that would be cool.

kaikreuzer:
Thank you guys - I didn't test it myself, but I assume that you have done this extensively already.

PR #1556, opened 2022-11-30
[MainUI] Add possibility to edit JSON variable keys in widgets

flo-mic:
@ghys this PR implements a possibility to work with nested JSON objects/arrays in widget variables. It allows to set the value of a custom key of the variable. The key can be nested in any depth to allow more complex scenarios. 

To realize this the PR adds a new action configuration `actionVariableKey` which allows to set the key of the object. The configuration syntax is like this:
```
actionVariableKey: name
actionVariableKey: user.name
actionVariableKey: user[0].address[1].street
actionVariableKey: [0]
actionVariableKey: [0].label
```

If you want to edit the key `user.name` but the key `user` is missing in the object, the parent hierarchy is automatically created. (Same applies if you want to add an value to an array but the array is not existing) This should make it easier for not so advanced users.

Signed-off-by: Florian Michel <florianmichel@hotmail.de>


flo-mic:
@ghys I extended the functionality, so now you can also use variable keys in oh-input type widgets

ghys:
Thanks, any improvement is good to have, however I would suggest to make the new "key" options "advanced" because I think this could become very confusing very fast if there's no documentation. Most people would use variables to carry scalar values so this new feature would be kind of confusing at first if presented like this.

flo-mic:
@ghys the new option is now advanced, thanks for the hint.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.js
```diff
@@ -122,6 +122,10 @@ export const actionParams = (groupName, paramPrefix) => {
         return ['variable'].indexOf(configuration[paramPrefix + 'action']) >= 0
       }),
     pt(paramPrefix + 'actionVariableValue', 'Variable Value', 'The value to set the variable to').a()
+      .v((value, configuration, configDescription, parameters) => {
+        return ['variable'].indexOf(configuration[paramPrefix + 'action']) >= 0
+      }),
+    pt(paramPrefix + 'actionVariableKey', 'Variable Key', 'The value to set the key of avariable').a()
```
ghys, line 128: Still not satisfied with the description here, it needs to be more detailed:

```suggestion
    pt(paramPrefix + 'actionVariableKey', 'Variable Key', 'Consider the variable value is an object and set the correponding deep property within that object using a key syntax. Examples: <code>user.name</code>, <code>user[0].address[1].street</code>, <code>[0]</code>, <code>[0].label</code>. The inner property and its parent hierarchy will be created if missing.').a()
```

Also remember to regenerate the components' documentation.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
```diff
@@ -15,5 +15,6 @@ export default () => [
   pb('useDisplayState', 'Use Display State', 'Use the formatted state as the value for the input control'),
   pb('showTime', 'Show time', 'Display time when type set to datepicker'),
   pt('defaultValue', 'Default value', 'Default value when not found in item state or variable'),
-  pt('variable', 'Variable', 'Name of the variable to set when the input changes')
+  pt('variable', 'Variable', 'Name of the variable to set when the input changes'),
+  pt('variableKey', 'Variable Key', 'Allows modifying a specific key of a variable object').a()
```
ghys, line 19: Use a description like above.

ghys (changes requested):
Thanks, some suggestion regarding the parameter's description to make it less confusing.

Also, not only oh-input supports setting a variable - oh-button, oh-knob, oh-slider, oh-stepper and oh-toggle do too, so why limit this feature extension to oh-input?

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.js
```diff
@@ -122,6 +122,10 @@ export const actionParams = (groupName, paramPrefix) => {
         return ['variable'].indexOf(configuration[paramPrefix + 'action']) >= 0
       }),
     pt(paramPrefix + 'actionVariableValue', 'Variable Value', 'The value to set the variable to').a()
```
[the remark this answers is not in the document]
flo-mic: done

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.js
flo-mic: done, also updated the docs

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
flo-mic: done

flo-mic:
@ghys thanks for the feedback. I implemented the property for all widgets you was mentioning (was even not aware that all of them can handle variables...) Also the description was updated. Anything else that needs to be changes?

bundles/org.openhab.ui/doc/components/oh-button.md
```diff
@@ -99,11 +99,21 @@ Button performing an action
...
+  <PropDescription>
+    Consider the variable value is an object and set the corresponding deep property within that object using a key syntax. Examples: <ul><li><code>user.name</code></li><li><code>user[0].address[1].street</code></li><li><code>[0]</code></li><li><code>[0].label</code></li></ul>The inner property and its parent hierarchy will be created if missing.
+  </PropDescription>
+</PropBlock>
 <PropBlock type="BOOLEAN" name="clearVariable" label="Clear Variable After Action">
   <PropDescription>
     Name of the variable to clear after performing the action
   </PropDescription>
 </PropBlock>
+<PropBlock type="BOOLEAN" name="clearVariable" label="Clear Variable After Action">
```
ghys, line 112: This doesn't seem right. Did you regenerate the docs using the provided script?

See https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/CONTRIBUTING.md#documentation--resources
(admittedly it's a little hidden)

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/index.js
```diff
@@ -2,13 +2,15 @@ import { WidgetDefinition, pt, pb } from '../helpers.js'
 import { actionGroup, actionParams } from '../actions.js'
 
 const VariableParameter = pt('variable', 'Variable', 'Name of the variable to set on input change')
+const VariableKeyParameter = pt('variableKey', 'Variable Key', 'Consider the variable value is an object and set the corresponding deep property within that object using a key syntax. Examples: <code>user.name</code>, <code>user[0].address[1].street</code>, <code>[0]</code>, <code>[0].label</code>. The inner property and its parent hierarchy will be created if missing.').a()
 const ClearVariableParameter = pb('clearVariable', 'Clear Variable After Action', 'Name of the variable to clear after performing the action')
+const ClearVariableKeyParameter = pb('clearVariableKey', 'Clear Key Propertoy Inside An Object After Action', 'Consider the variable value is an object and remove a corresponding deep property within that object using a key syntax. Examples: <code>user.name</code>, <code>user[0].address[1].street</code>, <code>[0]</code>, <code>[0].label</code>. The property will be removed after performing the action.').a()
```
ghys, line 7:
```suggestion
const ClearVariableKeyParameter = pb('clearVariableKey', 'Clear Object Variable Property After Action', 'Consider the variable value is an object and remove a corresponding deep property within that object using a key syntax. Examples: <code>user.name</code>, <code>user[0].address[1].street</code>, <code>[0]</code>, <code>[0].label</code>. The property will be removed after performing the action.').a()
```

This should update the related docs and solve the remarks in the process after you run the generate.js script.

ghys (approved):
I fixed the review issues above myself, LGTM now.
Very niche feature I think but still valuable, thanks!

Issue #1532, opened 2022-10-22
On requesting /robots.txt return `Disallow: /`


dilyanpalauzov:
## The problem

Requesting http://ip-addr:8080/robots.txt returns by default an html file with header `Content-Type:text/plain` and status code 200.

When the openHAB instance is exposed to Internet, search engines can crawl it and create load.

This applies also to https://demo.openhab.org/robots.txt.

## Your suggestion

Requesting http://ip-addr:8080/robots.txt shall return instead
```
User-agent: *
Disallow: /
```
This way when the openHAB instance is exposed to internet, search engines will not cause load by crawling it.

rkoshak:
> When the openHAB instance is exposed to Internet, search engines can crawl it and create load.

The Security docs state:

> Security Warning: It is vitally important that you MUST NOT directly expose your openHAB instance to the Internet (e.g. by opening a port in your firewall)!

If OH were deployed in a supported manner, I don't think this would be an issue because the search engine would need a username and password to even get to the robots.txt. 

While I don't think this is a bad idea (defense in depth is a good idea and this is really easy to implement), it's not something that should be needed except when someone brashly and against all warnings to the contrary throws common sense out and exposes their OH directly to the Internet without a reverse proxy and password (or other) protections.

ghys:
I still think it makes sense, could be copied the same way as the webmanifest. PR appreciated :)

flo-mic:
@dilyanpalauzov the feature is now available see #1557

lolodomo:
Closed by #1557

PR #1557, opened 2022-12-01
[MainUI] Add robots.txt to disallow search crawler

flo-mic:
This PR adds a robots.txt file to disallow search engines as requested here:  #1532 

**This implementation will not protect your instance on the public internet** nor block malicious crawlers from scanning you. The robots.txt file is considered by known "good" search engine crawlers which will not index the web server due to the disallow statement in the robots.txt file. Malicious crawlers don't care about your your privacy concerns and will scan your instance in any case!

Signed-off-by: Florian Michel <florianmichel@hotmail.de>


ghys (approved):
Non-brainer... thanks for taking care of it!

Issue #1494, opened 2022-09-18
MainUI can't control setpoint items (for example target temperature)


Novanic:
Unfortuantely MainUI can still not control setpoint items like the target temperature of a thermostat. A readOnly=false configured setpoint item is just displayed with an "Analyze" button and a read-only graph is opened with a click on it. That is completely useless. Instead there should for example be a slider widget.

This affects a lot of the bindings. A fix would be great to increase the usability of MainUI.

See also https://github.com/openhab/openhab-webui/issues/765#issuecomment-886258778

OpenHAB 3.3.0 is also affected.

Novanic:
Hello @ghys , could you please re-open the issue? It isn't fixed within OpenHAB 3.4...

ghys:
Can you provide any details? Is the item you're using as a setpoint tagged "Setpoint" (or "Control")?
It's a requirement if you want a stepper or slider by default: see https://github.com/openhab/openhab-webui/pull/1558#issuecomment-1345574684.
Binding authors should also add the appropriate tags to the channel type definitions so they are added when the link is created in the UI.

mkrathome:
> Can you provide any details? Is the item you're using as a setpoint tagged "Setpoint" (or "Control")?
It's a requirement if you want a stepper or slider by default: see #1558 (comment).
Binding authors should also add the appropriate tags to the channel type definitions so they are added when the link is created in the UI.

Hi,
@ghys I'm not quite sure, if @Novanic opens this after I got in struggle with creating sitemap out  of sematic model. I can not say, if it is solved for all the fancy UI stuff, because I'm only use sitemap. But I recognize, that now stepper-card is added at item configuration.

The issue raised up in the livisi-community and we found out, that also the UI-default-Widgets are not useful.

My problem is not solved: if I rebuild my home at semantic model and put some thermostate-equippment as group in a room, and add this group to the sitemap, I have no chance to change the target temperature. I tried many things: item types and all kind of metadata defining the only thing I get to run is a list-widget. 

For me the idea was to modelling the whole system and only but the top level in the sitemap.  I thought, that should be the way to get the System as simple as possible also for not engineers.

Hope you agree and can fix that 

All the Best  Manuel

Issue #399, opened 2020-10-15
[MainUI] Allow manually setting Item state from UI


rkoshak:
One challenge that many users need to overcome is boot strapping the state of some Items. For example, I have a bunch of DateTime Items that drive my Time of Day rules. Some are populated from Astro and others are "hard coded". For those "hard coded" Items I have to go to great lengths to initialize the Items. Options include:

- posting an update through the REST API
- posting an update through the Karaf console
- posting an update from a System started Rule
- encoding the value in metadata and have a rule initialize the Item based on that metadata (what I currently do)

It's not always possible to initialize these Items from the BasicUI. I don't know about HABPanel or MainUI but I would expect to be able to do this from MainUI for all Item types. 

Most of the time it would be so much easier for most use cases to have a way to update the Item's state from the UI. I can imagine being able to click on the field that shows the Item's current state and being able to set it manually from there.

However, if the official stance is that this sort of thing should only be done from the controlling UI I understand. But I would find it intuitive to be able to adjust the Item's state from the Item page where you can set the metadata and other properties.

ghys:
With #426 & the latest fixes in #457 you can achieve what you wish by setting the "Default Standalone Widget" metadata to `oh-input-card` (Input Card).

For the `type` parameter you can use either `date` (native control) or `datepicker` (Framework7 control):

![image](https://user-images.githubusercontent.com/2004147/97201633-f495af00-17b2-11eb-8ea3-26904be27254.png)

![image](https://user-images.githubusercontent.com/2004147/97201383-ada7b980-17b2-11eb-9739-f0e8d9982d5e.png)

In the config for the default widget make sure you enable the "Send Button" option to have the button to send the command.

And if you set the `type` to `texteditor` you can use a String item for rich text (HTML) input - could be useful for groceries list,  sticky notes and the like:

![image](https://user-images.githubusercontent.com/2004147/97202019-63730800-17b3-11eb-9e22-92c088bb9caf.png)

rkoshak:
Would using a Toggle Card work for Switches? What about Contacts and the Numbers? 

Don't get me wrong, the ability to enter text and Date Times is huge! But my intent with the above was to support all Item types as a way to boot strap their values and/or drive tests before a Page or Sitemap has been created.

ghys:
[...]

PR #1558, opened 2022-12-02
[MainUI] extend default standalone/item/cell widget selection

flo-mic:
This PR extends the default widget selection for standalone, cell and item default widgets to provide better default widgets on DateTime and Number items as requested here #399 and here #1494

Basically the PR adds the following default widgets:
- oh-input for datetime on items of type `DateTime` (standalone, list)
- oh-input for numbers on items of type `Number` (standalone, list)
- oh-stepper on `Number:Temperature` items as well as for items with the semantic class `Control` or `SetPoint` with semantic property `Temperature` (standalone, list)
- oh-slider on `Number:*` items with the semantic class `Control` or `SetPoint` and a semantic property from `ColorTemperature`, `Level`, `SoundVolume` (standalone, list, cell)
- oh-toggle on `Number*` items with the semantic class `Control` or `SetPoint` and a semantic property from `Light`, `Power`, `Energy` (standalone, list, cell)
- oh-toggle on `Number*` items with the semantic class `Switch` (standalone, list, cell)

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/standard/cell/default-cell-item.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/standard/cell/default-cell-item.js
@@ -92,9 +92,9 @@ export default function itemDefaultCellComponent (item, itemNameAsFooter) {
     //   }
     // }
 
-    if(item.type.startsWith('Number') && !stateDescription.readOnly) {
+    if (item.type.startsWith('Number') && !stateDescription.readOnly) {
       if ((semanticClass === 'Control' || semanticClass === 'SetPoint')) {
-        if (semanticProperty === 'ColorTemperature' || semanticProperty === 'Level' || semanticProperty === 'Temperature' || semanticProperty === 'SoundVolume') {
+        if (semanticProperty === 'ColorTemperature' || semanticProperty === 'Level' || semanticProperty === 'SoundVolume') {
           component = {
             component: 'oh-slider-cell',
             config: {
--- a/bundles/org.openhab.ui/web/src/components/widgets/standard/cell/default-cell-item.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/standard/cell/default-cell-item.js
@@ -3,10 +3,14 @@
    in the "cellWidget" metadata namespace of the item
  */
 
+import * as Semantics from '@/assets/semantics.js'
+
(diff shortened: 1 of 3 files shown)


ghys:
This looks great, I'll have a look shortly so we can eventually have it in the next release 👍

ghys:
Having second thoughts:

> oh-input for datetime on items of type DateTime (standalone, list)
> oh-input for numbers on items of type Number (standalone, list)

I think most DateTime and Number items are really statuses, even if they're not readonly (the default), and it would be wrong to display an input UI in those cases as well as have adverse side effects (like the state description's pattern being ignored).
I'd offer the input widget by default only if the item has a Control or Setpoint semantic class - the tags are easy to add.

> oh-toggle on Number* items with the semantic class Control or SetPoint and a semantic property from Light, Power, Energy (standalone, list, cell)
oh-toggle on Number* items with the semantic class Switch (standalone, list, cell)

Not sure about the rationale for these two, especially the first one. Their state wouldn't be displayed at all because the toggle doesn't allow it. If their class is Switch then it _could_ make sense to display a toggle in all cases (not only numbers but other types accepting OnOff commands too, like Color).

ghys:
Ok so I modified it as follows:

### Cards, list items
- if the semantic class is Control or Setpoint AND the item is not read only
  - if the type is DateTime => input
  
    image
   - if the type is Number => input
   
     image
   - if the type is Number:Temperature OR the semantic property is Temperature => stepper
   
     image
   - if the semantic property is ColorTemperature, Level or SoundVolume => slider
   
     image

- if the semantic class is Switch (whatever its type but the tag should only be applied to items compatible with OnOff commands) AND the item is not read only => toggle

  image

- else, no changes, fallback to a label or others depending on the type
  
  image

  image

### Cells
- if the semantic class is Control or Setpoint AND the item is not read only
  - if the semantic property is ColorTemperature, Level or SoundVolume => slider
  - _(nothing else because there are no input or stepper cells - we could perhaps use a knob for setpoints though)_
- if the semantic class is Switch (whatever its type but the tag should only be applied to items compatible with OnOff commands) AND the item is not read only => toggle
- else, no changes, fallback to a simple card or others depending on the type

ghys (approved):
I think it's satisfactory now, and (I hope) won't cause many regressions unless the model is wrong.

PR #1565, opened 2022-12-08
[MainUI] [WIP-With question to maintainer] For oh-stepper, allow custom value in the stepper

petero-dk:
Use case:

For my KNX thermostats the setpoint is controlled by a setpoint Offset item, this determines the change of the setpoint from the base set point.

My users are unaware of the base setpoint and just want to adjust the total setpoint.

Example: The base setpoint is 18 C, and the offset is 2 C, this means that the actual setpoint is 20 C

I want the text in the center of the stepper to show 20C but actually change the item that has 2 C.


ghys:
Nothing against it, but putting a value that's not the item's state is unusual, so you'll want to have the parameter 1) made "advanced" i.e. hidden by default and 2) clearly stating in the description that it's an override to the default behavior ("Override central value display"?)

> Obviously this can be expanded to the knob control also, and needs to be implemented into the dependent widgets i.e. oh-stepper-cell if the idea is approved

It's automatic, the standard library widgets (cards & cells) normally pass their own parameters to the underlying system library widget and get a parameter group from the latter parameters (see their definitions).
The knob control's parameters are based on the [vue-knob-control](https://github.com/kramer99/vue-knob-control), it only appears to allow customization of the value text through a function, which is not supported by openHAB widget expressions, so it might prove difficult.

florian-h05:
@petero-dk 
I don't know how your approach should work:

1. You could use an expression like `"=(Number(items.Setpoint.state) + 20)"` to display the real temperature and not the offset.
2. Your offset is 2, your setpoint therefore is 22. Now you press the + button, then the setpoint widget will send 23 to the offset Item.

I'd rather propose to have an `offset` parameter for the widget, that takes any float and adds this to the Item state for the display, but subtracts it for the sendCommand. I can open a PR for this in the next time.

florian-h05:
Superseded by #1875.

petero-dk:
@florian-h05 thank you for building on this, I have been swamped so never got the time to get back to this.

PR #1575, opened 2022-12-11
[MainUI] `oh-sipclient`: Fix microphone access stays active when foreground is left on iOS

florian-h05:
## Description

This PR fixes an issue on iOS devices, where microphone access stayed active when the `oh-sipclient` component left the foreground. Microphone access could only be stopped by reloading MainUI.
It also included a small change to avoid having the same issue when using a video call and accessing the local video (I haven't experienced this because I don't use video calls).

Moreover, comments are changed to uppercase and JSDoc is added for some methods.

## Testing

I have deployed this to my production system and it works fine now. Microphone access stops when I switch to another part of MainUI or switch, but keeps active when I leave Safari or change to another tab (I guess because `stopForegroundActivity()` is not called here). When I use the openHAB iOS, microphone access also stops when I put the app in the background.

@ghys This can be merged after the feature freeze since it's a bugfix, right?

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-sipclient.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-sipclient.vue
@@ -67,7 +67,8 @@ export default {
       remoteParty: '',
       phonebook: new Map(),
       loggerPrefix: 'oh-sipclient',
-      showLocalVideo: false
+      showLocalVideo: false,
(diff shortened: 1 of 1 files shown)


ghys (approved):
> @ghys This can be merged after the feature freeze since it's a bugfix, right?

Yes!
I'll trust you confirmed it fixed the actual problem since I can't test but LGTM.

PR #1576, opened 2022-12-11
[MainUI] Update JS Scripting tern definitions

florian-h05:
Reference
https://github.com/openhab/openhab-js/compare/v3.0.0...v3.1.2.

## Description

This updated the tern definitions for the current openhab-js version included in the addon (3.1.2).

The tern definitions for the `cache` namespace were updated to allow access to the `shared` and the `private` caches from core.

## Testing

I need to check whether autocompletion for `cache` works correctly.

@ghys This should go into 3.4.0. Since the feature freeze is already active, this should get merged as bugfix. 
It indeed fixes deprecation warnings that will be logged when someone tries to use something like `cache.get()` (which now would be `cache.private.get()` or `cache.shared.get()`).

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/openhab-js-tern-defs.json
+++ b/bundles/org.openhab.ui/web/src/assets/openhab-js-tern-defs.json
@@ -350,17 +350,31 @@
               }
         },
         "CacheNamespace": {
+            "shared": {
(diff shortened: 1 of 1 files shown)


ghys (approved):
> This should go into 3.4.0. Since the feature freeze is already active, this should get merged as bugfix.

Absolutely, feel free to submit any PR, the important thing is that the tern definitions match the library at release time.

florian-h05:
You are absolutely right 👍
 
I think there should be more more changes in the library because the feature freeze is active and all changes that I have for the library are enhancements.

PR #1574, opened 2022-12-11
Improved sitemap handling in MainUI

mherwege:
Signed-off-by: Mark Herwege [mark.herwege@telenet.be](mailto:mark.herwege@telenet.be)

Closes [#1538](https://github.com/openhab/openhab-webui/issues/1538)
Closes [#748](https://github.com/openhab/openhab-webui/issues/748)
Closes [862](https://github.com/openhab/openhab-webui/issues/862)

This PR improves sitemap handling for strings with blanks and special characters.
It also fixes attributes with value 0.

PR #1578, opened 2022-12-13
Normalize channel label before suggesting it as item name

Hilbrand:
Normalizing allows for channel names to contain subscript characters like in units; CO2.

Fixes https://github.com/openhab/openhab-addons/issues/13909
(Solution taken from https://codegolf.stackexchange.com/a/238478)

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/channel-list.vue
+++ b/bundles/org.openhab.ui/web/src/components/thing/channel-list.vue
@@ -89,8 +89,6 @@
 </style>
 
 <script>
-import diacritic from 'diacritic'
-
 import ChannelGroup from './channel-group.vue'
 import ChannelLink from './channel-link.vue'
 import ItemForm from '@/components/item/item-form.vue'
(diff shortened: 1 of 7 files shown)


ghys:
Thanks @Hilbrand!
Note that normally I would object to this cursed syntax (I mean, it's ugly), maybe `.normalize('NFKD')` is more palatable. Despite the project using Babel I refrained from using JS features that are too recent.

Another thing to consider is that this normalization happens in multiple places, not only in `add-link.vue`, so you might have missed some. For instance, you have similar things happening when you mass add items with the "Add Equipment to Model" feature and possibily in other places too.
So while it was a mistake to not doing this before, it might be time to refactor this "normalize any string into a valid item name" requirement into a pure JS function that would be testable.
I recently resurrected the unit tests for the UI, planned but never followed up on, here: https://github.com/openhab/openhab-webui/pull/1574#issuecomment-1346610271
I think the way to go is to implement this function, with unit tests, and use it wherever is needed, wdyt?

Hilbrand:
Ah. Didn't place to much attention to the syntax. Will change.
Sounds like a good idea to make it a util. What is your suggestion? I'm not too familiar with the structure. But was thinking about adding a `util.js` to `js/openhab` containing a function `normalizeLabel`. wdyt? I've already identified some other places this function needs to be applied.
Tests cases would be sweet. I'll have to look into that.

ghys:
> But was thinking about adding a util.js to js/openhab containing a function normalizeLabel. wdyt? 

Sounds good, looks like the proper place for it - openHAB related pure JS code.

Hilbrand:
I've introduced the util method and added a jest test.

ghys:
Great! Small cosmetic remark, since there's already `this.$f7.utils` in the prototype would you mind using the plural as well (`this.$oh.utils`) to match?

Hilbrand:
Renamed to `utils`

ghys (approved):
Excellent, much better now, this should be the way to go for such type of code from now on.
Thanks again!

PR #1587, opened 2022-12-17
Show Basic UI and CometVisu intros in Main UI

wborn:
Using these Markdown tweaks a small intro will show in Main UI instead of emptiness and a "more" button:

Screenshot from 2022-12-17 21-59-40


lolodomo (approved):
LGTM

PR #1600, opened 2022-12-22
Upgrade to Karaf 4.4.3

### LINKED ISSUE #1172: HTTP code 200 returned for invalid pages
When connection to the openHAB server and try to access to notifications api (it's only on cloud, so this should produce an error), I get http code 200:
````
$ curl http://openhabian:8080/api/v1/settings/notifications -v
[...]
< HTTP/1.1 200 OK
< ETag: "1624816166000"
< Content-Type: text/html
< Content-Length: 1215
< Server: Jetty(9.4.40.v20210413)
[...]
````

This causes the Android app to recognize the server as cloud instance and tries to parse the response as json, but it's html. IMO the server should respond with 404 for not found subpages.

Browser for reference:
grafik

Android issue: https://github.com/openhab/openhab-android/issues/2662

### PR DESCRIPTION
**Pull request by wborn**
* Rework Servlets to use Http Whiteboard annotations in favor of proprietary `org.openhab.core.io.http.servlet` classes

---

Depends on openhab/openhab-core#3252

To do:

* [x] Rebase and fix merge conflicts (most notably #1521)
* [x] Upgrade to the latest greatest Karaf 4.4 version
* [x] Check, fix add-on servlets that no longer work
* [x] Fix missing content type headers causing openHAB SVG logo not showing in main UI
* [x] Remove or fix UIErrorPageServlet

Fixes #1172

### CODE REVIEW

**Comment by J-N-K** [comment]
There is one issue with MainUI, the openHAB logo in the upper left corner is not properly loading. I couldn't figure out why. BasicUI is fine as far as I can see. 

I can't check the full build because as mentioned in https://github.com/openhab/openhab-webui/pull/1353 it's not possible to build HaBOT on Apple Silicon.

**Comment by J-N-K** [comment]
Did you get the openHAB logo to load?

**Comment by wborn** [comment]
> Did you get the openHAB logo to load?

it's still an issue so I'll add it to the todos.

**Comment by J-N-K** [comment]
Maybe @digitaldan can help us here?

**Comment by digitaldan** [comment]
>Did you get the openHAB logo to load?

Can someone post a screenshot from the chrome developer/debug window of the network request (including headers and response) failing ?

**Comment by J-N-K** [comment]
Missing logo:

Bildschirm­foto 2023-01-11 um 22 03 19

Element:

Bildschirm­foto 2023-01-11 um 22 05 38

HTTP request:

Bildschirm­foto 2023-01-11 um 22 04 19

HTTP response:

Bildschirm­foto 2023-01-11 um 22 04 39

**Comment by digitaldan** [comment]
Hmm, there's no content type on that like `content-type: image/svg+xml`, but the returned length looks right.    I'm assuming that content type  would be handled internally by Jetty as we don't do any custom mapping there.   Do you know if SVGs are working else where ?

**Comment by digitaldan** [comment]
Actually, that content length does not look right, way to big compared to my own system:
image

Can you curl that URL and see what the server is actually returning ?

**Comment by digitaldan** [comment]
Actually, never mind, it does look the right size,  i forgot i was going through my Caddy server which was doing some compression, here's the image directly to my OH, although it does contain a content-type which may be the missing element here.
image

**Comment by J-N-K** [comment]
Yes, two headers are missing:

```
Content-Type: image/svg+xml
Vary: Accept-Encoding
```

Requesting the file directly shows the correct file.

J-N-K:
I found that the reason for the missing content-type is that the MIME-type map in the ContextHandler is empty. I didn’t check if that was the case before.

wborn:
The content type is cached as well and based on the `Resource.toString` value. :upside_down_face:

See:

https://github.com/eclipse/jetty.project/blob/da9a0b30691a45daf90a9f17b5defa2f1434f882/jetty-server/src/main/java/org/eclipse/jetty/server/CachedContentFactory.java#L411

J-N-K:
And that works? I was already there with the debugger, but `_mimeTypes` has an empty map in my system.

wborn:
It looks empty but if it is empty it falls back to use `__dftMimeMap` which does not show in your debugger if you hover over it. :slightly_smiling_face:

wborn:
@openhab/webui-maintainers can you review this PR?
It's required for successful builds now that https://github.com/openhab/openhab-core/pull/3252 got merged.

kaikreuzer (approved):

Issue #1597, opened 2022-12-20
Migrate Blockly to JS Scripting (GraalJS)


florian-h05:
Awaiting #1601.

## The current situation

In #1170, a first solution how to make the Blockly blocks compatible with JS Scripting (GraalJS) was proposed, however this solution has two problems:

- it won't work anymore: GraalJS and Nashorn, which is currently used but will go away with openHAB 4, have seperate MIME types
- it doesn't present the "best" possible code to users

To port Blockly to JS Scripting, we therefore came to the conclusion (also in #1170), that Blockly should use the integrated [openhab-js](https://github.com/openhab/openhab-js) library instead of the raw Java APIs.

## Our approach/solution

Blockly maintainer @stefan-hoehn and I (openhab-js and JS Scripting addon maintainer) joined forces and decided on the following approach:

- Support both Nashorn and GraalJS at the same time (= support code generation for both)
- GraalJS will be available by default on openHAB 4 and will own the default MIME type, so generate for GraalJS by default
- Existing Blockly scripts will get automatically converted when save is clicked
- If Blockly is on GraalJS, provide additional functionality, e.g. https://github.com/openhab/openhab-js/issues/206
- If one wants to use Nashorn for Blockly, #1601 can be used to set Nashorn as the engine

## Concerns

- [x]  _GraalJS is more strict on string concatenation, e.g. when adding a string to an Item‘s state:_ This is no problem, since openhab-js avoids returning Java objects where possible and uses JS natives, e.g. for Item state and Thing status we use `string`.
- [x] _Notably differences in the implementation of timers and (multi-)threading:_ Yes, they exist, but nearly all entry points are synchronized, means all timer creation APIs (`setTimeout`, `setInterval`, `createTimer`) create threadsafe timers and `Thread.sleep` also works.
- [x] _Cancelling a running timer or a running `Thread.sleep` can to a thread/sleep interrupted warning:_ This is by design, but it will never occur with Blockly or generally from a single script (multi-thread synchronization makes it impossible to call `cancel` on a currently executing timer). This can only happen when a currently executing timer created by script a is cancelled by script b (to make this possible, the timer must be shared across the scripts).
- [ ] The above described Problem can occur when using `sharedScript`. We should address the log issue.
- [ ] _Running a rule from another rule can lead to multi-thread access requested but not allowed:_ This is still a problem, but it can be solved. I‘ll have a look at this.

## To Dos

[...]

PR #1601, opened 2022-12-22
[MainUI] Add tags & script language to script settings

florian-h05:
Reference #1597.
Reference https://github.com/openhab/openhab-addons/issues/14005.

## Description

This PR extends the script settings to allow the user to set tags and change the MIME type after creation of the script.
The ability to add and edit tags was requested several times (I think also by @rkoshak).

The ability to change the MIME type of a script after creation is **very important for openHAB 4**, because Nashorn will (very likely) swap it's MIME type with GraalJS, but Nashorn scripts do not run on GraalJS without some changes. With the extended script settings, the user can just change the MIME type to the new MIME type of Nashorn.

The "Script" tag that all scripts have is hidden in the settings.

### Script Creation

image

Note that the "Tags" section is newly added and doesn't show the "Script" tag although it is present.

### Editing a Script

Here has already been a settings popup. It has been extended with tags and the ability to change the MIME type:
image

Note that this script settings are from a Blockly script, therefore only Javascript versions (I've only installed GraalJS) are listed. If you open these settings for a normal script, you can change the MIME type to any of the available engines, e.g. Rules DSL.

### Editing a Script in a Rule

The settings are newly introduced for this case and limited to the MIME type:
image

/ping @ghys @rkoshak @stefan-hoehn


stefan-hoehn:
I love the idea. Again a step forward to easier allow to the new ECMAScript version.

regarding the tags:

I recently added some more search criteria in /settings/rules/ (see https://github.com/openhab/openhab-webui/pull/1540)
Maybe it is possible to show and search for the tags there (like in /settings/items/) as well?

florian-h05:
I've just checked it:
- searching by tags works
- searching by script states works
- searching by UID works
- searching by description works

FYI: The code to list rules and the one to list scripts is the same and scripts are in fact only a special type of rules.

florian-h05:
@ghys This is now ready for review, it's working fine on my system and doesn't print any errors to the console.

stefan-hoehn:
Does it not only search for the tags but also show the tags?. Our idea until now was that everything that is searchable is also shown on the list and vice versa.

florian-h05:
It also shows the tags -- it behaves exactly the same like the rule overview.

bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
```diff
@@ -243,6 +231,22 @@ export default {
         this.load()
       })
     },
+    loadScriptModules () {
+      this.$oh.api.get('/rest/module-types/script.ScriptAction').then((data) => {
```
ghys, line 235: One tiny general remark:
this component is also used to edit script conditions.
I understand you probably took this into account and figured there was no reason the list of languages would be different from the action to the condition, so good enough, but just pointing it out.

florian-h05: Thanks for that information. 
I wasn‘t aware of this, but as there is no difference in the list of available languages for actions and conditions, this should be fine.
If I see it correctly, this component also did not take into account whether it edits an action or a condition before this PR.

PR #1613, opened 2022-12-29
[MainUI] Script Editor: Block runRule and save while rule is running (or uninitialized)

florian-h05:
## Description

- Fixes blocking of the run rule command while rule is RUNNING or UNINITIALIZED. This was already implemented, but nor working.
- Additionally blocks save rule while a rule is RUNNING to avoid problems with automation engines like JS Scripting (GraalJS)
- Shows toast messages when run rule or save rule is blocked and explains why it is blocked

Note: Block means that a click on the button does not execute the action, the button is still visible and looks normal.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
@@ -329,7 +329,13 @@ export default {
     },
     runNow () {
       if (this.createMode) return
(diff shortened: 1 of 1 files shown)


ghys (approved):
LGTM, that conditional message could be a minor hurdle when (if) we finally implement l10n for the admin screens but for now it's okay!

florian-h05:
Yes, you are probably right but then we can introduce two different messages or something like this.

PR #1621, opened 2023-01-02
Limit alexa parameter state description options

jsetton:
Related to https://github.com/openhab/openhab-alexa/pull/540


ghys (approved):
I read through https://github.com/openhab/openhab-alexa/pull/540 and think I understood the problem.

Hope you know what you're doing, 150 seems like a pretty future-proof yet arbitrary limit ;)

jsetton:
> Hope you know what you're doing, 150 seems like a pretty future-proof yet arbitrary limit ;)

Per my testing 150 standard sized options produce a compressed endpoint cookie of ~4KB leaving room for additional capabilities in the case of a group endpoint. I initially was considering a 100 limit but I figure the likelihood of having two capabilities with 100+ options under the same group endpoint is highly unlikely. So I went with a higher limit.

PR #1623, opened 2023-01-03
show custom namespaces for items

stefan-hoehn:

depends on https://github.com/openhab/openhab-core/pull/3298

Shows even the custom namespaces created by a user in the UI of the item

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
@@ -1,25 +1,21 @@
 <template>
   <f7-card>
-    <f7-card-content v-if="item.metadata && Object.keys(item.metadata).filter((n) => n !== 'semantics').length > 0">
+    <f7-card-content v-if="item.metadata && Object.keys(item.metadata).filter((n) => n !== 'semantics2').length > 0">
       <f7-list>
         <ul>
           <f7-list-item
-            v-for="namespace in standardNamespaces" :key="namespace"
-            :link="'/settings/items/' + item.name + '/metadata/' + item.metadata[namespace].id"
-            :title="item.metadata[namespace].label"
-            :after="(item.metadata[namespace].id) ? item.metadata[namespace].id : 'Not Set'" />
+            v-for="namespace in wellKnownNamespaces" :key="namespace"
+            :link="'/settings/items/' + item.name + '/metadata/' + namespace.name"
+            :title="namespace.label"
+            :after="namespace.value || 'Not Set'" />
         </ul>
-      </f7-list>
-    </f7-card-content>
-    <hr>
-    <f7-card-content v-if="item.metadata && Object.keys(item.metadata).filter((n) => n !== 'semantics').length > 0">
-      <f7-list>
-        <ul>
+        <ul v-if="customNamespaces.length > 0">
+          <f7-list-item divider />
           <f7-list-item
             v-for="namespace in customNamespaces" :key="namespace"
-            :link="'/settings/items/' + item.name + '/metadata/' + item.metadata[namespace].id"
-            :title="item.metadata[namespace].label"
-            :after="(item.metadata[namespace].id) ? item.metadata[namespace].id : 'Not Set'" />
+            :link="'/settings/items/' + item.name + '/metadata/' + namespace.name"
+            :title="namespace.label"
+            :after="namespace.value || 'Not Set'" />
         </ul>
       </f7-list>
     </f7-card-content>
@@ -42,11 +38,39 @@ export default {
     }
   },
   computed: {
-    standardNamespaces () {
-      return Object.keys(this.item.metadata).filter((n) => !this.item.metadata[n].custom)
+    editableNamespaces () {
+      if (!this.item.metadata) return []
+      // TODO: determine somehow if other MetadataProviders are not editable
+      // for now we'll assume they're all editable except "semantics"
+      return Object.keys(this.item.metadata)
+        .filter((n) => n !== 'semantics')
+        .map((n) => {
+          return {
+            name: n,
+            value: this.item.metadata[n].value
+          }
+        })
+    },
+    wellKnownNamespaces () {
+      return this.editableNamespaces
+        .filter((n) => MetadataNamespaces.some((wk) => wk.name === n.name))
+        .map((n) => {
+          const wellKnown = MetadataNamespaces.find((wk) => wk.name === n.name)
+          return {
+            ...n,
+            label: wellKnown.label
+          }
+        })
     },
     customNamespaces () {
-      return Object.keys(this.item.metadata).filter((n) => this.item.metadata[n].custom)
+      return this.editableNamespaces
+        .filter((n) => !MetadataNamespaces.some((wk) => wk.name === n.name))
+        .map((n) => {
+          return {
+            ...n,
+            label: n.name
+          }
+        })
     }
   },
   methods: {
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
@@ -40,7 +40,8 @@ export default {
   computed: {
     editableNamespaces () {
       if (!this.item.metadata) return []
-      // TODO: determine somehow if other MetadataProviders are not editable
+      // TODO: determine somehow if other namespaces are not editable
+      // (non-managed MetadataProvider)
       // for now we'll assume they're all editable except "semantics"
       return Object.keys(this.item.metadata)
         .filter((n) => n !== 'semantics')
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
@@ -4,11 +4,10 @@
       <f7-list>
         <ul>
           <f7-list-item
-            v-for="namespace in item.metadata" :key="namespace"
-            v-if="!namespace.custom"
-            :link="'/settings/items/' + item.name + '/metadata/' + namespace.id"
-            :title="namespace.label"
-            :after="(item.metadata[namespace.id]) ? item.metadata[namespace.id].value : 'Not Set'" />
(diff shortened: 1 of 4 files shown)


stefan-hoehn:
Btw, if I view this right, there is no way currently to delete a metadata namespace, is there?

florian-h05:
There is a REST endpoint for this:
C2C234FC-E0DB-4DA8-B9F3-C86C04204279

bundles/org.openhab.ui/web/src/pages/settings/items/item-details.vue
```diff
@@ -184,9 +184,18 @@ export default {
...
-        this.item = data
-        this.iconUrl = (localStorage.getItem('openhab.ui:serverUrl') || '') + '/icon/' + this.item.category + '?format=svg'
+      let metadataNamespacesList = MetadataNamespaces.map((n) => n.name).join(',')
+      this.$oh.api.get(`/rest/items/${this.itemName}/metadata/namespaces`).then((itemNamespaces) => {
+        this.$oh.api.get(`/rest/items/${this.itemName}?metadata=${itemNamespaces.join(',')}`).then((data) => {
+          // map the labels of the known namespaces or add "custom:" to the user defined namespaces
+          Object.keys(data.metadata).forEach(namespace => {
+            let rs = MetadataNamespaces.find(ns => ns.name === namespace)
+            data.metadata[namespace].id = namespace
+            data.metadata[namespace].label = (rs === undefined) ? 'custom: ' + namespace : rs.label
```
splatch, line 194: Not sure if marking a namespace as custom as each of them is custom, I'd rather say others are "predefined" or "standard". Maybe not making any extra text there would be better because any prefix or suffix brings another implication - a translation.

bundles/org.openhab.ui/web/src/pages/settings/items/item-details.vue
```diff
@@ -184,9 +184,18 @@ export default {
       this.$store.dispatch('stopTrackingStates')
     },
     load () {
-      this.$oh.api.get(`/rest/items/${this.itemName}?metadata=semantics,${MetadataNamespaces.map((n) => n.name).join(',')}`).then((data) => {
-        this.item = data
-        this.iconUrl = (localStorage.getItem('openhab.ui:serverUrl') || '') + '/icon/' + this.item.category + '?format=svg'
+      let metadataNamespacesList = MetadataNamespaces.map((n) => n.name).join(',')
+      this.$oh.api.get(`/rest/items/${this.itemName}/metadata/namespaces`).then((itemNamespaces) => {
+        this.$oh.api.get(`/rest/items/${this.itemName}?metadata=${itemNamespaces.join(',')}`).then((data) => {
```
florian-h05, line 189:
```suggestion
        this.$oh.api.get(`/rest/items/${this.itemName}?metadata=.+`).then((data) => {
```

Does fetch all metadata for that Item and you can remove the call to `/rest/items/{metadata}/namespaces`.

bundles/org.openhab.ui/web/src/pages/settings/items/item-details.vue
stefan-hoehn: I hear what you are saying though I still think some marking makes sense as you are _creating_ it as a _custom_ name as well. These are kind of special. 

image

I was even thinking if we should kind of sort them to the bottom. With some effort we could maybe mark them with some kind of icon though I probably have to research bit how to do that here in vue.

ghys: > // map the labels of the known namespaces or add "custom:" to the user defined namespaces

Yeah I agree with @splatch here - I don't support the UI making assumptions or restrictions or enforcing prefixes on the namespaces you can edit.

stefan-hoehn: I am not sure if I brought across my intent. My idea was to somehow make visible the namespaces that are custom created by the user and the ones that are known by openHAB. I was't giving them prefixes in the name I only presented that in the UI. What about marking them with an icon or just separate them by a line?

image

florian-h05: I support the idea of somehow making visible that a namespaces is a custom user namespace that is not directly linked to any openHAB functionality (by this I mean for example the expire metadata that is read and processed by core). 

Prefixing what is shown IMO can be confusing for the end user, I’d rather put in a line together with some wording like „User defined metadata“.

stefan-hoehn: I optimised the loading via the ".*" param and separated standard and custom namespaces by a horizontal line (and removed the icon). See edited description of the PR.

florian-h05:
Fixes #1633.

ghys:
@stefan-hoehn could you please fix the builds i.e. the failing lint checks?

bundles/org.openhab.ui/web/src/pages/settings/items/item-details.vue
```diff
@@ -184,7 +184,13 @@ export default {
       this.$store.dispatch('stopTrackingStates')
     },
     load () {
-      this.$oh.api.get(`/rest/items/${this.itemName}?metadata=semantics,${MetadataNamespaces.map((n) => n.name).join(',')}`).then((data) => {
+      this.$oh.api.get(`/rest/items/${this.itemName}?metadata=.+`).then((data) => {
```
ghys, line 187: This fetch is also in the model page. It doesn't work there (search for MetadataNamespaces across the project):

image

bundles/org.openhab.ui/web/src/pages/settings/items/item-details.vue
```diff
@@ -184,7 +184,13 @@ export default {
...
+        Object.keys(data.metadata).forEach(namespace => {
+          let rs = MetadataNamespaces.find(ns => ns.name === namespace)
+          data.metadata[namespace].id = namespace
+          data.metadata[namespace].label = (rs === undefined) ? namespace : rs.label
+          data.metadata[namespace].custom = (rs === undefined)
+        })
```
ghys, line 193: I'm not a fan of changing the response from the API just to determine whether a namespace is well-known or not (and add fields with the label and a flag).
This logic must be enclosed in the item-metadata-menu.vue component where it doesn't depend on the parent component already having performed that logic before passing the data in the prop.

bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
```diff
@@ -4,10 +4,22 @@
...
+            v-for="namespace in standardNamespaces" :key="namespace"
+            :link="'/settings/items/' + item.name + '/metadata/' + item.metadata[namespace].id"
+            :title="item.metadata[namespace].label"
+            :after="(item.metadata[namespace].id) ? item.metadata[namespace].id : 'Not Set'" />
+        </ul>
+      </f7-list>
+    </f7-card-content>
+    <hr>
+    <f7-card-content v-if="item.metadata && Object.keys(item.metadata).filter((n) => n !== 'semantics').length > 0">
+      <f7-list>
```
ghys, line 16: That `<hr>` is ugly, use this:

```suggestion
        </ul>
        <ul v-if="customNamespaces && customNamespaces.length > 0">
          <f7-list-item divider title="Custom" />
          <f7-list-item
```
Filtering out the semantics namespace can happen in the computed property.

ghys (changes requested):

bundles/org.openhab.ui/web/src/pages/settings/items/item-details.vue
stefan-hoehn: I tried this but I ended up with an v-if statement in the v-for which is not allowed to be used (the linter told me), so I came up with this idea.

Also it is not only to detect if it is a custom namespace but also to enrich the API information with label that is known from the MetadataNamespaces and for those that are unknown, detect that it is a custom namespace. Also I need to detect that it is a custom namespace. 

The alternative I could think of could the be that the whole label enrichment would then be made in the computed section instead which to me seems to be more complicated than my approach and would require a completely different approach but I may be the wrong person then to do that, fair enough... maybe someone else takes over then.

bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
stefan-hoehn: Adding this instead of hr looks like 
image
but I think the "custom" is not looking good.

I would recommend to leave the title "Custom" out:

image

What about "semantics"? Should be add it to the known "MetadataNamespaces" or does it belong to "custom" or should we filter it out completely (not show it all)?

ghys: > I would recommend to leave the title "Custom" out

Okay, it looks slightly better indeed.

> What about "semantics"? Should be add it to the known "MetadataNamespaces" or does it belong to "custom" or should we filter it out completely (not show it all)?

This namespace is provided by a different MetadataProvider so it's not editable (in fact you already have the info in "Semantic Classification" on the item page). So it shouldn't be displayed at all in this list.
By the way, this PR allows the listing of every namespace (except "semantics") as they were editable, as there's no way to determine from which MetadataProvider they come. The problem is, if they're from a non-managed MetadataProvider (for example provided by an add-on) they will be presented as editable but saving will have adverse effects (they would then be defined twice on different providers I think, so I don't know what will happen). See https://github.com/openhab/openhab-core/issues/3281 for more info.

bundles/org.openhab.ui/web/src/pages/settings/items/item-details.vue
ghys: > The alternative I could think of could the be that the whole label enrichment would then be made in the computed section instead which to me seems to be more complicated than my approach and would require a completely different approach but I may be the wrong person then to do that, fair enough... maybe someone else takes over then.

No worries. It's indeed the more correct implementation as this logic is only for the display purposes of the item-metadata-menu component, so processing the data is this component's concern only, and you can simply pass it the item unmutated (mitigates mistakes like the missing change in model.vue as the component is used there too).

See https://github.com/openhab/openhab-webui/pull/1623/commits/e85af092bbae1b44d24c19b4d0bf5fe28ac2db95 for the implementation.
I fixed the model page and made some cleanups.

ghys (approved):

PR #1624, opened 2023-01-05
Update license headers to 2023

wborn:
I have updated the headers because it is now 2023. :slightly_smiling_face:

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/BasicUITile.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/BasicUITile.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010-2022 Contributors to the openHAB project
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information.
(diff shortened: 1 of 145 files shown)


splatch:
Isn't it better to make code headers rely on last change which reflects copyright? Change of copyright header does not extend copyright itself. Maybe https://github.com/diffplug/spotless/tree/main/plugin-maven#retroactively-slurp-years-from-git-history option would help?

Updating copyright headers may cause unnecessary conflicts for PRs and also cause a lot of false alarms with spotless itself (wrong date), even if code was authored before given year.

wborn:
I'm happy with the way it works. 🙂 There is no proper Git history for most repos because a lot of code got relocated and commits got squashed.

splatch:
> I'm happy with the way it works. slightly_smiling_face There is no proper Git history for most repos because a lot of code got relocated and commits got squashed.

Not sure if  people who don't know how to rebase their PR or run into unexpected conflicts can say same thing. I am afraid other group is more than just one person.

Other point you raise suggest that statement "Copyright (c) 2010-2022 Contributors to the openHAB project" might not be valid. Maybe not for this repository but for other ones.

wborn:
I'm no maintainer here, just helping them a bit where I can. 😉

kaikreuzer (approved):
Thanks, lgtm.
I am not aware of major issues in the past years wrt this process, so I'd also prefer to keep it as is.

Issue #1542, opened 2022-11-13
missing basic auth for basicui results in user errors & secvuln: information leakage


rdslw:
- [X] Basic UI

There is a dual problem related to basic auth implementation within basic ui:
* low security vuln: information leakage
* missing http Authorization header in basicui results inability to use basic auth for basic ui 

Detailed description:
1. Even if 'Implicit User Role' is disabled, basicui is able to retrieve and show to unathenticated user all relevant items, state etc. It's not possible (in such case) to write to items, but reading is possible. This is contrary to what documentation states, and contrary to user expecting what it shall do upon blocking it. 

2. if user disables option 'Implicit User Role' and enables option 'Allow Basic Auth', after going into http://openhab/basicui/app there is no Authorization header served by openhab, resulting in inability to use basicui to steer. Just showing this header would solve this problem.

ad 3. combination of 1&2 results in many people misunderstanding how authentication works (while in fact it does not) for basicui, resulting in numerous answers on community.openhab.org that 'enabling both implicit user role and basic auth' is important and proper for basicui authenticated usage. This is not the case. Just implicit user role enabled is enough and gives ANYBODY (basic auth having no effect) to full access to basicui.

## Expected behavior/fix proposed

a. If option 'basicauth' is enabled, all urls http://openhab/basicui/ should present http Authorization header and use this information for authentication
b. if option 'Implicit user role' is disabled, basicui shall NOT have ability to read and present items state

## Your environment

openhabian 1.7
openhab 3.3
tested on chrome/firefox@linux and safari@ios

lolodomo:
BasicUI is very old and authentication was added later in OH3 when MainUI became the main UI (and supported authentication) . 

Just using default authentication settings is OK to use BasicUI. 

If you or someone else wants to add support for authentication in BasicUI, he is welcome.

rdslw:
Thank you kindly for the insight about OH1/OH2/OH3 history.

I agree that full authentication support would be an enhancement.

In the meantime, currently case 1 as described above  is a result of bug in current code, and is a security vulnerability, allowing basicui to circumvent restapi security, and read values EVEN if implicit role is disabled. This scenario is currently  implemented while shall stop fully basicui, which is not the case, as now it only stops write, but not reads.

May I suggest, change this issue to the bug itself, as it's shorter to fix, and actually important, and does not need addingauthentication support to basicui.

florian-h05:
[...]

PR #1626, opened 2023-01-07
[WIP][basicui] Block the UI when implicit user role is disabled

lolodomo:
Related to #1542

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/servlet/CmdServlet.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/servlet/CmdServlet.java
@@ -54,22 +54,27 @@ public class CmdServlet extends BaseServlet {
     public static final String SERVLET_NAME = "CMD";
 
     private final EventPublisher eventPublisher;
+    private final ImplicitUserRoleSecurityHandler securityHandler;
 
     @Activate
     public CmdServlet(final @Reference HttpService httpService,
             final @Reference HttpContextFactoryService httpContextFactoryService,
-            final @Reference ItemRegistry itemRegistry, final @Reference EventPublisher eventPublisher) {
+            final @Reference ItemRegistry itemRegistry, final @Reference EventPublisher eventPublisher,
+            final @Reference ImplicitUserRoleSecurityHandler securityHandler) {
         super(httpService, httpContextFactoryService, itemRegistry);
         this.eventPublisher = eventPublisher;
+        this.securityHandler = securityHandler;
     }
 
     @Activate
     protected void activate(BundleContext bundleContext) {
+        securityHandler.addUriToChecks(WEBAPP_ALIAS);
         super.activate(WEBAPP_ALIAS + "/" + SERVLET_NAME, bundleContext);
     }
 
     @Deactivate
     protected void deactivate() {
+        securityHandler.removeUriFromChecks(WEBAPP_ALIAS);
         httpService.unregister(WEBAPP_ALIAS + "/" + SERVLET_NAME);
     }
 
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/servlet/ImplicitUserRoleSecurityHandler.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/servlet/ImplicitUserRoleSecurityHandler.java
@@ -1,0 +1,132 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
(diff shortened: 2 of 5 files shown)


lolodomo:
Here is the result:
image

The used page is the same as when listing the sitemaps.

lolodomo:
Here is a jar for testing **with a OH 3.4 server**:
[org.openhab.ui.basic-3.4.1-SNAPSHOT.zip](https://github.com/openhab/openhab-webui/files/10366907/org.openhab.ui.basic-3.4.1-SNAPSHOT.zip)

wborn:
Maybe you can use one servlet filter instead of adding the logic to each servlet?

lolodomo:
> Maybe you can use one servlet filter instead of adding the logic to each servlet?

Can you please point an example (to help me understanding what you are talking about) ?

lolodomo:
You mean javax.servlet.Filter ?
This could be added to the parent class and so be global for the 3 servlets.

wborn:
I had a look at the code and it seems that the you can also create a Component that implements the `Handler` interface which will then be used for checking security constraints, see: [OpenHABHttpContext.handleSecurity](
https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.io.http/src/main/java/org/openhab/core/io/http/internal/OpenHABHttpContext.java#L54). 

You can have a look at the [AuthenticationHandler](https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthenticationHandler.java#L48) implementation.
It might also be an idea to make the [AuthenticationHandler.isSecured](https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthenticationHandler.java#L134-L141) check more extensible so the same handler can be used for securing add-on Servlets. :slightly_smiling_face: 

If there wasn't any infrastructure in place already there are some Http Whiteboard annotations that can be used for creating filters, see [Registering Servlet Filters](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.whiteboard.html#d0e121055). 

The `BaseServlet` will be removed with #1600 so I wouldn't add anything to that class.

lolodomo:
> I had a look at the code and it seems that the you can also create a Component that implements the `Handler` interface which will then be used for checking security constraints, see: [OpenHABHttpContext.handleSecurity](https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.io.http/src/main/java/org/openhab/core/io/http/internal/OpenHABHttpContext.java#L54).
> 
> You can have a look at the [AuthenticationHandler](https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthenticationHandler.java#L48) implementation. It might also be an idea to make the [AuthenticationHandler.isSecured](https://github.com/openhab/openhab-core/blob/8d64ecfd8db99844d79a776fef37cbe527c1f0c6/bundles/org.openhab.core.io.http.auth/src/main/java/org/openhab/core/io/http/auth/internal/AuthenticationHandler.java#L134-L141) check more extensible so the same handler can be used for securing add-on Servlets. 🙂

Looks like a good approach ;)
I will try it.

lolodomo:
I created a Handler service, I see that it is properly started. Unfortunately the "handle" method is not called when my servlet is invoked.

lolodomo:
Ok, I was able to make it work by setting a higher priority to my new handler.
I really believe there is a bug in `AuthenticationHandler` because it does not call `context.execute(request, response)` when no exception is thrown. This is necessary to trigger the next handler.

lolodomo:
@wborn : for my curiosity, can you explain in what case the AuthenticateHandler has a real usage ? Because in practice, at least with the default settings, this handler is doing nothing because the parameter org.openhab.auth:enabled is not set to true. To be verified but I believe this setting is even not defined at all and is apparently not exposed in MainUI. I have the strange feeling it is something not used by default.

By the way, I was finally able to block Basic UI just by implementing a specific `handler`. I will push my changes tomorrow.

I also checked what are the servlets in core framework. I identified 3:

- /audio
- /icon
- /chart

These servlets are by default not protected by any authentication. This is probably not critical but good to know. Something similar to what I did for the Basic UI servlets could certainly be applied.

lolodomo:
Here is the new version implementing a new HTTP request handler.
Note that the 2 new classes have to be moved to the core framework. I will prepare a PR.
So this is still in progress.

ghys:
> These servlets are by default not protected by any authentication. This is probably not critical but good to know. Something similar to what I did for the Basic UI servlets could certainly be applied.

Beware that requiring authorization for /icon might break the icons because the authorization header is not currently added to the image requests, in oh-icon widgets and related, like it is for API requests.

lolodomo:
> Beware that requiring authorization for /icon might break the icons because the authorization header is not currently added to the image requests, in oh-icon widgets and related, like it is for API requests.

I had clearly in mind that such change could impact other UI.

lolodomo:
As there is in parallel work to add authentication to Basic UI, I am stopping my work which will become useless.

Issue #912, opened 2021-02-16
Aggregate using diff_first or diff_last will leave a gap at first entry


crnjan:
## The problem

Having a HP that logs usage time - incrementally, number of hours. Used this values to display number of hours HP was producing heat for either DHW or underfloor heating:

```
   ...
    - component: oh-aggregate-series
      config:
        name: DHW
        gridIndex: 0
        xAxisIndex: 0
        yAxisIndex: 0
        type: bar
        item: gBasement_BoilerRoom_HP_DHW_InOperation_HP
        dimension1: date
        aggregationFunction: diff_first
        service: influxdb
  ...
```

result:

<img width="893" alt="Screenshot 2021-02-16 at 21 16 37" src="https://user-images.githubusercontent.com/3818211/108119978-dbd24180-70a0-11eb-9991-2bf5ef3c7e36.png">

All good, just first bar is missing. Looking at the code, this is expected.

## Your suggestion

We could extend the period and fetch one item "before" first period and use that one for reference to calculate first diff - but might be problematic to define what time interval to use to fetch that item?

I used a bit different approach and introduced a new aggregator `diff` that just takes first and last value from within the group itself.

```
    case 'diff':
      aggregate = arr[1][arr[1].length - 1] - arr[1][0]
```

result:

<img width="872" alt="Screenshot 2021-02-16 at 21 23 08" src="https://user-images.githubusercontent.com/3818211/108120577-c873a600-70a1-11eb-9799-2543a8d5233a.png">

I look forward to any feedback or comments - and if `diff` aggregator makes sense to others as well, I can prepare a PR ...

ghys:
I'm not against a "diff" aggregation, but it could mislead people into believing the interval between 2 values is the exact aggregation period selected when it's not, i.e. with a hourly period if would be e.g. 59 minutes if you have data points every minute, 55 minutes if you have data points every 5 minutes etc.
The Persistence API has a `boundary` option:
_("Gets one value before and after the requested period.)_
I think I ran into problems when using it. Might be worth another look though - perhaps with a mechanism to only including it when querying data for aggregate series (not time series) with a diff_* aggregator.

crnjan:
Did a quick spike and seems as `boundary` does return additional value, thx for the hint. Will try to use that value for diff_* aggregator when I find time to do so ..

rumexcrisp:
is there any news on this issue? i would be very interested in a solution to this problem

seimenb:
[...]

PR #1634, opened 2023-01-11
Fix diff_last aggregate calculation

crnjan:
In order to calculate `diff_last` include `boundary=true` parameter when requesting datapoints to read datapoint before requested interval (before datestart), if available. Use that one to calculate diff for first value.

fixes https://github.com/openhab/openhab-webui/issues/912 - partly, only applies for `diff_last`

Signed-off-by: Boris Krivonog boris.krivonog@inova.si

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/chart-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/chart-mixin.js
@@ -135,6 +135,7 @@ export default {
         return Promise.resolve(getter([]))
       }
 
+      let boundary = seriesComponents[component.component].includeBoundary?.(component)
       const itemPromises = neededItems.map((neededItem) => {
         if (this.items[neededItem]) return Promise.resolve(this.items[neededItem])
         return this.$oh.api.get(`/rest/items/${neededItem}`).then((item) => {
@@ -154,7 +155,8 @@ export default {
         let query = {
           serviceId: component.config.service || undefined,
           starttime: seriesStartTime.toISOString(),
-          endtime: seriesEndTime.subtract(1, 'millisecond').toISOString()
+          endtime: seriesEndTime.subtract(1, 'millisecond').toISOString(),
+          boundary: boundary
         }
 
         return Promise.all([itemPromises[neededItem], this.$oh.api.get(url, query)])
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/series/aggregators.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/series/aggregators.js
@@ -20,7 +20,11 @@ export default (aggregationFunction, arr, idx, values) => {
       aggregate = idx < 1 ? NaN : arr[1][0] - values[idx - 1][1][0]
       break
     case 'diff_last':
-      aggregate = idx < 1 ? NaN : arr[1][arr[1].length - 1] - values[idx - 1][1][values[idx - 1][1].length - 1]
+      if (idx < 1) {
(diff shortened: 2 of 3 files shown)


ghys:
When testing this PR with an existing chart I get (the server is 3.4.0 release build):

image

And the response is:
`
"{\"error\":{\"message\":\"Invalid start/end time in fetch request: 1675205999 \\u003e 1673910752\",\"http-code\":500,\"exception\":{\"class\":\"java.lang.IllegalArgumentException\",\"message\":\"Invalid start/end time in fetch request: 1675205999 \\u003e 1673910752\",\"localized-message\":\"Invalid start/end time in fetch request: 1675205999 \\u003e 1673910752\"}}}"
`
I haven't been immediately able to track where this error comes from... Could it be the persistence service (rrd4j in this case)?

Against main without the boundary parameter:

image

cweitkamp:
The REST resource does not set an `endDate` on the `FilterCriteria`to query the value after given end time.

https://github.com/openhab/openhab-core/blob/8f9dafd1d68a92248cd39414145175952d61cd46/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java#L346-L351

RRD4J persistence service uses current timestamp for `endDate` if not set. This probably is the culprit for the IAE thrown. We should place an issue in [openhab-addons](https://github.com/openhab/openhab-addons/issues) to address this.

https://github.com/openhab/openhab-addons/blob/80094b5e9d293dfb73c81377325da13172d6baaf/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java#L283-L284.

cweitkamp:
One temporarily solution for RRD4J would be to catch the thrown IAE and return an empty List.

https://github.com/openhab/openhab-addons/blob/80094b5e9d293dfb73c81377325da13172d6baaf/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java#L333

cweitkamp:
I created https://github.com/openhab/openhab-addons/pull/14238

cweitkamp:
@ghys  May I ask you an additional question? What does happen if you call the `PersistenceExtensions.minimumBetween()` or `PersistenceExtensions.maximumBetween()` methods from rules with parameters `begin` > `end`? Do you see the IAE too?

ghys:
@cweitkamp it seems so:
```java
logInfo('interval', FGWP102WallPlug_ElectricmeterkWh.minimumBetween(now, now.minusDays(30)).getState().toString())
```
```
00:30:38.894 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'test_rrd4j_interval' failed: Invalid start/end time in fetch request: 1674171038 > 1671579038
```

```java
logInfo('interval', FGWP102WallPlug_ElectricmeterkWh.minimumBetween(now.minusDays(30), now).getState().toString())
```
```
00:31:40.941 [INFO ] [rg.openhab.core.model.script.interval] - 525.25 kWh
```

ghys:
Closing as superseded by #1649. The discussion may continue here or be moved to #1649.

cweitkamp:
@ghys Thanks for testing. I will take care of that and add checks in OHC `PersistenceExtension`.

The fix for RRD4J has been merged. It takes care of both issues (start > end) and additional boundary value after end of requested time range.

ghys:
Excellent, then let's merge #1649.

m-jarzebowski:
Shouldn't the fix be backported to 3.4.x?

Issue #1418, opened 2022-06-20
Support for expression evaluation at chart axis level


splatch:
## The problem

Currently axis elements such as labels, colors and visibility are static. This is a diverge from series components which accept expressions at config level. In order to provide a unified handling both axis and series slots should permit expressions. This will allow making some axes conditional for wide charts, if ie. mobile screen is detected.

## Your suggestion
Expression evaluation is part of widget mixin, however most of chart elements have indirect access to evaluation through instance of chart widget. With some small changes in axis handling (probably in other chart related components) visibility of these elements can be a bit less constrained.

## Your environment
Example yaml as an idea

```yaml
  yAxis:
    - component: oh-value-axis
      config:
        min: 0
        gridIndex: 0
        name: Enery
    - component: oh-value-axis
      config:
        gridIndex: 0
        name: Coverage
        min: "0"
        max: "100"
        nameGap: 40
        nameRotate: "90"
        nameLocation: center
        show: =device.desktop # currently it doesn't work
```

Issue #1295, opened 2022-02-02
Widget: Item from props cannot be resolved for oh-aggregate-series


eric1905:
I want to create a bar chart which can be used in multiple pages. 
Therefore I use the oh-aggregate-series.
The item can be defined in the props section, but this item is not evaluated in the widget code, so currently it is only possible to create one widget per item and it is not possible to create only one widget and set the item via the props section:

`
props:
  parameters:
    - context: item
      description: Main toggle item (use for single toggle or as main switch for group)
      label: Item
      name: item
      required: false
      type: TEXT

...

                    series:
                      - component: oh-aggregate-series
                        config:
                          name: Stromverbrauch
                          gridIndex: 0
                          xAxisIndex: 0
                          yAxisIndex: 0
                          type: bar
                          item: =props.item
                          dimension1: hour
`

It is described in this thread: https://community.openhab.org/t/custom-widget-item-defined-in-property-not-shown/132707/7?u=eric1905

marianorenzi:
Same for charting switches:

```
    - component: oh-time-series
      config:
        name: Light
        gridIndex: 0
        xAxisIndex: 0
        yAxisIndex: 0
        type: line
        areaStyle: {}
      slots:
        markArea:
          - component: oh-mark-area
            config:
              name: Light
              item: =props.light_item
```

ReachyBayern:
Is there already a plan to fix it?

Issue #1102, opened 2021-06-12
Chart widget don't support expressions


Belgadon:
Inside an oh-mark-area object the items doesn't get called from props. It only works when its hardcoded by its name.

I made a minimal example. See the last line - the item is not selected.
If I hardcode the item by its name its working, but thats not very flexible.
I tried _=props.window_item_ and _=items[props.window_item]_.

```
uid: test
tags: []
props:
  parameters:
    - context: item
      description: Window Sensor Item shown in chart
      label: Window Item
      name: window_item
      required: false
      type: TEXT 
component: f7-card
config:
  title: test
slots:
  default:
    - component: oh-chart
      config:
        label: test
        periodVisible: false
      slots:
        grid:
          - component: oh-chart-grid
            config:
              show: true
              left: 50
              right: 50
        xAxis:
          - component: oh-time-axis
            config:
              gridIndex: 0
        yAxis:
          - component: oh-value-axis
            config:
              gridIndex: 0
        series:
          - component: oh-time-series
            config:
              name: Fenster
              gridIndex: 0
              xAxisIndex: 0
              yAxisIndex: 0
              type: bar
              color: red
            slots:
              markArea:
                - component: oh-mark-area
                  config:
                    name: Fenster
                    item: =props.window_item #=items[props.window_item]
```

arjan-io:
This is also true for the ```label``` item in
```
- component: oh-chart
      config:
        label: test
```
It does not support expressions.
It does work however in a ```oh-time-series``` like this:
```
- component: oh-time-series
      config:
        areaStyle:
          opacity: 0.2
        gridIndex: 0
        item: =props.hum
        name: Luchtovchtigheid
        service: rrd4j
        smooth: true
        type: line
        xAxisIndex: 0
        yAxisIndex: 1
```
(Different chart example)

canufarm:
> Inside an oh-mark-area object the items doesn't get called from props. It only works when its hardcoded by its name.

+1 from me.

Expressions also do not work with "period:". Most of the echart-settings are not supported by the chart widget. Would be great if all echart-settings could be used with expressions, because otherwise one has to build separate widget for each item, although the widgets are exact copies.

Thanks.

crnjan:
Which version are you using? This should work with latest snapshots (and 3.4.0 Milestone 4), i.e. a slightly modified widget:

```
uid: test
tags: []
props:
  parameters:
    - context: item
      description: Window Sensor Item shown in chart
      label: Window Item
[...]

Issue #1072, opened 2021-05-30
Area chart to support shading CLOSED instead of OPEN


ssalonen:
## The problem

I use `Contact` items to measure relay status. `CLOSED` status is used to denote situation when relay is closed, i.e. relay is passing power, `OPEN` is used to denote situation when there is no power.

Area chart seems to highlight time periods when item status is `OPEN` -- this is inverse to what I would like to have.

<img width="302" alt="image" src="https://user-images.githubusercontent.com/315866/120093797-4aee1600-c125-11eb-8f9e-8d43ed6bb531.png">

```yaml
    - component: oh-time-series
      config:
        name: OH patterit
        gridIndex: 0
        xAxisIndex: 0
        yAxisIndex: 0
        type: line
        areaStyle: {}
      slots:
        markArea:
          - component: oh-mark-area
            config:
              name: LivingRoomRadiatorPower
              item: LivingRoomRadiatorPower
```

## Your suggestion

Allow 'inverting' area chart colors, i.e. highlighting CLOSED instead of OPEN.

Can I use `areaStyle` configuration property to override colors?

## Your environment

openHAB 3.0.2

```yaml

```

crnjan:
A new property was introduced `states`:

```
markArea:
          - component: oh-mark-area
            config:
              name: LivingRoomRadiatorPower
              item: LivingRoomRadiatorPower
              states: CLOSED
```

that should do the trick. Also more can be specified

```
markArea:
          - component: oh-mark-area
            config:
              name: LivingRoomRadiatorPower
              item: LivingRoomRadiatorPower
              states: 
                 - CLOSED
                 - OFF
```

PR #1638, opened 2023-01-13
charts - generalize evaluating expressions for charts

crnjan:
This is based on https://github.com/openhab/openhab-webui/pull/1419. Also contains fixes from https://github.com/openhab/openhab-webui/pull/1634.

fixes https://github.com/openhab/openhab-webui/issues/1418
fixes https://github.com/openhab/openhab-webui/issues/1295
fixes https://github.com/openhab/openhab-webui/issues/1102
fixes https://github.com/openhab/openhab-webui/issues/1072

It's still WIP, need to perform more tests, any help greatly appreciated 

EDIT: went through cases reported in above issues and seems to work now. Also, as far as I tested (so far) locally with my setup seems to work - but probably don't have all use-cases setup so probably missed something - therefore would be much appreciated if any volunteer tries it out too, thank you!

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
@@ -12,9 +12,11 @@ export default {
     let rollingState = null
     let currentArea = null
     itemPoints.forEach((p) => {
-      if (onStates.includes(p.state) && !onStates.includes(rollingState)) {
+      // eslint-disable-next-line eqeqeq
+      if (onStates.findIndex((s) => s == p.state) >= 0 && onStates.findIndex((s) => s == rollingState) < 0) {
         currentArea = [{ name: markArea.name, xAxis: new Date(p.time) }]
-      } else if ((!onStates.includes(p.state) && onStates.includes(rollingState)) && currentArea) {
+      // eslint-disable-next-line eqeqeq
+      } else if ((onStates.findIndex((s) => s == p.state) < 0 && onStates.findIndex((s) => s == rollingState) >= 0) && currentArea) {
         currentArea.push({ xAxis: new Date(p.time) })
         currentArea[0].name = markArea.name || markArea.item
         currentArea[0].value = rollingState
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
@@ -6,7 +6,7 @@ dayjs.extend(LocalizedFormat)
 export default {
   get (component, points, startTime, endTime, chart) {
     let markArea = chart.evaluateExpression(ComponentId.get(component), component.config)
-    const onStates = markArea.onStates || ["ON", "OPEN"]
+    const onStates = markArea.onStates || ['ON', 'OPEN']
     const itemPoints = points.find(p => p.name === markArea.item).data
     markArea.data = []
     let rollingState = null
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
@@ -6,17 +6,17 @@ dayjs.extend(LocalizedFormat)
 export default {
   get (component, points, startTime, endTime, chart) {
     let markArea = chart.evaluateExpression(ComponentId.get(component), component.config)
-    const onStates = (markArea.onStates && !Array.isArray(markArea.onStates) ? [markArea.onStates] : markArea.onStates) || ['ON', 'OPEN']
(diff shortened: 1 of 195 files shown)


ghys:
Thanks! Generalizing expressions in chart widgets is a plus for sure! It's looking good code wise but I'll have a test run soon.

crnjan:
Created a test jar, please run:
```
bundle:update org.openhab.ui https://github.com/crnjan/openhab-webui/raw/feature/diff_poc/bundles/org.openhab.ui/target/org.openhab.ui-4.0.0-SNAPSHOT_eval.jar
```
in order to try it out. If there is any case not working, please let me know - and provide a (minimal) working sample (yaml) so I can try locally.

If needed, revert back to official UI jar with

```
bundle:update org.openhab.ui https://ci.openhab.org/job/openHAB-WebUI/lastSuccessfulBuild/artifact/bundles/org.openhab.ui/target/org.openhab.ui-4.0.0-SNAPSHOT.jar
```

crnjan:
Accidentally pushed (potential) fix for https://github.com/openhab/openhab-webui/issues/1072, introducing `onStates`

```
        markArea:
          - component: oh-mark-area
            config:
              item: gBasement_BoilerRoom_HP_Compressor_Powered
              onStates: OFF 
```

I can revert if needed ...

ghys:
> I can revert if needed ... also feel free to suggest a better name ...

It's sort-of related in a general sense (allow customizing charts beyond some arbitrary/common sense defaults, for those who need this customization), so no need to open a separate PR. The name `onStates` is fine, maybe `markStates` alternatively or simply `on` or `states`. Not sure which I like better? Since oh-mark-area isn't currently described and won't appear in config sheets or https://next.openhab.org/docs/ui/components/#chart-page-components (probably something to remedy one day) it's kind of obscure anyway :)

By the way it accidentally works when you feed that parameter a string instead of an array (because `includes` is both in the prototype of Array and String) but it might not work 100% as the user expects (for instance `onStates: ONOFF` will match both `ON` and `OFF`, also `onStates: ="ABC123"` will match `AB`, `BC`, `C12`... but `onStates: 1` won't match `1`, if needs to be `onStates: [1]`). Maybe transforming it into a single-element array if it's not an array would be safest.

NB: I tried this with a scene Number item but noticed it's actually a string in the API response in this format in my case:

image

Since `"4.0" == 4` in JS maybe it would be valuable to use the non-strict equality somehow to check if there's a match?

bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
```diff
@@ -1,20 +1,22 @@
...
 import LocalizedFormat from 'dayjs/plugin/localizedFormat'
+import ComponentId from '../../component-id'
 dayjs.extend(LocalizedFormat)
 
 export default {
   get (component, points, startTime, endTime, chart) {
-    const itemPoints = points.find(p => p.name === component.config.item).data
-    let markArea = Object.assign({}, component.config)
+    let markArea = chart.evaluateExpression(ComponentId.get(component), component.config)
+    const onStates = markArea.onStates || ['ON', 'OPEN']
```
ghys, line 9:
```suggestion
    const onStates = (markArea.onStates && !Array.isArray(markArea.onStates) ? [markArea.onStates] : markArea.onStates) || ['ON', 'OPEN']
```

bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-mark-area.js
```diff
@@ -1,20 +1,22 @@
...
     markArea.data = []
     let rollingState = null
     let currentArea = null
     itemPoints.forEach((p) => {
-      if ((p.state === 'ON' && rollingState !== 'ON') || (p.state === 'OPEN' && rollingState !== 'OPEN')) {
-        currentArea = [{ name: component.config.name, xAxis: new Date(p.time) }]
-      } else if (((p.state !== 'ON' && rollingState === 'ON') || (p.state !== 'OPEN' && rollingState === 'OPEN')) && currentArea) {
+      if (onStates.includes(p.state) && !onStates.includes(rollingState)) {
+        currentArea = [{ name: markArea.name, xAxis: new Date(p.time) }]
+      } else if ((!onStates.includes(p.state) && onStates.includes(rollingState)) && currentArea) {
```
ghys, line 17:
```suggestion
      // eslint-disable-next-line eqeqeq
      if (onStates.findIndex((s) => s == p.state) >= 0 && onStates.findIndex((s) => s == rollingState) < 0) {
        currentArea = [{ name: markArea.name, xAxis: new Date(p.time) }]
      // eslint-disable-next-line eqeqeq
      } else if ((onStates.findIndex((s) => s == p.state) < 0 && onStates.findIndex((s) => s == rollingState) >= 0) && currentArea) {
```

ghys (commented):
Suggestions for the above comment.

The rest of the PR LGTM!

PR #1645, opened 2023-01-15
add Command-Click for MacOs to select list entries

stefan-hoehn:
Signed-off-by: Stefan Höhn <mail@stefanhoehn.com>

I only learned today from @JustinGeorgi that on Windows and Linux Ctrl-Click allows to directly entries of a list (usually to remove entries of that list). However, Ctrl-Click isn't supported on MacOS which is a pity now that I know that feature. Therefore I added Cmd-Click instead to allow the selection on MacOS.

Note that I used "click.meta" which is the Command-Key on MacOS. This would be also the Windows-Key on Windows which I think shouldn't be an issue (it is hardly used and only for very special case IMHO) but I wanted to mention it in case someone has doubts.


ghys (approved):
Cool, thanks!

PR #1647, opened 2023-01-16
allow customisation of temperature, humidity and luminance badges too

crnjan:
On `oh-locations-tab` it's not possible to override `temperature`, `humidity` and `luminance` since they are handled a bit differently, as per https://community.openhab.org/t/not-all-custom-badges-are-working-what-am-i-doing-wrong-temperature-and-humidity-wont-change/143434.

Signed-off-by: Boris Krivonog [boris.krivonog@inova.si](mailto:boris.krivonog@inova.si)


ghys (approved):
Thanks

PR #1468, opened 2022-08-18
Adapt to core changes (addon.xml)

J-N-K:
Depends on #1452 
Depends on https://github.com/openhab/openhab-core/pull/3050

PR #1651, opened 2023-01-18
Fix thing-type retrieval and add-on management

J-N-K:
Related to #1468
Depends on https://github.com/openhab/openhab-core/pull/3293

- The bindingId was not properly determined for thing-type retrieval for bindings added by non-distrubtion addon services.
-Add-ons from community marketplace and JSON 3rd party add-on-service could not be installed


kaikreuzer (approved):

ghys (approved):
Okay, finally have a test 4.0 instance running to test.

LGTM, but there are still some regressions that would have to be handled:

1. installing add-ons on a fresh instance using the setup wizard is still broken
2. "featured" add-ons (i.e. those arbitrary selected to appearing as cards on top of store sections) need to be adjusted
3. installed add-ons from the distribution have (short) descriptions now so it's what appears when going to their page (before pressing "more").

These are relatively minor so still merging as it fixes the general brokenness.

PR #1662, opened 2023-01-26
Add a Scene settings menu entry and editor

### LINKED ISSUE #1528: Add support for Scenes in OpenHab
Following up on the debate on [this thread](https://community.openhab.org/t/scenes-openhab-vs-home-assistant-in-2022/138782)

# Scenes

## Definition

A scene is a preset that will change all items that are effected, so if you set a scene in a room, all lights in that room will be changed to the settings of that scene.

## Scenes in Hue

Scenes is the main user flow in the Hue interface. Other features such as triggers, buttons and other integrations are secondary in the user interface.
image

## Scenes in Home Assistant

Since version 0.15, Home Assistant has a support for scenes. See [this link](https://www.home-assistant.io/integrations/scene/) for more details. A scene in HA has a name, icon and a series of entities. Each entity in each scene has any number parameter values.

image

HA also supports transitions for scenes to gradually transition the values when possible. HA enables to create snapshots of multiple entities into a scene using their scene service.

To clarify some of the debate in the thread above, the HA concept and features [don't seem to be debated by users in their github](https://github.com/home-assistant/core/issues?q=label%3A%22integration%3A+scene%22+).

# Implementation

## Rules

A series of addon Widgets provide multiple features to support scenes. See [this](https://community.openhab.org/t/scene-control-suite-scene-control-widget/128259/2)

*Pros:* 
- No change to the core OH are required. 
- Each scene is captured in this model in a separate rule.

*Cons:* 
- Not sure editing scenes is supported. 
- Also going to assume that if the rules are manually edited, it could easily break any built-in editor.
- The number of scenes would generate a large number of rules which would need to be managed somehow (tags? new UI sections?)

## Core OH Addition

If OH did follow the HA model  - scenes can be added in the core model.
Scenes could be applied at the group level and provide a way for a group to switch between multiple scenes.
`livingRoomLights.SetScene("Morning")`

*Pros:* 
- Adding scenes to the core could support better UX for adding and editing scenes
- The scene data would have a separate conceptual space and not clog the rule section
- Applying scenes to group would make it easy to integrate them with rules

*Cons:* 
- Changes in core
- Need for new UI for capture and editing of scenes

# Other considerations

One additional consideration for implementing rules would be to integrate with existing scene systems. Almost every light system already have their own built-in scene support, it would be nice if the API could leverage them transparently: `livingRoomLights.SetScene("Morning")` would call scene "Morning" inside my Hue lights

ghys:
This picks up #1520 where it was left off, and (finally) adds the features that I wanted and described in https://github.com/openhab/openhab-webui/pull/1520#issuecomment-1369892928.

Kudos to @J-N-K for doing the groundwork.

Closes #1528.


stefan-hoehn:
I am pretty sure  that this is something that quite a number of people are looking for. Hence, I am wondering if it wouldn't make sense to write some documentation on it how to use it. From the code and the documentation and the forum entry it is not so clear to me what the final solution looks like.
I am willing to help here if someone of you gave me guidance how to actually use it and try it out (or where the entry point in the UI is and how it is supposed to work) and where it would make sense in the openhab docs to write something about it (even maybe with screenshots or screen recording). Let me know if I can help here.

florian-h05:
@ghys Please let me know when I should review this.

ghys:
Sure, this is ready for reviewing.

As for the docs, by all means. It's basically just a simplified rule editor, found in the Automation section of the settings (alongside Rules, Scripts and Schedule), which abstracts the "send command to" modules, asking just for a set of items and their target state. You also have buttons to update the target state from/to the current state, this is helpful when designing.

bundles/org.openhab.ui/web/src/components/config/editor/hint-rules.js
```diff
@@ -144,9 +149,18 @@ function buildModuleStructure (cm, moduleType) {
   return ret
 }
 
+function hintSceneItems (cm, line, parentNr) {
```
florian-h05, line 152: What I have noted during testing this, is that code hinting does not open up if you add a new line to the end of the section. I know how to fix this, but let's discuss this in another PR since this may have some side effects.

I also wondered if we could provide the user with some information how to "force" hinting (Ctrl + Space).

bundles/org.openhab.ui/web/src/pages/settings/rules/scene/scene-edit.vue
```diff
@@ -0,0 +1,613 @@
...
+          this.$set(module.configuration, 'command', state)
+        })
+      })
+    },
+    updateCommandFromCurrentState (ev, module) {
+      if (ev) ev.cancelBubble = true
+      const itemName = module.configuration.itemName
+      this.$oh.api.getPlain('/rest/items/' + itemName + '/state').then((state) => {
+        this.$set(module.configuration, 'command', state)
+        // this.$f7.toast.create({
```
florian-h05, line 536: I agree that a toast notification is not required for updating from current state. Not sure if this out-commented code should be removed.

bundles/org.openhab.ui/web/src/pages/settings/rules/scene/scene-edit.vue
```diff
@@ -0,0 +1,613 @@
...
+        //   closeTimeout: 2000
+        // }).open()
+      })
+    },
+    testCommand (ev, module) {
+      if (ev) ev.cancelBubble = true
+      const itemName = module.configuration.itemName
+      const command = module.configuration.command
+      this.$oh.api.postPlain('/rest/items/' + itemName, command, 'text/plain', 'text/plain').then((state) => {
+        // this.$f7.toast.create({
```
florian-h05, line 548: IMO a toast notification on commands send is always useful, so I'd vote for displaying a toast notification here.
I like to get some feedback in the UI when I interact with physical equipment, even though I am usually besides my equipment when creating scenes.

florian-h05 (approved):
LGTM, huge thanks for this!!

I have tested this both on PC and mobile phone, works fine. I love the code view as well, it allows super fast definition of Item and command.

Just two minor notes and one question regarding toast notifications.

bundles/org.openhab.ui/web/src/components/config/editor/hint-rules.js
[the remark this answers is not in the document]
ghys: > We have to add Enter to the extraKeys of CodeMirror in script-editor.vue

Yes, agreed. That could be valuable, if there are no immediate drawbacks. This can be handled in another PR.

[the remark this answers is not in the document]
florian-h05: Great. I think I can take care of that and try to fix the problem that page/widget editor is reaaally slow in iPadOS in this PR as well.

bundles/org.openhab.ui/web/src/pages/settings/rules/scene/scene-edit.vue
ghys: You're right, I was on the fence on these toasts and finally decided against, but the one for commands still makes sense. I have removed the commented code for the "current state to module" method and re-enabled this one.

florian-h05 (approved):
LGTM, thanks!

Let‘s wait for the checks to complete and merge then 👍

stefan-hoehn:
I was finally able to test it on the latest snapshot and thanks for the great contribution by all of you (in particular @J-N-K who started it) ❤️.  

I will start writing some documentation soon, though for those who want to get a glimpse what it looks like and how it can be found, here are two screenshots:
image
image
Example using it in a floorplan marker:
image


**During trying it out, I think I found two little glitches/bugs:**

1) Doesn't run without save
- create a scene
- add one (or more) item with its setting (preferable to something that is different from now)
- save it
- the you'll get this bar 
- run the rule, it works
- now change a setting and run the rule via the |> icon - it doesn't work
- save it via cmd/ctrl-s and run the rule - it works

Surely the reason is that running in the scene editor doesn't automatically save it before like in the rules section. Does it make sense to auto-save the scene when running it?

2) Ctrl/Cmd-R isn't working
- even though the tooltip on the run |> icon says "Run now (Ctrl-R)", it doesn't. Instead the window is reloaded (tested on MacOS)

Do like me to create an issue for that?

stefan-hoehn:
After some conversation between Florian and me: saving happens automatically during run when you are in the "script" editor for example of blockly, js, dsl rules but Scenes are one screen "before" which is one step earlier which explains why it behaves not the way I expected it.

From a UX perspective I would still recommend that running would save before, what you think?

> Ctrl/Cmd-R isn't working

Yep, I am happy to put that on my list 👍

stefan-hoehn:
I found another small glitch (@ghys) which requires to click twice on "Select Items" before the popup opens. Here's how to reproduce.

- click on any menu item first
- then click  "scenes" menu item
- choose one of the scenes
- then directly click on "Select Items", the popup doesn't open, click again it opens.
Note that clicking on image works right away

stefan-hoehn:
@ghys While I am documenting it:

In the code it does have triggers and conditions. Am I correct that even though these properties are there and could be edited in YAML they are not supported?

e.g.

```
triggers:
  - id: "2"
    configuration:
      cronExpression: 0 * * * * ? *
    type: timer.GenericCronTrigger
conditions: []
```

I did try it out and had no effect.

stefan-hoehn:
The  PR for the tutorial can be found here: https://github.com/openhab/openhab-docs/pull/2009

ghys:
> In the code it does have triggers and conditions. Am I correct that even though these properties are there and could be edited in YAML they are not supported?
> I did try it out and had no effect.

Hm, no, triggers & conditions should work, if they appear and are saved properly; it's for addressing the concern about having scenes put in sitemaps as a selection; can't say I haven't tested much though.

florian-h05:
> I found another small glitch (@ghys) which requires to click twice on "Select Items" before the popup opens. Here's how to reproduce.

I cannot reproduce this (iPadOS 16.3).

> Hm, no, triggers & conditions should work, if they appear and are saved properly

I have just tried this, you can add them and click save, but they aren‘t saved.

ghys:
> I cannot reproduce this (iPadOS 16.3).

Me neither, on different desktop and mobile browsers (I would have noticed otherwise I think).

> I have just tried this, you can add them and click save, but they aren‘t saved.

Indeed. Not sure at first sight why that happens, but it seems there were some more things that were removed with the duplication of the rules screens to add the scenes feature:
- the status of the rule/scene isn't updated in real time in the scene editor (there isn't code to monitor the status through SSE)
- the list of scenes don't display the rules/scenes statuses.

I noticed part of these before, and I think this was intentional, but not all.

florian-h05:
> the status of the rule/scene isn't updated in real time in the scene editor (there isn't code to monitor the status through SSE)

Status of scene IMO doesn't make sense to display, and I'd also propose to remove the disable/enable button.

florian-h05:
PR #1739 fixes the problem with triggers and conditions not working and removes the not-correctly working enable/disable actions as well as the rule status.

PR #1664, opened 2023-01-27
Blockly metadata blocks

stefan-hoehn:
Here are new **metadata blocks** that provide a very easy way of accessing custom namespace metadata information both for reading and writing. Even accessing config metadata on hierarchical levels as well as variable support that allows iteration over computed names is supported (see the examples below). 

Note: Metadata blocks are not available for Nashorn but only for graalvm / jsscripiting

I have used these blocks already myself for a number of rules for pretty complex usecases (rollershutter scheduling, window open notification with increased notification times...) and it works really well. Easy usage of Metadata configuration opens up a huge number of nice applications!

Here is a simple metadata definition rollerschutter_schedule
````
value: active
config:
  opening: 23:00
  closing: 22:00
````
with the blocks that access it:

image

I intentionally kept the value blocks from the config blocks. The value block is the main one value of the namespace which is also directly shown in the ui list of an item's metadata. The intent is something like "ON/OFF" or "enabled/disabled" while the config part holds complex configuration of that metadata. 

Here is a definition of a more complex metadata namespace:

```
value: ON
config:
  ooone: "123"
  oone:
    two: "456"
  one:
    two:
      three: "756"
  field1: 1
  field2: 2
  field3: 3
```

which allows deep hierarchy property access via configkey-dot-notation like one.two.three

image

Using variables and loops properties like field1, field2, field3 can even be accessed very easily:

image

~~**item name resilience**~~

~~From 4.0 on, new blocks that require items as names will be resilient in blockly in way that either the item name or the item object itself can be provided which reduces a lot of confusion. The meta blocks are the first that allow this (originally that was part of the blockly code but was then moved into the jsscripting library by).~~

~~All metadata blocks can deal with receiving either the name of the item or the item itself:~~

~~The real beauty is leveraged when dealing with blocks that return a collection of items via tags or from members.~~

~~However, this approach will be introduced even further _for both_ Nashorn and graalvm/jsscripting to *all blocks* that currently only can handle the name as an input which reduces a lot of confusion for the user of the blocks.~~

~~The first block that is also able to receive either the item name or item object is the __oh_event_ block_ (send command, post update) as part of this PR. Blockly automatically detects what is provided and converts the input according to the required input for the openHAB API. A follow up PR will provide this to all other applicable blocks.~~

---

Here are a number of test cases that exemplify how the blocks can be used:

image

image

image

### CODE REVIEW

File: bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-metadata.js [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+      this.setHelpUrl('https://www.openhab.org/docs/configuration/blockly/rules-blockly-metadata.html#get_meta_value')
+    }
+  }
+
+  javascriptGenerator['oh_get_meta_value'] = function (block) {
+    const theItem = javascriptGenerator.valueToCode(block, 'theItem', javascriptGenerator.ORDER_ATOMIC)
+    const namespace = javascriptGenerator.valueToCode(block, 'namespace', javascriptGenerator.ORDER_ATOMIC)
+
+    if (isGraalJs) {
+      return [`(items.metadata.getMetadata(${theItem}, ${namespace}) !== null) ? (items.metadata.getMetadata(${theItem}, ${namespace}).value) : 'undefined'`, javascriptGenerator.ORDER_CONDITIONAL]
```
florian-h05, line 38: Can you please explain why you `null` check here and set `undefined` as a string?

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-eventbus.js
```diff
@@ -29,9 +35,11 @@ export default function (f7, isGraalJs) {
     const itemName = javascriptGenerator.valueToCode(block, 'itemName', javascriptGenerator.ORDER_ATOMIC)
     const value = javascriptGenerator.valueToCode(block, 'value', javascriptGenerator.ORDER_ATOMIC)
     if (isGraalJs) {
-      return `items.getItem(${itemName}).${eventType}(${value});\n`
+      addItemObject(isGraalJs)
+      return `_itemObject(${itemName}).${eventType}(${value});\n`
```
florian-h05, line 39: In which case do you pass an Item object to the send command/post update block?

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
```diff
@@ -201,3 +201,42 @@ export function addDateComparisonSupportGraalVM () {
     ])
   return graalZdtCompare
 }
+
+// always return the item name no matter whether the name is given already or the item object
+export function addItemName (isGraalJs) {
+  let itemNameAccess = (isGraalJs) ? 'name' : 'getName()'
```
florian-h05, line 207: For GraalJS, I'd prefer that this method gets never used. Instead of using this method, please let me know where you need to be able to pass in both the Item and the Item name and I'll extend the openhab-js library to make that possible.

florian-h05 (approved):
LGTM, thanks!
This is a great addition for those who use Blockly ;-)

I have two questions and one comment, would be nice if you could answer them.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-eventbus.js
stefan-hoehn: The problem is that the naming of the blocks, which we cannot change, is kind of confusing, so it isn't necessarily clear whether you should apply the item-block, which is basically only a name or the while getItem-Block, which returns the whole item. This approach allows to use either one. This has often been an issue in forums, so by this approach the blocks that take an item become more forgiving for beginners.

 Also, there would be the usecase to get all members of an item and then iterate over them and send a particular command to all of them.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/utils.js
stefan-hoehn: Yes, I appreciate that: go ahead and do so. I will then tell you for the upcoming PR for all places where this would be applicable ahead of time, so we can jointly optimize that.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-metadata.js
stefan-hoehn: I had situations where the code would crash because the namespace didn't exist (or I misspelled it), so returning an undefined allows me to return the value undefined in case the namespace doesn't exist.

ghys:
>###  item name resilience
>From 4.0 on, new blocks that require items as names will be resilient in blockly in way that either the item name or the item object itself can be provided which reduces a lot of confusion. The meta blocks are the first that allow this (originally that was part of the blockly code but was then moved into the jsscripting library by @florian-h05 ).
>All metadata blocks can deal with receiving either the name of the item or the item itself:
>The real beauty is leveraged when dealing with blocks that return a collection of items via tags or from members.
>However, this approach will be introduced even further for both Nashorn and graalvm/jsscripting to all blocks that currently only can handle the name as an input which reduces a lot of confusion for the user of the blocks.
> The first block that is also able to receive either the item name or item object is the oh_event block (send command, post update) as part of this PR. Blockly automatically detects what is provided and converts the input according to the required input for the openHAB API. A follow up PR will provide this to all other applicable blocks.

Okay, I hate to be the eternal party pooper/contrarian and to be honest, and I'm quite tired of these endless, pointless debates so just know I won't fight this (I'll leave the next maintainer to their own pet peeves), and this PR will eventually get merged, but as you know I have expressed my opinion against this already, and I'm entitled to it, so I will reiterate.

To be blunt I don't see the "real beauty" of it at all. In fact I find it very inelegant. It feels like a patch to tackle a self-inflicted problem we can solve.
I'm very unimpressed that this means, for instance, a simple "send X command to item Y", one of the two basic blocks I made:

image

which is currently a nice one-liner which resembles the actual, simple block:
```js
 // GraalVM
items.getItem('MyItem').sendCommand('value');

 // Nashorn
events.sendCommand('MyItem', 'value');
```
would now be an ugly monstrosity like this:
```js
// GraalVM
function _itemObject(item) {
  return (typeof item !== "string") ? item : items.getItem(item);
}


_itemObject('MyItem').sendCommand('value');

// Nashorn
function _itemName(item) {
  return (typeof item === "string") ? item : item.getName();
}


events.sendCommand(_itemName('MyItem'), 'value');

```

So all my scripts would now have this foolproof function on the off chance that someone, somewhere might feed the block the wrong type? It's almost insulting, even. I care about the code Blockly generates for me, I know what I'm doing, I never asked for this and don't want it.

## Counter-proposal:

In retrospect (and I know, hindsight is 100%) the fundamental mistake IMO was to introduce blocks with the item object as an output. If you look closely, all blocks output basic types (I'll consider QuantityTypes and ZonedDateTime as basic types), and the only block which currently accepts an item object is the "get attribute of item" block.
So I would propose this: https://github.com/openhab/openhab-webui/commit/0782e415f7e1141985bacbcc4ee45c7e095f12dc

image

- the "get item" block disappears
- the "get members of group" and "get items with tag" blocks returns arrays of **Strings** (the names of the items)
- the "get attribute from item" accepts the item name (a String) and does a lookup in the registry, the "name" and "state" dropdown options are removed as they become unnecessary. (or in fact, "state" might stay and make the "get state of item" block redundant.

Yes this means more lookups to the registry, but those are pretty cheap. And I don't mind the added mapping if it means the user doesn't have to care about what's in the variable - it will always be a string, not an item object they might not have a clue about anyway if they are beginners. The green color of these "item" shadow blocks gives the clue that we're only manipulating text.

image

```js
// GraalVM
var i;

var i_list = items.getItem('MyItem').members.map((item) => item.name);
for (var i_index in i_list) {
  i = i_list[i_index];
  console.log(items.getItem(i).label);
  console.log(items.getItem(i).type);
  console.log(items.getItem(i).state);
}

// Nashorn
var i;

var i_list = Java.from(itemRegistry.getItem('MyItem').members).map(function (item) { return item.name; }));
for (var i_index in i_list) {
  i = i_list[i_index];
  print(itemRegistry.getItem(i).getLabel());
  print(itemRegistry.getItem(i).getType());
  print(itemRegistry.getItem(i).getState());
}

```

I believe it would restore some sanity, instead of adding more insanity to the insanity.
And I'm aware that it would break some scripts, but as we're on the road to a major version, we have an unique opportunity to break things.

stefan-hoehn:
After thinking the last two hours what I would answer about that, I keep it short: I know how important code generation for you is and I put a lot of thought into it to make it as elegant as possible before creating the PR. I will remove it from the PR and we keep it as it is.

stefan-hoehn:
"item name resilience" has been removed.

rkoshak:
What if the complex part were handled in the openhab-js library itself? If the Item name/Item Object stuff is handled there transparently for both to Blockly and to code based users it feels like it's a net benefit over all.

I can't speak for the library but it seems reasonable to me to make additions to it that make the code produced by Blockly cleaner and clearer, not just for something like this but in other areas too. 

I just want to throw that out in case it's not been considered. Blockly does not need to stand on its own and work around quirks without help. Personally, I would have expected that function to have been a part of the openhab-js library in the first place. Blockly should need to generate it.

That doesn't help Nashorn though. :-(

florian-h05:
> What if the complex part were handled in the openhab-js library itself? If the Item name/Item Object stuff is handled there transparently for both to Blockly and to code based users it feels like it's a net benefit over all.

The library tries to handle both the Item name and the Item object itself where possible. This is not supported everywhere in the library (yet), but I have already introduced it in parts, e.g. the metadata API which is especially useful for Blockly.

The problem in the case of "send command / post update"  is that the code for this method is different from most generated code in GraalJS:

```javascript
// send command or post update:

// with Item object:
var itemObj = items.getItem(...)
itemObj.sendCommand('ON')

// with Item name:
var itemName = 'my_item'
items.getItem(itemName).sendCommand('ON')

// metadata API for comparison:

items.metadata.getMetadata(itemOrName, namespace)
```

As you can see, in GraalJS sendCommand / postUpdate are instance methods on the Item object, therefore we need that object. The metadata method is a static method that takes Item or Item name as parameter.

> Personally, I would have expected that function to have been a part of the openhab-js library in the first place. Blockly shouldn't need to generate it.

IMO the library should make the life as easy as possible for Blockly, but it is no place for such helper functions since they are only Blockly related. When manually writing code, you know whether you have a Item object it Item name (string) when writing your code.

BUT: I have some good news. I was able to make Blockly generate different code based on the input type. Although Blockly doesn't provide direct access to the input type, you can determine it by comparing some type checks. Note that I have taken the logic for this from the Blockly source code itself.

florian-h05:
@ghys Regarding your feedback:

Whilst I agree that the current way with having both` oh_item` (Item name as string) and `oh_itemtype` is not optional, @stefan-hoehn explained me why we cannot introduce breaking changes to the Blockly blocks.

With user-written scripts/code, I would be perfectly fine with breaking changes, since it is a simple adjustment of the code that the user needs to do. 

But with Blockly this is completely different: You cannot change or remove a block because if you do, this will likely avoid the Blockly editor from opening at all. You cannot adjust a Blockly to some breaking changes when you cannot open this Blockly anymore. And completely recreating it is something we MUST avoid, since this isa huge amount of work for some scripts, and again: you cannot open your original Blockly script, so you have no chance to open your old Blockly scripts.

So we have no other chance than working somehow around that problem, but I also don't like helper/utility functions in the generated code, so I was wondering if there is a better solution - there is one:

My proposal is https://github.com/openhab/openhab-webui/pull/1664/commits/592f9e9f08d2c364b5a3026ea432b5db351939a2: This enables Item name resilience without requiring any helper function. With this change, Blockly checks the type of the input during code generation and generated customized code depending on the type of the input. See my post above.

rkoshak:
> As you can see, in GraalJS sendCommand / postUpdate are instance methods on the Item object, therefore we need that object.

I always suspected this is why there was a separate `sendCommand/postUpdate` action in Rules DSL (or maybe that was just a side effect of how it's injected into the Item Object). Would it make sense to add these to `items`?

```
items.sendCommand(itemorname, state);
```

The Item object would retain the methods and continue to work as always.

Yes, I'm aware I could use `actions.BusEvent.sendCommand(itemname, statestr)` but that's no shorter and even harder to read and understand than pulling the Item from the registry using `items`.

> but it is no place for such helper functions since they are only Blockly related. When manually writing code, you know whether you have a Item object it Item name (string) when writing your code.

True but I find myself bothered by how long some of the most common lines in all my rules have to be because I have to pull the Item from the registry to update it or command it. I find myself missing even Rules DSL and looking to jRuby with envy in this case. The number one thing we do in rules is interact with Items. Making that more flexible could only be an improvement IMO.  So I don't think this would only benefit Blockly alone and I meant to emphasize that in my reply above. I doubt many helpers made to support Blockly would only benefit Blockly users.

And we are now pretty solidly off topic for this thread. I'll open an issue on the library for further discussion.

florian-h05:
@ghys 
Even though https://github.com/openhab/openhab-js/pull/233 has been merged and therefore static Item lookup by name on the `items` namespace is possible, I would prefer to get this merged first. 

I will then open a follow up PR to update for https://github.com/openhab/openhab-js/pull/233 and remove the prepended code that is only required for the PercentType volume in Audio and Voice actions. I have a core PR open to remove this "dependency" on PercentType.

florian-h05 (approved):
LGTM, thanks!

I will leave merging to @ghys, so he can have a final look.

florian-h05 (approved):
Latest changes LGTM as well. Whilst they are not directly related to this PR, I have no problem with having this small enhancement here.

ghys (approved):
Thanks for the work!

PR #1670, opened 2023-01-28
[MainUI] Fix redirection for reverse proxy with authentication not working

### LINKED ISSUE #1207: Reverse proxy with authentication though redirection not working with service worker
## The problem

MainUI does not work behind a reverse proxy performing authentication through redirection (not Basic Auth). When the user tries to access openHAB landing page, if not authenticated but if the service worker has been previously cached, there is no redirection to the authentication page as it is the API call that are redirected and fail.

When authenticated, the setup works flawlessly (and in fact much better than Basic Auth where some browser may repeatedly ask for credentials).

## Expected behavior

In case of redirection of an API call, the entire browser window should be reloaded to the current URL, providing an opportunity for the browser to follow the redirection to the external authentication page.

As suggested in this other issue  #1018, a configuration option to disable service worker could also be a solution.

## Steps to reproduce

1. Install and configure reverse proxy with authentication through redirection
2. Log in with reverse proxy
3. Access openHAB page
4. Log out from reverse proxy
5. Refresh openHAB page

## Your environment

openHAB stable 3.1.0
Web UI in browser (desktop: Safari 14, Chrome 94, mobile Safari 14)
nginx SSL reverse proxy with [auth_request](http://nginx.org/en/docs/http/ngx_http_auth_request_module.html) module
[vouch-proxy](https://github.com/vouch/vouch-proxy) authentication proxy

## Browser console

Example error message when displaying openHAB from service worker cache.

```txt
[Error] XMLHttpRequest cannot load https://REDACTED.auth0.com/authorize?client_id=REDACTED&redirect_uri=https%3A%2F%2FREDACTED_OH_HTTP_NAME%2Fsso%2Fauth&response_type=code&scope=openid+email+profile&state=6F9uv2BorAr86SRC5Gc8oc3GrZBsIBZ due to access control checks.
```

## Browser network traffic

Screenshot 2021-11-15 at 17 31 56

Screenshot 2021-11-15 at 17 32 45

## Additional information

None

florian-h05:
Fixes #1207.

## Description

When using a reverse proxy like nginx to perform authentication through redirection (by using an authentication service like Auth0, Google etc.), the service-worker can cause problems. On intial load of MainUI, the user is correctly redirected to the login page, but after the service-worker is loaded and cached, the user will get an empty page when his session expired or he logged out. In this case, no redirection to the login page will happen, because it's not the MainUI load but "only" the initial REST API call that fails.

This PR solves this problem by handling the above described case for the initial REST API call (error handling logic was already in place for 401/Unauthorized (Basic Auth) but was missing for redirection). If that initial request fails, the user will get redirected to the login page.

## Approach

This PR is the way it is, because due to some limitations and problems I had discovered while solving the issue, I found no better solution:

- Handling of the 302/Found redirection code is not possible, because `XhrHttpRequest` follows the redirection, which in fact is rejected by the auth0 servers because of their CORS policy (which is the right behaviour)
- Disable following redirects is not possible for `XhrHttpRequest`. Using the `fetch` API ignoring redirects is possible, but due to the CORS policy it is not possible to access the `location` header from the code and therefore no redirection is possible.
- Catching the error that gets thrown when `XhrHttpRequest` follows the rejected redirection is possible and my solution: in case of such an error (code 0 and message 0, no headers), the cached service-worker is unloaded and then the site is reload. This redirects the user to the login page.

## Thoughts

When working on this PR, I wondered if we should provide error handling for all REST calls and not only for the initial call. This is clearly out of scope for this PR, but I'd like to get your opinion on this @ghys.

## Testing

I have tested this using `nginx` as reverse proxy, `vouche-proxy` for the `auth_request` and `auth0` as authentication provider. On my 3.4.1 test system, the fix is working great :smiley: 

/cc @tarag Can you please verify that the fix works?

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -441,7 +441,8 @@ export default {
                 const parts = line.split(':\t')
                 headersObj[parts[0]] = parts[1]
               })
-              window.location.href = headersObj['location']
+              // Redirect according to location header but modify URL arguments to redirect back to the UI and not the REST API after authentication
+              window.location.href = headersObj['location'].replace(window.location.href + 'rest', window.location.href)
             }
           } else {
             this.$f7.dialog.alert('openHAB REST API connection failed with error ' + err.message || err.status)
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -251,6 +251,8 @@
 </style>
 
 <script>
+import Framework7 from 'framework7/framework7-lite.esm.bundle.js'
+
 import cordovaApp from '../js/cordova-app.js'
 import routes from '../js/routes.js'
 import PanelRight from '../pages/panel-right.vue'
@@ -402,8 +404,10 @@ export default {
     loadData (useCredentials) {
       const useCredentialsPromise = (useCredentials) ? this.setBasicCredentials() : Promise.resolve()
       return useCredentialsPromise
-        .then(() => { return this.$oh.api.getWithUnwrappedErr('/rest/') })
+        .then(() => { return Framework7.request.promise.json('/rest/') })
         .catch((err) => {
+          console.error('Initial API request failed with error:')
+          console.info(err)
           if (err.message === 'Unauthorized' || err.status === 401) {
             if (!useCredentials) {
               // try again with credentials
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -402,9 +404,11 @@ export default {
     loadData (useCredentials) {
       const useCredentialsPromise = (useCredentials) ? this.setBasicCredentials() : Promise.resolve()
       return useCredentialsPromise
-        .then(() => { return this.$oh.api.get('/rest/') })
+        .then(() => { return Framework7.request.promise.json('/rest/') })
```
florian-h05, line 407: I chose to use the unwrapped request promise here to get access to the error object, which is logged. This may prove really helpful for other cases where MainUI doesn't load with the cached service-worker due to some API request failure.

ghys (approved):
Approving this after doing some tests. If there are regressions in this sensitive code then we'll deal with them.
Thanks for the improvement!

florian-h05:
Thanks for testing this! I will try to keep an eye on that and see if there are any regressions (I guess if there are, they cannot be overlooked).

kaikreuzer:
@florian-h05 When cherry-picking commits to the patch release branches, please remember to add the "patch" label to the PR, because otherwise the fix is not picked up by the release notes script. Thanks!

PR #1682, opened 2023-02-05
[MainUI] Analyzer: Use Item label instead of name if available

### LINKED ISSUE #1679: Chart page title doesn’t use Item label
The title bar of the chart page is using the Item name(s) instead of the Item label(s), which is quite “unpretty”. 

I guess this code needs some tweaks to use the label if defined:

https://github.com/openhab/openhab-webui/blob/952392225a78c3b35f7503525697de2799017908/bundles/org.openhab.ui/web/src/pages/analyzer/analyzer.vue#L542-L546

FYI: I think I can solve that issue myself.

florian-h05:
Fixes #1679.

Signed-off-by: Florian Hotze <florianh_dev@icloud.com>


ghys (approved):
I don't remember why I didn't do this in the first place, there must have been a reason but can't recall! 🤷  So LGTM.

PR #1688, opened 2023-02-06
[habpanel] Use commandDescription instead of stateDescription for selection list

### LINKED ISSUE #1686: [habpanel] Inconsistency in how state/command options are handled
I noticed an inconsistency in how state options and command options are handled between BasicUI and HABpanel.

Here you can see where the Selection widget in BasicUI calls `getCommandDescription().getCommandOptions()` to retrieve the info to populate the list for the Selection widget. This works irrespective of whether the item has state options or command options.
https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SelectionRenderer.java#L91

I believe it's because the REST API returns state options AND command options when only state options are defined for the item. I believe it's a result of this, where state options are copied to command options when no command options exist.
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java#L425

Here you can see where the Selection widget in HABpanel uses the state options to populate the list for the Selection widget. This works when state options are defined for the item, but doesn't work if only command options are defined.
https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.habpanel/web/app/widgets/selection/selection.widget.js#L55

So, should HABpanel be changed to work like BasicUI (i.e. should it use command options to populate the list of options for its Selection widget?

Note that HABot might have the same issue as HABpanel.
https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.habot/web/src/components/HbSelect.vue#L37

mhilbush:
Fixes #1686

To be consistent with how Basic UI works, use commandDescription to populate the list used by the Selection widget. This will ensure that a selection list will be presented to the user if the item has state options or command options. This works because openHAB internally copies the state options to the command options when command options don't exist on the item.

Signed-off-by: Mark Hilbush <mark@hilbush.com>

### DIFF
--- a/bundles/org.openhab.ui.habpanel/web/app/widgets/selection/selection.widget.js
+++ b/bundles/org.openhab.ui.habpanel/web/app/widgets/selection/selection.widget.js
@@ -52,7 +52,7 @@
             if (!vm.choices) {
                 switch (vm.widget.choices_source) {
                     case 'server':
-                        vm.choices = vm.item.stateDescription.options.map(function (option) {
+                        vm.choices = vm.item.commandDescription.commandOptions.map(function (option) {
                             return { cmd: option.value, label: option.label };
                         });
                         break;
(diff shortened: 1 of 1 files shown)


mhilbush:
@lolodomo Would this be a candidate to back port to 3.4 if there's another patch release.

kaikreuzer:
I have cherry-picked this fix to the `3.4.x` branch.

PR #1694, opened 2023-02-09
Fix typo in comment

theHacker:
Was just reading a little bit into the code, and found a typo :-)


ghys (approved):
Thanks, I think we can waive the DCO check for this one.

PR #1695, opened 2023-02-10
Improve SIP Client component docs

### LINKED ISSUE #1692: SIP Widget: there is no need for two distinct dialogs to provide the password
With openHAB 3.4 configuring the SIP widget can be done:
* by clicking on “Local Settings”, where one can adjust SIP Credentials (username, password, Own SIP address); and
* by clicking on “Configure Widget”, where the available settings are: WebSocket URL, Domain, Username, Password, Hide Caller ID, Phonebook, Enable Video, Enable Local Videw View, and “advanced settings”.

Since the “Local Settings” provide no space for domain, I assume that the username must contain the domain.  But since the “Configure Widget” asks separately for Domain and Username, I assume that the username shall not be included the domain.  So what is the purpose of having the Domain as separate value?

I think it is very confising to offer some settings over two different dialogs, and some over a single dialog.  password can be entered over both dialogs above, but Websocket URL and Own SIP address only over a single (distinct) dialog.  In fact, after configuring “Local Settings” I have assumed that I am ready and nothing more needs to be configured.  Please move the Own SIP Address property to the “Configure Widged” dialog and remove the “Local Settings” dialog.

“Own SIP address” with comment “ SIP Address (phone number) of this client. Used by the client to remove itself from the dial popup.” is not documented at https://www.openhab.org/docs/ui/components/oh-sipclient.html and https://www.openhab.org/docs/ui/components/oh-sipclient-card.html and I do not understand its purpose.

florian-h05:
Closes #1692.

Signed-off-by: Florian Hotze <florianh_dev@icloud.com>

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-map-page.md
+++ b/bundles/org.openhab.ui/doc/components/oh-map-page.md
@@ -52,12 +52,12 @@ Displays markers on a map
 </PropBlock>
 <PropBlock type="TEXT" name="tileLayerProvider" label="Provider for the background tiles">
   <PropDescription>
-    The provider of tiles to use for the background of the map. Use one from <a class="external text-color-blue" target="_blank" href="https://leaflet-extras.github.io/leaflet-providers/preview/">Leaflet Providers</a>, Some providers will not work until you set options, like access tokens, in the <code>tileLayerProviderOptions</code> parameter (in Code view). See <a class="external text-color-blue" target="_blank" href="https://github.com/leaflet-extras/leaflet-providers#providers-requiring-registration">here</a> for more info. The default is CartoDB, the variant depending on the dark mode setting.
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.ui/doc/components/oh-map-page.md
```diff
@@ -52,12 +52,12 @@ Displays markers on a map
...
+    The provider of tiles to use for the background of the map. Use one from <a class="external text-color-blue" target="_blank" href="http://leaflet-extras.github.io/leaflet-providers/preview/">Leaflet Providers</a>, Some providers will not work until you set options, like access tokens, in the <code>tileLayerProviderOptions</code> parameter (in Code view). See <a class="external text-color-blue" target="_blank" href="https://github.com/leaflet-extras/leaflet-providers#providers-requiring-registration">here</a> for more info. The default is CartoDB, the variant depending on the dark mode setting.
```
ghys, line 55: These were arguably "regressions" as I've remarked elsewhere but I've since then understood they were added by a PRs which ignored the "automatically generated" nature of these parts of the docs.
I'd say maybe we can accept the regression and then fix those URLs in the proper way (as in, fixing those in the source JS i.e. https://github.com/openhab/openhab-webui/blob/deb8e195fc106ee4f9a602fe1d9642abcadc50b3/bundles/org.openhab.ui/web/src/assets/definitions/widgets/map/index.js#L22-L25)

ghys (approved):

bundles/org.openhab.ui/doc/components/oh-map-page.md
florian-h05: I will try to provide a PR to fix these URLs in the JS ASAP 👍

florian-h05: @ghys See #1704.

PR #1703, opened 2023-02-12
Add new capabilities to oh-knob component using vue-round-slider

### LINKED ISSUE #1640: Add a round slider
While the existing knob is not what I really like for widget creation, I would rather like to have something like this
https://vue.roundsliderui.com/

@ghys Any chance to have this implemented ?

stefan-hoehn:
fixes #1640

Enhanced the oh-know via the [rounded-slider component](https://github.com/soundar24/vue-round-slider) and add several new capabilities.

See details at #1640 and examples there.

image


florian-h05 (approved):
Overall LGTM, thanks!

The `dottedPath` option for the slider-control does also apply to the knob which causes some design problems - I will take care of that.
And I also have a few parameter naming improvements in mind.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-knob.vue
```diff
@@ -1,6 +1,16 @@
 <template>
-  <knob-control v-bind="config" :text-color="config.textColor || ($f7.data.themeOptions.dark === 'dark') ? '#ffffff' : undefined" :value="value"
-                @input="sendCommandDebounced($event)" @click.native.stop="sendCommandDebounced(value, true)" @touchend.native.stop="sendCommandDebounced(value, true)" />
+  <div :style="dottedPath">
+    <knob-control v-if="!config.sliderType" v-bind="config" :text-color="config.textColor || (($f7.data.themeOptions.dark === 'dark') ? '#ffffff' : undefined)" :value="value"
+                  @input="sendCommandDebounced($event)" @click.native.stop="sendCommandDebounced(value, true)" @touchend.native.stop="sendCommandDebounced(value, true)" />
+
+    <round-slider v-else :value="value" :min="config.min" :max="config.max" :step="config.stepSize" :radius="config.size/2"
```
florian-h05, line 6: Using `v-bind` here would have been nice, but I see the reason why you did not: several paramaters have different names compared to the prop names.

stefan-hoehn: yes, exactly, that's why I couldn't use v-bind.

hmerk:
Thanks guys !!!!!!

ghys:
@florian-h05 @stefan-hoehn A word of advice when adding new libraries.

The RelativeCI bot was added (https://github.com/openhab/openhab-webui/pull/1703#issuecomment-1427098231) to track the size of the overall app and the "initial JS/CSS".

Here we can see that the Initial JS has increased (`1.86MiB(+7.53%)`) from the `1.73MiB` baseline - this is before minification but still - the baseline is the `app.[hash].js` and `app.[hash].css` which your browser will load no matter what your pages contain when you bring up the UI. If you explore more of the UI, additional JS and CSS will be loaded on demand using webpack's [code splitting](https://webpack.js.org/guides/code-splitting/). But in this case, since widgets are in the baseline, all libraries that are imported with `import RoundSlider from 'vue-round-slider'` will be part of the baseline, meaning they will be loaded for everyone regardless of whether they use the round sliders or not.

This is illustrated if you open the "report" artifact from the GitHub build results: https://github.com/openhab/openhab-webui/actions/runs/4157806937

image

The purple block in the top-left corner in the JS entry point and the added weight of vue-roundslider is highlighted).
In some cases it might be better to lazy load a component when it's too heavy.

See for example:

https://github.com/openhab/openhab-webui/blob/5bf98457f4b293b3c513b4292084f873df4886bd/bundles/org.openhab.ui/web/src/components/widgets/system/oh-video.vue#L29-L32

In this case maybe it's acceptable (or maybe not, in fact) but do keep that in mind.

florian-h05:
Thanks for the advice.

Wasn‘t aware of that size increase — I guess it‘s time for a follow-up PR to split off that component.

ghys:
No worries, the reasons I'm telling you this is because while the overall size of the UI has increased over the years, I've tried to keep the baseline to a minimum. For instance compare https://app.relative-ci.com/projects/ZNG5hy4VeSJQVQcq1Kvu/jobs/9-85f49c87-0ed2-470c-95e5-a0a4edad8145:

image

to a job from 6 days ago https://app.relative-ci.com/projects/ZNG5hy4VeSJQVQcq1Kvu/jobs/761-CfxT1mQIUU2wyVZ1P1hW:

image

The overall size has indeed increased from ~10MB to ~16MB as more features were added, but the "initial JS" has been kept to a minimum: 1.73MB vs. 1.67MB!
So the more we can keep (or carve) out of it (with common sense, some things belong in it), the better 🙂!

florian-h05:
@ghys PR #1718 implements lazy loading for the about 130 kB large `vue-round-slider` package (https://bundlephobia.com/package/vue-round-slider@1.0.1).

stefan-hoehn:
thanks, @florian-h05 , you are too fast for me ;-) I wanted to do it today, so thanks for taking care!

PR #1706, opened 2023-02-13
[basicui] Add support for iconify and Google material icons

lolodomo:
Make usage of iconify an option

Also make BasicUI compatible with other OH iconsets than classic


lolodomo:
Example with the following sitemap:
```
sitemap test label="Tests"
{
        Frame {
                Default item=CurrentTemp
                Default item=SetpointTemp
                Text item=CurrentTemp icon="if:wi:day-light-wind"
                Text item=CurrentTemp icon="iconify:wi:day-sunny-overcast"
                Group item=GTest
        }
}
```
image

bundles/org.openhab.ui.basic/src/main/resources/snippets/main.html
```diff
@@ -68,6 +68,7 @@
 				</div>
 			</div>
 		</script>
+		<script src="https://code.iconify.design/iconify-icon/1.0.3/iconify-icon.min.js"></script>
```
lolodomo, line 71: @ghys : what will happen if the user has no internet connection? Will it block the loading of the page? 
I could add a setting to enable or disable usage of iconify.

lolodomo:
I am going to also add support of material-icons which are already packaged and used by BasicUI (home and back icons for example).

bundles/org.openhab.ui.basic/src/main/resources/snippets/main.html
lolodomo: I will add this option.

ghys: Depends but since you're using the Iconify web component, it has a cache feature which seems to be enabled by default.
https://docs.iconify.design/iconify-icon/enable-cache.html

> By default, localStorage is enabled, sessionStorage is disabled.

This means if an icon has already been cached locally it will still work even without an internet connection. But if it isn't in the cache I don't think it will block the loading, you might just get a placeholder.

lolodomo: I was thinking about the script loading. 
In fact, the best would be to test BasicUI without internet connection. But not so easy to cut internet access while keeping local network.

ghys: Ah right, I didn't look at the change. I don't suppose everyone will be happy with having a JS file loaded from the internet in their Basic UI...

lolodomo: Ok so I will add an option.

lolodomo: Option now implemented. Default is to not use iconify.

lolodomo:
PR is now finished, ready for a review.

bundles/org.openhab.ui.basic/src/main/resources/snippets/main.html
lsiepel: What about embedding the file locally as resource?

lolodomo: Thank you @lsiepel , it works well with the file put as local resource.

lolodomo:
@kaikreuzer @ghys @florian-h05 : any chance to have a review of this enhancement?

florian-h05:
I am only MainUI maintainer, so you'll need Kai or Yannick for review.

kaikreuzer (approved):
lgtm, although I haven't tested it - but I assume you did that extensively anyhow already. 😄

PR #1718, opened 2023-02-17
Refactor `oh-knob` component to remove `vue-knob-control` & Lazy load dependencies

florian-h05:
Follow-up PR for #1703.

This:
- lazy loads `vue-knob-control`
- refactors `vue-round-slider` to only use `vue-round-slider` and remove `vue-knob-control`

A compatiblity layers ensures that existing `oh-knob` configurations still work and default values of properties have been adjusted so that `vue-round-slider` looks like the removed `vue-knob-control`.
Paremeter description and docs were updated accordingly to `vue-round-slider`.


ghys:
<html>
<body>


1.73MiB(-6.88%) | 1.86MiB
-- | --



</body>
</html>

Much better, thanks! I didn't notice vue-round-slider was a jQuery plugin, so this also moves the jQuery dependency to the separate bundle - it's definitely not needed in the entry point.
There's also an argument that it would be fair to move the 2nd library for rendering oh-knobs, vue-knob-control, to the new bundle as well, even if it's not as heavy, wdyt?

image

florian-h05:
IMO makes sense, I have implemented that. 
I wondered if we could get rid of `vue-knob-control` and entirely replace it with `vue-round-slider` — I will try this out and probably create a PR. WDYT?

ghys:
> I wondered if we could get rid of vue-knob-control and entirely replace it with vue-round-slider — I will try this out and probably create a PR. WDYT?

I don't mind, except we should indeed mind about those who already rely on the vue-knob-control, as the parameters are mostly different (though I've already said that the upcoming 4.0 could mean having a clean slate on some things, even if it involves breaking changes).

florian-h05:
We could keep the compatibility layer that to keep compatibility with the current parameters and avoid breaking changes …

florian-h05:
@ghys @stefan-hoehn 

I have refactored the `oh-knob` component to use `vue-round-slider` by default and remove `vue-knob-control`. See the PR description above.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-knob.vue
```diff
@@ -1,32 +1,24 @@
...
-                  :startAngle="config.startAngle" :endAngle="config.endAngle"
-                  :borderWidth="config.borderWidth" :borderColor="config.borderColor" :circleShape="config.circleShape"
-                  @input="sendCommandDebounced($event)" @click.native.stop="sendCommandDebounced(value, true)" @touchend.native.stop="sendCommandDebounced(value, true)" />
-  </div>
+  <round-slider v-bind="config" :value="value" :min="config.min" :max="config.max" :step="config.stepSize" :radius="(config.responsive) ? config.size/2 + '%' : config.size/2"
+                :style="`stroke-dasharray: ${(config.dottedPath) ? config.dottedPath : 0}`"
+                :rangeColor="config.primaryColor" :pathColor="config.secondaryColor"
+                :tooltipColor="config.textColor" mouseScrollAction="true"
+                :disabled="config.disabled" :width="config.strokeWidth" :line-cap="config.lineCap"
+                :startAngle="config.startAngle ? config.startAngle : -50" :endAngle="config.endAngle ? config.endAngle : -130"
```
ghys, line 7: instead of all that maybe it would be better to have perhaps a `resolvedConfig` computed property and do a `v-bind`  of that, instead of ` config`? This would lookup the former properties but still fall back to the actual properties if defined.

I'd like to offer the fact that the vue-round-slider config could be fed as-is instead of having some arbitrary mappings (for example, _rangeColor_ is mapped to _primaryColor_ but it means that a user can't set the original property _rangeColor_, I think it's rather confusing).

florian-h05: So you propose to v-bind a computed resolvedConfig, where this resolvedConfig is falling back to the old config if the new properties aren’t set, correct?

florian-h05:
@ghys I have addressed your review and introcuded a `resolvedConfig`, that falls back on the old parameters if the new ones aren't defined. The resolvedConfig also sets a few defaults (e.g. start and end angle) to make round-slider look like knob-control by default.
I have updated the parameter definition for the round-slider parameters.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-knob.vue
ghys: Yeah, what I meant is that if a user wants to set e.g. `tooltipColor` they should be able to use either `tooltipColor` or `textColor` (the former property name) in config, probably in that order.

florian-h05: That's possible with the latest commits :+1:

florian-h05: From my side this is finished now.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-knob.vue
```diff
@@ -1,33 +1,35 @@
...
+      return {
+        ...cfg,
+        step: cfg.step ? cfg.step : cfg.stepSize,
+        radius: cfg.radius ? cfg.radius : (cfg.responsive ? cfg.size / 2 + '%' : cfg.size / 2),
+        rangeColor: cfg.rangeColor ? cfg.rangeColor : cfg.primaryColor,
+        pathColor: cfg.pathColor ? cfg.pathColor : cfg.secondaryColor,
+        tooltipColor: cfg.tooltipColor ? cfg.tooltipColor : cfg.textColor,
+        width: cfg.width ? cfg.width : cfg.strokeWidth,
+        startAngle: !cfg.circleShape ? (cfg.startAngle ? cfg.startAngle : -50) : null,
+        endAngle: !cfg.circleShape ? (cfg.endAngle ? cfg.endAngle : -130) : null
```
ghys, line 30: Please be aware of the pitfalls of these ternary syntaxes for fallback values, any falsy value doesn't necessarily have to fallback. It's not "if undefined then.... else", it's "if falsy then... else".

For instance if your `cfg.radius` is 0 for some reason then the fallback would take effect, even if 0 would have been a valid and reasonable value.
Perhaps you're actually aware and I apologize for being pedantic, but I did bite me a few times so just wanted to make sure.

ghys (approved):
LGTM with the below remark but I don't see it having an impact given the relative nonsense of setting these properties to falsy values.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-knob.vue
florian-h05: I am aware of false and it should not end up in these paramaters, but I haven't thought of `0`. I will fix the fallbacks, thanks.

florian-h05: I've fixed the fallbacks and now explicitely check for `undefined` see https://github.com/openhab/openhab-webui/pull/1718/commits/6b6523e721c926a53973b6be152ab971ab6ac11d.
Note that for `responsive` the ternary operator is fine using falsy.

florian-h05:
@ghys Can you please have a look at the small, last change so we can merge this?

digitaldan:
I have this deployed on my production system and works great FYI.  

image

florian-h05:
> I have this deployed on my production system and works great FYI.

Cool :+1: 
Looks good.

Merge conflicts are resolved and this is finished now, so this is ready for merging now.

digitaldan:
@florian-h05 i opened one more PR against your branch, last one i promise ! 🤞

hmerk:
@florian-h05 @ghys 
could we have this merged before M1 will be generated this weekend ?

florian-h05:
The M1 build will start at 5pm CET (in 9 minutes) and I am not allowed to merge my own PR, so we need @ghys for merging.

hmerk:
Ups, did not know that it will be done today at 5:00 CET…..

hmerk:
Well, last time i checked the release buil, it took a couple of hours until it started to build the ui stuff.
Perhaps @ghys will have a chance to merge it before then….

hmerk:
@florian-h05 
M1 build started now …..

ghys (approved):
Sorry I wasn't available at all yesterday for the build. LGTM!

PR #1712, opened 2023-02-16
[basicui] Remove Roboto font as unused by the app

lolodomo:
Fix #1273

Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #1719, opened 2023-02-17
[basicui] Remove roboto font leftovers

florian-h05:
Follow-up PR for #1712.

/cc @lolodomo


bundles/org.openhab.ui.basic/NOTICE
```diff
@@ -18,213 +18,3 @@ Material Design Lite 1.0.2
...
-* License: Apache License 2.0
-* Source: https://github.com/google/roboto
-
-== Third-party license(s)
-
-=== Apache License, 2.0
-
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
```
lolodomo, line 32: @kaikreuzer : can you please confirm that we don't need to include the full license NOTICE from Apache License but just mention the license ?

lolodomo:
@florian-h05 : thank you a lot to finish that, that is ok for me, I have just a little doubt about the license notice (even if I think you took the right option).

bundles/org.openhab.ui.basic/NOTICE
florian-h05: Oh, now I understand what you mean.

Apache is used by material design lite as well …
I’m not sure what the rules are exactly, in case you want to finish this PR and change something, you should be able to push to my branch.

kaikreuzer: I think originally the requirement (from Eclipse Foundation times) was to include the full license texts at that place. But afaik, we didn't do this for other add-ons and I don't think that there is any legal need to do so. I'd hence be fine with removing it from this file.

kaikreuzer (approved):

PR #1723, opened 2023-02-17
[Blockly] Add workspace search

stefan-hoehn:
this adds a really helpful plugin that allows to search within the blockly workspace via Cmd/Ctrl-F

it is based on https://github.com/google/blockly-samples/tree/master/plugins/workspace-search

it opens up a small search window at the top right and blackens the blocks that match the search string

image


florian-h05 (approved):
LGTM, thanks!

Just some minor style issues with the search box in dark mode - I took care of them.
Only thing left is the color of the action buttons in the search box, but we cannot change the color because the buttons are static images.

PR #1726, opened 2023-02-21
Add step parameter to oh-input

mherwege:
When using number type in an oh-input card or oh-input-item list item, a default step value of 1 is always respected. This leads to confusion with users not being able to enter fractional numbers.
See [forum post](https://community.openhab.org/t/problems-with-numerical-values-in-input-cards/134774).

This PR makes the step parameter visible, and uses a default of any value (no step check) if no step value is provided. This is a small breaking change, but less confusing to users.

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-input-card.md
+++ b/bundles/org.openhab.ui/doc/components/oh-input-card.md
@@ -120,6 +120,11 @@ Display an input in a card
     Use the formatted state as the value for the input control
   </PropDescription>
 </PropBlock>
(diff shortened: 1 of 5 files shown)


ghys:
Thanks, could you please regenerate the docs to make the new parameter appear?

https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/CONTRIBUTING.md#documentation--resources

mherwege:
Updated the doc

ghys (approved):

PR #1727, opened 2023-02-22
Handle NOT_YET_READY in thing status

J-N-K:
See https://github.com/openhab/openhab-core/issues/3394


florian-h05 (approved):
LGTM, thanks!

digitaldan:
Could we possibly modify the BaseThingHandler to remove then add and persist channels when updated ?  So handle this behind the scenes so bindings don't need to change ?  

maybe here? 
https://github.com/openhab/openhab-core/blob/f48200caa6425324d0d5929004b8a148a0f0cbf7/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java#L433

PR #1729, opened 2023-02-24
[basicui] input widget

mherwege:
Depends on [PR #3398](https://github.com/openhab/openhab-core/pull/3398)

This adds support for an input widget in Basic UI. See Core PR for more extensive description.

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
@@ -33,9 +33,7 @@ import org.osgi.service.component.annotations.Reference;
  * This is an implementation of the {@link WidgetRenderer} interface, which
  * can produce HTML code for Input widgets.
  *
- * @author Kai Kreuzer - Initial contribution and API
- * @author Vlad Ivanov - BasicUI changes
- * @author Mark Herwege - Input widget
+ * @author Mark Herwege - Initial contributor
  */
 @Component(service = WidgetRenderer.class)
 @NonNullByDefault
@@ -57,15 +55,14 @@ public class InputRenderer extends AbstractWidgetRenderer {
         String snippet = getSnippet("input");
 
         snippet = preprocessSnippet(snippet, w);
-        snippet = snippet.replace("%id%", itemUIRegistry.getWidgetId(w));
 
         String dataState = getValue(w);
         State state = itemUIRegistry.getState(w);
         if (state == null || state instanceof UnDefType) {
-            snippet = snippet.replace("%empty_state%", !dataState.isEmpty() ? dataState : "-");
+            snippet = snippet.replace("%undef_state%", !dataState.isEmpty() ? dataState : "-");
             snippet = snippet.replace("%data_state%", "");
         } else {
-            snippet = snippet.replace("%empty_state%", "");
+            snippet = snippet.replace("%undef_state%", "");
             snippet = snippet.replace("%data_state%", dataState.isEmpty() ? state.toString() : dataState);
         }
 
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
@@ -59,11 +59,11 @@ public class InputRenderer extends AbstractWidgetRenderer {
         String dataState = getValue(w);
         State state = itemUIRegistry.getState(w);
         if (state == null || state instanceof UnDefType) {
-            snippet = snippet.replace("%undef_state%", !dataState.isEmpty() ? dataState : "-");
+            snippet = snippet.replace("%undef_state%", dataState);
             snippet = snippet.replace("%data_state%", "");
         } else {
             snippet = snippet.replace("%undef_state%", "");
-            snippet = snippet.replace("%data_state%", dataState.isEmpty() ? state.toString() : dataState);
+            snippet = snippet.replace("%data_state%", dataState);
         }
 
         String dataType = state instanceof Number ? "Number" : "Text";
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
@@ -66,6 +66,9 @@ public class InputRenderer extends AbstractWidgetRenderer {
             snippet = snippet.replace("%data_state%", dataState.isEmpty() ? state.toString() : dataState);
         }
 
+        String dataType = state instanceof Number ? "Number" : "Text";
+        snippet = snippet.replace("%data_type%", dataType);
+
         // Process the color tags
         snippet = processColor(w, snippet);
 
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
@@ -1,0 +1,79 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.ui.basic.internal.render;
+
+import org.eclipse.emf.common.util.ECollections;
+import org.eclipse.emf.common.util.EList;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.i18n.LocaleProvider;
+import org.openhab.core.i18n.TranslationProvider;
+import org.openhab.core.model.sitemap.sitemap.Input;
+import org.openhab.core.model.sitemap.sitemap.Widget;
+import org.openhab.core.types.State;
+import org.openhab.core.types.UnDefType;
+import org.openhab.core.ui.items.ItemUIRegistry;
+import org.openhab.ui.basic.render.RenderException;
+import org.openhab.ui.basic.render.WidgetRenderer;
+import org.osgi.framework.BundleContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+
+/**
+ * This is an implementation of the {@link WidgetRenderer} interface, which
+ * can produce HTML code for Input widgets.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ * @author Vlad Ivanov - BasicUI changes
+ * @author Mark Herwege - Input widget
+ */
+@Component(service = WidgetRenderer.class)
+@NonNullByDefault
+public class InputRenderer extends AbstractWidgetRenderer {
+
+    @Activate
+    public InputRenderer(final BundleContext bundleContext, final @Reference TranslationProvider i18nProvider,
+            final @Reference ItemUIRegistry itemUIRegistry, final @Reference LocaleProvider localeProvider) {
+        super(bundleContext, i18nProvider, itemUIRegistry, localeProvider);
+    }
+
+    @Override
+    public boolean canRender(Widget w) {
+        return w instanceof Input;
+    }
+
+    @Override
+    public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) throws RenderException {
+        String snippet = getSnippet("input");
+
+        snippet = preprocessSnippet(snippet, w);
+        snippet = snippet.replace("%id%", itemUIRegistry.getWidgetId(w));
+
+        String dataState = getValue(w);
+        State state = itemUIRegistry.getState(w);
+        if (state == null || state instanceof UnDefType) {
+            snippet = snippet.replace("%empty_state%", !dataState.isEmpty() ? dataState : "-");
+            snippet = snippet.replace("%data_state%", "");
+        } else {
+            snippet = snippet.replace("%empty_state%", "");
+            snippet = snippet.replace("%data_state%", dataState.isEmpty() ? state.toString() : dataState);
+        }
+
+        // Process the color tags
+        snippet = processColor(w, snippet);
+
+        sb.append(snippet);
+        return ECollections.emptyEList();
+    }
+
+}
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
@@ -37,7 +37,7 @@ import org.slf4j.LoggerFactory;
  * This is an implementation of the {@link WidgetRenderer} interface, which
  * can produce HTML code for Input widgets.
  *
- * @author Mark Herwege - Initial contributor
+ * @author Mark Herwege - Initial contribution
  */
 @Component(service = WidgetRenderer.class)
 @NonNullByDefault
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
@@ -75,5 +75,4 @@ public class InputRenderer extends AbstractWidgetRenderer {
         sb.append(snippet);
         return ECollections.emptyEList();
     }
-
 }
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
@@ -17,6 +17,8 @@ import org.eclipse.emf.common.util.EList;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.i18n.LocaleProvider;
 import org.openhab.core.i18n.TranslationProvider;
+import org.openhab.core.items.Item;
+import org.openhab.core.items.ItemNotFoundException;
 import org.openhab.core.model.sitemap.sitemap.Input;
 import org.openhab.core.model.sitemap.sitemap.Widget;
 import org.openhab.core.types.State;
@@ -28,6 +30,8 @@ import org.osgi.framework.BundleContext;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This is an implementation of the {@link WidgetRenderer} interface, which
@@ -38,6 +42,8 @@ import org.osgi.service.component.annotations.Reference;
 @Component(service = WidgetRenderer.class)
 @NonNullByDefault
 public class InputRenderer extends AbstractWidgetRenderer {
+
+    private final Logger logger = LoggerFactory.getLogger(InputRenderer.class);
 
     @Activate
     public InputRenderer(final BundleContext bundleContext, final @Reference TranslationProvider i18nProvider,
@@ -66,7 +72,18 @@ public class InputRenderer extends AbstractWidgetRenderer {
             snippet = snippet.replace("%data_state%", dataState);
         }
 
-        String dataType = state instanceof Number ? "Number" : "Text";
+        Item item = null;
+        try {
+            item = itemUIRegistry.getItem(w.getItem());
+        } catch (ItemNotFoundException e) {
+            logger.debug("Failed to retrieve item during widget rendering: {}", e.getMessage());
+        }
+        String dataType;
+        if (item != null && item.getAcceptedDataTypes().stream().anyMatch(o -> Number.class.isAssignableFrom(o))) {
+            dataType = "Number";
+        } else {
+            dataType = "Text";
+        }
         snippet = snippet.replace("%data_type%", dataType);
 
         // Process the color tags
--- a/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
+++ b/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
@@ -6,7 +6,7 @@
 		%label%
 	</span>
 	<div
-		class="input__control mdl-form__control mdl-textfield mdl-js-textfield"
+		class="mdl-form__control mdl-form__input mdl-textfield mdl-js-textfield"
 		data-control-type="input"
 		data-item="%item%"
 		data-widget-id="%widget_id%"
@@ -16,10 +16,10 @@
 		<input
 			%valuestyle% type="text"
 			id="oh-input-%item%"
-			class="input__value input mdl-textfield__input"
+			class="mdl-textfield__input"
 			value="%data_state%"
 		/>
-		<label for="oh-input-%item%" class="input__label input mdl-textfield__label">
+		<label for="oh-input-%item%" class="mdl-textfield__label">
 			%undef_state%
 		</label>
 	</div>
--- a/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
+++ b/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
@@ -6,7 +6,7 @@
 		%label%
 	</span>
 	<div
-		class="mdl-form__control mdl-textfield mdl-js-textfield"
+		class="input__control mdl-form__control mdl-textfield mdl-js-textfield"
 		data-control-type="input"
 		data-item="%item%"
 		data-widget-id="%widget_id%"
@@ -15,11 +15,11 @@
 		<input
 			%valuestyle% type="text"
 			id="oh-input-%item%"
-			class="mdl-form__input--value mdl-form__input mdl-textfield__input"
+			class="input__value input mdl-textfield__input"
 			value="%data_state%"
 		/>
-		<label for="oh-input-%item%" class="mdl-form__input--label mdl-form__input mdl-textfield__label">
-		    %empty_state%
+		<label for="oh-input-%item%" class="input__label input mdl-textfield__label">
+			%empty_state%
 		</label>
 	</div>
 </div>
--- a/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
+++ b/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
@@ -1,6 +1,6 @@
 <div class="mdl-form__row mdl-cell mdl-cell--6-col mdl-cell--8-col-tablet %visibility_class%">
-	<span class="mdl-form__icon">
-		<img data-icon="%category%" src="../icon/%category%?state=%state%&format=%icon_type%&anyFormat=true" />
+	<span %iconstyle% class="mdl-form__icon">
+		%icon_snippet%
 	</span>
 	<span %labelstyle% class="mdl-form__label">
 		%label%
--- a/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
+++ b/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
@@ -10,6 +10,7 @@
 		data-control-type="input"
 		data-item="%item%"
 		data-widget-id="%widget_id%"
+		data-item-type=%data_type%
 		for="oh-input-%item%"
 	>
 		<input
--- a/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
+++ b/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
@@ -19,7 +19,7 @@
 			value="%data_state%"
 		/>
 		<label for="oh-input-%item%" class="input__label input mdl-textfield__label">
-			%empty_state%
+			%undef_state%
 		</label>
 	</div>
 </div>
--- a/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
+++ b/bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
@@ -1,0 +1,25 @@
+<div class="mdl-form__row mdl-cell mdl-cell--6-col mdl-cell--8-col-tablet %visibility_class%">
+	<span class="mdl-form__icon">
+		<img data-icon="%category%" src="../icon/%category%?state=%state%&format=%icon_type%&anyFormat=true" />
+	</span>
+	<span %labelstyle% class="mdl-form__label">
+		%label%
+	</span>
+	<div
+		class="mdl-form__control mdl-textfield mdl-js-textfield"
+		data-control-type="input"
+		data-item="%item%"
+		data-widget-id="%widget_id%"
+		for="oh-input-%item%"
+	>
+		<input
+			%valuestyle% type="text"
+			id="oh-input-%item%"
+			class="mdl-form__input--value mdl-form__input mdl-textfield__input"
+			value="%data_state%"
+		/>
+		<label for="oh-input-%item%" class="mdl-form__input--label mdl-form__input mdl-textfield__label">
+		    %empty_state%
+		</label>
+	</div>
+</div>
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -284,7 +284,19 @@
 			padding: 0 10px 0 10px;
 		}
 	}
-	&__text {
+	&__input {
+		font-weight: normal;
+		font-size: 14px;
+		text-align: right;
+		&--value {
+			margin: 0 0;
+			padding: 0 0 0 0;
+		}
+		&--label {
+			margin: 0 -12px;
+			padding: 0 12px 0 0;
+		}
+}	&__text {
 		overflow: hidden;
 		flex-shrink: 2;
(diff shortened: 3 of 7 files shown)


lolodomo:
I will review this OR during the weekend. 
Did you check the rendering with the two themes and with the condensed or not condensed layout?

mherwege:
I played with various sizes of the UI, thereby forcing a switch to/from condensed mode. i didn’t check the themes.
I am not good in CSS styling and there is a lot to learn for me on that. Always keen to get some feedback to help me learn.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1491,6 +1492,40 @@
...
+	/* class ControlInput extends Control */
+	function ControlInput(parentNode) {
+		Control.call(this, parentNode);
+
+		var
+			_t = this;
+
+		_t.input = _t.parentNode.querySelector("input[type=text]");
+
+		_t.valueNode = _t.parentNode.parentNode.querySelector(o.formValue);
```
lolodomo, line 1504: I should normally find "mdl-form__value" in your snippet but I don't.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
```diff
@@ -0,0 +1,25 @@
...
+	>
+		<input
+			%valuestyle% type="text"
+			id="oh-input-%item%"
+			class="input__value input mdl-textfield__input"
+			value="%data_state%"
+		/>
+		<label for="oh-input-%item%" class="input__label input mdl-textfield__label">
+			%empty_state%
+		</label>
```
lolodomo, line 23: I am not sure to understand what you want to display there ?
This is a part that apparently is not dynamically updated when receiving state updates.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1491,6 +1492,40 @@
...
+
+		function onChange() {
+			_t.parentNode.dispatchEvent(createEvent("control-change", {
+				item: _t.item,
+				value: _t.input.value
+			}));
+		}
+
+		_t.setValuePrivate = function(value) {
+			_t.input.value = value;
```
lolodomo, line 1514: In Java, we are considering value nad item state. Not there ?
If necessary, you can consider this:
```
		_t.setValuePrivate = function(value, itemState) {
		}
```

lolodomo (changes requested):

lolodomo:
Could you please show some screenshots for everyone ?

lolodomo:
Here are some captures:
image
image
with the following sitemap:
```
sitemap test3 label="Tests"
{
        Frame {
                Default item=CurrentTemp
                Input item=CurrentTemp
        }
}
```

lolodomo:
If I enter `20,0  °C`, 200 is sent !
```

15:18:26.747 [INFO ] [openhab.event.ItemCommandEvent       ] - Item 'CurrentTemp' received command 200 °C
15:18:26.763 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'CurrentTemp' changed from 19 °C to 200 °C
```
If I enter `18 °C`, it works:
```
15:21:47.489 [INFO ] [openhab.event.ItemCommandEvent       ] - Item 'CurrentTemp' received command 18 °C
15:21:47.500 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'CurrentTemp' changed from 15 °C to 18 °C
```
Could be a problem with the comma.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -0,0 +1,78 @@
...
+    public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) throws RenderException {
+        String snippet = getSnippet("input");
+
+        snippet = preprocessSnippet(snippet, w);
+        snippet = snippet.replace("%id%", itemUIRegistry.getWidgetId(w));
+
+        String dataState = getValue(w);
+        State state = itemUIRegistry.getState(w);
+        if (state == null || state instanceof UnDefType) {
+            snippet = snippet.replace("%empty_state%", !dataState.isEmpty() ? dataState : "-");
```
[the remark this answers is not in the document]
mherwege: state.toString() was returning NULL in some cases. The logic to create value replaces that with one or multiple -, but that is not always available, so needed to do go this way when there is no value.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
mherwege: If there is no value yet, this label will be displayed as a placeholder. As soon as something is in the item state, or something is entered, it disappears. See mdl light text input widget.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -0,0 +1,78 @@
...
+    public boolean canRender(Widget w) {
+        return w instanceof Input;
+    }
+
+    @Override
+    public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) throws RenderException {
+        String snippet = getSnippet("input");
+
+        snippet = preprocessSnippet(snippet, w);
+        snippet = snippet.replace("%id%", itemUIRegistry.getWidgetId(w));
```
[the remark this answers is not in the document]
mherwege: Indeed, doesn't look like it is needed. I copied it from the text widget, and didn't remove it.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
lolodomo: Ok, I will.
And I will test with a NULL/UNDEF state.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: I have a specific class input__value for the value. I didn't get the expected result combining mdl-form__value with mdl-textfield__input, so created input, input__value and input__label to complement mdl-textfield. The main issue was with alignment.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: The one thing that one cannot do at the moment is make the field completely empty if the underlying item type is numeric or quantityType. One would need to be able to post UNDEF, but I don't think that is useful, and it would block the possibility to update Text items with the string UNDEF. So I don't think there is real need to use itemState here.

mherwege: At least, in my testing, it behaved as I expected without.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
lolodomo: Color and position look no perfect in that case:
image

mherwege:
Thanks for the screenshots.

> Could be a problem with the comma.
> 
> Edit: if I enter 21.0 °C, it works and the value is finally transformed again in 21,0 °C (certainly when the state update is handled).
> 

That's indeed unfortunate. The code just passes the value on and doesn't do any interpretation. The event is a string and OH core interprets it. Do you know if there is a way to improve on this?

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
lolodomo: You probably have to add `%valuestyle%` also on the label tag (input).

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: But `o.formValue` is ".mdl-form__value".  I do not understand how `_t.valueNode` is properly initialized.

lolodomo:
> That's indeed unfortunate. The code just passes the value on and doesn't do any interpretation. The event is a string and OH core interprets it. Do you know if there is a way to improve on this?

If I use 22.5 in my console command to send an update, it works and I see 22,5 as new state.
If I use 25,4 in my console command to send an update, it does not work and I see 254,0 as new state.
I am not 100% sure but I think commands have to be send using `.` as decimal separator, while the displayed value are using the language to decide between .` `and `,`. In France, the decimal separator is `,`.
I think but I am not yet sure that the UI should convert the value in US/java standard, that is a `.`) before sending the value to the server.
Maybe there is JavaScript function for doing that ?

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
mherwege: I don’t think I should, because label can only represent an empty value and is grayed by default.

lolodomo: Look at the column at left, the value in that case is not grayed.

mherwege: Sorry, didn’t see screenshots. Will check again.

mherwege:
> Maybe there is JavaScript function for doing that ?

The key will be knowing what the underlying item type is. If it is Text, it shouldn't be interpreted. If numeric, it should be, but units should be separated. I will see what I can do on that.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
mherwege: Could it have something to do with the dark theme?

lolodomo:
> The key will be knowing what the underlying item type is. If it is Text, it shouldn't be interpreted. If numeric, it should be, but units should be separated.

Correct.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
lolodomo: No this is the same with the default theme (white theme).

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: You are right. I changed to a specific class, but didn't change this. It didn't work anymore.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
mherwege: It was due to the dark theme. I provided a fix.

mherwege:
It looks like there is no easy solution in javascript to parse a number string in a locale. Here is what I could find: https://stackoverflow.com/questions/55364947/is-there-any-javascript-standard-api-to-parse-to-number-according-to-locale

mherwege:
I implemented some basic logic to try to derive the number and thousand separator from the number value in the input. It does not try to base this on the browser locale, just looking at the number and erring on the side of dot as a decimal separator.
I also now revert the value on change if an empty number is entered.

mherwege:
@lolodomo Can you test again? I think comma should be handled more gracelfully now.

lolodomo:
> @lolodomo Can you test again? I think comma should be handled more gracelfully now.

I tested it quickly and yes it seems to be working now using either point or comma. I even tried with a negative value and it seems also OK. Using "+" does not work but I don't thing this is important.

mherwege:
@lolodomo I didn't test, but I think the + should be fixed now (forgot + has a special meaning in the regular expression).

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: I removed item state from java, only kept a check for null or UNDEF.

lolodomo:
I just tested again with very good results:
image
For the number, I tested with `.` and `,`, also without `.` nor `,`, and also with `-` and `+`.
I also tried with `32 °F` and the resulting value was `0,0 °C`, as expected.
I have still to test when value is or switch to NULL or UNDEF.
I will start a new review.

lolodomo:
Note that it should be a good point to add an icon for "input" in our classic iconset. Could be in a separate PR.

lolodomo:
For NULL value, this is now properly (vertically) aligned, there is only the color which is ignored in that case.
image
Unfortunately, I cannot test when state is set to UNDEF while the page is still opened, I can't send UNDEF state through the console command.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
```diff
@@ -0,0 +1,26 @@
+<div class="mdl-form__row mdl-cell mdl-cell--6-col mdl-cell--8-col-tablet %visibility_class%">
+	<span class="mdl-form__icon">
+		<img data-icon="%category%" src="../icon/%category%?state=%state%&format=%icon_type%&anyFormat=true" />
```
lolodomo, line 3: Please replace this line by `%icon_snippet% (recent change on all other snippet).
`

bundles/org.openhab.ui.basic/web-src/_layout.scss
```diff
@@ -218,6 +218,22 @@
 			box-shadow: none;
 			-webkit-box-shadow: none;
 		}
+		.mdl-textfield {
+			&__input,
+			&__label {
+				font-weight: normal;
+				font-size: 14px;
```
lolodomo, line 225: I see that the font is different than the one for Text widget. I was asking myself if we should not use the same ?
And the size should be the same, so to be checked.

bundles/org.openhab.ui.basic/web-src/_theming.scss
```diff
@@ -7,6 +7,8 @@ body[data-theme="default"] {
 	--container-text-color: #616161;
 	--switch-on-track-bg: #9fa8da;
 	--border-color: #ccc;
+	--input-border-color: rgba(0,0,0,.12);
```
lolodomo, line 10: Couldn't we use an existing variable for border color ? Maybe "--border-color" (to be checked where it is used) ?

bundles/org.openhab.ui.basic/web-src/_theming.scss
```diff
@@ -7,6 +7,8 @@ body[data-theme="default"] {
...
+	--input-undef-color: rgba(0,0,0,.26);
```
lolodomo, line 11: Couldn't we use an existing variable for undef color ? Why a different color ? Should'nt it be simply the default color used for the Text widget for example ? I am not sure there is a specific color already defined for UNDEF value.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1491,6 +1511,79 @@
...
+		_t.setValuePrivate = function(value) {
+			if (_t.verify) {
+				_t.verify.cancel();
+			}
+			_t.input.value = value;
+			lastValue = value;
+		};
+
+		_t.setValueColor = function(color) {
+			_t.input.style.color = color;
```
lolodomo, line 1576: You should also update the color of the input label here in case the state is NULL/UNDEF but you receive an update of value color.

lolodomo (changes requested):

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: Isee that it is similar for the Text widget, so I think it is OK.

lolodomo:
I discovered a little bug. If my current value is `19,5 °C` and if I only remove the space between the value and the unit, the new value is 195,0 °C.
```
13:53:17.958 [INFO ] [openhab.event.ItemCommandEvent       ] - Item 'CurrentTemp' received command 195 °C
13:53:17.974 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'CurrentTemp' changed from 19.5 °C to 195 °C
```
Maybe you should reject any value (valueArray[0]) if it contains other thing than + - . , 0 1 2 3 4 5 6 7 8 9 ?

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
```diff
@@ -0,0 +1,26 @@
...
+		data-item-type=%data_type%
+		for="oh-input-%item%"
+	>
+		<input
+			%valuestyle% type="text"
+			id="oh-input-%item%"
+			class="mdl-textfield__input"
+			value="%data_state%"
+		/>
+		<label for="oh-input-%item%" class="mdl-textfield__label">
```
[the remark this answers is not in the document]
mherwege: That would do it. But I am not convinced we should. You cannot set NULL or UNDEF through the UI. If you set anything, it would always obey the color rules. I on purpose refrained from setting a color for NULL or UNDEF to make it very apparent that the - means no value is set. The MDL input element also on purpose uses a different color for the placeholder, and the - is just a placeholder. The color of the placeholder is the same as the borderline color. I tweaked these to be visible, but not too strong on light and dark themes.

bundles/org.openhab.ui.basic/web-src/_theming.scss
mherwege: See above. We could look for an existing color, but I do think it should be distinct from a real value.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: No, because it will not be shown anymore when there is a value. The input label represents the placeholder.

mherwege:
> I discovered a little bug. If my current value is `19,5 °C` and if I only remove the space between the value and the unit, the new value is 195,0 °C.

I don’t immediately see an easy way to solve this. The code splits into an array on spaces. If there is no space, it will pass the value to core as is and core accepts it, and then makes the . vs , mistake again.
I will see if there is anything I can do about this.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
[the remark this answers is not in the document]
lolodomo: NULL/UNDEF can be set by the binding itself.

My idea is only to be consistent with all other existing widgets, starting with the Text widget.
When the user defines a default color (through the valuecolor attribute), this color will be used when the item state is NULL or UNDEF.
So I am asking myself whether we should introduced something different here.

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: > And as you can never replace a value by a NULL or UNDEF from the input element, it will never be required.

This will never be triggered by the UI but can be triggered by a binding or a rule for example.
We can even try it with a rule that will set the item state to UNDEF for example.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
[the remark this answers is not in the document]
lolodomo: What I try to explain is that until now the user has a full control on label/value colors (depending on item(s) state). I think you are introducing a case where value color is hardcoded. Are we OK on that point ?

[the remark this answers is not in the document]
lolodomo: Maybe your option with hardcoded color is the best option in that case, I am just not sure.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: But then it will be done in the Java code that updates the widget snippet, not the javascript code.

lolodomo: When the page is already opened, the Java code is no more used, the JavaScript code (front-end) will then receive directly sitemap SSE events from the OH server and use them to update widget values and colors.

bundles/org.openhab.ui.basic/src/main/resources/snippets/input.html
mherwege: Done

bundles/org.openhab.ui.basic/web-src/_layout.scss
mherwege: Font and font size should be the same, only the weight (not bold). I did it on purpose, but will change. I will still keep the placeholder without bold, again to differentiate clearly.

lolodomo:
> I don’t immediately see an easy way to solve this. The code splits into an array on spaces.

It is easy I believe, you just have to add an additional test to check that the value you get as first value in your array only contains "valid" characters. Valid means for example no letters.

mueller-ma:
You can consider to make the unit a fixed text view after the input field. I'm not sure if it's a real use case to change the unit here.
This will also allow to display a numeric-only keyboard on mobile. Maybe add a flag later if users want to change the unit?

mherwege:
> You can consider to make the unit a fixed text view after the input field. I'm not sure if it's a real use case to change the unit here. This will also allow to display a numeric-only keyboard on mobile. Maybe add a flag later if users want to change the unit?

I was considering doing this through the inputHint parameter as suggested here: https://github.com/openhab/openhab-core/pull/3398#issuecomment-1449546243. If one would give an inputHint of number, a numeric keypad could be shown and the unit would not be changeable.
I want to include this in a follow-up PR, but have created a base PR in core to make this possible already, see here: https://github.com/openhab/openhab-core/pull/3418

mherwege:
> It is easy I believe, you just have to add an additional test to check that the value you get as first value in your array only contains "valid" characters. Valid means for example no letters.

That would mean rejecting number and unit without space in between. It seems OH does accept it without space in between. I went for another solution, that first isolates the number and treats everything afterwards as a unit.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: The labelcolor and iconcolor for the row is updated in the parent class (Control), so no need to do it here.
The label tag for the input element is the placeholder, and I prefer not to give that a specific color, but a neutral representation showing it is not set. If a rule changes it to UNDEF or NULL through a postUpdate, it should be the placeholder (label) that is shown, and not the value anymore, so in the neutral color. I tested this, but I did need to refresh the BasicUI to see the update to NULL or UNDEF. Otherwise it did not update. It looks like the SSE subscription did not pass through a state update to NULL or UNDEF. A sendCommand will not work as it does not accept NULL or UNDEF as valid commands. Other widgets have the same issue, proof in point, a text widget with the same value also did not update to NULL or UNDEF.

bundles/org.openhab.ui.basic/web-src/_theming.scss
mherwege: --border-color will do. It is used for the lines between the rows, but it does not harm if the underline in the input field has the same color.

bundles/org.openhab.ui.basic/web-src/_theming.scss
mherwege: Not sure what to do with the undef color. The best match might be to also use the border color for this, but I am not sure.

mherwege:
@lolodomo Ready for a new review.

One thing I am still struggling with is how to be sure about the item type. The InputRenderer class now sets it in the snippet from the state, but that only gives relevant info if the state is not UNDEF. If it is, it will default to Text and the number processing will not be done properly. I couldn't find a way to get the real item type from the ItemUIRegistry.
It works well as soon as you refresh the UI after updating from something UNDEF to a real value. But it will never work without a refresh. I also don't see anywhare in the javascript code a REST call to get item type. That would be a way to do it, but seems to me this Input Widget cannot be the only one that would have an interest in knowing the real item type.

mherwege:
> One thing I am still struggling with is how to be sure about the item type. The InputRenderer class now sets it in the snippet from the state, but that only gives relevant info if the state is not UNDEF. If it is, it will default to Text and the number processing will not be done properly. I couldn't find a way to get the real item type from the ItemUIRegistry.
> It works well as soon as you refresh the UI after updating from something UNDEF to a real value. But it will never work without a refresh. I also don't see anywhere in the javascript code a REST call to get item type. That would be a way to do it, but seems to me this Input Widget cannot be the only one that would have an interest in knowing the real item type.

OK, I think I have found the solution for this, see last commit.

mherwege:
> Note that it should be a good point to add an icon for "input" in our classic iconset. Could be in a separate PR.

Not sure if I will be any good at that. I am not an icon designer and I find it difficult to create something that fits in with what is already there.

lolodomo:
> Not sure if I will be any good at that. I am not an icon designer and I find it difficult to create something that fits in with what is already there.

Same for me. 
Maybe we should just an issue for that.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -0,0 +1,95 @@
...
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an implementation of the {@link WidgetRenderer} interface, which
+ * can produce HTML code for Input widgets.
+ *
+ * @author Mark Herwege - Initial contributor
```
lolodomo, line 40: contribution

If not, you have this:
```
[INFO] Individual report appended to summary report.
[WARNING] Code Analysis Tool has found:
 0 error(s)!
 1 warning(s)
 0 info(s)
[WARNING] org.openhab.ui.basic.internal.render.InputRenderer.java:[40]
First javadoc author should have "Initial contribution" contribution description.
```

bundles/org.openhab.ui.basic/web-src/_layout.scss
```diff
@@ -217,6 +217,25 @@
...
+			&__input,
+			&__label {
+				font-family: "Helvetica","Arial",sans-serif;
+				font-size: 14px;
+				text-align: right;
+			}
+			&__input {
+				font-weight: 700;
+			}
+			&__label {
```
lolodomo, line 230: I am not at all familiar with CSS. Why do you create several "blocks" for &__input et for &__label, rather than defining all properties in one "block" for each of them ?

lolodomo:
I will do a final review and tests before the end of the week-end.

Note for myself: Javascript to be reviewed again + check if helvetica is the standard font family used in Basic UI + test the behaviour when item state is changed to UNDEF.

bundles/org.openhab.ui.basic/web-src/_layout.scss
lolodomo: Ok, thank you for the explanation.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
mherwege: Changed

lolodomo:
As I have supposed by reading your code, the state update to NULL/UNDEF is not yet handled, this is confirmed by my tests.
I will work on a fix of your code (in the JavaScript part).

mherwege:
> I will work on a fix of your code (in the JavaScript part).

Thank you for looking into this. Can it be this is not handled anywhere? I also did not see an update of the text widget when changing state to NULL/UNDEF.

mherwege:
@lolodomo, changing the value to UNDEF externally (through a rule) is passed on. I did miss to properly treat this. However, sending NULL through a rule does not update the value at all. The reason is here: https://github.com/openhab/openhab-webui/blob/d25871d4500df96194ace0be2123ebf22325e6b4/bundles/org.openhab.ui.basic/web-src/smarthome.js#L2079

lolodomo:
> However, sending NULL through a rule does not update the value at all.

Yes. 
At the origin of openHAB, NULL was the initial state of an item and was not supposed to be used to update an item. 
Many years later, there was a discussion that sometimes UNDEF was less appropriate than NULL, and we started accepting bindings to use NULL as state update. I believe Basic UI was then not updated. 
I think it is out the scope of this PR to fix that. 
But at least UNDEF should be property supported.

mherwege:
@lolodomo The last commits I did should make it accept UNDEF and change the state accordingly.

lolodomo:
> @lolodomo The last commits I did should make it accept UNDEF and change the state accordingly.

Still not perfect. Your code can lead to display an empty string rather than "-".
When first opening the page:
image
Now opening the page while there was a non UNDEF state and then trigger a state update to UNDEF:
image

I am also surprised you do not play with your two elements like in the java code. That is certainly the source of this small issue.

I believe the fix is not too much difficult to do, I will try to push it to you today (I hope).

mherwege:
@lolodomo I pushed a commit that hopoefully fixes it, but did not test this yet. I am a bit short on time. I will try to do this by the end of the day. You can already have a look if you want.
In principle, the undefined value (placelholder) stays constant and is in another field in the html. So there would be no need to update. It would be possible to set the placeholder by default to -. But we then run into issues with an empty string being an allowed value, that should not show the placeholder. And as soon as the input field is set to an empty string, the placeholder is shown. So it is better to update the placeholder indeed.

mherwege:
@lolodomo Tested and working now.
It needed more than just updating the placeholder dynamically. It also required calling mdl state update and validation functions.

lolodomo:
> @lolodomo Tested and working now.
> It needed more than just updating the placeholder dynamically. It also required calling mdl state update and validation functions.

All my tests are not OK, thank you @mherwege .
I am going to have a final look on the code now.

bundles/org.openhab.ui.basic/web-src/_layout.scss
```diff
@@ -217,6 +217,25 @@
...
+				@media screen and (max-width: $layout-tablet-size-threshold) {
+					margin: 0 -12px;
+					padding: 0 12px 0 0;
+				}
+				margin: 0 -16px;
+				padding: 0 16px 0 0;
```
lolodomo, line 236: I am not expert but I think you should probably reverse the order, that is first set margin and padding and finally the `@media`

bundles/org.openhab.ui.basic/web-src/_theming.scss
lolodomo: Ok, let's do it as you propose.

bundles/org.openhab.ui.basic/web-src/_theming.scss
lolodomo: Let's keep it as you did.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -259,6 +260,25 @@
...
+		_t.wait = function() {
+			args = arguments;
+			timeoutId = setTimeout(function() {
+				callback.apply(null, args);
+				this.timeoutId = undefined;
+			}, waitingTime);
+		};
+
+		_t.cancel = function() {
+			clearTimeout(timeoutId);
```
lolodomo, line 278: You could check first that timeoutId is not undefined.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1503,6 +1523,91 @@
...
+						changeValue = changeValue.replace(/\./g, "").replace(",", ".");
+					}
+					if (unitValue.length > 1) {
+						changeValue = changeValue + " " + unitValue;
+					}
+				}
+			}
+
+			if (!changed) {
+				_t.setValuePrivate(lastValue);
```
lolodomo, line 1564: In case the item state was UNDEF and you enter an invalid number, this will lead to the field displaying an empty string rather than displaying `- °C` for example. Probably because the second parameter you do not provide is set to null.
To handle this case properly, this would require that you store the last value and + the last item state.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1503,6 +1523,91 @@
...
+			} else {
+				_t.parentNode.dispatchEvent(createEvent("control-change", {
+					item: _t.item,
+					value: changeValue
+				}));
+				// We don't know if the sent value is a valid command and will update the item state.
+				// If we don't receive an update in 1s, revert to the previous value.
+				_t.verify = new WaitingTimer(function() {
+					_t.setValuePrivate(lastValue);
```
lolodomo, line 1573: Very probably the same thing here.

lolodomo (changes requested):
Last comments

lolodomo:
@mherwege : I submitted a proposal to your repo to address my 3 last comments.
Ideally, if you could check my proposal before the end of the weekend, this would allow to merge this PR and include it in the coming milestone 2.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: In that case, the placeholder should show again as it online gets reset when something is received. The advantage of the placeholder is that it is a separate field, at the same place, only showing when the value is empty. The correction on empty string (removing the placeholder), is only relevant when receiving values.

bundles/org.openhab.ui.basic/web-src/_layout.scss
mherwege: Not an expert myself, but I believe the more specific definition takes precedence, whatever the sequence in the style sheet file. I did try playing with screen sizes in the browser debug tool.

lolodomo: But if you see how it is done in this file, you will see that this is always done in the reverse order.
Maybe it is not important, but just for consistency for the whole file, I would prefer the reverse order.

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: The problem is that you are resetting the input label when not expected, 
My solution avoids that. 
An even better solution would be to have a different method that only changes the input value.

lolodomo: I will update my proposal later today.

lolodomo: My new idea was not working.
In fact, I believe the code I proposed to you is just working perfectly.

mherwege:
@lolodomo I didn’t test, but having another look at the code, I understand the problem and I believe your proposed solution is correct. I included it.

lolodomo (approved):
LGTM, thank you

PR #1732, opened 2023-02-24
Developer sidebar: Extend search & Sort search results

florian-h05:
Extends the developer sidebar search for:

- Items:
  - metadata namespaces (requires exact match)
  - tags (requires exact match)
- rules/scenes/scripts:
  - itemName & thingUID of triggers, actions, conditions
  - script code (e.g. JavaScript, Rules DSL, Blockly)
  - script MIME types (requires exact match)
- Pages:
  -  label

All searches are non case-intensive unless other stated.

Search results are now alphabetically sorted and rules, scenes and scripts divided into their own sections.
Normal searches on settings pages and the Item detail page show a hint for the developer sidebar search if the screen is width enough.

The required screen size for the usage of the developer sidebar has been reduced from 1280 to 1000 pixels to allows usage on smaller screens or inside smaller windows.

Related to https://community.openhab.org/t/openhab-4-0-wishlist/142388/104?u=florian-h05.

### DIFF
--- a/bundles/org.openhab.ui/web/package-lock.json
+++ b/bundles/org.openhab.ui/web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "openhab",
-  "version": "3.0.0",
+  "version": "4.0.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "openhab",
-      "version": "3.0.0",
+      "version": "4.0.0",
       "license": "EPL-2.0",
       "dependencies": {
         "@blockly/field-slider": "^4.0.3",
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -132,7 +132,7 @@
     <!-- <f7-view url="/panel-right/"></f7-view> -->
     </f7-panel>
 
-    <f7-panel v-if="showDeveloperSidebar" right :visible-breakpoint="1000" resizable>
+    <f7-panel v-if="showDeveloperSidebar" right :visible-breakpoint="1280" resizable>
       <developer-sidebar />
     </f7-panel>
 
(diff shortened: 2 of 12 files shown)


florian-h05:
@ghys This is ready for review, but please wait with merging, I'll probably add searching for script MIME types as well.

florian-h05:
@ghys Finally this is ready for review and merging.

florian-h05:
@rkoshak
I’ve received your ping for the community discussion. Currently, a search inside UI pages is not possible because I had no idea how to efficiently search the UI page object, which can have varying structure. However, I just had the idea to stringify the object and perform a search against this string.

florian-h05:
@ghys This is finally ready for review.

ghys:
> The required screen size for the usage of the developer sidebar has been reduced from 1280 to 1000 pixels to allows usage on smaller screens or inside smaller windows.

This 1280px minimum width was actually carefully chosen :slightly_smiling_face:  - it allows the default view  for the UI once you increase the width to cohabit with both the left sidebar and the developer sidebar, and never experience an horizontal scrollbar (which are ugly and to be avoided!)

```css
@media (min-width: 1024px) {
  .block-narrow .block, .block-narrow .col, .block-narrow .row {
      width: 720px;
      max-width: 100%;
  }
```

florian-h05:
> This 1280px minimum width was actually carefully chosen slightly_smiling_face 

Okay, I will revert :+1: 

> which might result in both a significant performance hit

On my PC (Ryzen 7, so very powerful) as well on my iPad Air 2019, search results are displayed almost instantenous for the first typed character, even though the data has to load over WiFi. After the first charactater is typed, there is no loading time.

> search results which might be too broad when you start typing - for instance, when you type "a" you will basically match anything (this was arguably also the case before when matching only UIDs and labels but less so when you add some letters?)

This is still the case, when I start typing `f`, I get hundreds of results (to be fair, I have about 900 Items), but adding a second letter `Fl` reduces results to just 100, which in this case is perfectly fare because I have 85 Items with `Florian` in the Item name. Given that search results are sorted after category and only 5 each are shown by default, I don't see a problem here.

ghys (approved):
Alright, thank you!

PR #1733, opened 2023-02-25
[basicui] Support primary and secondary as value for label/value colors

### LINKED ISSUE #1716: Allow using theme colors as label/value colors
Hi!
For the Android Client primary and secondary will be added to the [the list of known colors](https://github.com/openhab/openhab-android/blob/main/mobile/src/main/res/values/arrays.xml#L93-L113).
So that the theme-colors can be used as label or value color as well.
https://github.com/openhab/openhab-android/issues/3248

Maybe this would be an idea for the webUI as well? Or at least to check that this option is not causing any issue?

Thanks!
Tim

lolodomo:
Closes #1716


lolodomo:
Considering the following sitemap:
```
sitemap test2 label="Tests"
{
        Frame {
                Default item=CurrentTemp
                Default item=CurrentTemp valuecolor=[<20="orange","primary"] labelcolor=[<20="red","secondary"]
                Default item=CurrentTemp valuecolor=[<20="red","secondary"] labelcolor=[<20="orange","primary"]
        }
}
```
the result is the following:
image
image

kaikreuzer (approved):

PR #1734, opened 2023-02-25
Configuration & iPhone style fixes for the `options` widget action

### LINKED ISSUE #1210: oh-button actionOptions formatting
When spaces are used following the commas in the list of options for oh-button actionOptions, all but the first option may fail when selected due to embedded spaces in the POSTed data. For example, in "actionOptions: 0=Off, 25=25%, 50=50%, 75=75%, 100=Full" "Off" will succeed while the others will fail. Changing to "actionOptions: 0=Off,25=25%,50=50%,75=75%,100=Full" will work.

The problem is that "split(',')" retains the embedded spaces (https://github.com/openhab/openhab-webui/blob/0bcbd8c8ad81011b37a923f82235b25b3c913ea9/bundles/org.openhab.ui/web/src/components/widgets/widget-actions.js#L82):

`const actionsPromise = new Promise((resolve, reject) => { if (actionCommandOptions && typeof actionCommandOptions === 'string') { resolve(actionCommandOptions.split(',').map((o) => { const parts = o.split('=')`
Adding trim() solves the problem:
`const actionsPromise = new Promise((resolve, reject) => { if (actionCommandOptions && typeof actionCommandOptions === 'string') { resolve(actionCommandOptions.split(',').map((o) => { const parts = o.trim().split('=')`

florian-h05:
- Fixes the style of the actions popup on iPhones in the openHAB iOS app. It did not respect the safe-area-inlets and therefore large action popups possibly placed behing the notch or outside of the display.
- Fixes #1210: trim whitespaces when parsing the `actionOptions` to avoid having whitespaces in commands sent to the server.

### DIFF
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -141,13 +141,8 @@ html
      --f7-bars-link-color #ff3b30
 
 // Fix actions-modal/popup is placed behind the notch or outside of the screen on iPhone in openHAB iOS app
-@media (orientation: portrait)
-  .ios .actions-modal
-    max-height: calc(100% - env(safe-area-inset-top, 0px) - env(safe-area-inset-bottom, 0px))
-
-@media (orientation: landscape)
-  .ios .actions-modal
-    max-height: calc(100% - env(safe-area-inset-left, 0px))
+.device-ios .actions-modal
+  max-height: calc(100% - var(--f7-safe-area-top) - var(--f7-safe-area-bottom))
 
 .md .contextual-toolbar
   --f7-theme-color #474747
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -140,6 +140,14 @@ html
    .delete
      --f7-bars-link-color #ff3b30
 
+@media (orientation: portrait)
+  .ios .actions-modal
+    max-height: calc(100% - env(safe-area-inset-top, 0px) - env(safe-area-inset-bottom, 0px))
+
+@media (orientation: landscape)
+  .ios .actions-modal
+    max-height: calc(100% - env(safe-area-inset-left, 0px))
+
 .md .contextual-toolbar
   --f7-theme-color #474747
   --f7-bars-bg-color #474747
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -140,6 +140,7 @@ html
    .delete
      --f7-bars-link-color #ff3b30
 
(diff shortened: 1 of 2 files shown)


florian-h05:
Here are some screenshots from the openHAB iOS app on iPhone XS:

### Portrait mode:

Before:


After:


### Landscape mode:

Before:


After:

florian-h05:
@ghys Can you please review this PR with some „priority“ Yannick? Should be a fast review, only a few changed lines, and it would be nice to be able to use that actions menu in the iOS app.

bundles/org.openhab.ui/web/src/css/app.styl
```diff
@@ -140,6 +140,15 @@ html
    .delete
      --f7-bars-link-color #ff3b30
 
+// Fix actions-modal/popup is placed behind the notch or outside of the screen on iPhone in openHAB iOS app
+@media (orientation: portrait)
+  .ios .actions-modal
```
ghys, line 145: The class `.ios` denotes the ios _theme_, you probably want this applied for iOS _devices_ regardless of the theme. So that would be `.device-ios` instead.

bundles/org.openhab.ui/web/src/css/app.styl
```diff
@@ -140,6 +140,15 @@ html
...
+    max-height: calc(100% - env(safe-area-inset-top, 0px) - env(safe-area-inset-bottom, 0px))
```
ghys, line 146: There are f7 CSS variables (https://v5.framework7.io/docs/css-variables#app-core) for that: instead of
`env(safe-area-inset-*)`
I would prefer
`var(--f7-safe-area-*)`

bundles/org.openhab.ui/web/src/css/app.styl
```diff
@@ -140,6 +140,15 @@ html
...
+
+@media (orientation: landscape)
+  .ios .actions-modal
+    max-height: calc(100% - env(safe-area-inset-left, 0px))
```
ghys, line 150: Are you sure there's a need for two different implementations for portrait & landscape?
And it doesn't seem right to use the left safe area to configure a height (when you rotate into landscape mode the top/bottom safe areas should automatically reflect the new orientation, see https://webkit.org/blog/7929/designing-websites-for-iphone-x/)

ghys (changes requested):

bundles/org.openhab.ui/web/src/css/app.styl
florian-h05: Thanks for the hint. I've must overread that.

florian-h05:
@ghys I've addressed your review and tested it on my real hardware, looks even better now!

florian-h05:
I just need to rebase to get the CI build working (Karaf upgrade...)

ghys (approved):

PR #1735, opened 2023-02-27
Support configuring Input element for sitemap model

mherwege:
https://github.com/openhab/openhab-core/pull/3398 added an Input element to sitemaps.
A PR for BasicUI to support input elements is in progress (https://github.com/openhab/openhab-webui/pull/1729).


florian-h05 (approved):
LGTM, thanks!

Should I merge now or do you prefer waiting for #1729 to get merged?

mherwege:
Configuring it is already supported in text files, although no UI can use it yet. So I don’t think we have to wait for merging this. I am hopeful BasicUI will support it soon.

ghys:
I'm late here but it might "unbalance" this popup here:
image

Maybe we can remove "List", what do those even do?

florian-h05:
> I'm late to this party but it might "unbalance" this popup here:

It indeed does, but I don't see a problem in this fact.

> Maybe we can remove "List", what do those even do?

Seems like they display a list by splitting the Item state using a user-configured separator.

PR #1736, opened 2023-02-27
Improvements for Eclipse provisioned add-ons

wborn:
This improves the UI when working with add-ons that are provisioned using Eclipse.

* Show these add-ons in the "openHAB Distribution" section
* Show in the add-on details when add-ons are provisioned with Eclipse
* Show the add-on documentation and issue links when add-ons are provisioned with Eclipse
* Hide Install/Uninstall buttons when add-ons are provisioned with Eclipse

Related to openhab/openhab-core#3406

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/addon-store.js
+++ b/bundles/org.openhab.ui/web/src/assets/addon-store.js
@@ -25,6 +25,7 @@ export const Formats = {
   'json_download_url': 'Linked JSON File',
   'jar_download_url': 'Linked JAR file',
   'kar_download_url': 'Linked KAR file',
+  'eclipse': 'Eclipse',
   'karaf': 'Karaf'
 }
 
--- a/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
@@ -5,7 +5,7 @@
         <div>{{ headline || autoHeadline || "&nbsp;" }}</div>
       </div>
       <div class="addon-card-title">
-        <div class="addon-card-title-after">
+        <div v-if="showInstallActions" class="addon-card-title-after">
           <f7-preloader v-if="addon.pending" color="blue" />
           <f7-button v-else-if="addon.installed" class="install-button prevent-active-state-propagation" text="Remove"
                      color="red" round small @click="buttonClicked" />
@@ -146,6 +146,10 @@ export default {
       let docsBranch = 'final'
       if (this.$store.state.runtimeInfo.buildString === 'Release Build') docsBranch = 'final-stable'
       return `https://raw.githubusercontent.com/openhab/openhab-docs/${docsBranch}/images/addons/${this.addon.id.substring(this.addon.id.indexOf('-') + 1)}.png`
+    },
+    showInstallActions () {
+      let splitted = this.addon.uid.split(':')
+      return splitted.length < 2 || splitted[0] !== 'eclipse'
     }
   },
   mounted () {
--- a/bundles/org.openhab.ui/web/src/components/addons/addon-info-table.vue
+++ b/bundles/org.openhab.ui/web/src/components/addons/addon-info-table.vue
@@ -46,7 +46,7 @@ export default {
       let sourceName = 'openHAB Distribution'
       if (source === 'marketplace') {
         sourceName = 'Community Marketplace'
-      } else if (source !== 'karaf') {
+      } else if (source !== 'eclipse' && source !== 'karaf') {
         sourceName = '3rd Party (' + source + ')'
       }
       info.push({
@@ -82,8 +82,12 @@ export default {
         value: ContentTypes[this.addon.contentType] || this.addon.contentType
       })
 
-      let format = Formats.karaf
-      if (source !== 'karaf' && Object.keys(this.addon.properties).length > 0) {
+      let format
+      if (source === 'eclipse') {
+        format = Formats.eclipse
+      } else if (source === 'karaf') {
+        format = Formats.karaf
+      } else if (Object.keys(this.addon.properties).length > 0) {
         for (const property in this.addon.properties) {
           if (Formats[property]) format = Formats[property]
         }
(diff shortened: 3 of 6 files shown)


ghys:
Thanks, LGTM code-wise, but I just have to wonder:

What is the point of adding Eclipse-provided add-ons to an add-on store if it's just for listing them and you can't do anything about them?

wborn:
It helps to see that the add-on is available and it is also used to get to the add-on specific configuration options.

ghys (approved):
Right, thanks!

mvalla:
this is a useful improvement.
However it look like the binding gets duplicated in the "Bindings" view:
bindings

looking into details they appear differently, first one:

1st

while if I click the second one its details remain uncomplete:

2nd

Two versions of the same bindings can also be found when starting a InBox search:

inbox

What shall I look to understand while the same binding being developed in Eclipse gets duplicated?

mvalla:
also it looks like the `eclipse:` instance does not get translated, while the other gets the localization (italian)

wborn:
Works fine for me @mvalla. I do not have any `jar:` based add-on entries when adding that add-on as usual:

settings

choose

mvalla:
ok @wborn thanks for the check.
But then there should be some jar of the same binding poppoing out from somewhere from the IDE environment ... but from where?
I am on Eclipse 2022.03, Windows 10.
Where should I check?

wborn:
The JAR add-ons are provided by the [JarFileAddonService](https://github.com/openhab/openhab-core/blob/850a9239de078a1b52cd489fd56e7ebc59e74d5d/bundles/org.openhab.core.addon/src/main/java/org/openhab/core/addon/internal/JarFileAddonService.java). So you could set a break point in it, or add some more logging to it, to see why it thinks some bundles are relevant. I'm using the same Eclipse version but on Ubuntu.

mvalla:
some more context:
- `JarFileAddonService:addingBundle()`, gets called with the jar built by the Eclipse project of the local openhab-addons repo :
    - `file:/C:/   (...path to...)  /openhab/git/openhab-addons/bundles/org.openhab.binding.openwebnet/target/org.openhab.binding.openwebnet-4.0.0-SNAPSHOT.jar`
- only one openwebnet biding is present when calling `lb` on Apache Felix:
```
Module               osgi.identity; osgi.identity="org.openhab.binding.openwebnet"; type="osgi.bundle"; version:Version="4.0.0.202303261754" [id=168]
Location             file:/C:/ (...path to...)  /openhab/git/openhab-addons/bundles/org.openhab.binding.openwebnet/target/org.openhab.binding.openwebnet-4.0.0-SNAPSHOT.jar
State                32
Version              4.0.0.202303261754
Bundle                 168|Active     |    1|org.openhab.binding.openwebnet (4.0.0.202303261754)
LastModified         1679853615547
Headers              [Service-Component=OSGI-INF/org.openhab.binding.openwebnet.internal.OpenWebNetHandlerFactory.xml,OSGI-INF/org.openhab.binding.openwebnet.internal.discovery.BusGatewayUpnpDiscovery.xml,OSGI-INF/org.openhab.binding.openwebnet.internal.discovery.UsbGatewayDiscoveryService.xml, Bundle-License="Eclipse Public License 2.0";link="https://www.eclipse.org/legal/epl-2.0/", 
[....] 
Automatic-Module-Name=org.openhab.binding.openwebnet, Bundle-ManifestVersion=2, Bundle-SymbolicName=org.openhab.binding.openwebnet, Bundle-Version=4.0.0.202303261754, Bundle-Developers=openhab;email="info@openhab.org";name=openHAB;organization=openHAB;organizationUrl="https://www.openHAB.org", Built-By=mvalla, Require-Capability=osgi.service;filter:="(objectClass=org.openhab.core.io.transport.serial.SerialPortManager)";effective:=active,osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.5.0)(!(version>=2.0.0)))",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))", Tool=Bnd-6.4.0.202211291949]
BundleContext        org.eclipse.osgi.internal.framework.BundleContextImpl@1c4053aa
SymbolicName         org.openhab.binding.openwebnet
BundleId             168
ServicesInUse        [ThingTypeRegistry, ConfigDescriptionRegistry, SerialPortManager]
RegisteredServices   [ThingHandlerFactory, UpnpDiscoveryParticipant, ConfigStatusProvider, DiscoveryService, DiscoveryService]
```
- if i call `http://localhost:8080/rest/addons` i get:
```
{
    "uid": "jar:binding-openwebnet",
    "id": "openwebnet",
    "label": "Binding OpenWebNet (BTicino/Legrand)",
    "version": "4.0.0.202303261754",
    "compatible": true,
[....]   
  }
```
- I am develping in Eclipse the openwebnet addon which is alreay part of the official disitribution. Is this case taken into consideration when adding the Eclipse Add-on Service?
 It looks like in this case the addon is already present as jar and the REST/UI shows it (under "Other Add-ons" section) , while the Eclipse Add-on Service provides another view but at the end it's the same addon. 
Maybe the "Other Add-ons" section in this case should be removed? I think these caseas are possibile: 
    1. developing in Eclipse an addon which is not official yet
    1. developing in Eclipse an addon which is already part of the official distribution (my case)
    1. provisioned via Eclipse an addon which is already part of the official distribution

mvalla:
the same duplication problem occurs if I provision via Eclipse an official addon for which I do not have a project configured in Eclipse, for example Hue Binding:

hue

wborn:
> Location             file:/C:/ (...path to...)  /openhab/git/openhab-addons/bundles/org.openhab.binding.openwebnet/target/org.openhab.binding.openwebnet-4.0.0-SNAPSHOT.jar

I have a different bundle location which explains why I don't see these jar add-ons:

`reference:file:/home/wouter/git/openhab/openhab-addons/bundles/org.openhab.binding.openwebnet/target/org.openhab.binding.openwebnet-4.0.0-SNAPSHOT.jar`

wborn:
Perhaps it doesn't use these reference locations on Windows because it doesn't work well, see:

https://bnd.bndtools.org/chapters/300-launching.html

>Windows and reference URLs do not work well because Windows keeps files locked.

mvalla:
In any case the view of the jar add on instance is broken, while the Eclipse one has documentation and is well rendered so the latter is the only one to be shown in UI.

But why the Eclipse add-on Service addons do not show up in the lb command and in the /rest/addons call? If this was the case then UI could choose the Eclipse one in case of same addon id…
Should I open an issue?
Which repo?

wborn:
You can create an issue for it in openhab-core which has the code of all add-on services.

PR #1739, opened 2023-02-28
Scene editor: Fix triggers/conditions & Remove enable/disable and status

florian-h05:
This:

- Fixes an issue where adding triggers and conditions to scenes using the YAML code tab was not possible (reported in https://github.com/openhab/openhab-webui/pull/1662#issuecomment-1435019).
- Fixes an issue where `null` was entered as triggers or conditions and saving the scene failed.
- Fixes SSE event subscription for rule status inside scene editor.
- Removes the disable and enable actions from the scene the scenes list as there are no status badges.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
@@ -21,10 +21,10 @@
       <f7-link color="red" v-show="selectedItems.length" v-if="!$theme.md" class="delete" icon-ios="f7:trash" icon-aurora="f7:trash" @click="removeSelected">
         Remove {{ selectedItems.length }}
       </f7-link>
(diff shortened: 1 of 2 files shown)


ghys:
>    Removes the disable and enable actions from the scene editor and the scenes list.
>    Removes the rule status from the scene editor.

> Two words about the removals of enable/disable and the rule status: these were not working due to missing SSE update subscription, and as scenes are stateless they aren't needed.

I'll have to disagree here, I'm not in favor of withholding information or crippling functionality that exists. Scenes are still rules. Suppose you found a way to disable a scene (for example with the API explorer or from a rule) or it's in error somehow,  you will have a hard time understanding what's going on or re-enabling it. Besides I think it's valuable to be able to disable a scene if for one reason or another you don't want it to run, including manually (with a button) or from another rule.

florian-h05:
Okay, I understand your points. I will revert the removals and instead fix the SSE updates.

florian-h05:
@ghys 
I reverted the removal of the rule status and enabled/disable actions and fixed SSE inside scene editor.
I decided to keep enable/disabled removed from scenes list, as there is no status badge which displays the state.

From my side this is finished now.

PR #1740, opened 2023-03-01
Update Material icons to 6.7.0 & Remove tff fonts

### LINKED ISSUE #1708: Some missing Material Icons
In my openHAB 3.4.2 UI, some material icons are not working.

This is my YAML for a oh-label-card widget:

```yaml
component: oh-label-card
config:
  trendItem: SenecHome_BatteryPower
  title: Solarspeicher
  footer: =items.SenecHome_BatteryCurrent.state
  item: SenecHome_BatteryFuel
  icon: material:solar_power
  taphold_action: navigate
  taphold_actionPage: page:PV_Speicher
  action: navigate
  actionPage: page:PV_Speicher
  actionPageTransition: f7-circle
```

The rendered result looks like this:

image

And the HTML like this:

image

The expected icon for `solar_power` would be:

> solar_power_FILL0_wght400_GRAD0_opsz48

(See https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Asolar_power%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048)

Also weird, that the incorrect rendered plug icon is misplaced.

For other oh-label-cards this works fine:
image

florian-h05:
Fixes #1708.
Material icon source: https://github.com/jossef/material-design-icons-iconfont/tree/v6.7.0/dist/fonts.

This:

- Removes tff fonts as all modern browsers support woff (e.g. Firefox added support in 2010) and woff2 (see https://www.w3schools.com/css/css3_fonts.asp), which both have better compression.
- Sets CSS to temporarily show system font while loading fonts (even though we talk about icons here). Recommendation from Chrome Lighthouse audit, see https://developer.chrome.com/docs/lighthouse/performance/font-display/?utm_source=lighthouse&utm_medium=devtools#how-to-avoid-showing-invisible-text.


lolodomo:
> Removes tff fonts as all modern browsers support woff (e.g. Firefox added support in 2010) and woff2 

I have no particular knowledge about fonts but had some readings. It looks like you are fully right.
That being said, TrueType Fonts format is required on old Android phones (before Android 4.4.4) so maybe we could keep it ?

I was curious to see what formats for material design are packaged in BasicUI. In fact, we are packaging other formats than TTF woff and woff2 (6 different formats), some formats that will never be downloaded because not mentioned in `@font-face`. I am going to remove them. The SVG format is the biggest of all these formats.

florian-h05:
> TrueType Fonts format is required on old Android phones (before Android 4.4.4) so maybe we could keep it ?

I am not sure if Android 4.4.4 is official supported by MainUI: 

https://github.com/openhab/openhab-webui/blob/b70d959d7c0982439703bfe69397abbb340ee16a/bundles/org.openhab.ui/web/package.json#L54-L55

lolodomo:
If minimum supported by MainUI is Android 5, removing ttf format should be safe, I believe.

ghys (approved):
> If minimum supported by MainUI is Android 5, removing ttf format should be safe, I believe.

Absolutely, I believe we have to make a cutoff at some point, if you're using some hardware that's still using Android KitKat then you'll have to make an upgrade at some point. Besides I don't believe the main UI will perform very well, so on this kind of hardware you maybe should consider sticking to Basic UI or the app.

ghys (approved):

florian-h05:
@ghys Can we merge this?

Issue #1026, opened 2021-04-28
[Main UI] Widget expression tester shows `undefined` as empty string


matthijskooijman:
## The problem

When you use the widget expression tester and enter an expression that results in an undefined value, an empty result is shown.

## Expected behavior

It would be more useful if the tester would show `undefined`, making it more clear to the user that their expression is valid, but returns undefined (just showing nothing at all is vague, I initially suspect I had a syntax error or some other hidden problem).

## Steps to reproduce

![image](https://user-images.githubusercontent.com/194491/116379939-cbaeaf00-a813-11eb-990d-8e1130b9eb10.png)

You can confirm that the value is undefined by forcing it to string:

![image](https://user-images.githubusercontent.com/194491/116380022-dff2ac00-a813-11eb-8eaf-81fcfc17d114.png)

## Your environment

```yaml
runtimeInfo:
  version: 3.0.2
  buildString: Release Build
locale: en-NL
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 11.0.11
  javaVendor: Azul Systems, Inc.
  javaVendorVersion: Zulu11.48+21-CA
  osName: Linux
  osVersion: 5.10.17-v7+
  osArchitecture: arm
  availableProcessors: 4
  freeMemory: 49622120
  totalMemory: 194838528
bindings:
  - chromecast
  - dsmr
  - ipp
  - mqtt
  - spotify
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: true
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    pixelRatio: 1
    prefersColorScheme: dark
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 4
    language: en-US
    languages:
      - en-US
      - en
    onLine: true
    platform: Linux x86_64
  screen:
    width: 1280
    height: 1024
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: dark
    filled: true
[...]

PR #1747, opened 2023-03-02
Fix undefined in widget expression evaluation

florian-h05:
Regression from #1696.

The widget expression evaluator returns undefined a string to make it appear in the widget expression tester etc. 
This has unattended side effects because the UI internally relies on undefined checks.

This fixes the regression by reverting the change and instead extracting the widget-expression tester to its own component, which will display `undefined` as string.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
+++ b/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
@@ -187,24 +187,24 @@
           <f7-list>
             <f7-list-item divider title="Things" />
             <f7-list-button href="/settings/things/add" color="blue" :animate="false">
-              Add Thing
+              Add thing
             </f7-list-button>
             <f7-list-button @click="quickAddThing" color="blue">
-              Add Thing (quick)
+              Add thing (quick)
             </f7-list-button>
             <f7-list-button href="/settings/things/inbox" color="blue" :animate="false">
               Inbox
             </f7-list-button>
             <f7-list-item divider title="Items" />
             <f7-list-button href="/settings/items/add" color="blue" :animate="false">
-              Create Item
+              Create item
             </f7-list-button>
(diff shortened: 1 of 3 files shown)


florian-h05:
@ghys This is critical as it fixes a regression.

bundles/org.openhab.ui/web/src/components/widgets/generic-widget-component.vue
```diff
@@ -11,10 +11,10 @@
   <generic-widget-component v-else-if="componentType && componentType.startsWith('widget:') && visible" :context="childWidgetContext()" @command="onCommand" />
   <component v-else-if="componentType && componentType.startsWith('oh-') && visible" :is="componentType" :context="context" @command="onCommand" />
   <div v-else-if="componentType && componentType === 'Label' && visible" :class="config.class" :style="config.style">
-    {{ config.text }}
+    {{ (config.text) ? config.text : 'undefined' }}
```
ghys, line 14: At first sight this would transform all [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) expressions (i.e. also `false`, `0`, `null`, empty strings etc.) into `undefined`  which is surely not what we want.

Honestly I don't understand what #1696 was supposed to be fixing, #1026 was about the widget expression tester so that's probably where a fix is needed, no need to make global changes to handle this case I believe.

ghys (changes requested):

bundles/org.openhab.ui/web/src/components/widgets/generic-widget-component.vue
florian-h05: Ooops, you are right.

#1696 tried to fix #1026, but it was indeed the wrong way. I've just rechecked the widget expression tester, it is possible to make the required change in the developer-sidebar. I'll fix that.

ghys (approved):
Yes! Thanks for the fix (and for isolating the tester into its component!).

PR #1753, opened 2023-03-04
[BasicUI] Fix label color initialization for a color widget

### LINKED ISSUE #1752: [BasicUI] Label color for color widget is wrong when opening the page
Using these items
```
String TestString "Test string [%s]"
Color TestColor "Test color [%s]"
```
and this sitemap
```
sitemap test6 label="Tests"
{
        Frame {
                Default item=TestString
                Default item=TestColor labelcolor=[TestString=="test"="yellow", "orange"]
        }
}
```
the color of the label is not set when you open the page (it should be yellow rather than white when TestString is equal to "test":
image
While the page is opened, it is then properly updated:
image
The bug is only the initial building of the page.

lolodomo:
Fix #1752


lolodomo:
Here is the result now when opening the page:
image

PR #1755, opened 2023-03-04
[BasicUI] Remove support for List element

lolodomo:
Related to openhab/openhab-android#3285

This sitemap widget is not documented.

It is not implemented neither in Android app nor iOS app.

It was partially implemented in BasicUI but does not work well (it breaks the layout).

Without any opened issue, we can be relatively sure that noone is using it.


kaikreuzer (approved):
I guess you are right that this isn't used.
IIRC, I introduced this 13 years ago for being able to render log entries and did that for some demos. But we have better ways nowadays and BasicUI isn't meant for that.

Shall we also remove the widget from the sitemap DSL syntax?

florian-h05:
Since `List` is not supported by any BasicUI client (web, iOS, Android) it IMO should be removed from the Sitemap DSL syntax as well as from MainUI:

- https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/pages/settings/pages/sitemap/sitemap-edit.vue
- https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley

lolodomo:
> Shall we also remove the widget from the sitemap DSL syntax?

Yes, I will do it tomorrow.

lolodomo:
> Since `List` is not supported by any BasicUI client (web, iOS, Android) it IMO should be removed from the Sitemap DSL syntax as well as from MainUI:

Could you please manage that?
Once removed from MainUI, I will remove it from sitemap syntax.

Issue #1365, opened 2021-02-12
Persistence Aggregate group item not showing 'Analyze' option


baku104788:
I've got RRD4j installed as default persistence and with it I have some group switches that I'd like to track / be able to analyze history of.  I've updated my .persist file to include the group names but the analyze option is not available. Checking the rest API I get an empty list when I request a summary if items within the RRD4j persistence service however if I request data for the item itself I get the history data.  I'm running 3.1.0M1 and RRD4j is my default persistence service though I also have JDBC configured.  My persist file is included below. 

Strategies {
        everyMinute : "0 * * * * ?" 
        everyHour : "0 0 * * * ?"
        everyDay  : "0 0 0 * * ?"
       default = everyMinute,everyChange,restoreOnStartup
}

Items {
    * : strategy=everyMinute,everyChange,restoreOnStartup
    tracker_occupancy : strategy = everyMinute,everyChange
    presence_elle : strategy = everyMinute,everyChange
    presence_jared : strategy = everyMinute,everyChange
}

ghys:
Could probably be solved here...
https://github.com/openhab/openhab-webui/blob/b7500b55519f17798b5ff2676496a1762092a019/bundles/org.openhab.ui/web/src/pages/analyzer/analyzer.vue#L352

```js
if ((item.type.indexOf('Number') >= 0 ...
```
so that items of type `Group:Number` wouldn't be considered "discrete".

BigGeorgeTx:
Seeing the same issue using 3.4.0 SNAPSHOT 3033

blue-kaleidoscope:
Yes, can confirm. I would also be interested in a solution.

miccgn:
I can confirm this with OH 3.3. and MariaDB as the only persistence service. Values for group items show up in MariaDB tables, but won't be charted.

kristofejro:
On openHAB 3.4.0.M3 it is also still present.

PR #1758, opened 2023-03-05
Fix missing diagram lines for groups

chrros95:
Like others,  I've noted that the analyzer will not show the data if the source is a group that derives its state from its members (#1365 ). 
With this PR the analyzer will also respect groups of numbers or dimmers.

Thanks to @ghys for the hint on where to look at.


florian-h05:
Can you please rebase towards main to get rid of the one commit in the change list that doesn't belong to you?

PR #1765, opened 2023-03-05
Fixes WebAudio sink not playing on Safari

### LINKED ISSUE #1717: Web Audio Sink on iOS devices not working
## The problem
Main UI within Safari on iPad / iPhone does not play the sound using Web Audio Sink.

On my iPad Pro with iOS 16.3 and on my iPhone with iOS 16.3 audio sink is not working. Same on my other iPad (non por model).
I have Main UI running and activated “Enable Web Audio sink support” but no sound is played.
I have tried it with a HABPanel but also here no sound is played.
I upgraded to OH 3.4.1 to use this “Enable Web Audio sink support” feature in the Main UI.

When testing the same (Main UI and enable audio sink) on a Windows 10 machine within Chrome brwoser: the sound is played successfully.

As I have no Mac I cannot test if the issue is related to iOS or to Safari.

I tested it via a virtual switch item that triggers a rule that plays the sound "Barking" and uses sink "Web Audio" and volume = 100

## Expected behavior
I should hear the sound

## Your environment
```yaml
runtimeInfo:
  version: 3.4.1
  buildString: Release Build
locale: en-DE
systemInfo:
  configFolder: /etc/openhab
  logFolder: /var/log/openhab
  javaVersion: 11.0.11
  javaVendor: Azul Systems, Inc.
  javaVendorVersion: Zulu11.48+21-CA
  osName: Linux
  osVersion: 5.10.103-v7+
  osArchitecture: arm
  availableProcessors: 4
  freeMemory: 114643720
  totalMemory: 324403200
  startLevel: 100
```

## Browser console
On my iPad I activated in the Developer Tools in the Debug tab the SSE Streaming. If the audio is pushed I saw in the streaming list the record

```txt
openhab/webaudio/playurl PlayURLEvent
“/audio/…a uuid…”
```

digitaldan:
Fixes #1717.

Safari (and Chrome possibly as well) requires a user interaction after a web audio stream has been requested, which means audio will not play until at least one stream has been started and then the user clicks somewhere in the web app.
This implements a workaround (https://www.mattmontag.com/web/unlock-web-audio-in-safari-for-ios-and-macos) which unlocks and stores the AudioContext globally, so that after the first stream only a single interaction is required for the lifetime of that browser session.


digitaldan:
This is my quick and dirty workaround for Safari being difficult.

florian-h05 (approved):
LGTM, thank you very much!!

I have tested this, works fine with openHAB's `barking.mp3`.
Does not work with the `doorbell.mp3`, my BrowserStack iOS debugger (thx @ghys) says `Unhandled Promise Rejection: EncodingError: Decoding failed`, but this is related to the encoding and we can't help there.

@digitaldan Is this ready for merging? It still has WIP.

digitaldan:
Let me clean it up a bit, i believe there are some formatting violations, plus i want to make the variable naming more consistent.   I kinda just threw it together to be honest ;-)

florian-h05:
I have fixed the linting errors, so you need to update before cleaning up.

wborn:
When it's fixed and works, it would also be nice when the same changes are applied (in additional PRs) to HABPanel and Basic UI. Those UIs use very similar code for handling webaudio. 🙂

florian-h05:
@digitaldan
How does the clean-up go? Can you please rebase if you work on this, I think Jenkins fails due to some core changes and this outdated branch.

digitaldan:
Unfortunately i'm having some trouble getting this working now on safari, i'll play a little more this afternoon.

digitaldan:
Ok, all cleaned up.  My Safari issues were my own (helps to point Safari at your dev server when testing).  I'll look at updating the other UI's in separate PR.

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -624,29 +625,35 @@ export default {
...
+          if (typeof (window.AudioContext) !== 'undefined') {
+            this.audioContext = new AudioContext()
+            unlockAudioContext(this.audioContext)
+          }
         }
         console.log('Playing audio URL: ' + audioUrl)
         this.$oh.api.getPlain(audioUrl, '', '*/*', 'arraybuffer').then((data) => {
-          context.decodeAudioData(data, function (buffer) {
-            let source = context.createBufferSource()
+          this.audioContext.decodeAudioData(data).then((buffer) => {
```
florian-h05, line 638: When using the Promise based version of `decodeAudioDate`, we should be aware that the Promise version was added to Safari just in 2021 (see https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/decodeAudioData#browser_compatibility), which will cause issues with older iOS devices. WDYT?

digitaldan: ahh, was not aware of that, i'll change it shortly.

florian-h05: Thx 👍

digitaldan: Very good catch, fixed

florian-h05 (approved):
LGTM, thank you very much!
Works fine on my iPad.

Issue #1769, opened 2023-03-05
Some classic icons are hard to see in dark mode


mueller-ma:
## Which UI are you reporting an issue for?

- [ ] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

None: Issue is in the classic icon set.

## The problem

Some icons from the classic icon set are hard to see when the UI is in dark mode. This could be Basic UI, Android app or the [website](https://www.openhab.org/docs/configuration/iconsets/classic/#icons).

* https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/network.png
* https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/motion.png
* https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/movecontrol.png

## Expected behavior

These icon should be easy to see on all backgrounds.

Suggestion: At least for the SVG version this issue could be fixed by replacing `fill="#323232"` with `fill="currentColor"`. The Android app and Basic UI (hopefully soon) will then color this icon light gray or dark gray depending on the background.

- https://github.com/openhab/openhab-webui/issues/1743
- https://github.com/openhab/openhab-android/pull/3282

## Steps to reproduce

1. Open https://www.openhab.org/docs/configuration/iconsets/classic/#icons when the browser is set to dark mode.

lolodomo:
In our classic iconset, I believe all the icons have colors and the renderihould remain the same for SVG and PNG icons. So your idea with currentColor could be not suitable for this iconset. 

Regarding support in BasicUI of color control for OH icons (delivered by the OH icon server), I am not yet sure that is easily doable, this would be a big change in the way to render these icons. We cannot assume this will be done until someone contributee something.

lolodomo:
But I agree that we have to change the colors of these icons.

lolodomo:
And for other iconsets containing only SVG, I think your solution is good and currentColor should then be the preference in these SVG icons.

lolodomo:
@mueller-ma : for motion icons, here is the current result in dark mode and bright mode:
![image](https://github.com/openhab/openhab-webui/assets/10186704/7483b9a4-0290-4903-bf92-bf3aad11a4cb)
[...]

PR #1777, opened 2023-03-06
Update icon network

mueller-ma:
The white border makes it easier to see on dark backgrounds.

See #1769


lolodomo:
Would it be possible for you to make captures with this icon used in Basic in the two different themes ?

Note that you will have to also adjust network-on and network-off and finally use the `.convert.sh` script to generate the corresponding PNG files.

lolodomo:
In fact, no need to change network-on and network-off, they looks good in white and dark themes.

The new icon looks a little strange on a dark theme:
image

mueller-ma:
> The new icon looks a little strange on a dark theme

Agreed. I looked at it in a larger version and then it was fine.

I guess someone with better design skills should work on this.

PR #1778, opened 2023-03-06
[mainui] Adds the "handleSize" config for the knob control.  

digitaldan:
By default it looks like the handleSize uses the strokeWidth if not set.

So with `handleSize` set:

```yaml
strokeWidth: 5
handleSize: 20
```
image

and with no `handleSize` (so how it works now)
```yaml
strokeWidth: 5
```

image


florian-h05:
See #1718, which refactors oh-knob and makes the vue-round-slider much better configurable.

digitaldan:
Gotcha, so you are suggesting to add this config option to that PR instead, correct ?

florian-h05:
Yes.
With this PR most round-slider config options will become available and the old options vor knob-control are replaced.
However there is a compatibility layer to avoid breaking existing widgets.

You can open a PR to my PR‘s branch, or wait until my PR is merged. Just don’t forget to regenerate the docs then.

digitaldan:
kool, see https://github.com/florian-h05/openhab-webui/pull/7

PR #1787, opened 2023-03-10
[basicui] Use sitemap input hint

mherwege:
This PR builds on https://github.com/openhab/openhab-webui/pull/1729 and requires https://github.com/openhab/openhab-core/pull/3418.

The PR will be rebased once the final version of  https://github.com/openhab/openhab-webui/pull/1729 has been merged. While it is feature complete from my perspective, testing and feedback are much appreciated.

This PR uses the sitemap Input widget's inputHint parameter to adjust the characteristics of the input field. The inputHint parameter will have the following impact:

- `number`: a QuantityType will be represented as a number field with a non-changeable unit field. Updates will be assumed to be in the unit of the old value.
- `date`: a datepicker is provided. This datepicker uses the standard html input type=date parameter. Therefore behaviour will be browser specific. Older browsers may revert to a text field. Also the representation is dependent on system or browser (for Chrome) locale.
- `time`: similarly a time picker is provided.
- `datetime`: uses the html input type=datetime-local, provided a combined date and time picker, will also be browser dependent.

The date and time pickers use standard html input type parameters. That restricts the formatting and will result in a browser dependent representation. This however, is the solution that can be achieved with least effort and least code. I also did not find good date/time pickers for Material Design Lite, that could easily be brought into the code.


mherwege:
Filling out a few fields:
image
Notice the comma in the Number input text field, vs the dot in the Number input number field. The input number fields respects the browser setting, which I have set to English GB, while the input text field uses the value from OH as is. My regional settings use a comma.

mherwege:
Entering a date in the Date input date field:
image
Results in:
image

mherwege:
The `Date input datetime` field also shows a time picker (as would the `Date input time`) field:
image

mherwege:
I have tested this on Chrome and Firefox on Windows, and to a lesser extend on Edge.
Firefox is a bit specific with different controls and a clear button on the date, time and datetime input fields I cannot hide. Using the clear button will however always reset to the previous value.
It also should look fine in dark them and condensed mode.

mherwege:
@lolodomo It is up to you how you want to proceed with this review. This PR already includes https://github.com/openhab/openhab-webui/pull/1729, with all review adjustments from the PR. There is some extra refactoring in this one to properly accomodate support for the inputHint parameter, https://github.com/openhab/openhab-core/pull/3418.
You could first continue the review on https://github.com/openhab/openhab-webui/pull/1729, and after merging that one look at this one, or I can also close  https://github.com/openhab/openhab-webui/pull/1729 and we continue here (after merging  https://github.com/openhab/openhab-core/pull/3418).
Let me know what you prefer.

lolodomo:
Please keep #1729 , we are going to first finalize and merge #1729.

lolodomo:
@mherwege : I let you rebase (the other PR is now merged) and then I will test and review this PR.

mherwege:
@lolodomo Rebase done. I struggled a bit as things had diverged a bit.
I did not include your last suggestion on https://github.com/openhab/openhab-webui/pull/1729 I included in that PR. I already had a working correction for this in this PR, and kept that. I did try if it still works correctly for the case the correction was meant for.

lolodomo:
@mherwege : can you please rebase it again, there is a conflict in _theming.scss. I will then start the tests and the review.

mherwege:
@lolodomo Rebase done.

mherwege:
@lolodomo In the review of the Android PR, a change in behaviour was made for dateTime type items.

1. If the item is of type dateTime, set a default inputHint to datetime (avoids text input for dateTime items)
2. If only date is set through input element, append time value from previous
3. If only time is set through input element, set the date part to the previous value

I have not made a change to this PR (yet), but I am unsure if, and how, I can.

1: I need text input for datetime anyway to allows for browsers not supporting the HTML5 input types.
2 and 3: The Android app has direct access to the item state. Here I only have access to the formatted state at the moment of doing a change (in the js onChange function).

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -58,38 +69,127 @@ public boolean canRender(Widget w) {
...
+            inputType = "datetime-local";
+            inputPattern = "pattern=\"[0-9]{4}-[0-9]d{2}-[0-9]d{2} [0-9]{2}:[0-9]{2}(:[0-9]{2})?\"";
+        }
+        snippet = snippet.replace("%input_type%", inputType);
+        snippet = snippet.replace("%input_pattern%", inputPattern);
+
+        String displayState = getValue(w);
+        State state = itemUIRegistry.getState(w);
+
+        String undefState = "-";
```
lolodomo, line 120: That means `"%undef_state%"` will now be replaced by "-" when the item is a number or string item with a state ?
It was before set to "" which made sense.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -58,38 +69,127 @@ public boolean canRender(Widget w) {
...
+            if ("number".equals(inputHint)) {
+                String[] stateArray = displayState.trim().split(" ");
+                if (stateArray.length > 0) {
+                    undefState = stateArray[0];
+                }
+            }
+        }
+        snippet = snippet.replace("%undef_state%", undefState);
+
+        String dataState = displayState;
```
lolodomo, line 137: No need to assign a value as you assign it in the if and in the else.
Or you could assign it to "" and then avoid an else in your if.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -58,38 +69,127 @@ public boolean canRender(Widget w) {
...
+        if (state == null || state instanceof UnDefType) {
+            dataState = "";
+        } else {
+            dataState = displayState.isEmpty() ? state.toString() : displayState;
+            if ("number".equals(inputHint)) {
+                String[] stateArray = displayState.trim().split(" ");
```
lolodomo, line 143: Why `displayState.trim()` and not `dataState.trim()` ?

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -58,38 +69,127 @@ public boolean canRender(Widget w) {
...
+                }
+            } else if (state instanceof DateTimeType) {
+                dataState = displayState.isEmpty() ? ((DateTimeType) state).format("%1$tY-%1$tm-%1$td %1$tT")
+                        : displayState;
+                if ("date".equals(inputHint)) {
+                    dataState = ((DateTimeType) state).format("%1$tY-%1$tm-%1$td");
+                } else if ("time".equals(inputHint)) {
+                    dataState = ((DateTimeType) state).format("%1$tT");
+                } else if ("datetime".equals(inputHint)) {
+                    dataState = ((DateTimeType) state).format("%1$tY-%1$tm-%1$tdT%1$tT");
```
lolodomo, line 155: Why using a different format than when no hint is provided ?

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -45,6 +50,12 @@ public class InputRenderer extends AbstractWidgetRenderer {
 
     private final Logger logger = LoggerFactory.getLogger(InputRenderer.class);
 
+    static final Pattern NUMBER_PATTERN = Pattern.compile("^(\\+|-)?[0-9\\.,]+");
+    static final Pattern DOT_SEPARATOR_PATTERN = Pattern
+            .compile("^-?(([0-9]{1,3}(,[0-9]{3})*)|([0-9]*))?(\\.[0-9]+)?$");
+    static final Pattern COMMA_SEPARATOR_PATTERN = Pattern
+            .compile("^-?(([0-9]{1,3}(\\.[0-9]{3})*)|([0-9]*))?(,[0-9]+)?$");
```
lolodomo, line 57: Please add "private"
Please also move the static before the non static

lolodomo (changes requested):
Partial review

lolodomo:
> 2 and 3: The Android app has direct access to the item state. Here I only have access to the formatted state at the moment of doing a change (in the js onChange function).

Just because you removed the code that was providing this information. You have this information each time an item state is updated through the function `_t.setValuePrivate`. In the current version (before your changes), the item state is saved in a variable named `lastItemState` that can be then used in function `onChange`.

lolodomo:
First tests.

For any kind of item, if I enter a value while the current state is NULL, the item state is updated in the server but "empty/NULL" values are restored in UI.

For number with dimension, whatever input hint (unset, set to text or set to number), if I enter a number, I see the command correctly send, the item state updated in server but the value is restored to the previous value in UI.

For number item without dimension with a state pattern set to `[%f]`, I can see that the ending 0 are removed when input hint is set to number while not when set to text or unset. As an example, if I enter value 123.45, the display value is finally 123.45 with hint set to number while 123,450000 when not.  Looks like the state pattern is not respected ?

For number item without dimension with a state pattern set to `[]` and input hint set number, no unit is displayed. I would expect to see the unit from the item state (which is °C in my use case).

For date time item and input hint set to date, selecting a new date is correctly sent to the server but then the old date is restored.

For date time item and input hint set to time, if I enter the hour, I see just after the hour disappearing and remains only 2 fields.

More generally, all stuff with date time seems to not work at all in Firefox, I was not able to fill a proper date & time.

lolodomo:
I guess the biggest problem should be in function `setValuePrivate` as it is the function that updates the UI when an item state is received. It is mostly not working.
I will have a look to the JavaScript and try to find where is the problem.

mherwege:
@lolodomo Not behind my computer at the moment, but for many of your tests, it looks like the SSE connection was not working properly. Values should be restored if the SSE connection does not send the final update.

mherwege:
> For number item without dimension with a state pattern set to `[]` and input hint set number, no unit is displayed. I would expect to see the unit from the item state (which is °C in my use case).

This is expected. The state pattern is not a unit, it could be anything. When inputHint is number, only number or number with dimension is allowed. For number with dimension, the unit is shown. Any state description beyond number formatting is ignored. I don’t see an easy way around this. State descriptions were meant for visualization, never for input patterns.

mherwege:
> You have this information each time an item state is updated through the function `_t.setValuePrivate`.

But not when the servlet is initially created.

lolodomo:
> > You have this information each time an item state is updated through the function `_t.setValuePrivate`.
> 
> But not when the servlet is initially created.

That is true but if necessary you can pass it from Java code to client code in a data field when the page is created.

lolodomo:
> > For number item without dimension with a state pattern set to `[]` and input hint set number, no unit is displayed. I would expect to see the unit from the item state (which is °C in my use case).
> 
> This is expected. The state pattern is not a unit, it could be anything. When inputHint is number, only number or number with dimension is allowed. For number with dimension, the unit is shown. Any state description beyond number formatting is ignored. I don’t see an easy way around this. State descriptions were meant for visualization, never for input patterns.

You did not understand my use case. I have disabled the state pattern and U am expecting to see the unit of the item state. Look at one of my review comments in the Java file, you apparently used the wrong variablevariable and it explains this wrong result.

lolodomo:
> but for many of your tests, it looks like the SSE connection was not working properly. Values should be restored if the SSE connection does not send the final update. And it looks like that is what is happening.

Yes, it could explain what I saw. 
What is strange is that I never encountered any problem with SSE and I used Ctrl+F5 to clear my cache. 
I will try again.

lolodomo:
No, it was not SSE and it was working for one item.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -58,38 +69,127 @@ public boolean canRender(Widget w) {
...
+                } else if ("datetime".equals(inputHint)) {
+                    dataState = ((DateTimeType) state).format("%1$tY-%1$tm-%1$tdT%1$tT");
+                }
+            }
+        }
+        snippet = snippet.replace("%data_state%", dataState);
+
+        String unitSnippet = "";
+        if ("number".equals(inputHint)) {
+            String[] stateArray = displayState.trim().split(" ");
```
lolodomo, line 163: You should consider item state when displaySratte is empty. 
It explains one issue I found

lolodomo (changes requested):

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
lolodomo: It leads to no value displayed when no state pattern ([]). I saw this issue but did not mention it in my quick report.

lolodomo:
Just to show the two little bugs that I can see when there is not state description:
Second line at right, missing value:
image
The corresponding sitemap:
```
        Frame label="Input number hint number" {
                Input item=TestNumber inputHint="number" labelcolor=["green"] valuecolor=["red"]
                Input item=TestNumber label="Test number no state descr []" inputHint="number"
        }
```

Second line at right, missing value and unit:
image
The corresponding sitemap:
```
        Frame label="Input number:dimension hint number" {
                Input item=CurrentTemp inputHint="number" labelcolor=["green"] valuecolor=["red"]
                Input item=CurrentTemp label="Current Temp no state descr []" inputHint="number"
        }
```

The good point is that I know how to fix that, it is in the review comment from yesterday. I am going to test the fix.

lolodomo:
With such item:
```
Number TestNumber "Test number [%.3f]" (GTest1)
```
and this sitemap:
```
        Frame label="Input number no hint" {
                Input item=TestNumber labelcolor=["green"] valuecolor=["red"]
                Input item=TestNumber label="Test number no state descr []"
        }
        Frame label="Input number hint number" {
                Input item=TestNumber inputHint="number" labelcolor=["green"] valuecolor=["red"]
                Input item=TestNumber label="Test number no state descr []" inputHint="number"
        }
        Frame label="Input number hint text" {
                Input item=TestNumber inputHint="text" labelcolor=["green"] valuecolor=["red"]
                Input item=TestNumber label="Test number no state descr []" inputHint="text"
        }
````
When I update the state to value 234.5
```
11:47:18.476 [INFO ] [openhab.event.ItemCommandEvent       ] - Item 'TestNumber' received command 234.5
11:47:18.492 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'TestNumber' changed from 123.45 to 234.5
```
the display of value does not respect the state pattern while input hint is set to number. It is ok when no input set or when input set is set to text. See second frame:
image

File: bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java (hunk 1 above)
Comment (mherwege) [reply bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java:120]: In principle, undef_state can be set once and never touched anymore. It will only show when the data value is empty. So in almost all cases, it would be fine to have "-". But indeed, we ran into the issue that for text values, an empty string is acceptable, and would show the undef_value. I forgot about it here. So went back to having default "" and updating undef_state from the SSE return value in js.

File: bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java (hunk 4 above)
Comment (mherwege) [reply bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java:155]: The HTML5 input types date, time and datetime-local expect a ISO formatted datetime string. This will not be visible to the user when these input types are used. If the inputHint is text (or none), not HTML5 input type will be used, only a plain text field. The pattern in that case removes the "T" separator.
In the Android app, we ultimately forced an input type when the itemType is DateTime, not leaving the option for a text input. For consistency, we could take this approach here as well.

File: bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java (hunk 5 above)
Comment (mherwege) [reply bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java:57]: Done.

File: bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java (hunk 6 above)
Comment (mherwege) [reply bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java:163]: Changed.

**Comment by mherwege** [comment]
@lolodomo With the TestNumber example you provided above, I also tried creating a Text widget:
```
Text item=TestNumber label="Test number no state descr []"
```
FWIW, the number also does not show in this case.

**Comment by lolodomo** [comment]
> FWIW, the number also does not show in this case.

Yes, you're right and that is expected. 
But for an input field, I believe you should have the current state to be able to update it?

**Comment by lolodomo** [comment]
> Your suggestions do make sense though. I am working through them.

So I do not start updating your code and I am waiting for your next push. 
I hope you will be a little available during next week as I would like to merge this PR next weekend to have it included in milestone 3.

**Comment by mherwege** [comment]
> the display of value does not respect the state pattern while input hint is set to number. It is ok when no input hint or when input hint is set to text.

I think I know what is causing this. The reason is the state pattern `%.3f`. It will interpret something like "2,450" as 2450, as it contains 3 digits after the comma and therefore settles on English number format in the parseNumber method (java) and function (js). That's the limit of trying to be intelligent in interpreting the numbers. If you change the pattern to `%.2f` I believe it shows correctly. I don't see any easy way out, apart from trying to compare the displayState and item state. When using the itemState, all formatting is lost. The displayState delivers a format that may be in comma format (depending on international settings) and needs to be parsed. But I don't know what the locale format is. And even if I get it on the Java side, the browser may use a different locale, so it may still be out of sync. I don't see an easy solution yet.

File: bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java (hunk 3 above)
Comment (mherwege) [reply bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java]: I changed this.

**Comment by mherwege** [comment]
@lolodomo Ready for a next review.

2 issues you may want to look into, and I would appreciate your view on.

1. Number parsing for %.3f formats
2. When starting BasicUI with a Numer:Dimension item with state NULL or UNDEF, the unit field is not initialized, and therefore, setting a value leads to a js exception. A refresh solves it. I didn't find a why to figure out if a non-initialized item is of type Number:Dimension, and therefore I should create an empty placeholder for the unit in the Java code.

File: bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java (hunk 4 above)
Comment (mherwege) [reply bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java:155]: I have now forced a datetime inputHint for DateTime items, but still need to keep some of the code as older browsers may not support the input types and default to text.

**Comment by mherwege** [comment]
> Sometimes, for hour/minute/second, only 2 fields are displayed and I can't understand how I can enter the 3 fields in that case.

That is a constraint of the HTML5 time input type. They only allow setting hour and minute, not seconds. How relevant is this in the context of OH?

**Comment by lolodomo** [comment]
> > Sometimes, for hour/minute/second, only 2 fields are displayed and I can't understand how I can enter the 3 fields in that case.
> 
> That is a constraint of the HTML5 time input type. They only allow setting hour and minute, not seconds.

But when seconds are defined in the state, I was able to display and edit them with success.

> How relevant is this in the context of OH?

Not important I believe. Would it possible and a good idea to not display them ?

**Comment by mherwege** [comment]
> But when seconds are defined in the state, I was able to display and edit them with success.
> 
> > How relevant is this in the context of OH?
> 
> Not important I believe. Would it possible and a good idea to not display them ?

OK, I see the same thing. It should be possible to cut after minutes and not display seconds. I will look into it.

**Comment by mherwege** [comment]
> Not important I believe. Would it possible and a good idea to not display them ?

Last commit puts in logic to remove the seconds from visualisation and input.

File: bundles/org.openhab.ui.basic/web-src/smarthome.js [hunk 7]
```diff
@@ -1601,17 +1655,51 @@
...
+					}
+				} else {
+					var undefArray = undefValue.split(" ");
+					if (undefArray.length > 0) {
+						undefValue = undefArray[0];
+					}
+				}
+			} else if (_t.itemType === "datetime") {
+				newValue = ((lastItemState !== "NULL") && (lastItemState !== "UNDEF")) ? itemState : value;
+				newValue = newValue.trim().split(".")[0];		// drop millis
```
lolodomo: The problem is here.
It is possible that value does not contain a "." (state pattern defined as `[%s]`):
image
Is there a reason to not consider only itemState here ?
Are you trying to consider the state pattern defined for the item ? Is it a good idea for date& time ?

lolodomo (changes requested):

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: I was not understanding that line ;)

mherwege: Already fixed, I came to the same conclusion, only using itemState for datetime.

mherwege: Apparantly a %s state pattern drops the millis (but still shows the time offset). The events always have millis.
Note that in all the logic, I drop the time offset. I send without offset and drop the offset from received values.

lolodomo:
For me, the version before the last one was OK for live update of dates.

For text and number items, if the item state switches to UNDEF, I see "UNDEF" displayed when,there is no state description (right column).
image
But if you open the page while the state is UNDEF, "UNDEF" is then not displayed.
image
It is not consistent. I think a blank would be OK in that case.

lolodomo:
I think you are not handling the color of the new "unit" field:
image
Should it be colored when the item state is NULL/UNDEF ?

For the date, the widget gets the defined color even when the item state is NULL/UNDEF.
image
Is it what you would like ? It looks to me not fully consistent.

lolodomo:
Does no more compile with your last change:
```
[INFO] --- frontend-maven-plugin:1.9.0:npm (gulp build) @ org.openhab.ui.basic ---
[INFO] Running 'npm run build' in D:\dev\openhab3\git\openhab-webui\bundles\org.openhab.ui.basic
[INFO]
[INFO] > org.openhab.ui.basic@4.0.0 build
[INFO] > gulp default
[INFO]
[INFO] [15:20:19] Using gulpfile D:\dev\openhab3\git\openhab-webui\bundles\org.openhab.ui.basic\gulpfile.js
[INFO] [15:20:19] Starting 'default'...
[INFO] [15:20:19] Starting 'css'...
[INFO] [15:20:19] Starting 'copyFontLibs'...
[INFO] [15:20:19] Starting 'eslint'...
[INFO] [15:20:19]
[INFO] D:\dev\openhab3\git\openhab-webui\bundles\org.openhab.ui.basic\web-src\smarthome.js
[INFO]   1662:21  error  Expected '===' and instead saw '=='  eqeqeq
[INFO]   1694:18  error  Expected '!==' and instead saw '!='  eqeqeq
[INFO]
[INFO] ✖ 2 problems (2 errors, 0 warnings)
[INFO]
[INFO] [15:20:19] 'eslint' errored after 519 ms
[INFO] [15:20:19] ESLintError in plugin "gulp-eslint"
[INFO] Message:
[INFO]     Failed with 2 errors
[INFO] Details:
[INFO]     domainEmitter: [object Object]
[INFO]     domainThrown: false
[INFO]
[INFO] [15:20:19] 'default' errored after 535 ms
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  14.756 s
[INFO] Finished at: 2023-05-08T15:20:19+02:00
```

mherwege:
@lolodomo Sorry for the build failure.
I added color handling for undef values. It was not necessary before as the color of the input widget label is fixed, but unit and datetime placeholders are not put there but in the value fields. So it needs handling.

lolodomo:
It is better after each commit ;)

You still have to handle the colors when opening the page and item state is NULL/UNDEF.
It is OK when state switches to UNDEF while the page is already opened.

Regarding rendering of states and live updates of states, remains only one problem: number item with NULL/UNDEF state, input hint number and no state description. When the item state is updated, the field remains blank. I believe you mentioned yesterday that you have no idea how to handle this case.
image
It is OK for number without dimension.

Regarding date inputs, I like that the format is the French format.

Regarding number input with hint number, I see that the value is displayed with a comma as decimal separator. It uses the French format. Good.
image
But let see what happens if the item state is 56 and 15 °C:
image
Strangely it displays "56.00" and "15.0 °C" instead of "56,00" and "15,0 °C".

lolodomo:
And regarding time filling, there is no time picker available ?

PS: all these tests were run in Firefox. I will test later with Chrome.

lolodomo:
> And regarding time filling, there is no time picker available ?

Ok, I see there is one with Chrome ;)
For firefox, is there something missing or does Firefox not support itme picking ?

mherwege:
> And regarding time filling, there is no time picker available ?
> 
> PS: all these tests were run in Firefox. I will test later with Chrome.

Chrome shows a time picker, Firefox does not.

lolodomo:
> Chrome shows a time picker, Firefox does not.

Yes and it is better !
Tests with date & time also OK using Chrome on Windows.

mherwege:
> Regarding filling input field for date & time, it works now perfectly. Just one question: I see that the command is sent each time you update one element like the month for example. So if you want to set a new date & date, in fact this will trigger several commands, including wrong (intermediate) values. I was asking myself if the command should not be sent only when the user pushes the ENTER key in the field ? Is it at least possible ?

I don't know if there is a way to control this. However, I don't think it sends multiple updates when you use the picker, only when you edit the field directly.
This is the disadvantage of using the standard HTML5 input types. While relatively easy to use and available in all modern browsers, it doesn't allow much control.

mherwege:
> You still have to handle the colors when opening the page and item state is NULL/UNDEF.
> It is OK when state switches to UNDEF while the page is already opened.

That is done now.

lolodomo:
> > Chrome shows a time picker, Firefox does not.
> 
> Yes and it is better ! Tests with date & time also OK using Chrome on Windows.

Date & time picker of Microsoft Edge is stricly identical to the Chrome one ! So works well.

lolodomo:
> > You still have to handle the colors when opening the page and item state is NULL/UNDEF.
> > It is OK when state switches to UNDEF while the page is already opened.
> 
> That is done now.

Does not compile ;)

mherwege:
> Regarding rendering of states and live updates of items, remains only one problem: number item with NULL/UNDEF state, input hint number and no state description. When the item state is updated, the field remains blank. I believe you mentioned yesterday that you have no idea how to handle this case.

Indeed, I am struggling with that one. The real solution should be to have the unit field defined in the snippet, but I don't know how to identify an Item that can have a unit in the Java code. `QuantityType.class.isAssignableFrom(o)` will also be true for pure Number items that do not have a dimension, and where I should make room for the unit field.

mherwege:
> Does not compile ;)

Have to run gulp before each smallish change ;). These are all linter errors that slip in.

lolodomo:
> > Regarding rendering of states and live updates of items, remains only one problem: number item with NULL/UNDEF state, input hint number and no state description. When the item state is updated, the field remains blank. I believe you mentioned yesterday that you have no idea how to handle this case.
> 
> Indeed, I am struggling with that one. The real solution should be to have the unit field defined in the snippet, but I don't know how to identify an Item that can have a unit in the Java code. `QuantityType.class.isAssignableFrom(o)` will also be true for pure Number items that do not have a dimension, and where I should make room for the unit field.

With the change of UoM merged yesterday, in that case, I believe you must consider the default unit of this number dimension which will become the unit of the item state when it will have a first value.

lolodomo:
> > You still have to handle the colors when opening the page and item state is NULL/UNDEF.
> > It is OK when state switches to UNDEF while the page is already opened.
> 
> That is done now.

Confirmed to work properly in all conditions.

lolodomo:
What should happen if the user sets input hint to date or time or datetime on a String item or a number item ?
They are unexpected cases I guess.
Is it handled properly by this PR (not yet tested) ?

lolodomo:
> With the change of UoM merged yesterday, in that case, I believe you must consider the default unit of this number dimension which will become the unit of the item state when it will have a first value.
> Edit: in fact, it could be also the unit set to the item through the new `unit` metadata.

Sorry, it is not yet merged but will be very soon. So put this problem in standby and don't losse your time.

mherwege:
> What should happen if the user sets input hint to date or time or datetime on a String item or a number item ?
> They are unexpected cases I guess.
> Is it handled properly by this PR (not yet tested) ?

I don’t expect it to work, and I don’t have special treatment for it. I didn’t do much testing on it.
We could log warnings in the Java code and default the inputHint to text if there is a mismatch.

lolodomo:
> We could log warnings in the Java code and default the inputHint to text if there is a mismatch.

I agree.

Isn't it something we should have done in the sitemap syntax ?

mherwege:
> > We could log warnings in the Java code and default the inputHint to text if there is a mismatch.
> 
> I agree.
> 
> Isn't it something we should have done through a control in the sitemap syntax ?

I was thinking the same thing. It could go [here](bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/validation/SitemapValidator.xtend) and in the corresponding checks in the mainUI sitemap editor. I am just not sure to have easy access to the item type at these places.
And what happens if you change the item type after creating the sitemap? We would still need a check here to cope with that anyway.

lolodomo:
I summarize the status. Here are the remaining "problems", almost everything concerns the input hint number:

1. Input hint number for number item with dimension: display of value and unit when there is no state pattern (in standby until new UoM stuff is merged)
2. Input hint number: display of value not respecting the number of decimals defined in the state pattern
3. Input hint number: display value sometimes with "," as decimal separator, sometimes with"."
4. Adding handling of inconsistent combinations of item types and input hint values

We should wait the merge of openhab/openhab-core#3481 and check again number with dimension when item and unit in state pattern are different.

And I plan run additional tests of the input widget for number items.
And of course finalize the code review.

lolodomo:
> I was thinking the same thing. It could go [here](bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/validation/SitemapValidator.xtend) and in the corresponding checks in the mainUI sitemap editor. I am just not sure to have easy access to the item type at these places.
> And what happens if you change the item type after creating the sitemap? We would still need a check here to cope with that anyway.

Let's go with controls in each UI.

mherwege:
> Let's go with controls in each UI.

With the latest commit, it writes warnings in the log when using. I am still unsure about checks in the UI configuration. So far, we nowhere do a check on item type and I think it is not easy to do. That is partially because there is an easy direct match between item type and widget. It also always leaves the problem of defining (or changing) the item after the sitemap, which is now possible, but would become impossible if we impose a strict control there. Anyway, I think that would be subject to another PR.

mherwege:
> 1. Input hint number for number item with dimension: display of value and unit when there is no state pattern (in standby until new UoM stuff is merged)

I solved that part. There is no error anymore, it updates correctly. Only thing is no unit may be displayed at start with UNDEF value of the widget state description is empty, but it updates as soon as a state is set. With the new Unit handling, it will probably show the default unit from the start.

mherwege:
> 4\. Adding handling of inconsistent combinations of item types and input hint values

Done

mherwege:
The SAT failure is in a line of code copied verbatim from core, [here](https://github.com/openhab/openhab-core/blob/6af9870c506f62d74aad46bfcf7d6c22f1353843/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L1327). Can it be the SAT version has not been updated in the webui repository yet? The line is using valid Java 17 code.

mherwege:
The SAT failure is in a line of code copied verbatim from core, [here](https://github.com/openhab/openhab-core/blob/6af9870c506f62d74aad46bfcf7d6c22f1353843/bundles/org.openhab.core.ui/src/main/java/org/openhab/core/ui/internal/items/ItemUIRegistryImpl.java#L1327). Can it be the SAT version has not been updated in the webui repository yet? The line is using valid Java 17 code.

mherwege:
> * Input hint number: display of value not respecting the number of decimals defined in the state pattern
> * Input hint number: display value sometimes with "," as decimal separator, sometimes with"."

I have been testing this a bit. It looks to me like the issue is with Firefox. I don't have problems with Chrome or Edge. In my tests, I have Chrome set to English language and Firefox and Edge set to Dutch (which also uses a comma decimal separator). Chrome shows . as expected, with the right number of decimals. Edge shows , with the right number of decimals. Firefox is hit and miss. It always cuts extra zeros in the decimal places. And it sometimes shows . inso ,
Not nice, but not sure there is much I can do in the code.

lolodomo:
> I have been testing this a bit. It looks to me like the issue is with Firefox. I don't have problems with Chrome or Edge. In my tests, I have Chrome set to English language and Firefox and Edge set to Dutch (which also uses a comma decimal separator). Chrome shows . as expected, with the right number of decimals. Edge shows , with the right number of decimals. Firefox is hit and miss. It always cuts extra zeros in the decimal places. And it sometimes shows . inso ,
> Not nice, but not sure there is much I can do in the code.

Ok, I will check too.

mherwege:
> > the display of value does not respect the state pattern while input hint is set to number. It is ok when no input hint or when input hint is set to text.
> 
> I think I know what is causing this. The reason is the state pattern `%.3f`. It will interpret something like "2,450" as 2450, as it contains 3 digits after the comma and therefore settles on English number format in the parseNumber method (java) and function (js). That's the limit of trying to be intelligent in interpreting the numbers. If you change the pattern to `%.2f` I believe it shows correctly. I don't see any easy way out, apart from trying to compare the displayState and item state. When using the itemState, all formatting is lost. The displayState delivers a format that may be in comma format (depending on international settings) and needs to be parsed. But I don't know what the locale format is. And even if I get it on the Java side, the browser may use a different locale, so it may still be out of sync. I don't see an easy solution yet.

Last commit should solve this. It is highly unlikely someone will input commas as thousand separators. And it can only make a mistake on this when entering something like 12,345 with 3 places after the comma. In this case, the comma will be interpreted as decimal separator and not thousand separator. For 12,345,678 the commas will be interpreted as thousand separators.

lolodomo:
I just made tests with the new UoM stuff:
```
Number:Length TestNumberDim1 "Test number dim 1 [%.1f cm]" (GTest1) { unit="mm" }
Number:Length TestNumberDim2 "Test number dim 2 [%.1f cm]" (GTest1)
```
```
	Frame label="Input number:dimension no hint" {
		Input item=TestNumberDim1 labelcolor=["green"] valuecolor=["red"]
		Input item=TestNumberDim1 label="Test number dim 1 no state descr []"
		Input item=TestNumberDim2 labelcolor=["green"] valuecolor=["red"]
		Input item=TestNumberDim2 label="Test number dim 2 no state descr []"
	}
	Frame label="Input number:dimension hint number" {
		Input item=TestNumberDim1 inputHint="number" labelcolor=["green"] valuecolor=["red"]
		Input item=TestNumberDim1 label="Test number dim 1 no state descr []" inputHint="number"
		Input item=TestNumberDim2 inputHint="number" labelcolor=["green"] valuecolor=["red"]
		Input item=TestNumberDim2 label="Test number dim 2 no state descr []" inputHint="number"
	}
	Frame label="Input number:dimension hint text" {
		Input item=TestNumberDim1 inputHint="text" labelcolor=["green"] valuecolor=["red"]
		Input item=TestNumberDim1 label="Test number dim 1 no state descr []" inputHint="text"
		Input item=TestNumberDim2 inputHint="text" labelcolor=["green"] valuecolor=["red"]
		Input item=TestNumberDim2 label="Test number dim 2 no state descr []" inputHint="text"
	}
```
If I open the page while the item state is NULL/UNDEF, the units displayed on the left column are wrong for input hint number (and correct for other input hints):
image
After the state update, the displayed unit is then OK:
image
And the unit is not displayed when there is no unit in state description (should be m).

If IU open the page while the item state is neither NULL nor UNDEF, the units displayed on the left column are wrong for input hint number. This is the internal item unit rather than the unit defined in state description. It is OK on the right column (no state description). Units are fixed after a item state update.
image

lolodomo:
And here is a message from @J-N-K :

> If you call .getUnit or .getUnitSymbol on the item and get what the item currently uses internally. This of course can change if add/update/remove the unit metadata.

You can certainly use that to get internal item unit, even when the item state is NULL or UNDEF. It should help for the cases without state description.

mherwege:
Sorry for the time it took to adjust the code based on your feedback. I had a busy period. I hope this new version is getting closer to what we would want to achieve. I tested with all of your examples, and I believe the results are better. Still, I would appreciate you to test as well. It is clear this is not a straightforward thing.

Getting and using the correct unit is challenging. I tried various combinations and finally settled on the new code. All is about what takes precedence. What if the sitemap label has an empty state description, should it use the item state description or the item unit from unit metadata? It now uses unit metadata.
Note that there exists a method in AbstractWidgetRenderer, [getUnitForWidget](https://github.com/openhab/openhab-webui/blob/0b555580baff63f2c44e2dd996107e7105b13102/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java#L446), that does not yield the unit needed here. This code does call the methods referred to by @J-N-K and I was hoping to be able to use that. So I had to duplicate some of that code and create some specific logic. I did not want to touch the original code in fear of breaking things somewhere else.

> Units are fixed after a item state update.

This is expected for inputHint = number. With inputHint number, a numeric keyboard would be shown on a mobile. If you would be able to change the unit, you would need a full keyboard. In most cases this is fine as one would typically always enter the values in the same input. If changing units is desired, no inputHint or inputHint text should give the desired behaviour.

lolodomo:
I will have a new look. 

I was thinking for unit you should be able to reuse the same Java code used for slider or setpoint widgets.

mherwege:
> I was thinking for unit you should be able to reuse the same Java code used for slider or setpoint widgets.

And I think that is using the getUnitForWidget method. But I couldn’t get that to work in all cases, especially with empty labels or null/undef values.

There is another issue with more complex state descriptions, for example:
`Input item=test label=“test [test prefix %.0f %unit% test postfix]`
This is a perfectly acceptable state description, but would fail in working as a good input pattern. I am trying to work on a good solution for that (stripping pre- and postfix and eventually putting it in extra fixed fields).

mherwege:
https://github.com/openhab/openhab-webui/pull/1923 starts from this PR and uses proposed changes in https://github.com/openhab/openhab-core/pull/3644.

lolodomo:
I tried again your PR but I discovered that your Baisc UI from your PR is not working in a recent snapshot (3519).
Would it be possible for you to rebase your PR ?

lolodomo:
I can apparently do it myself in GitHub but I would prefer that you do it yourself to be sure that I do not break your branch.
I will do it myself locally to confirm that it solves the problem.

lolodomo:
Even after merging locally main into your branch, Basic UI does not work. Nothing happens when I push a button in UI that should trigger a rule. If I rather use Basic UI packaged with the snapshot 3519, then it works.
It is as if your PR totally broke Basic UI, I do not understand what is happening.

lolodomo:
Very strange, it looks my update method using bundle:update is the source of the problem ! If I put your jar in my addons folder, then it seems to work. I will start again from your branch without rebase.

lolodomo:
If I just check how it behaves when state updates are received, I see that displayed values are OK but the unit is wrong with input hint number (internal unit displayed instead of pattern unit). As soon as a state update is received, then the unit is updated properly. So wrong unit when you open the page. That is annoying.

lolodomo:
Looking at my last message with captures May the 12th, this bug was already mentioned.
Is it so difficult to fix that ? At least when the state is not NULL/UNDEF, it should be easy, you just have to take the unit from the label, if present.

mherwege:
I just tested again with your item and sitemap definitions from https://github.com/openhab/openhab-webui/pull/1787#issuecomment-1545526789

Starting BasicUI with values already set in the items gives me:

image

Do you see the same thing when you test? I don't see anything wrong with this.

When I start with UNDEF values I get this:

image

Again, this looks correct to me.

This was tested in Eclipse IDE with core rebased to today.

lolodomo:
No, this is not the result I got this evening when I use the jar built with from your PR inside snapshot 3159.

mherwege:
I don't know what's wrong here. I have the core code from the main branch today, did a maven update in Eclipse on all I have open, forcing download of dependencies. I did not rebase the PR. It looks like somehow there is still old code running in your install.

lolodomo:
> Is there any chance an older jar is running?

No, I started with a fresh install of OH (3519). 
And I run "git pull mherwege" and then created a new local branch tracking your remote branch.
Can you point a particular code line that is new and I will compare if it matches in my local branch. 
And please also provide your jar, I will test it.

lolodomo:
And look at this:
```
$ git pull mherwege sitemap_input_hint
From https://github.com/mherwege/openhab-webui
 * branch              sitemap_input_hint -> FETCH_HEAD
Already up to date.
```

mherwege:
Is that the full git log? The 2 commits before the ones in your list are the ones that fix the May 12th bug.

mherwege:
The jar compiled by me can be found [here](https://www.dropbox.com/scl/fi/vks6yz3xfww06x3mlbd3e/org.openhab.ui.basic-4.0.0-SNAPSHOT.jar?dl=0&rlkey=fotubuwayou8p3p949bn84svy).

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -58,38 +74,271 @@ public boolean canRender(Widget w) {
...
+
+    private String cleanValue(String value, Widget w, Item item) {
+        String prefix = getPrefix(w, item);
+        String postfix = getPostfix(w, item);
+        String newValue = value.startsWith(prefix) ? value.substring(prefix.length()) : value;
+        newValue = value.endsWith(postfix) ? newValue.substring(0, newValue.lastIndexOf(postfix)) : newValue;
+        return newValue.trim();
+    }
+
+    private String getUnit(Widget w, NumberItem item) {
```
mherwege, line 254: This is the new method that is key to making it work, and was not there before May 22nd.

lolodomo:
> Is that the full git log? The 2 commits before the ones in your list are the ones that fix the May 12th bug.

That is only the beginning,

lolodomo:
I certainly have your last code. 

> The jar compiled by me can be found [here](https://www.dropbox.com/scl/fi/vks6yz3xfww06x3mlbd3e/org.openhab.ui.basic-4.0.0-SNAPSHOT.jar?dl=0&rlkey=fotubuwayou8p3p949bn84svy).

I will try your jar but this will be only this evening. 
But I doubt the problem could be my way to build the jar, I just ran "mvn clean install".

mherwege:
I now tested myself on a new install of the snapshot with my jar. And I found the difference between your and my test.

If you define the items and sitemap in the UI (just copying the test items and test sitemap you provided), it works. That is what I always did so far.
If you define them in files, it doesn't.
I was not expecting this difference. So there clearly is a difference in behaviour between these 2 ways of configuring.

mherwege:
Testing in a debugger:

- When using file item definitions, `item.getLabel()` returns `Test number dim 1`
- When using item definitions in UI, `item.getLabel()` returns `Test number dim 1 [%.1f cm]`

Inspecting the NumberItem item object also shows this difference in the label field.

lolodomo:
I confirm that my tests were done with everything defined in config files.

lolodomo:
Core UI probably handles item label but I believe the code in Basic UI should not do it directly. It should rather read widget (sitemap) label.

mherwege:
> Core UI probably handles item label but I believe the code in Basic UI should not do it directly. It should rather read widget (sitemap) label.

The logic I use:

- First try the widget label
- If it is null (no widget label defined), go for the item label
- Special case, if it is empty (empty widget label defined), go for UOM in metadata or default UOM
- After trying the widget label, when it is still empty go for the item label, same logic as above

So far, none of the widgets needed to get the unit explictly. It was always derived from the state. As we need to edit here, we need it explicitly. Note that the widget label is **not** set to the item label when it is not defined.

I can go back to the behaviour of using the unit from the state, but that brings back the problem of the unit for UNDEF or NULL states. There is no way to retrieve that when the label is defined on the item (or at least when it is defined in files), as you suggest me to do here: https://github.com/openhab/openhab-webui/pull/1787#issuecomment-1545533304. It does work when it is defined in the sitemap label, the way I tested.

All of this was a surprise to me. It is very clear though that there is a difference in behaviour between setting the label field of the GenericItem coming from a text file or the UI. The UI does not strip the value formatting. And as this still works with BasicUI, it is a strong indication that that is the right behaviour. It is, to say the least, inconsistent. It was never a problem because nobody has been interested in the pure format string so far (to extract the unit). BasicUI is clearly capable coping with label with or without the value formatting string. I don't know why it was stripped in the original code in the first place and I may be overlooking something, of course.

Just copying the textual definitions of items and sitemaps in the respective import parts of the UI should result in exactly the same behaviour as using text files. We discovered here it does not. Sitemaps are OK. Items are not. File definitions strip the value part from the label in the GenericItem. UI definitions don't. So far, I am not aware of this having consequences in the UI, except for what I try to do here. I cannot provide a solution that will work for definitions in text files without a change to core.

mherwege:
I found a way to make it work in both cases, but it is very ugly. Here is what the code would look like:

```
    private String getUnit(Widget w, NumberItem item) {
        String unit = "";
        if (w.getLabel() != null) {
            unit = getUnitFromLabel(item, w.getLabel());
        }
        // required when item is defined through UI and the label contains the value format pattern
        if ((unit != null) && unit.isBlank() && (item.getLabel() != null)) {
            unit = getUnitFromLabel(item, item.getLabel());
        }
        // required when the value pattern is in the state description (text definition or meta data)
        if ((unit != null) && unit.isBlank()) {
            StateDescription stateDescription = item.getStateDescription();
            if (stateDescription != null) {
                String pattern = stateDescription.getPattern();
                if (pattern != null) {
                    unit = getUnitFromLabel(item, "[" + pattern + "]");
                }
            }
        }
        if (unit == null || unit.isBlank()) {
            unit = item.getUnitSymbol();
        }
        unit = (unit == null) ? "" : unit;
        return unit;
    }
```
I can commit this if you want, but it is still a workaround for different behaviour between text definitions and UI definitions.

lolodomo:
> I found a way to make it work in both cases, but it is very ugly. Here is what the code would look like:

OK, let's go with that method. Can you please commit that ?

In all cases, I can continue my tests setting the unit in the widget label.

mherwege:
> OK, let's go with that method. Can you please commit the change ?

Committed this work around.

lolodomo:
Regarding the behaviour when receiving state updates (including UNDEF), it is now good even if not perfect.
I detected a different rendering when opening while the current state is NULL/UNDEF when opening the page:
image
and while the current state is defined when opening the page and then switched to UNDEF:
image
It only concerns the case where the user disabled the display of value with `[]` in the label.
I will not block on that.

lolodomo:
The build is failing in HABot because your branch is outdated.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
```diff
@@ -58,38 +75,282 @@ public boolean canRender(Widget w) {
...
+        List<Class<? extends State>> dataTypes = item.getAcceptedDataTypes();
+        if (("number".equals(inputHint)
+                && !(dataTypes.contains(DecimalType.class) || dataTypes.contains(PercentType.class)))
+                || (("date".equals(inputHint) || "time".equals(inputHint) || "datetime".equals(inputHint))
+                        && !dataTypes.contains(DateTimeType.class))) {
+            logger.warn("Invalid inputHint {} for item {} of type {}, set to default", inputHint, item.getName(),
+                    item.getType());
+            inputHint = null;
+        }
+        if ("datetime".equals(dataType) && ((inputHint == null) || "".equals(inputHint) || "text".equals(inputHint))) {
```
lolodomo, line 113: Would be better to use `inputHint.isEmpty()` instead of `"".equals(inputHint)`

lolodomo (commented):
Review of Java code and HTML snippet is OK. I have just one small comment.

lolodomo:
Remains to review CSS and JavaScript code and do more real tests.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/InputRenderer.java
mherwege: Adjusted and rebased.

mherwege:
As the relevant core PR https://github.com/openhab/openhab-core/pull/3644 has now been merge, I also updated that one with the review feedback in this PR. It may be easier to check out that one instead. I didn't test yet, but I believe the issue with difference between UI config and file config impact would not appear there, as it is covered in core.

lolodomo:
Do you mean this PR should be closed in favor of #1923 ?

mherwege:
> Do you mean this PR should be closed in favor of #1923 ?

Yes, I think so. #1932 builds contains everything in this one, but simplified as core offers better support.

lolodomo:
So I let you close it. 
I will switch to the other PR. 
My code review was almost finished. 
I will test the other PR with a fresh new server starting tomorrow.

mherwege:
Replaced by #1923.

PR #1788, opened 2023-03-11
Use channelType label & description as placeholder for channel details

florian-h05:
Use the `channelType`’s label and description as placeholder inside the `channel-general-settings` component if they are available and no explicit label or description is set.

Remove the `ready` var from the `channel-edit` component as is is only used by the `channel-general-settings` component, for which all required data are passed to `channel-edit` and therefore independent from the API request inside `channel-edit`.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/channel/channel-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/channel/channel-edit.vue
@@ -11,7 +11,7 @@
     <f7-block class="block-narrow">
       <f7-col v-if="channel">
         <f7-block-title>Channel</f7-block-title>
-        <channel-general-settings :channel="channel" :channelType="channelType" :createMode="false" :ready="ready" />
+        <channel-general-settings :channel="channel" :channelType="channelType" :createMode="false" ready="true" />
       </f7-col>
       <f7-col v-if="channelType != null">
         <f7-block-title v-if="configDescription.parameters">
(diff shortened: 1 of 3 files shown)


florian-h05:
Example:

IMG_2062

bundles/org.openhab.ui/web/src/pages/settings/things/channel/channel-edit.vue
```diff
@@ -11,7 +11,7 @@
     <f7-block class="block-narrow">
       <f7-col v-if="channel">
         <f7-block-title>Channel</f7-block-title>
-        <channel-general-settings :channel="channel" :createMode="false" :ready="ready" />
+        <channel-general-settings :channel="channel" :channelType="channelType" :createMode="false" ready="true" />
```
ghys, line 14: I didn't notice it before but I think it's quite weird to have a `ready` prop - in my mind it's a component-local flag and should remain so, normally you would only need to switch from a "loading" state to a "ready" state in your component, if you have data to load, and replace the UI with `v-if`/`v-else`; not have the subcomponent handle the "not ready" state.

florian-h05: You are right, I will implement this.

florian-h05:
@ghys Ready for re-review.

ghys (approved):
This seems more logical and cleaner, thanks!

Issue #1129, opened 2021-07-09
widget development: Unable to correctly create custom list widgets


wuschi:
## The problem

Coming from OH 1.8, I'm currently trying to get my head around building custom widgets.
I am currently trying to build a custom list widget that renders _exactly_ like the predefined widgets, e.g. the oh-rollershutter-item.
At the current state of the widget editor, this seems to be impossible. For example, for the f7-list-item widget, i cannot define a slot "footer" that renders within `<div class="item-inner"><div class="item-footer">`. Same applies to slot "media". 

## Expected behavior

I would have expected that the widget editor lets me define custom list widgets in yaml using the f7 widget components with the respective slots renderings correctly.

## Steps to reproduce

doesn't really apply here, but here's my current content of the widget editor:

```yaml
uid: custom_list_item
tags:
  - list
  - blinds
props:
  parameters:
    - context: item
      description: Control Item
      label: Control Item
      name: item
      required: false
      type: TEXT
  parameterGroups: []
timestamp: Jul 9, 2021, 2:30:05 PM
component: f7-list-item
config:
  mediaItem: true
  title: Test-Title
slots:
  content:
         - component: f7-row
           config:
             class:
               - align-items-center
             style:
               width: 100%
               flex-wrap: nowrap
           slots:
             default:
               - component: f7-col
                 config:
                   class:
                     - display-flex
                     - flex-direction-row
                     - align-items-center
                     - margin-top
                     - margin-bottom
                   style:
                     height: 100%
                     width: fit-content
                     min-width: 100px
                 slots:
                    default:
                      - component: oh-button
                        config:
                          icon-f7: lightbulb_fill
                          style:
                            --f7-button-bg-color: transparent
                            --f7-button-hover-bg-color: transparent
                            --f7-button-pressed-bg-color: transparent
               - component: f7-col
                 config:
                   class:
                     - display-flex
                     - flex-direction-row
                     - align-items-center
                     - margin-top
                     - margin-bottom
                   style:
                     width: fit-content
                     min-width: 100px
                 slots:
                   default:
                     - component: oh-rollershutter-item
                       config:
                         item: =props.item
  
```

## Your environment

```yaml
runtimeInfo:
  version: 3.0.2
  buildString: Release Build
locale: de-DE
systemInfo:
  configFolder: /openhab/conf
  userdataFolder: /openhab/userdata
  logFolder: /openhab/userdata/logs
  javaVersion: 11.0.10
[...]

PR #1798, opened 2023-03-15
[MainUI] Pass slot names down in generic-widget-component (#1129)

michalsrb:
Without this custom widgets based on f7-list-item can not pass components to slots other than inner.

This is based on @arroyoj's  suggestion https://github.com/openhab/openhab-webui/issues/1129. I've been using it in my openhab installation for about month and half, I've seen nothing break in the standard widgets and it allowed some custom widgets that weren't possible before.


arroyoj:
@michalsrb, thanks for confirming the fix from #1129 is working for you and submitting the PR!

ghys (approved):
Thanks indeed!
Also many thanks to @arroyoj for the investigation and detailed explanation of the problem in https://github.com/openhab/openhab-webui/issues/1129#issuecomment-922174803, this part of the code is very sensible to changes but in this instance it seems very logical and benign and the confirmed bugfix makes it worth it. Eventual (but probably unlikely) regressions will be addressed when/if they arise.

Issue #1743, opened 2023-03-01
Change color of "colorless" SVG icons depending on design 


mueller-ma:
## Which UI are you reporting an issue for?

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

Let me copy the issue of @moltke69 which he opened in the Android repo (https://github.com/openhab/openhab-android/issues/3204):

If you use own SVG icons without any "color information" (like fill, stroke, ...) the used color is always "black" independent to the the design you use (light or dark). For light design this maybe ok, but for dark design, you cannot see the icon anymore.

![Screenshot_20230207-182555](https://user-images.githubusercontent.com/96469391/217319805-19aa3750-ae5c-40aa-adba-825dfbe005e1.png)  ![Screenshot_20230207-182635](https://user-images.githubusercontent.com/96469391/217319810-0c5b6c87-3d6a-4dc9-a633-814259f0b927.png)

Of course, it is possible to add colors in the SVG files, but if you want to be flexible (one user at home is using the light design, the other the dark design), you cannot use the same icons.

It would be great, if you can choose the default color (for colorless SVG icons) on the basis of the design or perhaps on the "dynamic colors based on the wallpaper".

## Your suggestion

My solution: SVG can set the color to the keyword `currentColor` which can be set by CSS to color the icon. The Android app will set the color to black for the light theme and white for the dark theme (https://github.com/openhab/openhab-android/pull/3282).

## Your environment

```yaml

```

lolodomo:
In fact, I would like to use the icon color attribute. I have a pending PR already working for non OH sources. 
For OH icons source, BasicUI uses a HTML img tag to render the icon and at my knowledge it is not possible to control the image color.

mueller-ma:
Using the icon color attribut isn't flexible enough, e.g. you can set the icon to black, but then all UIs must be in light mode.

lolodomo:
There is probably already a default color set in CSS for icon as the colors for non OH icons are different depending on the theme. I am going to check that.

mueller-ma:
> For OH icons source, BasicUI uses a HTML img tag to render the icon and at my knowledge it is not possible to control the image color.

Right, it seems that the CSS is only applied, when the SVG is included in a `svg` tag: https://codepen.io/smashingmag/pen/MWQjEKN

> In fact, I would like to use the iconcolor attribute.

What's the exact behavior of this new attribute? Tint over the icon?

[...]

PR #1792, opened 2023-03-13
[BasicUI] Consider the correct iconset when reloading the icon

lolodomo:
Fix #1790

Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #1799, opened 2023-03-17
[BasicUI] Use inline SVG for "colorless" SVG icons

lolodomo:
Related to #1743

Use a HTML svg element (inline SVG) instead of a img element
when the icon servlet returns a SVG icon containing "currentColor".

Then it is possible to adjust the color of these icons by using the
iconcolor sitemapo attribute.


lolodomo:
This PR is currently based on PR #1792. I will rebase it when #1792 is merged.

The change is working well but I have still to check if cache is really working when using "fectch". It looks like it is not with Firefox.

The disadvantage of this change is that:

1. each icon is now retrieved twice when a new page is opened
2. the weight of the page when using inline SVG is bigger than when using img tags

This change will add nothing really new when using the classic iconset as all (or almost all ? - I have not checked all icons) icons have hardcoded colors. It starts being very interesting when adding custom colorless SVG icons.

I could also make this new feature an option of Basic UI.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -349,36 +355,78 @@
...
+			_t.icon = _t.parentNode.parentNode.querySelector(o.formIconSvg);
+		};
+
+		_t.getSVGIconAndReplaceWithInline = function(srcUrl, checkCurrentColor, defaultSVG) {
+			// fetch(srcUrl, { cache: "force-cache" }).then(function(response) {
+			fetch(srcUrl).then(function(response) {
+				if (response.ok && response.headers.get("content-type") === "image/svg+xml") {
+					response.text().then(function(data) {
+						if (!checkCurrentColor || data.indexOf("currentColor") !== -1) {
+							_t.replaceIconWithInlineSVG(data);
```
dilyanpalauzov, line 392: Instead of `currentColor`, aren’t `Canvas` or `CanvasText` also valid, relative (system) colors, which shall be considered when embedding?

lolodomo: Do you mean a canvas can be embedded inside a SVG icon? 
Can you please provide an example?

dilyanpalauzov: https://developer.mozilla.org/en-US/docs/Web/CSS/system-color contains a list of so called “system colors”.  With `Canvas` I mean one of these system colors, not a surface for drawing.

That is: `style="background:Canvas;color:CanvasText"` and `style="background:CanvasText;color:Canvas"` are valid in SVG and can be used, when the embedding system does not handle `prefer-color-scheme` or `color-scheme`. ( I have tried the to send emails that are supposed to be rendered in dark mode, when the email client does not understand the notion of dark mode.  I had limited success).

lolodomo: So you would like an inline SVG in case the SVG contains Canvas or CanvasText in addition to currentColor?

dilyanpalauzov: I mean in theory SVGs can contain any “system color“.  It is up to you to handle the system colors, as they are intended.

Whether SVGs are embedded or not, I do not know.

lolodomo: But support of system colors is not the purpose of this PR. 
It looks like an additional feature request to be coveted by another PR. 
Please fill a feature request if not yet done.

kaikreuzer:
@lolodomo You asked for this to be reviewed, but the PR is still in DRAFT mode. Is this by accident?

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -349,36 +355,78 @@
...
+				newIconElement.setAttribute("data-icon", dataIcon);
+			}
+
+			// Replace the current icon element with the built inline SVG
+			_t.iconContainer.replaceChild(newIconElement, _t.icon);
+			_t.icon = _t.parentNode.parentNode.querySelector(o.formIconSvg);
+		};
+
+		_t.getSVGIconAndReplaceWithInline = function(srcUrl, checkCurrentColor, defaultSVG) {
+			// fetch(srcUrl, { cache: "force-cache" }).then(function(response) {
```
kaikreuzer, line 387: You might want to remove this line.

kaikreuzer (changes requested):
This looks like some voodoo, but I trust you have tested it. 😄 
Just one small comment below.

lolodomo:
> @lolodomo You asked for this to be reviewed, but the PR is still in DRAFT mode. Is this by accident?

My wish was to merge it before new milestone.
But I will add an option to let the user choose if he wants this feature or not. I will finish the PR probably Friday.

lolodomo:
@kaikreuzer : PR is now fully ready for a merge.

With this sitemap:
```
	Frame label="Text icon color" {
		Default item=TestNumber icon="mic" iconcolor=[==56="blue", "yellow"]
		Default item=SetpointTemp icon="temperature" iconcolor=["yellow"]
	}
```
and the mic.svg a custom icon containing currentColor as the fill color.

With the feature enabled: => inline SVG and color considered
image
image

With the feature disabled: => IMG and color ignored
image

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: Done

kaikreuzer (approved):
Thanks!

PR #1807, opened 2023-03-27
Add a SSE reconnection mechanism

florian-h05:
Currently, on SSE failure there is no reliable reconnection mechanism (browsers might try something), e.g. on network failure, and therefore SSE connection is lost forever once an error occurs. To reconnect, the user has to change/reload the page.

This implements a SSE reconnection mechanism, that attempts a SSE reconnection after SSE failure.
The time is increasing from 1 -> 2 -> 4 -> 8 -> 10.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/sse.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/sse.js
@@ -73,6 +73,9 @@ function newSSEConnection (path, readyCallback, messageCallback, errorCallback, 
       eventSource.clearKeepalive()
       eventSource.keepaliveTimer = setTimeout(() => {
         console.warn('SSE timeout error')
+        if (heartbeatCallback) {
+          heartbeatCallback(false)
+        }
       }, (seconds + 2) * 1000)
     }
 
--- a/bundles/org.openhab.ui/web/src/js/openhab/sse.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/sse.js
@@ -5,6 +5,7 @@ let openSSEClients = []
 
 function newSSEConnection (path, readyCallback, messageCallback, errorCallback, heartbeatCallback) {
   let eventSource
+  let reconnectSeconds = 1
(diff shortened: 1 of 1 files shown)


ghys:
This is critical and very sensitive code but it also seems like with those changes it received proper care and should therefore be tested properly, so I cannot do that right now, but will do!
Thanks for the time spent on this. I especially like this:
> The time is increasing from 1 -> 2 -> 4 -> 8 -> 10.

bundles/org.openhab.ui/web/src/js/openhab/sse.js
```diff
@@ -17,61 +18,74 @@ function newSSEConnection (path, readyCallback, messageCallback, errorCallback,
...
     }
-  }
 
-  eventSource.setKeepalive = (seconds = 10) => {
-    console.debug('Setting keepalive interval seconds', seconds)
-    eventSource.clearKeepalive()
-    eventSource.keepaliveTimer = setTimeout(() => {
-      console.warn('SSE timeout error')
-      if (heartbeatCallback) {
-        heartbeatCallback(false)
```
digitaldan, line 63: Looks like `heartbeatCallback(false)` is now not called at all, should this be moved into the `onerror` function below?

florian-h05: I just realized that I had the wrong train of thought when I removed it.
I will revert this change.

I don’t think it should be moved to onerror. AFAIK onerror is triggered on something different then heartbeat timeout.

florian-h05: @digitaldan I have reverted that change, can you please have another look and approve if everything is okay?

digitaldan (approved):
LGTM

ghys (approved):
I'll admit I didn't do the exhaustive testing I wanted, but I don't think we should postpone it any further either since it seems quite the improvement - and the code looks good :+1:

PR #1808, opened 2023-03-27
Show warning if SSE connection or send command fails

florian-h05:
Depends on #1807.

Currently, SSE connection and sending commands fail silently, which makes the user think that everything is fine, but instead outdated values are displayed and commands are not sent.

This implements user warnings for such cases, a large openHAB orange toast "Communication failure" with the option to reload is displayed at the bottom center: 
- as long as SSE connection is broken
- when sending a command to an Item failed for 5 sec

If the Item does not exist (error code 404), "<Itemname> not found" is displayed for 5 sec

### DIFF
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -286,7 +286,7 @@ html
 .disable-user-drag
   user-drag none
 
-.toast-sse-connection-failed
+.failure-toast
   --f7-toast-bg-color var(--f7-theme-color)
   --f7-toast-text-color white
   --f7-button-text-color white
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -287,7 +287,8 @@ html
   user-drag none
 
 .failure-toast
-  --f7-toast-bg-color var(--f7-theme-color)
+  background-color #e64a19 !important
+  font-size 22px !important
   --f7-toast-text-color white
   --f7-button-text-color white
   --f7-button-border-color white
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -287,7 +287,7 @@ html
   user-drag none
 
 .failure-toast
-  background-color #e64a19 !important
+  background-color var(--f7-theme-color) !important
   font-size 22px !important
   --f7-toast-text-color white
   --f7-button-text-color white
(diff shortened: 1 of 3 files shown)


digitaldan:
Hi @florian-h05 , thanks for picking this up!  Sounds like a good approach, one thing i would want to check is to make sure the Mobile client is not also displaying an error message on top of this one.  I'm fine either having the IOS not show something, or possibly have the UI check to see if a mobile client is handling this message (like we do for showing or not showing the top right "other apps" button) .  I 'll find some time to play  around as i have not really thought this through.

florian-h05:
I’ve already thought of this, the SSE warning is not displayed if `window.OHApp` is available. The command warning is displayed, but if the iOS displays its own message, the iOS toast is displayed on top, so I don’t see a problem there. 

FYI #1807 is required to have this PR properly work (the SSE warning is only persistent if the reconnection mechanism is in place).

florian-h05:
Screenshots:

image

image

bundles/org.openhab.ui/web/src/css/app.styl
```diff
@@ -285,3 +285,10 @@ html
 // Disable user dragging of background images in canvas layouts
 .disable-user-drag
   user-drag none
+
+.failure-toast
+  background-color #e64a19 !important
+  font-size 22px !important
```
ghys, line 291: I would really prefer finding a Framework7 CSS variable (https://v5.framework7.io/docs/css-variables) instead of hardcoding a color, also I'm not sure about this 22px font size :sweat_smile: 

There are other options for example a centered toast if the app is effectively unusable when the connection is broken (from https://v5.framework7.io/docs/toast#examples):

Screenshot from 2023-04-05 21-21-06

Or maybe a Notification (https://v5.framework7.io/docs/notification#examples)?

Screenshot from 2023-04-05 21-33-10

florian-h05: This color is the openHAB orange, I have switched to using the `--f7-theme-color` CSS variable instead of hard-coding it.
I have chosen the 22px font size to make the warning larger.

Regarding the centered toast: I really don‘t like its look without having an icon, and it seems that we can‘t have both icon and reload button. A toast at the bottom seemed the visually best solution to me.

I haven‘t really looked into the notifications, but IMO we should not use them and instead „reserve“ them for real notifications when MainUI supports Web Push one day.

ghys: Thanks for the answer, I don't mind your choices but had to express mine ;)

ghys (approved):

PR #1824, opened 2023-04-03
Sitemap editor: Support `inputHint` config of Input element

mherwege:
This adds support for configuring the inputHint parameter of the Input sitemap element in the sitemap editor of MainUI.
See https://github.com/openhab/openhab-core/pull/3418 and the BasicUI implementation #1787.


florian-h05 (approved):
LGTM, except that the `inputHint` parameter validation is not working. I see the problem, I'll fix that myself.

Issue #1831, opened 2023-04-09
[BasicUI] More contrast in dark mode


dilyanpalauzov:
I use openHAB 3.4 and Basic UI in Dark mode with Firefox.  This is a screenshot at the top level (outside of any groups)

![base1](https://user-images.githubusercontent.com/4992947/230761924-f45b29b3-26e3-4f36-bb46-8bbf6100a06d.png)

This is a screen shot of a group:

![base2](https://user-images.githubusercontent.com/4992947/230761942-845e3eca-2414-402d-8d74-6318aca99e91.png)

To be able easily to perceive the text the contrast between text color and text background shall be high.  In the first screenshot the contrast is low.  In the second screenshot the text is lighter and therefore the contrast is better. In the sitemap file I have no settings which change the color on that particular items.

The second screenshot shows (on the left) a switch and on the right 
```
Switch item=rd_manual mappings=[ON="Ръчно", OFF="Автоматично"] labelcolor=[OFF="green", ON="red"]
Setpoint item=rd_on_watts minValue=400 maxValue=3500 step=100
```
The UP and DOWN buttons, and the ON state (Ръчно) are displayed with black text on dark gray background.  This has very low contrast and therefore is hard to read.

* In the main sitemap (first screenshot, outside of any group) increase the contrast between text color and text background
* In the Setpoint and Switch mappings=[ON="…", OFF="…"] items (second screenshot right half), and Switch without mapping (second screenshot left half) increase the contrast between text and background for the buttons.

lolodomo:
I will try to customize the colors.

lolodomo:
> * In the main sitemap (first screenshot, outside of any group) increase the contrast between text color and text background

I do not understand what is this case ? Is it outside any group or outside any frame ?
Can you provide a very small sitemap showing that ?
I have good contrast in my main sitemap page.
Maybe you are using an unexpected sitemap feature ???

dilyanpalauzov:
Sample zz.sitemap:
```
sitemap zz label="A3" {
    Text item=inv1_Battery_capacity
    Group item=r1_power label="1 Реле" visibility=[r1_online==OPEN] valuecolor=[r1_color=="green"="green", r1_color=="lime"="lime", r1_color=="red"="red", r1_color=="maroon"="maroon"] {
      Text item=inv1_Battery_capacity
    }
}
```
Main view
[...]

PR #1834, opened 2023-04-10
[BasicUI] Use more contrasted color buttons in widgets when in dark mode

lolodomo:
Applied to colopicker, setpoint, switch/rollerblind and switch/mappings widgets.

Also change the track color of the switch widget when state is OFF

Related to #1831


lolodomo:
Here is the new rendering:
image
image

Compared to before:
image
image

kaikreuzer (approved):
Thanks!

PR #1838, opened 2023-04-11
[BasicUI] Change assumption in case icon value contains only 2 segments

lolodomo:
The first segment is now expected to be the icon source.
The icon set is determined: "classic" if source is "oh" or "" in other cases.


kaikreuzer:
For my understanding: What does it mean if the iconSet is set to ""? Shouldn't it rather be kept to be `DEFAULT_ICON_SET` in that case?

lolodomo:
For other sources than openHAB, I can't guess what could be a good default icon set.
I chosen "" because for material icons there is no icon set. But thinking about it, yes I could use DEFAULT_ICON_SET in all cases, it will make no difference.

kaikreuzer (approved):
Thanks!

PR #1842, opened 2023-04-14
[BasicUI] Fix text color when the page does not contain any frame

lolodomo:
Use the same text color (depending on theme) as a frame was present.

Related to #1831


lolodomo:
This was fast :)

Issue #1839, opened 2023-04-12
Items icons not displayed in all pages


lolodomo:
## The problem

The syntax for item icon has been extended in core framework with PR openhab/openhab-core#3539.

The icon value can now contain until 3 segments separated by a colon (":").
First segment is the icon source. Example: oh, if, iconify, material, f7, ...
Second segment is the icon set. Example: classic
Third segment is the icon name (and can contain hyphen). Example: temperature
In case only two segments are provided, the first segment is the icon source and the second the icon name. "classic" icon set is assumed.
In case only one segment is provided, the icon source is assumed to be the openHAB server and its classic icon set and the value is then the icon name.

MainUI already supports this new syntax in certain pages but not all.

Here are examples of item icon values that should work everywhere in MainUI:

1. temperature
2. oh:humidity
3. oh:classic:pressure
4. material:arrow_drop_down
5. material:favorite
6. iconify:wi:day-sunny-overcast
7. if:wi:day-thunderstorm

Only "temperature" icon is displayed in the equipment tab of MainUI (opening the "Web services" pop-up). The others should be displayed too.
![image](https://user-images.githubusercontent.com/10186704/231412107-2a12c57f-9a14-48f9-aed7-dd59b1dfa8e6.png)

Icon "oh:classic:pressure" is shown nowhere, neither in the items page, nor the item page, nor the edit item page and nor the model page.

Icon "material:favorite" is not shown in the item page (but OK in the items page,  the edit item page and the model page):
![image](https://user-images.githubusercontent.com/10186704/231532361-7dcfe2bf-b646-4846-8579-d4ef2534a2e4.png)

Icons "iconify:wi:day-sunny-overcast" and "if:wi:day-thunderstorm" are not shown in the item page (but OK in the items page,  the edit item page and the model page):
![image](https://user-images.githubusercontent.com/10186704/231532809-38e0319c-fed0-4c1f-8eae-3d4374c3fd6d.png)

## Expected behavior

MainUI should support the new icon syntax with 3 possible segments in all places.
It includes:

1. the support of "oh:classic:value" as icon value
2. the display of the icon in the item page for iconify and material icons
3. the display of the icon in the equipment tab when an icon source is part of the icon value

## Steps to reproduce

Here are example items I used for my tests:
```
Group GTest2 "Test items group 2" [ "WebService" ]

[...]

PR #1843, opened 2023-04-15
Sitemap and item config parsing adjustments

mherwege:
- Adapt sitemap and item lexers to changes in icon name syntax
- Restrict the elements that can be added to a sitemap
- Added extra sitemap validations (in line with xtext validation in core)
- Added test for sitemap parser and validation

This solves the issue in main UI created by https://github.com/openhab/openhab-core/pull/3539 and https://github.com/openhab/openhab-core/pull/3378.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/pages/sitemap/__tests__/sitemap-edit_jest.spec.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/pages/sitemap/__tests__/sitemap-edit_jest.spec.js
@@ -54,6 +54,7 @@ describe('SitemapEdit', () => {
     wrapper.vm.selectWidget([wrapper.vm.sitemap.slots.widgets[0], wrapper.vm.sitemap])
     await wrapper.vm.$nextTick()
     wrapper.vm.addWidget('Frame')
+    await wrapper.vm.$nextTick()
     wrapper.vm.selectWidget([wrapper.vm.sitemap.slots.widgets[0].slots.widgets[0], wrapper.vm.sitemap])
     await wrapper.vm.$nextTick()
     localVue.set(wrapper.vm.selectedWidget.config, 'label', 'Frame Test')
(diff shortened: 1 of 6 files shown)


florian-h05 (changes requested):
I have tried to add Items from textual definition with the example from the core PR, however the icon value is not recognized and therefore not set. It seems that the functions to process the value are missing in items-lexer.nearley as well as for sitemap-lexer.nearley.

lolodomo:
Look at #1839. Would be great if it could be fixed before tomorrow.

florian-h05:
#1839 is a different issue, the problem with the Item lexer is that whatever you type as icon name is not saved to the Item, because the lexer does not work properly.

mherwege:
I will need to check again. I thought it was working for me with the examples from the core PR and this this PR.  If it is not for you, there may still be something wrong. Unfortunately, I will not be able to do it before tomorrow night. I don’t have access to my PC at the moment.

J-N-K:
Is this PR only needed to make the new feature work or is it needed to fix something?

lolodomo:
> Is this PR only needed to make the new feature work or is it needed to fix something?

Probably only if you want to import items containing the extended syntax with several segments for the category/icon ?

florian-h05:
> Is this PR only needed to make the new feature work or is it needed to fix something?

It seems it is only needed to to allow the new icon syntax in the „Add Items from textual definition“ and Sitemap editor’s codetab.

J-N-K:
Ok, so it‘s not blocking M2.

florian-h05:
BTW, I‘ve just checked what the transformation editor in the UI does after the recent core changes. It needs some adjustment to be fully operational, but IMO it‘s non-blocking since nothing really broke. I‘ve already worked a bit on the adjustment, should be able to open a PR soon.

mherwege:
I know it is a bit late for the milestone, but I committed a fix to the parser.
Note that it also contains a fix for the syntax of an identifier in the items parser (bringing it in line with the one in the sitemap parser). The items parser was accepting more than it should have. This wasn't a real issue as long as the provided items file was valid, but would also import with invalid identifiers.

florian-h05:
@ghys Feel free to take over the review, I won’t have the time to finish it anytime soon.

bundles/org.openhab.ui/web/src/pages/settings/pages/sitemap/__tests__/sitemap-edit_jest.spec.js
```diff
@@ -186,39 +350,4 @@ describe('SitemapEdit', () => {
     wrapper.vm.validateWidgets()
     expect(lastDialogConfig).toBeFalsy()
   })
-
-  it('validates valuecolor', async () => {
```
florian-h05, line 190: Why did you remove that test?

florian-h05 (commented):
Code LGTM & everything works fine now!

I have just one minor question regarding the unit tests, thanks for improving them!

bundles/org.openhab.ui/web/src/pages/settings/pages/sitemap/__tests__/sitemap-edit_jest.spec.js
mherwege: It shouldn't have been removed. I don't know what happened there. I put it back in.

florian-h05 (approved):
LGTM, thanks!

PR #1847, opened 2023-04-18
Remove redundant modifiers from interfaces

wborn:
This removes redundant modifiers from the interfaces similar to openhab/openhab-core#3560.


kaikreuzer (approved):
lgtm, thanks!

PR #1850, opened 2023-04-18
Dynamically load the list of semantic tags

jimtng:
Depends on https://github.com/openhab/openhab-core/pull/3559 (already merged now)

This is the first time I've ever worked with Vue and modern javascript, so I'd appreciate any tips, thanks!

Adding custom semantic tags is now possible: https://github.com/openhab/openhab-core/pull/3519 

This PR loads and and uses the custom tags in the Main UI. As a bonus, the hard coded semantic tags + all the translations are removed from mainui, so they no longer need to be maintained separately. They are now loaded from core through the REST API.

I could use some help in testing this PR because I am not familiar with everything that may be affected. To test, install this jar file from here: https://github.com/jimtng/openhab-webui/releases/tag/semantics-0.0.1 onto the latest snapshot

Even if you don't have any custom semantics, at least test to make sure that everywhere the semantic tags are used, work as expected. I am not entirely familiar with the MainUI especially around its page builder.

image

image


jimtng:
Rebased

florian-h05:
@jimtng Can you please rebase again?

Before starting to review this, I'd like to get @ghys opinion regarding your taken approach because I am not sure if this is the way to go or if using a VueX store would be a better solution.

jimtng:
Rebased. If there's a better way, I'll be happy to close this. I really don't know anything about Vue.

ghys:
- I was once quite rudely made aware that "Equipments" was not a word - "equipment" has no plural form: https://www.britannica.com/dictionary/eb/qa/equipments-equipment-noncount-mass-noun-singular-plural#:~:text=Answer,is%20no%20word%20%22equipments.%22
- if the semantic tags are now fetched dynamically, then architecturally-wise they don't belong in the `assets` folder anymore, theoretically that is for non-dynamic things bundled with the UI code; it would be better to make it a service of sorts, perhaps in the `js/openhab` folder.
- as @florian-h05 says making it part of the Vuex store is best if it's ultimately user-controllable

jimtng:
> * if the semantic tags are now fetched dynamically, then architecturally-wise they don't belong in the `assets` folder anymore, theoretically that is for non-dynamic things bundled with the UI code; it would be better to make it a service of sorts, perhaps in the `js/openhab` folder.

I agree with this and I suppose this would be a trivial change.

> * as @florian-h05 says making it part of the Vuex store is best if it's ultimately user-controllable

This part I have zero knowledge of. Is there an example of other things that have implemented Vuex store? I can only find `// TODO use a Vuex store` scattered here and there.

florian-h05:
Unfortunately I haven’t much experience with Vuex either, but the Vuex Stores are defined here:

https://github.com/openhab/openhab-webui/tree/main/bundles/org.openhab.ui/web/src/js/store

You can access the store with `this.$store`.

jimtng:
> Unfortunately I haven’t much experience with Vuex either, but the Vuex Stores are defined here:
> 
> https://github.com/openhab/openhab-webui/tree/main/bundles/org.openhab.ui/web/src/js/store
> 
> You can access the store with `this.$store`.

Yes, I've already found that, but nothing has been implemented using it to model from.

florian-h05:
@jimtng I am currently didding into Vuex, I think I can open a PR soon.

jimtng:
Thank you! I'll close this PR.

PR #1853, opened 2023-04-19
[habpanel] Document scene activation using the selection widget

osama-salah:
Added a description of how to use the "Selection" widget for changing scenes


stefan-hoehn (changes requested):
The text in general is fine, however based on our formatting rules, every line should start in a new line (they will still rendered as one paragraph).

cc: @florian-h05

bundles/org.openhab.ui.habpanel/doc/habpanel.md
```diff
@@ -208,6 +208,8 @@ The knob widget is similar in essence to the slider, but in a rotary fashion. It
...
 
+To display different scenes on a Selection widget (like the image above), you could create an item with type `String` (e.g. call it modes). Then create different rules for you different scenes (i.e. a rule per scene) such as 'meditation', 'work', 'sleep' ... etc. The Trigger (i.e. When) of every rule should be `an item receives a command` and every rule's triggering Item should be the one that you have just created (the one we called 'modes') and the Command string could be any string of your choice (e.g. meditation, work, sleep). Finally, in your HABPanel Selection widget, you could send any of these commands to that item by setting the `List of comma-separated choices` to the list of the rules that you have just created. Selecting any of them will trigger the corresponding rule. To display a different label for a command, you could use the form 'command=label' (e.g. meditation=Meditation,work=Work). Please, refer to [this link](https://www.openhab.org/docs/tutorial/rules_basic.html#create-the-rule) for how to create a rule.
```
florian-h05, line 211:
```suggestion
To display different scenes on a Selection widget (like the image above), you could create an item with type `String` (e.g. call it modes).
Then create different rules for you different scenes (i.e. a rule per scene) such as 'meditation', 'work', 'sleep' ... etc. 
The Trigger (i.e. When) of every rule should be `an item receives a command` and every rule's triggering Item should be the one that you have just created (the one we called 'modes') and the command string could be any string of your choice (e.g. meditation, work, sleep).
Finally, in your HABPanel Selection widget, you could send any of these commands to that item by setting the `List of comma-separated choices` to the list of the rules that you have just created.
Selecting any of them will trigger the corresponding rule.
To display a different label for a command, you could use the form `command=label` (e.g. `meditation=Meditation, work=Work`).
Please refer to [this link](https://www.openhab.org/docs/tutorial/rules_basic.html#create-the-rule) for how to create a rule.
```

florian-h05 (approved):
LGTM, thanks!

PR #1865, opened 2023-04-29
Use right UID attribute in Thing picker

stefan-hoehn:
Thing picker sort uses wrong UID attribute as fallback during sort of thing list.

Follow-up for #1837.


florian-h05 (approved):
Thanks for the fix!

PR #1868, opened 2023-05-01
[blockly] Toggle between item name or label

stefan-hoehn:
This PR allows to show either the name or the label of the item.

Since the beginning the item picker would show the label during selection but the item block itself then would show the name (hence the "id" that is used internally). This always confused me because I rather remember the labels rather the "internal" names of the item. From the implementation perspective it is now clear to me why this was done (it was much easier to generate the code) and also it is consistent with what is used for code generation. 

This new way allows the user of the blockly editor to switch between showing the label or the name while the code generation always uses the name.

Note that this requires both the name and the label stored in the blockly code. Because the current code does not store the label, it needs to repicked once, via the item picker, to retrieve that label and then store it. I may add automatic retrieval for the label in the future but due to some internal limitations it doesn't seem that easy to provide clear code separation for that.

The switch can be found on the lower bar of the editor

image

The functionality behaves as follows

https://user-images.githubusercontent.com/5937600/235449639-8cbef4c5-5f19-4d33-a591-2791cfc6f52d.mov

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
@@ -21,10 +21,9 @@ export default function (f7, isGraalJs) {
       this.setColour(160)
       this.setInputsInline(true)
 
-      let thisFieldpicker = this.fieldPicker
       this.setTooltip(() => {
         let tooltip = 'Pick an Item from the Model'
-        const itemData = thisFieldpicker.data
+        const itemData = this.fieldPicker.data
         if (itemData[0] !== itemData[1]) {
           tooltip = itemData[0]
         }
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
@@ -7,6 +7,7 @@ import Blockly from 'blockly'
 import { javascriptGenerator } from 'blockly/javascript'
 import { FieldItemModelPicker } from './fields/item-field'
 import { FieldThingPicker } from './fields/thing-field'
+import api from '@/js/openhab/api'
 
 export default function (f7, isGraalJs) {
   /* Helper block to allow selecting an item */
@@ -40,6 +41,11 @@ export default function (f7, isGraalJs) {
 
       if (!this.fieldPicker.data) { // "migrate" old storage
         this.fieldPicker.data = [this.fieldPicker.value_, this.fieldPicker.value_]
+        if (this.fieldPicker.value_ && this.fieldPicker.value_ != 'MyItem') {
+          api.get(`/rest/items/${this.fieldPicker.value_}?metadata=^$`).then((data) => {
+            this.fieldPicker.data = [this.fieldPicker.value_, data.label]
+          }).catch()
+        }
       }
       this.fieldPicker.value_ = (this.workspace.showLabels) ? this.fieldPicker.data[1] : this.fieldPicker.data[0]
 
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
@@ -41,7 +41,7 @@ export default function (f7, isGraalJs) {
 
       if (!this.fieldPicker.data) { // "migrate" old storage
         this.fieldPicker.data = [this.fieldPicker.value_, this.fieldPicker.value_]
-        if (this.fieldPicker.value_ && this.fieldPicker.value_ != 'MyItem') {
+        if (this.fieldPicker.value_ && this.fieldPicker.value_ !== 'MyItem') {
           api.get(`/rest/items/${this.fieldPicker.value_}?metadata=^$`).then((data) => {
             this.fieldPicker.data = [this.fieldPicker.value_, data.label]
           }).catch()
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
@@ -11,20 +11,49 @@ import { FieldThingPicker } from './fields/thing-field'
 export default function (f7, isGraalJs) {
   /* Helper block to allow selecting an item */
   Blockly.Blocks['oh_item'] = {
+    fieldPicker: null,
     init: function () {
+      this.fieldPicker = new FieldItemModelPicker('MyItem', null, { f7 })
       this.appendDummyInput()
         .appendField('item')
-        .appendField(new FieldItemModelPicker('MyItem', null, { f7 }), 'itemName')
+        .appendField(this.fieldPicker, 'itemName')
       this.setColour(160)
       this.setInputsInline(true)
-      this.setTooltip('Pick an item from the Model')
+
+      let thisFieldpicker = this.fieldPicker
+      this.setTooltip(function () {
+        let tooltip = 'Pick an item from the Model'
+        const itemData = thisFieldpicker.data
+        if (itemData[0] !== itemData[1]) {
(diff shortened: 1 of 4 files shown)


florian-h05 (changes requested):
Overall LGTM, please have a look at my comments.

Before merging this, I'd also like to have a look at getting the Item labels for the "migration".

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
```diff
@@ -11,20 +11,49 @@ import { FieldThingPicker } from './fields/thing-field'
...
       this.setOutput(true, 'oh_item')
+    },
+    update_: function (name, label) {
+      this.fieldPicker.data = [name, label]
+    },
+    mutationToDom: function () {
+      let container = Blockly.utils.xml.createElement('mutation')
+
+      if (!this.fieldPicker.data) { // "migrate" old storage
+        this.fieldPicker.data = [this.fieldPicker.value_, this.fieldPicker.value_]
```
florian-h05, line 42: We could get the label from the REST API:

```suggestion
        api.get(`/rest/items/${this.fieldPicker.value_}?metadata=^$`).then((data) => {
          this.fieldPicker.data = [this.fieldPicker.value_, data.label]
        }).catch()
```

You need this import on top:

```javascript
import api from '@/js/openhab/api'
```

BUT there are currently other issues when trying to make this work, so please address my review in one commit and for this here an additional commit.

bundles/org.openhab.ui/web/src/components/config/controls/blockly-editor.vue
```diff
@@ -1191,6 +1198,12 @@ export default {
         this.workspace.registerToolboxCategoryCallback('LIBRARY_' + definition.uid, defineLibraryToolboxCategory(definition, this.$f7))
       })
     },
+    showHideLabels (showLabels) {
+      this.workspace.showLabels = showLabels
+      this.workspace.refresh()
+      this.workspace.loadItem('test')
+      debugger
```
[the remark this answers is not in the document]
stefan-hoehn: no, for sure, how embarrassing! 🤦

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
stefan-hoehn: To be honest, this wasn't even meant to be committed at all because I failed to get it running. So I will uncomment it for the time being for you to jump in.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
```diff
@@ -11,20 +11,49 @@ import { FieldThingPicker } from './fields/thing-field'
...
+        const itemData = thisFieldpicker.data
+        if (itemData[0] !== itemData[1]) {
+          tooltip = itemData[0]
+        }
+        return tooltip
+      })
       this.setHelpUrl('https://www.openhab.org/docs/configuration/blockly/rules-blockly-items-things.html#item')
       this.setOutput(true, 'oh_item')
+    },
+    update_: function (name, label) {
```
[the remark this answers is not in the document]
stefan-hoehn: ok, then we need this to be updated in domMutation as well.

ghys:
I'm thrilled for this change but I invite you to reconsider how you want to design it:

The bottom toolbar is already cluttered on phone screen dimensions and won't accept more additions IMO:

image

florian-h05:
@ghys Regarding the retrieval of Item labels: I see one problem with storing the labels in Blockly, because changes to the Item label would not be reflected. 
What do you think of having a Vuex Store for Items? That would also save us many requests for the Item pickers? When your core PR for cacheable Items is finished and merged, we could check for changes to the Items everytime the Vuex store getter is called, and refresh if there are any changes.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
```diff
@@ -7,24 +7,59 @@ import Blockly from 'blockly'
...
+      this.fieldPicker = new FieldItemModelPicker('MyItem', null, { f7 })
       this.appendDummyInput()
         .appendField('item')
-        .appendField(new FieldItemModelPicker('MyItem', null, { f7 }), 'itemName')
+        .appendField(this.fieldPicker, 'itemName')
       this.setColour(160)
       this.setInputsInline(true)
-      this.setTooltip('Pick an item from the Model')
+
+      let thisFieldpicker = this.fieldPicker
```
florian-h05, line 24: Awwor functions don't have their own `this` binding, so you can directly access `this.fieldPicker` and don't need another reference.

```suggestion

```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
```diff
@@ -7,24 +7,59 @@ import Blockly from 'blockly'
...
+      this.setTooltip(() => {
+        let tooltip = 'Pick an Item from the Model'
+        const itemData = thisFieldpicker.data
```
florian-h05, line 27:
```suggestion
        const itemData = this.fieldPicker.data
```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/fields/item-field.js
```diff
@@ -18,7 +18,8 @@ export class FieldItemModelPicker extends Blockly.FieldTextInput {
   showEditor_ (options) {
     if (this.f7) {
       const itemsPicked = (value) => {
-        this.setEditorValue_(value.name)
+        this.data = [value.name, value.label]
+        this.setEditorValue_(value.label)
```
florian-h05, line 22:
```suggestion
        this.setEditorValue_(value.name)
```

To have the default behavious of showing the Item also for newly picked Items.

bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
```diff
@@ -25,6 +25,9 @@
                  :tooltip="rule.status.description" />
       </span>
       <span class="display-flex flex-direction-row align-items-center">
+        <f7-segmented v-if="!newScript && isBlockly" class="margin-right">
```
florian-h05, line 28: Let's remove the `segmented` as it is not needed here:

```suggestion

```

bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
```diff
@@ -25,6 +25,9 @@
...
+          <f7-button outline small :active="blocklyShowLabels" icon-f7="square_on_circle" :icon-size="($theme.aurora) ? 20 : 22" class="no-ripple" @click="toggleBlocklyItemLabelId" :tooltip="'Toggle to show either item labels or IDs'" />
```
florian-h05, line 29: ... and hide the button in code preview as it is not working there (error thrown):

```suggestion
        <f7-button v-if="!newScript && isBlockly && !blocklyCodePreview" outline small :active="blocklyShowLabels" icon-f7="square_on_circle" :icon-size="($theme.aurora) ? 20 : 22" class="no-ripple margin-right" @click="toggleBlocklyItemLabelId" :tooltip="'Toggle to show either item labels or IDs'" />
```

florian-h05 (changes requested):
Just a few minor comments, then we are done.

Have you considered @ghys concerns with layout?
I could imagine having the two-button thing for the code preview replaced by a single button with changing icon to save space and don't run into problems on phone screens, even though I don't think many use Blockly on their phone.

florian-h05 (approved):

florian-h05 (commented):
LGTM, thanks!

The buttons in the bottom bar are usable on phones, even though the bottom bar is „full“, but I don‘t see a big problem here since Blockly IMO is not really usable on small phone screens and therefore unlikely to be used there.

PR #1870, opened 2023-05-01
[blockly] Support storing timers in shared cache

### LINKED ISSUE #1858: [Blockly] Option to save a Timer in the shared cache
## The problem

Many use cases, particularly for users migrating from Rules DSL, are addressed by having one Timer accessed and controlled from multiple rules (e.g. a condition that won't run the rule actions if the timer exists). 

## Your suggestion

Provide some way to choose to save and retrieve Timers in the shared cache, not just the private cache. I'm hoping there's a way to make this possible using the exiting blocks, maybe just providing an option to tell the block which cache to use. I don't like the idea of duplicating all the Timer blocks.

## Your environment

```yaml

```

stefan-hoehn:
fixes #1858 

Allows to decide if a timer is either private or shared. Shared timers can be used in different rule than it was created

image


florian-h05 (approved):
LGTM, thanks!

I have improved the labelling of the create timer block to be inline with the NashornJS version:

image

PR #1849, opened 2023-04-18
Fix icons not displayed & not dynamic on all pages

florian-h05:
Fixes #1839.
Fixes #574.
Closes #1860.

* Fixes Item icon styling on the Item detail page.
* Add support for openHAB iconsets.
* Fix `oh:` icons not properly displayed in `default-list-item.vue`.
* Refactor `oh-icon` config & style binding.
* Enable real-time state on the semantic model page.
* Enable state for the channel link edit page.
* Enable dynamic icons in the settings where missing and possible (semantic model page, channel link edit page, Item edit page when setting category, Items list page).
* Enable dynamic icons in the default list widget for most Items except some and update the docs accordingly.
* Add a refresh button to the Item list page.

PR #1874, opened 2023-05-06
Enable dynamic icons only for some icon types in default list widget

florian-h05:
Follow-up for #1849.


florian-h05:
@ghys This sets the iconUseState to true only for Item types that have a "predictable" state. For groups, the group type is checked.

florian-h05:
@ghys 
So this is the only PR left for now that I can’t review myself, but I’ll try keeping you busy with more PRs 😉

ghys (approved):
Thanks! It's far better than enabling it in all cases IMO.
The added inline documentation is also very valuable so we know when these are enabled or disabled by default, and the user can make their own choice.

lolodomo:
For everyone, it is now enabled by default only for Contact', Dimmer, Rollershutter and Switch item types.

PR #1875, opened 2023-05-06
`oh-knob` & `oh-stepper`: Add offset functionality

florian-h05:
Supersedes #1565.

This adds an offset parameter to the knob and stepper components, that allows to add or subtract a number value to the displayed state of the component. 

This is useful for thermostats which have a fixed temperature of n and are controlled with an offset to that temperature n, e.g. you have a KNX HVAC system which has a fixed temperature of 20 °C: -2 would mean the target temperature is 18 °C, +2 would mean that the target temperature is 22 °C. By setting the offset parameter to 20, the UI will display the real target temperature instead of the -2 or 2.

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-knob-card.md
+++ b/bundles/org.openhab.ui/doc/components/oh-knob-card.md
@@ -81,6 +81,11 @@ Use the advanced properties to change the appearance from a knob to a rounded sl
     Minimum interval between values (default 1)
   </PropDescription>
 </PropBlock>
+<PropBlock type="DECIMAL" name="offset" label="Offset">
+  <PropDescription>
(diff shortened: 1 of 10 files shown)


florian-h05:
@ghys While implementing this, I noticed that the setpoint has an input field, which however does not work when setting min/max values (e.g. min is 14, max is 26: you want to input 22, but the moment you remove the second digit the value will become 14). 
I then decided to completely disable the input, WDYT of that decision?

ghys:
> I then decided to completely disable the input, WDYT of that decision?

Maybe it could upset some users who liked the feature for items which don't have min/max constraints. I'd say maybe make it optional, opt-in by default, but it's probably better to still have the option.

florian-h05:
Okay, thanks for the feedback. I will let you know when I'm finished :+1:

florian-h05:
@ghys Done!

ghys (approved):
Thanks, LGTM!

PR #1878, opened 2023-05-07
`foreground-service`: Listen to tab visibility change event

### LINKED ISSUE #1872: SIP does not unregister when page is left
I have a SIP widget. When I change within the Main UI from Overview to Location/Equipment/Property the SIP widget does unregister (sends REGISTER with Expires:0).  When I load the page, where thr SIP widget is shown, the SIP widget does again register.

The SIP widget shall unregister also when the page is left: tab is closed or a completely different page is typed in the URL.

openHAB 3.4.2, openhab-android 3.4.6

florian-h05:
Fixes #1872.

This makes the foreground service also listen to the `visibilitychange` event to stop and resume foreground activity when the tab is changed or the browser is closed.

See https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-foreground-service.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-foreground-service.js
@@ -21,6 +21,14 @@ export default {
 
     this.$f7.on('pageAfterIn', this.onPageAfterIn)
     this.$f7.on('pageBeforeOut', this.onPageBeforeOut)
+
+    document.addEventListener('visibilitychange', () => {
+      if (document.visibilityState === 'visible') {
+        this.startForegroundActivity()
+      } else if (document.visibilityState === 'hidden') {
+        this.stopForegroundActivity()
+      }
+    })
   },
   beforeDestroy () {
     this.$f7.off('pageAfterIn', this.onPageAfterIn)


florian-h05:
FYI @digitaldan

ghys (approved):
Cool! I'm all for avoiding things happening in the background when they're not useful, especially when the UI is used on a battery-powered device, like a phone or tablet.

florian-h05:
Totally agreed 👍 

I noticed this issue especially when using the SIP client in the iOS app: When I switched to another app, mic access stayed active. This is quite confusing and I don’t feel so good when the yellow indicator is displaying that something is listening to me.

ghys:
It was even a bigger problems for UIs that just listen to `/rest/events` SSEs unfiltered (like HABPabel, HABot or OH2's Paper UI) - in any decent-sized installation there's a lot of traffic there and most of it isn't useful for displaying the current page in the UI, so that's why I introduced the `POST /rest/events/states/*` endpoints for end-user pages and most settings pages maintain their own SSE connections to filter out unwanted events.

florian-h05:
> It could even be useful to close the SSE connection and resume it (with the current values) on visibilitychange.

I was not thinking about SSE when I created this PR, but very good points! I'm not sure what SSE currently does when the tab is switched, however thanks to my PRs with the toast message and the SSE reconnection mechanism I have noticed that once I leave Safari on iOS and come back, SSE has to reconnect so it seems that it is not always connected. 

> With this change, will the SIP widget remain active, when the user changes the tab, and later returns? It shall stay active.

With this change, it will unregister when the user changes the tab, and then reconnect once he returns. As long as the user is in another tab or has closed the browser window, the SIP widget is disconnected.

dilyanpalauzov:
> As long as the user is in another tab ... the SIP widget is disconnected.

This is not good. It shall be possible to receive calls in inactive tabs and it shall be possible during a call to change the tabs.

florian-h05:
There are also good arguments to have that behaviour. 

If you want the ability to not have this, feel free to implement something.

PR #1661, opened 2023-01-24
Use cacheable lists when possible

ghys:
Depends on https://github.com/openhab/openhab-core/pull/3335.

Will fetch items, things and rules with the `cacheable` parameter set
when appropriate.

Closes #1650.

Signed-off-by: Yannick Schaus <github@schaus.net>

PR #1882, opened 2023-05-08
Dynamically load the list of Semantic tags and store them in Vuex

### LINKED ISSUE #1822: Get the list of semantic tags from REST
## The problem

Currently the semantic tags are hard coded into semantics.js along with the translations in the i18n asset. Not only would this need to be (manually?) synchronized with the core, but it also prevents the ability to add additional semantic tags on the fly, a feature that currently doesn't exist, but can be done in the future. I have implemented such a feature in the jruby library, and perhaps one day it can be merged into core.

## Your suggestion

Create REST endpoints in core for providing the list of semantic tags + labels + relevant locales e.g. `/semantics/` `/semantics/locale_code` to replace the current hard coded list.

It will also get tag synonyms and descriptions

florian-h05:
Supersedes #1850.
Closes #1822.

Depends on https://github.com/openhab/openhab-core/pull/3559 (already merged now).
Adding custom semantic tags is now possible: https://github.com/openhab/openhab-core/pull/3519.

This PR loads the Semantic tags when MainUI is loaded the first time and stores them in Vuex.
This allows the removal of the hard-coded Semantic tags and the translations from the assets and therefore makes the initial JS smaller.


florian-h05:
There seems to be an issue that causes the translations not to be loaded, because the Language header is missing. I'll check how the rest of the UI loads correct translations and how to do it here.

ghys:
> and therefore makes the initial JS smaller.

Yes but it's usually put in cache whereas the additional REST query might not be :)

I see a lot of translations removed in this PR but I would like to have them transferred somehow to their new place, probably in openhab-core, so they are not lost.

florian-h05:
> whereas the additional REST query might not be :)

Yeah, but to put them in cache we’d need caching support in core (last-modified etc)

> I see a lot of translations removed in this PR but I would like to have them transferred somehow to their new place

There are already translations in openHAB core (see https://github.com/openhab/openhab-core/tree/main/bundles/org.openhab.core.semantics/src/main/resources), however I’m not sure if all translations from webui are there.

ghys:
Yes these are the synonyms for a tag (as you can see there could be more than one), and they could be complemented by the "synonyms" metadata.
The semantic-tag-to-localized-card-label is a different matter.

florian-h05:
> The semantic-tag-to-localized-card-label is a different matter.

Core seems to take the translations from those synonyms files, so when removing our translations here we should make sure that all of them are included in the synonyms translations.

florian-h05:
I will check which translations are available at openHAB Core:

- [ ] Bg
- [ ] Ca
- [ ] Cs
- [ ] Da
- [ ] De
- [ ] El
- [ ] En
- [ ] Es
- [ ] Fa
- [ ] Fi
- [ ] Fr
- [ ] He
- [ ] Hu
- [ ] It
- [ ] Iw
- [ ] Lb
- [ ] Lv
- [ ] Nl
- [ ] No
- [ ] Pl
- [ ] Pt-BR
- [ ] Pt
- [ ] Ru
- [ ] Sk
- [ ] Sl
- [ ] Sv
- [ ] Uk
- [ ] Zh

florian-h05:
@ghys Please see my comment. IMO we cannot block (UI) progress because of a few translations.

lolodomo:
@florian-h05 : openhab/openhab-core#3646 has been merged. 
Can you please check that it does not impact your PR?

florian-h05:
@lolodomo It required adjustments, your core PR was breaking.

@ghys This is ready for review and AFAIK the REST API for tags is finished, so we should not need any more updates/adjustments here. Can you please review and merge?

lolodomo:
> loads the Semantic tags when MainUI is loaded the first time

What do you mean by the first time? 
Isn't it each time MainUI is opened in a browser? 
As it is now possible to update the list of tags, this loading should happen at least at each MainUI startup.

florian-h05:
AFAIK the `loadData` method in app.vue is only called when MainUI loads or some events are emitted, so everything there (e.g. pages, sitemaps etc.) is not automatically refreshed from the server as long as you don‘t reload the browser tab or reload MainUI from the „Help & About“ page.

To automatically make those refresh in MainUI, they need to be loaded every time. We could use ETags or Last-Modified do cache responses to avoid unnecessary traffic, but I‘d rather do this in another PR and it probably also requires core changes. There already is a PR to use this for Items, Things and rules, see #1661.

ghys (approved):
> For Luxembourgish, you can probably translate?

Ech kann e bisschen lëtzebuergesch schwätzen... but it's not my forte...! (My German is better actually).

I'm good with the changes in the end, let's get it to OH4!

lolodomo:
I will test the next snapshot to see how it behaves with user semantic tags.

lolodomo:
@florian-h05 : it looks like it works well. I added (using REST API) two new locations and 1 new equipment. After restarting MainUI, I can see these new semantic tags when I create a new group in Main UI and finally the new locations and the new kind of equipment appear in Location and Equipment tabs. Very good.

Do you think you will find the time before OH4 is released to add a new page to manage semantic tags in Main UI ?

ghys:
A quick reminder that cache control should be implemented in core for the `GET /rest/tags` endpoint like in https://github.com/openhab/openhab-core/pull/3335 as the list is requested before the UI is even displayed. Caching the tags makes a lot of sense as it doesn't change often (or at all), to avoid about 17kb of avoidable traffic and a few milliseconds of delay. Every bit helps when using the UI remotely (OH Cloud or other).

See
https://github.com/openhab/openhab-core/commit/6e83d3f8def866af727e30c59b031a86ccbad18a#diff-4656a36e72716d544cfc5692a3e5409cd59b7f26ae9f8d97af4a9e18c43adc9a
for an example.
No `staticDataOnly` parameter is even necessary in this case as all data is static. This means no change should be needed in the UI code.

florian-h05:
> it looks like it works well.

👍 

> Do you think you will find the time before OH4 is released to add a new page to manage semantic tags in Main UI ?

Unfortunately I’m very busy this week and away next week (10th to 14th July), so I fear that I won’t make it before the feature freeze. I’m hoping to at least finish the UoM stuff for the new release.

florian-h05:
FYI: I have contacted the CrowdIn contributors a few seconds ago and asked them to re-contribute their translations at openhab-core.

@kaikreuzer I also noted that the removed translations from MainUI are still in CrowdIn (https://crowdin.com/translate/openhab-webui/384), can you please remove them?

kaikreuzer:
@florian-h05 Could you tell me what exactly you want me to remove?

florian-h05:
With this PR, MainUI loads the semantic tags together with the translations from the REST API and we therefore do not use the translations at openhab-webui anymore. This PR also removed the transation files from the repo, but seems like this was not synced back to CrowdIn.

Issue #1463, opened 2022-08-07
Persistence configuration via UI


J-N-K:
## The problem

I'm currently working on adding the possibility to configure the whole openHAB system via UI. Parts of this effort are #1448 and #1452. The next step is adding UI support for persistence service configuration. 

Currently persistence is configured in two places. The default persistence service is configured on the right side of the settings page under "System Services/Persistence" and individual configuration (like URLs or access data) is done under "Other Services/<service name>". The configuration of items and strategies can only be done via files.

The core part of configuring items/strategies (and filters as new feature) is in https://github.com/openhab/openhab-core/pull/2871.

I tried to implement UI support for the latter part but I'm unsure how to achieve the best user experience.

## Your suggestion

There are three possible places for the configuration page:

- On the "System Services/Persistence" page with a gear wheel in the selection list. The wheel will then open a new page for configuration. It feels wrong to configure individual services under "System Services". It also requires a special handling for `org.openhab.persistence` which is currently handled by a generic config-sheet.

- On the "Other Services/<service name>" page. This is slightly better because it groups all configuration for a service in one place. It's not so good because the list will get longer an longer the more persistence services are installed and there is no page currently for services that do not require configuration (like rrd4j), so we would need to add support for that, too. Also IMO the configuration done here should move to the add-on page once https://github.com/openhab/openhab-core/pull/3050 is implemented and good support for configuring all add-ons is provided. That makes item/strategy configuration IMO too invisible. 

- Via a new "Persistence" section (like the "Transformation" section). This would be my first choice but having three places where persistence is configured (default service, service configuration and items/strategies) in different parts of the UI could be confusing.

WDYT?

kaikreuzer:
> IMO the configuration done here should move to the add-on page

I'd agree on this. The "other services" currently are configurations for various add-ons of different types. We should instead have a dedicated place where for each add-on its configuration can be found easily. This is then imho the right place to also put the persistence add-on configurations.

ghys:
> It's not clear why certain things are listed under "Other Services" while others are under "System Services:".

Quite simple from a technical POV: the response to the `GET /rest/services` made in this page distinguishes those services with `category` set to `system` and the rest (I didn't deem it necessary to have more than these 2 groups, the system services are usually the majority anyway).

![image](https://user-images.githubusercontent.com/2004147/183727151-7e92c0a1-690c-4c96-af34-fb057f1a1fce.png)

[...]

PR #1891, opened 2023-05-12
Remove "Other Services" section from main settings page

J-N-K:
Add-on service settings are configurable from the respective add-on's page and therefore the "Other Services" section of the main settings page is not needed anymore.
Services provided by openHAB core that were previously listed unter "Other Services" were moved to "System Services".

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
@@ -134,14 +134,6 @@
               :link="'services/' + service.id"
               :title="service.label" />
           </f7-list>
-          <f7-block-title>Other Services</f7-block-title>
-          <f7-list class="search-list">
-            <f7-list-item
-              v-for="service in otherServices"
-              :key="service.id"
-              :link="'services/' + service.id"
-              :title="service.label" />
-          </f7-list>
         </f7-col>
       </f7-row>
       <f7-block-footer v-if="$t('home.overview.title') !== 'Overview'" class="margin text-align-center">
@@ -161,7 +153,6 @@ export default {
       servicesLoaded: false,
       addonStoreTabShortcuts: AddonStoreTabShortcuts,
       systemServices: [],
-      otherServices: [],
       objectsSubtitles: {
         things: 'Manage the physical layer',
         model: 'The semantic model of your home',
@@ -201,7 +192,6 @@ export default {
       // can be done in parallel!
       servicesPromise.then((data) => {
         this.systemServices = data.filter(s => s.category === 'system')
-        this.otherServices = data.filter(s => s.category !== 'system' && s.category !== 'persistence')
         this.servicesLoaded = true
       })
     },
--- a/bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
@@ -201,7 +201,7 @@ export default {
       // can be done in parallel!
       servicesPromise.then((data) => {
         this.systemServices = data.filter(s => s.category === 'system')
-        this.otherServices = data.filter(s => s.category !== 'system')
+        this.otherServices = data.filter(s => s.category !== 'system' && s.category !== 'persistence')
         this.servicesLoaded = true
       })
     },


florian-h05:
I wonder how we should handle the settings for other addons, because those are also reachable from the addon page. The “problem” I see with removing them here, is that accessing them via the addon page are several clicks more.

J-N-K:
True, there are some more clicks. On the other hand it is far more consistent, since - as you mentioned - other add-ons are also configured from their add-on page. 

I think removing them from the right columns was also the result of the discussion in #1463 (e.g. https://github.com/openhab/openhab-webui/issues/1463#issuecomment-1208283538).

florian-h05:
> On the other hand it is far more consistent, since - as you mentioned - other add-ons are also configured from their add-on page.

I meant that when persistence add-ons are removed from the "Other Services" section other add-ons are still there, however reading your linked post, I wonder if the other services section should be removed completely. Everything that can be configured there can be done through the add-on page.

J-N-K:
In general: yes. But for some reason it is currently not possible to configure e.g. basicui or javascript scripting from the add-ons page, I guess the config-description URI is probably not correctly set.

There are also other setting like the LSP that don't have an add-on page.

lolodomo:
Take care before merging. I don't know if this is the same for persistence services, but I see problems for openhabcloud and voiceRSS with I go to the setting from the addon page:

For openhab cloud, no settings are shown in bindings/addons page:
image

For VoiceRSS, the setting is shown but the value is not filled (while it is from the main settings):
image

So, for persistence services, please check that it is working from addons before removing the access from the main settings.

J-N-K:
It works with InfluxDB, and with the fix I just provided for JDBC.

lolodomo:
> It works with InfluxDB, and with the fix I just provided for JDBC.

Fine ;)

Even if I am a little off-topic, maybe you have an idea why it does not work for openhab cloud and voicerss ?

J-N-K:
I also checked DynamoDB, JPA and MongoDB and they work. RRD4J has no configuration.

J-N-K:
For openhabcloud it's due to the issue that was fixed today by @holgerfriedrich: the config-description-uri was wrong. For voicerss the service-id is wrong.

florian-h05:
@J-N-K What is the state here?

lolodomo:
I checked with success configuration access from add-on page for openHAB cloud, BasicUI, VoiceRSS and Vosk.

J-N-K:
IMO persistence services are fine, I did not check the others.

florian-h05:
I've checked IO (openHAB Cloud), UI (BasicUI), MISC (Metrics Service), automation (JS Scripting).
Problematic is the rule-based speech intepreter (`org.openhab.rulehli`) because is is provided by core and therefore no add-on, but has the category `voice`. Even though it belongs to voice, it should be changed to `misc` or `system` because it is provided by openHAB core.

bundles/org.openhab.ui/web/src/pages/settings/settings-menu.vue
```diff
@@ -201,7 +201,7 @@ export default {
       // can be done in parallel!
       servicesPromise.then((data) => {
         this.systemServices = data.filter(s => s.category === 'system')
-        this.otherServices = data.filter(s => s.category !== 'system')
+        this.otherServices = data.filter(s => s.category !== 'system' && s.category !== 'persistence')
```
florian-h05, line 204: FYI I tested this to filter out all add-ons:

```suggestion
        this.otherServices = data.filter(s => !['system', 'automation', 'bindings', 'io', 'persistence', 'transformations', 'ui', 'voice'].includes(s.category))
```

J-N-K: > Problematic is the rule-based speech intepreter (`org.openhab.rulehli`) because is is provided by core and therefore no add-on, but has the category `voice`. Even though it belongs to voice, it should be changed to `misc` or `system` because it is provided by openHAB core.

I agree. I also think that LSP should be system, because that is probably the only non-add-on service left.

lolodomo: Maybe the rule-based speech intepreter could become an add-on?

florian-h05: LSP and rule-based speech intepreter are the only non-add-on services left (AFAIK), for LSP I agree that it should be `system` and for rule-bases sp...ter I would say either `system` or `misc`. These two options would make keep them visible in MainUI. 

Rule-based sp..ter would be in the system settings section with `system` or the only one in the Other Services section with `misc`.

lolodomo: The most important is to not loose settings in existing user setups. 
Then showing them in system or misc is a detail.

J-N-K:
How do we proceed here?

florian-h05:
I haven't checked whether Rule HLI and LSP are now displayed under the "System Settings" tab, but if they now are I'd finally remove the "Other Services" section.

florian-h05:
@J-N-K Do you want to contribute in this PR? I could review and merge …

florian-h05 (approved):
LGTM, thanks!

J-N-K:
I get you point, but it seems a bit awkward and hard to understand that you configure a persistence (or IO) add-on in one place (on the main setting page) and it's logging at another place (on the add-on page) while both settings are in the same place (the add-on page) for bindings.

florian-h05:
As an iPhone user myself I have already noticed the similarities your original design Yannick, however I agree that it is a bit confusing to have the settings split over a few places. 
Unfortunately, most iOS apps have settings inside the app besides in the system Settings, so having to configure the addon from the addon‘s page is no real problem IMO.

ghys:
I'm not really arguing, just explaining :hugs: .
In fact maybe having the add-on's (formerly only binding's) configuration in the add-on store was a bad idea after all, as we could just have add-ons add entries to the settings menu like on iOS. So you would have e.g. "Philips Hue" directly under Settings and users wouldn't have to look for it in the add-on store, which was in the end really meant for browsing, installing and uninstalling, nothing more.
As for the logging, in fact it could have been a single consolidated page where you would configure levels for all loggers at once.

The removal of "Other Services" however offers an opportunity to adjust the Settings menu to keep it balanced when presented on two columns, as we could layout it like this:

- Left column
  * 1st-tier Configuration: Things, Model, Items, Page
  * 2nd-tier Configuration: Persistence, Tags (?), Transformations - no need to have a section header title for those, just separate them from the 1st tier
  * Automation: Rules, Scenes, Scripts, Schedule
- Right column
  * Add-ons: Bindings, Automation, UI, Other
  * Services (formerly System Services)

florian-h05:
Sound like a great idea!

I have one comment:

> 2nd-tier Configuration: Persistence, ...

I'd probably keep persistence out of the main settings and only have it under the add-on settings in the right column.

Could you probably open a PR so we can have a look?

florian-h05:
@ghys I have opened a new issue to discuss and coordinate the redesing of add-on configuration, see #1935.

PR #1893, opened 2023-05-13
`oh-knob`: Fix `releaseOnly` not working

florian-h05:
Regression from #1875.


florian-h05:
@ghys This is critical and fixes a regression from a previous PR.

florian-h05:
@kaikreuzer FYI: I would consider this regression blocking for M3, however the fix is a one liner and adds something back that was removed in #1875 and therefore should be quick to review when @ghys is available.

ghys (approved):

Issue #1862, opened 2023-04-27
[Main UI] Slider widget causes unsolicited handleCommand() calls to bindings when channel value is PercentType(BigDecimal)


andrewfg:
There is a bug in the Main UI page script as follows..

1. Main UI is displaying a slider value for a PercentType channel, and..
2. OH core sends an SSE update of the channel value comprising float PercentType(BigDecimal) value (not an integer), then..
3. Main UI makes a rounding error between the slider's supposed integer position, and the float value of the channel value, so..
4. Main UI wrongly sends a POST to OH core with the new (wrongly rounded) integer value, with the result that..
5. OH core therefore is forced to (wrongly) call handlecommand() on the binding..

The issue was first reported here:  https://github.com/openhab/openhab-core/issues/3571

lolodomo:
@openhab/webui-maintainers : maybe something easy to fix? 
This issue looks relatively critical.

florian-h05:
I am currently working on the evcc binding, but I can have a look at this.

andrewfg:
@florian-h05 the posts in the thread starting [here](https://github.com/openhab/openhab-core/issues/3571#issuecomment-1523794004) will probably give you a better idea what is happening..

florian-h05:
Have you considered adjusting the `step` property of the slider widget to something smaller? E.g. `0.1`?

andrewfg:
I am not reporting a bug in a widget; it is a bug in Main UI
![image](https://github.com/openhab/openhab-webui/assets/893994/ecabd9e6-aef2-431d-9113-71e26ff875af)

florian-h05:
MainUI is internally using the same components that you could use in widgets. Your Item is a dimmer right, and I guess it is using `oh-slider` as a control. 
The slider widget is rounding to a given number of decimals (default 0 decimals when `step` is 1), you can influence that behaviour by setting the `Default Standalone Widget` metadata and setting a step size like `0.25`, but I see your problem.

if my guess with `oh-slider` is correct, I can modify the slider so that it does not trigger commands when a state update is received but only when the user actually interacts with it.

florian-h05:
@andrewfg Was a bit more complicated than expected, but here is the fix: #1894.

You can follow https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/CONTRIBUTING.md#npm-scripts to set up a MainUI dev environment, then check out my PR and start up the dev server. Note that it is not finished yet.
[...]

PR #1894, opened 2023-05-13
`oh-slider`: Fix command sent on state update due to rounding

florian-h05:
Fixes https://github.com/openhab/openhab-webui/issues/1862.

When a state update like 40.545 was received, the slider component commanded the Item to 41 because of the rounding used.
Unfortunately, the slider component has no built-in way to avoid that.
This decouples the slider value from the actual Item value and only send command when the difference between them is larger or equal than step size.

This is no perfect solution, because it can be that the slider does not accept input when the above checks don't pass. 
Example: Your Item has a value of 40.545, then the slider does not accept 40.5 and 41.

I've checked `oh-setpoint` and `oh-knob`, they are not affected by this bug.


ghys (approved):
I'll trust you (gladly!) on this one as I've not experienced the problem, but the fix looks sound.

florian-h05:
I have never experienced it with my Items, but it was super easy to reproduce when putting a slider and an input on one page and linking both with the same number Item. When you typed something like `40.5`, it changed to `41` because of the slider widget.

ghys:
Yes I understand, there were problems with colorpickers too IIRC, that's because there's only a "change" event provided in these components and it's difficult to figure out if the change comes from a user interaction or an external change to the item's state. Either way the change led to a new command sent which is not what's desired in the latter case and it's also hard to track (I have had to deal with infinite state-update->command loops too).

PR #1897, opened 2023-05-13
[BasicUI] Handle properly state updates to NULL or UNDEF

### LINKED ISSUE #1886: [BasicUI] Should handle properly NULL as live state update
When a page is already opened, Basic UI correctly supports UNDEF as new item state update but ignores NULL states.
NULL state as a live state update should be handled properly.

lolodomo:
Live update to NULL state is no more ignored.
Live update to NULL or UNDEF is now properly handled by setpoint and slider widgets.

Fix #1886


kaikreuzer (approved):
Thanks!

PR #1898, opened 2023-05-13
[blockly] Add dictionary loop

### LINKED ISSUE #1711: Blockly Dictionary loop block with value and key from dictionary
Currently there is no loop block in Blockly, to iterate over a Dictionary and get the key and the value from each dictionary entry.

https://community.openhab.org/t/blockly-dictionary-loop/144354/4 describes a workaround, utilising the under the hood generated code.

Having such an iteration block would be very nice.

stefan-hoehn:
fixes #1711
This adds a new loop type which allows to iterate over a dictionary as follows

image

It allows to either pass a variable (trusted to be of type Dictionary) or a Dictionary-Block itself.


florian-h05 (approved):
LGTM, thanks!

PR #1899, opened 2023-05-14
[blockly] Show toast instead of error when saving / running from code preview

stefan-hoehn:
Since the very beginning the following error dialog always annoyed me...

image

which happens when you try to save in the blocky code review.

Instead we can detect that situation and show a nice hint that saving (and running) is not supported when in the code review page:

image


bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
```diff
@@ -383,6 +394,7 @@ export default {
           }).open()
         })
       })
+        .catch(() => {})
```
florian-h05, line 397: Is this really required? I don't think so.

stefan-hoehn: As discussed offline, catch is required to suppress ugly promise.reject message but I don't like catch-alls either. I added a specific reject-reason and cleanly catch it and rethrow everything else now.

florian-h05 (approved):
LGTM, very nice!

PR #1900, opened 2023-05-15
Item & Thing code tabs: Minor fixes

florian-h05:
- Item code tab: Fix lock icon always displayed.
- Thing code tab: Fix missing warning when leaving page but there are unsaved changes.


ghys (approved):
Thanks!

PR #1901, opened 2023-05-15
Items: Add `unit` metadata for UoM (`Number:`) Items

### LINKED ISSUE #1890: Unit as pre-defined metadata
With the new UoM in core separating the internal unit from the display unit, the user will need to frequently define a "unit" metadata if he does not want the default unit as internal unit.

A new entry has to be added in that pop-up:
image

I suggest to place it just after "State Description", so in second position.

florian-h05:
Fixes #1890.

This adds the `unit` metadata from https://github.com/openhab/openhab-core/pull/3481 to the pre-defined metadata namespaces for UoM Items (Item type `Number:`) and provides a metadata edit page for it.

Once https://github.com/openhab/openhab-core/pull/3611 is merged, the Item create page will be adjusted to set the `unit` metadata to the system default on Item creation.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
@@ -1,5 +1,6 @@
 export default [
   { name: 'stateDescription', label: 'State Description' },
+  { name: 'unit', label: 'Internal Unit' },
   { name: 'commandDescription', label: 'Command Options' },
   { name: 'synonyms', label: 'Synonyms' },
   { name: 'widget', label: 'Default Standalone Widget' },
--- a/bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
@@ -1,6 +1,6 @@
 export default [
+  { name: 'unit', label: 'Unit' },
   { name: 'stateDescription', label: 'State Description' },
-  { name: 'unit', label: 'Internal Unit' },
   { name: 'commandDescription', label: 'Command Options' },
   { name: 'synonyms', label: 'Synonyms' },
   { name: 'widget', label: 'Default Standalone Widget' },
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-unit.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-unit.vue
@@ -5,7 +5,7 @@
     </f7-block-header>
     <f7-list>
       <f7-list-input
-        label="Internal Unit"
+        label="Unit"
         name="value"
         ref="value"
         type="text"
@@ -14,7 +14,13 @@
         @blur="(evt) => metadata.value = evt.target.value" />
     </f7-list>
     <f7-block-footer class="param-description">
-      <small>Unit used internally, e.g. in events and for persistence.<br>The unit of the item is independent of the state description. The state description is used for display purposes only, it can contain any compatible unit and will not affect the internal unit.</small>
+      <small>All processed values are internally normalized to the specified <code>unit</code>. The normalized value is used to propagate the value to external integrations (e.g. persistence, REST API, WebSocket) so these values will always have the specified unit and scale.
+        <br>
+        Changing the <code>unit</code> will change the values that these systems receive.
+        <br>
+        <br>
+        Additionally the <code>unit</code> will be used in UIs if no unit is specified for the state description. The state description is used for display purpose only, it can contain any compatible unit and will not affect the internal <code>unit</code>.
+      </small>
     </f7-block-footer>
   </div>
 </template>
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-unit.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-unit.vue
@@ -1,0 +1,26 @@
+<template>
+  <div>
+    <f7-block-header>
+      <b style="color: var(--f7-theme-color) !important;">WARNING: Changing the internal unit can corrupt your persisted data!</b>
+    </f7-block-header>
+    <f7-list>
+      <f7-list-input
+        label="Internal Unit"
+        name="value"
+        ref="value"
+        type="text"
+        placeholder="leave empty to use system default"
+        :value="metadata.value"
+        @blur="(evt) => metadata.value = evt.target.value" />
+    </f7-list>
+    <f7-block-footer class="param-description">
+      <small>Unit used internally, e.g. in events and for persistence.<br>The unit of the item is independent of the state description. The state description is used for display purposes only, it can contain any compatible unit and will not affect the internal unit.</small>
+    </f7-block-footer>
+  </div>
+</template>
+
+<script>
+export default {
+  props: ['itemName', 'metadata', 'namespace']
+}
+</script>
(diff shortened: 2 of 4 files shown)


bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
```diff
@@ -1,5 +1,6 @@
 export default [
   { name: 'stateDescription', label: 'State Description' },
+  { name: 'unit', label: 'Internal Unit' },
```
florian-h05, line 3: @J-N-K WDYT about the labelling?

bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-unit.vue
```diff
@@ -0,0 +1,26 @@
...
+        label="Internal Unit"
+        name="value"
+        ref="value"
+        type="text"
+        placeholder="leave empty to use system default"
+        :value="metadata.value"
+        @blur="(evt) => metadata.value = evt.target.value" />
+    </f7-list>
+    <f7-block-footer class="param-description">
+      <small>Unit used internally, e.g. in events and for persistence.<br>The unit of the item is independent of the state description. The state description is used for display purposes only, it can contain any compatible unit and will not affect the internal unit.</small>
```
florian-h05, line 17: @J-N-K Please let me know what you think about that description and the warning above in line 4.

J-N-K:
Looks good. Please note that the metadata `unit` is optional, if it is not set, the system default is used (state description still is independent).

florian-h05:
There is a notice that system default is used when `unit` is not set:

image

I think that makes it clear enough that it is optional, WDYT?

bundles/org.openhab.ui/web/src/assets/definitions/metadata/namespaces.js
ghys: I think it could be labelled "Measurement Unit" or even "Unit" and promoted to the top (stateDescription and commandDescription are somehow linked so it feels wrong to list a metadata namespace between them).

florian-h05: Partly agreed, let's put it to the top.
What I don't like about "Unit" or "Measurement Unit" that they don't make clear that it is a internal unit. The user could think that they are for display purpose (even though the description says they aren't, but I'm not sure if every user always reads those descriptions carefully).

ghys: > they don't make clear that it is a internal unit

Yet I (and I suppose many users who will make decisions based on these descriptions) am not sure what an "internal" unit is.

J-N-K: If no state description is set (which is the default), it is also used for displaying the values.

florian-h05:
@spacemanspiff2007 wrote on the linked issue:

> When creating a UoM item the unit should be mandatory, see core issue.
The system default should be (pre-)selected but the unit should be visible directly on the site with the create button.

I also had a look at the linked core issue https://github.com/openhab/openhab-core/issues/3615.

I will provide a field to set the Item metadata on Item creation and as soon as the core PR for UoM information is merged, we can populate the unit metadata field with the system default.

WDYT?

spacemanspiff2007:
Sounds great, as long as the field will always be automatically populated so creating an UoM item will always result in the corresponding `unit` metadata. 
It should not be possible to create a UoM item through the GUI without the metadata.
That way it's ensured that future changes (e.g. to system defaults, locale, etc.) will not lead to wrong data/graphs in external systems.

lolodomo:
If we don't want to postpone again the milestone 3, the merge of this PR would be greatly appreciated (if it works).

lolodomo:
> Additionally the unit will be used if no unit is specified for the state description.

Additionally the unit will be used `in UIs` if no unit is specified for the state description.

> ... to external systems (e.g. persistence, RestAPI, HABApp) so  ...

We could maybe avoid mentioning specifically a tool that is not part of the official OH distribution ?
Just a remark.

J-N-K:
I would suggest to use `external integrations (e.g. persistence, REST API, WebSocket)`.

florian-h05:
> If we don't want to postpone again the milestone 3, the merge of this PR would be greatly appreciated (if it works).

This is not ready yet and for setting the unit metadata when creating a UoM Item, https://github.com/openhab/openhab-core/pull/3611 has to be merged first.
Unfortunately I’m on school trip until Sunday morning, but I think I can finish this on Sunday.

lolodomo:
IMHO, if we have this PR allowing to easily add the unit metadata, the change in item creation page could be delayed.

florian-h05:
If we agree to delay the change in Item creation page this doesn’t need the core PR and I should be able to finish this (it’s just about naming and descriptions then) from my iPad.

florian-h05:
@ghys Can you please review and merge this for M3? Thanks!

spacemanspiff2007:
> We could maybe avoid mentioning specifically a tool that is not part of the official OH distribution (and not in OH documentation) ?
> Just a remark.

It's shown in the [rule technology overview](https://www.openhab.org/docs/tutorial/rules_overview.html). Also NodeRed, jRuby, Mycroft AI are mentioned multiple times through the whole docs and these are all not part of the official OH distribution.
So the argument "not part of the official OH distribution" is invalid.

rkoshak:
> Also NodeRed, jRuby, Mycroft AI are mentioned multiple times multiple times through the whole docs and these are all not part of the official OH distribution.

I'm not arguing against the sentiment (I have no problems with mentioning tools outside the OH project in the docs where it make sense). But I wanted to correct this list of examples of external projects.

- NodeRed is mentioned in only three places I can find through search. In addition to the link you provided, one is in openHABian's docs where it's explained that openHABian installs it and the second is an example for how a third party tool can use an openHAB API Token. The first is very appropriate since it's part of what openHABian does and openHABian is part of official openHAB. The second is arguable whether NodeRed is the best choice but *something* needs to be picked as an example there and the screenshots from NodeRed are very clear so I would argue it's a good choice.

- jRuby has an official automation add-on and it's helper library is part of the openHAB project too which put it at the same level as at least Rules DSL and GraalVM JS Scripting (maybe Groovy and Java too?). It is an official part of OH in every meaningful way. 

- Mycroft.AI is a third party system which openHAB officially supports integration with, similar to Alexa, Google Assistant, and HomeKit. And it's only mentioned in the Mycroft.AI pages in the docs, as is appropriate.

All three of these examples *are* a part of official openHAB.

But again, I'm not arguing that HABApp shouldn't be mentioned in the docs anywhere because it isn't part of the openHAB project. In fact, isn't it an option in openHABian to install? If not it should be. That would make it as "official" as NodeRed at least. 

However, because HABApp isn't discussed anywhere in the docs except that one table in Getting Started, it's a poor choice to use as an example in the description of the field built into MainUI. So I agree with @lolodomo in this one case but not as a general rule to apply everywhere.

spacemanspiff2007:
Nicely written and I mostly agree with you. 
I was just showing that we have parts that are not official OH distribution and thus I think the argument was void. If you or @lolodomo  write you think it's not a good fit _here_ that's totally fine and I'll gladly accept it.

>  In fact, isn't it an option in openHABian to install? If not it should be.

It has been since over two years - option `2B` in the menu (I know it by heart now :wink: ).

J-N-K (approved):
LGTM for the description part. Can't comment on the code.

ghys (approved):
We wouldn't want a milestone release with the core change but no UI support. LGTM, thanks for the implementation!

PR #1903, opened 2023-05-18
[blockly] provide persistence type to blocks

### LINKED ISSUE #1625: Blocky persistence add parameter for persistence service
I have been converting some old jython rules to Blocky. I am impressed with the power that is already there.

I do have code in my old rules that retrieves data from a non-default persistence. I have rrd4j as a default persistence, but also store a subset in influxDB. Some of the data is also fed from outside directly into influxDB, but mapped to items in OH. So I have more history then created directly through OH.
It would be nice to have the persistence service as a (optional) parameter to the persistence blocks in Blocky.
The other part I am missing is support for historicState (on the non-default service). I had seen some request for that before, but cannot locate it anymore.

stefan-hoehn:
fixes #1625 

Allows to provide the persistence type to persistence blocks
default can be provided instead of selecting a particular one (is backward compatible to old blocks)
Either item name or item can be provided (note that variables must have the item not the item name as the type of variables cannot be detected)

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
@@ -4,9 +4,9 @@
 */
 import Blockly from 'blockly'
 import { javascriptGenerator } from 'blockly/javascript'
-import { addDateSupport } from './utils'
+import { addDateSupport, blockGetCheckedInputType } from './utils'
 
-export default function defineOHBlocks_Persistence (f7, isGraalJs) {
+export default function defineOHBlocks_Persistence (f7, isGraalJs, persistenceServices) {
   /*
   * Provides a number of different (non-)statistical metrics for an item according to the given date
   * Blockly part
@@ -16,18 +16,21 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs) {
       this.appendDummyInput()
         .appendField('get the')
         .appendField(new Blockly.FieldDropdown([
-          ['average', 'averageSince'], ['delta', 'deltaSince'],
-          ['deviation', 'deviationSince'], ['variance', 'varianceSince'], ['evolution rate', 'evolutionRate'],
-          ['minimum', 'minimumSince'], ['maximum', 'maximumSince'], ['sum', 'sumSince'],
+          ['state average', 'averageSince'], ['state delta', 'deltaSince'],
+          ['state deviation', 'deviationSince'], ['state variance', 'varianceSince'], ['evolution rate', 'evolutionRateSince'],
+          ['state minimum', 'minimumSince'], ['state maximum', 'maximumSince'], ['state sum', 'sumSince'],
           ['previous state value', 'previousState'], ['previous state value time', 'previousStateTime'],
           ['historic state', 'historicState']
         ], this.handleTypeSelection.bind(this)
         ), 'methodName')
       this.methodName = this.getFieldValue('methodName')
       this.appendValueInput('itemName')
-        .appendField('of the state of item named ')
+        .appendField('of item ')
         .setAlign(Blockly.ALIGN_RIGHT)
-        .setCheck(['String', 'oh_item'])
+        .setCheck(['String', 'oh_item', 'oh_itemtype'])
+      this.appendValueInput('persistenceName')
+        .appendField('from')
+        .setCheck(null)
       this.updateShape()
       this.setInputsInline(false)
       this.setOutput(true, null)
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
```diff
@@ -16,18 +16,21 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs) {
       this.appendDummyInput()
         .appendField('get the')
         .appendField(new Blockly.FieldDropdown([
-          ['average', 'averageSince'], ['delta', 'deltaSince'],
-          ['deviation', 'deviationSince'], ['variance', 'varianceSince'], ['evolution rate', 'evolutionRate'],
-          ['minimum', 'minimumSince'], ['maximum', 'maximumSince'], ['sum', 'sumSince'],
+          ['state average', 'averageSince'], ['state delta', 'deltaSince'],
+          ['state deviation', 'deviationSince'], ['state variance', 'varianceSince'], ['evolution rate', 'evolutionRateSince'],
+          ['state minimum', 'minimumSince'], ['state maximum', 'maximumSince'], ['state sum', 'sumSince'],
```
florian-h05, line 21: IMO average state, minimum state, maximum sound better.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
```diff
@@ -60,6 +63,10 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs) {
       }
     },
     updateShape: function () {
+      let persistenceNameInput = this.getInput('persistenceName')
```
florian-h05, line 66:
```suggestion
      const persistenceNameInput = this.getInput('persistenceName')
```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
```diff
@@ -93,35 +100,39 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs) {
   */
   javascriptGenerator['oh_get_persistvalue'] = function (block) {
     const itemName = javascriptGenerator.valueToCode(block, 'itemName', javascriptGenerator.ORDER_ATOMIC)
-    const methodName = block.getFieldValue('methodName')
+    const inputType = blockGetCheckedInputType(block, 'itemName')
+    let itemCode = toItemCode(itemName, inputType)
```
florian-h05, line 104: I'd put this line before `let code` and make it `const itemCode`.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
```diff
@@ -200,20 +233,61 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs) {
...
     } else {
       const { dtf, zdt, getZonedDatetime } = addDateSupport()
       const persistence = addPersistence()
-      let code = `${persistence}.lastUpdate(itemRegistry.getItem(${itemName}))`
+      let code = `${persistence}.lastUpdate(${itemCode}${persistenceExtension})`
       return [code, 0]
     }
   }
 
+  function toItemCode (itemName, inputType) {
```
florian-h05, line 252: Would it make sense to put this into the `utils.js` because there are other blocks that can use this as well?
And I'd rename it to `generateItemCode`.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
```diff
@@ -165,15 +184,21 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs) {
   */
   javascriptGenerator['oh_persist_changed'] = function (block) {
     const itemName = javascriptGenerator.valueToCode(block, 'itemName', javascriptGenerator.ORDER_ATOMIC)
+    const inputType = block.getInput('itemName').connection.targetBlock().outputConnection.getCheck()[0]
```
florian-h05, line 187: Why do you not use `blockGetCheckedInputType` here?

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
stefan-hoehn: This is the same like before only that I have moved it into the dropdown (it was part of the label before the dropdown). IMHO, it is actually the state average not the average state. Also it is consistent to the others.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
stefan-hoehn: I am not surprised you ask.  Because it doesn't work.  I did but there is a particular condition in 

> if (!thisBlock || !connectedBlock) return ''

which in these cases will return nothing.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
stefan-hoehn: I will eventually if in the future it would be used somewhere else but until then I would recommend to keep it here. I changed the name.

florian-h05: The functionality provided by this function is also used in the eventbus blocks and probably others, would be nice if you could provide a follow-up PR to refactor those to use this function.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
florian-h05: I don't understand why it shouldn't work, for me it works fine:

```javascript
    const inputType = blockGetCheckedInputType(block, 'itemName')
```

stefan-hoehn: Strangely, it does work with my unit tests and I don't know why...
I replaced it with blockGetCheckedInputType.

florian-h05 (approved):
LGTM, thanks!

PR #1911, opened 2023-05-28
fix blockly uom var support

stefan-hoehn:
fixes variable support for the blockly unit of measurement blocks which did not work at all -> var defaults to item object 

Blockly cannot detect the type that is contained in a variable so it expects an item object.
This allows to iterate over a group of item members which returns a list of item objects


florian-h05 (approved):
I haven’t tested this since I’m away from home, but looks good!

PR #1917, opened 2023-06-06
Add a persistence configuration page

### LINKED ISSUE #1463: Persistence configuration via UI

florian-h05:
Closes #1463.
Refs https://github.com/openhab/openhab-core/pull/2871.
Refs https://github.com/openhab/openhab-core/pull/3642.

It is accessible from the add-on settings page and has both a design and a code tab.

The design tab allows to set persistence strategies for Items, define cron strategies and set the default strategies. It does not duplicate names for (cron) persistence strategies and filters as well as configs for the same set of Items.
All four filters provided by openHAB core (treshold, time, equals/not equals, include/exclude) can be configured.
When the user removes a cron strategy or a filter, it is automatically removed from all configs so that there is no API failure (400 Bad Request).

No code completion is not provided, but required attributes for filters are automatically set on save to avoid API failure (500 Internal Server Error).

A few words about order and sorting:
- openHAB Core seems to sort the cron strategies.
- Configurations itself are unsorted, they could be sorted alphabetically by the UI.
- Items of configuration are sorted by their type (groups before normal Items) as well as alphabetically.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
@@ -42,6 +42,19 @@
               <!-- Default Strategies -->
               <strategy-picker title="Default Strategies" class="skeleton-text skeleton-effect-blink" />
             </div>
+            <div>
+              <f7-block-title medium style="margin-bottom: var(--f7-list-margin-vertical)">
+                Filters
+              </f7-block-title>
+              <div v-for="ft in filterTypes" :key="ft.name">
+                <f7-block-title>
+                  {{ ft.label }}
+                </f7-block-title>
+                <f7-list class="skeleton-text skeleton-effect-blink">
+                  <f7-list-item />
+                </f7-list>
+              </div>
+            </div>
           </f7-col>
         </f7-block>
 
@@ -52,6 +65,7 @@
             </div>
           </f7-col>
           <f7-col class="modules">
+            <!-- Configuration -->
             <div>
               <f7-block-title medium style="margin-bottom: var(--f7-list-margin-vertical)">
                 Configuration
@@ -80,6 +94,7 @@
                 </f7-list-item>
               </f7-list>
             </div>
+            <!-- Strategies -->
             <div>
               <f7-block-title medium style="margin-bottom: var(--f7-list-margin-vertical)">
                 Strategies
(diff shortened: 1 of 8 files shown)


florian-h05:
The addon settings page looks for persistence addons like this:

image

J-N-K:
Looks great. Do you think we can add configurations for filters, too?

florian-h05:
It should be possible, I will have a look tomorrow or in the next days. 

It would be great if you could provide me real examples what the REST API returns for the filters. 
Swagger helps, but does not tell everything ;-)

florian-h05:
Thanks.

Out of interest: Would there be any way that openHAB core provides config descriptions for the filters?
That would allow filter configuration from the UI without having to hard-code the filters into MainUI.

hmerk:
Any Progress here @florian-h05 ??

florian-h05:
Unfortunately not yet, I was really busy and my free time went into openhab-js.

I will try to finish it this weekend.

florian-h05:
@J-N-K @hmerk I am finished now, all filters can be configured by UI :tada: 
I updated the screenshots in https://github.com/openhab/openhab-webui/pull/1917#issuecomment-1579208349, you can open them under the screenshots menu.

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
```diff
@@ -0,0 +1,688 @@
...
+      currentConfiguration: null,
+      currentCronStrategy: null,
+      currentFilter: null,
+
+      predefinedStrategies: ['everyChange', 'everyUpdate', 'restoreOnStartup'],
+      filterTypes: [
+        {
+          name: 'thresholdFilters',
+          label: 'Threshold',
+          configDescriptionParameters: [
```
florian-h05, line 257: @J-N-K Can you please have a look at the config description parameters, especially the descriptions?

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
```diff
@@ -0,0 +1,688 @@
...
+      loading: false,
+      currentTab: 'design',
+      currentConfiguration: null,
+      currentCronStrategy: null,
+      currentFilter: null,
+
+      predefinedStrategies: ['everyChange', 'everyUpdate', 'restoreOnStartup'],
+      filterTypes: [
+        {
+          name: 'thresholdFilters',
```
J-N-K, line 255: maybe `thresholdFilter`? It's the description of a single filter. Also below.

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
```diff
@@ -0,0 +1,688 @@
...
+            }
+          ]
+        },
+        {
+          name: 'timeFilters',
+          label: 'Time',
+          configDescriptionParameters: [
+            {
+              advanced: false,
+              description: 'Amount of time that must have passed since the last persist',
```
J-N-K, line 290:
```suggestion
              description: 'Amount of time that must have passed since the last value has been persisted',
```

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
```diff
@@ -0,0 +1,688 @@
...
+            }
+          ]
+        },
+        {
+          name: 'equalsFilters',
+          label: 'Equals/Not Equals',
+          configDescriptionParameters: [
+            {
+              advanced: false,
+              description: 'Enter values seperated by comma (use point <code>.</code> as decimal point), e.g. <code>one, two, three</code>, to be persisted',
```
J-N-K, line 312:
```suggestion
              description: 'Enter values separated by comma (use point <code>.</code> as decimal point), e.g. <code>one, two, three</code>, to be persisted',
```

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
florian-h05: This doesn't work, because the filter type name and label are used by the UI to render the sections and determine where to read from and where to store filters. And the property on the persistence is named `thresholdFilters`.

I have implemented the UI part for filters in a way, that once additional filters are added you only need to update the filterTypes definition and the code tab.

stefan-hoehn:
@hmerk I am currently testing this and we found a few minor quirks that Florian already fixed. So, almost done and looks pretty good to me. Probably can be merged by tonight.

stefan-hoehn:
@florian-h05 I thoroughly tested it including the latest changes - thx 4 adding my request on the validation and allowing names with _ and numbers <3 

And it looks good to me from a functional perspective.

florian-h05:
@ghys This is finally ready for review! Let's get this into 4.0 🥳

ghys (approved):
> Probably can be merged by tonight.

What Stefan wants, Stefan gets I suppose :wink: 
Jokes aside I have no quirks with this code and I think it's an important feature and therefore should be merged as early as possible so that it could be "battle-tested" before the release.

hmerk:
@florian-h05 
Thanks, great job.
But I have to admit, that I sruggled a bit to find the config. It was not really intuitive that the blue healine "Persistence Cinfiguration" is a link to the edit menu....

florian-h05:
> It was not really intuitive that the blue healine "Persistence Cinfiguration" is a link to the edit menu....

I'm very open for suggestions, I am not super happy with the UX to get there.
Please also have a look at #1935 where the redesign of the settings page to allow fast access to addon setting is discussed.

stefan-hoehn:
Just to add on that: the documentation for the filters has been nicely documented by @J-N-K here: https://github.com/openhab/openhab-docs/pull/2088

PR #1919, opened 2023-06-08
`oh-repeater`: Fix `key` property does not refresh repeater

florian-h05:
Fixes #1918.
Regression from #1904.

PR #1920, opened 2023-06-08
`oh-slider` & `oh-knob`: Fix user configured unit (state description) ignored

### LINKED ISSUE #1913: [UoM] Main UI Color Temperature Slider displays Celsius rather than Kelvin
See https://github.com/openhab/openhab-core/issues/3643 where @J-N-K says it is not an OH Core problem.

On OH 4.0 snapshot, for things with channels resp. items of type system.color-temperature-abs the Main UI default color temperature slider for those items displays the values using the system default temperature measurement Celsius (as shown below). Expected would be that Main UI would display color temperature in Kelvin (even in the USA)..

(It also sends a QuantityType(Celsius) to the binding when the slider is moved. Which is not a problem since the binding can convert the UoM to Kelvin).

I suppose that this is related to recent UoM changes in OH core??

image

Note: I did not check on a US system but I suppose the default color temperature slider would display in Fahrenheit..

florian-h05:
Fixes #1913. Fixes https://github.com/openhab/openhab-core/issues/3643.
Depends on https://github.com/openhab/openhab-core/pull/3647.

While fixing this, I also noticed a minor regression from PR #1894, that I also fixed.

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-slider-card.md
+++ b/bundles/org.openhab.ui/doc/components/oh-slider-card.md
@@ -107,7 +107,7 @@ Display a slider in a card to control an item
 </PropBlock>
 <PropBlock type="TEXT" name="unit" label="Unit">
   <PropDescription>
-    Text to append to the label while dragging the cursor
+    Unit for the command sent and also append to the label while dragging the cursor, leave empty to use Item's unit
   </PropDescription>
 </PropBlock>
 <PropBlock type="BOOLEAN" name="releaseOnly" label="Send command only on release">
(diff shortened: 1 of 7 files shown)


florian-h05:
@ghys Thanks for merging #1919.
This PR also fixes a minor regression, so it would be great to get a review here as well.

The semantic tags PR is not time critical, but required as a first step when I add the possibility to user-configure the semantic model (core PRs by @lolodomo are on their way). If I get the time, I want to aim for openHAB 4.0 for the UI support.

My persistence configuration PR is not finished yet, but definitely should go into 4.0.

ghys (approved):

PR #1922, opened 2023-06-12
[BasicUI] Adjust slider unit handling

mherwege:
Adjust BasicUI slider unit handling.
So far, the unit was not passed on to the slider. With proposed changes in core https://github.com/openhab/openhab-core/pull/3644, unit would be passed and needs to stripped for the slider to work properly.

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1680,7 +1680,6 @@
 		}, 200);
 
 		_t.setValuePrivate = function(value, itemState) {
-			// itemState contains only value in the display unit (in case unit is set in label pattern)
 			if (_t.hasValue) {
 				_t.valueNode.innerHTML = value;
 			}
@@ -1695,7 +1694,7 @@
 			if (itemState === "NULL" || itemState === "UNDEF") {
 				_t.input.value = 0;
 			} else {
-				_t.input.value = itemState;
+				_t.input.value = itemState.split(" ")[0] * 1;
 			}
 			_t.input.MaterialSlider.change();
 		};


mherwege:
See: https://github.com/openhab/openhab-core/pull/3644#issuecomment-1587095260

> In file smarthome.js (BasicUI), line 1698 has to be fixed and comment at line 1683 has to be removed. Line 1659 has to be checked too (I am not sure of the behaviour of parseInt) .

Changes in 1698 and 1683 done, no need to change 1659 as parseInt drops anything after the number.

mherwege:
One issue could still remain in line 1688 (`_t.reloadIconColor(ItemState)`). I could strip the unit there as well. But it makes me wonder about the behaviour with other widgets that carry a unit. Would the expectation be that the state conditions are defined without unit for slider widget and with unit for other widgets?

mherwege:
@lolodomo For your review and comment.

lolodomo:
No need to strip the unit when calling reloadIcon. 
First, it is not done for other widgets. 
Second, the icon server is doing the strip itself:
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.ui.icon/src/main/java/org/openhab/core/ui/icon/AbstractResourceIconProvider.java#L71

lolodomo (approved):
LGTM, thank you

mherwege:
@lolodomo I believe this PR does not depend on core https://github.com/openhab/openhab-core/pull/3644. It would be required to merge if https://github.com/openhab/openhab-core/pull/3644 gets merged, but doesn't break anything if merged before. It will only try to strip a (non-existing) unit in that case.

lolodomo:
You're right.

PR #1923, opened 2023-06-13
[basicui] Use sitemap input hint v2

mherwege:
This PR is a continuation of https://github.com/openhab/openhab-webui/pull/1787, using the extra functionality provided by core if https://github.com/openhab/openhab-core/pull/3644 gets merged in core. It would be an alternative to https://github.com/openhab/openhab-webui/pull/1787 with the same functionality, but reduced complexity.

Requires https://github.com/openhab/openhab-core/pull/3644.


lolodomo:
@mherwege : for your information, I have done tests with number and I discovered 2 issues:

1. it does not work at all with number without dimension, the entered value is apparently rejected
2. If I clear the field and press enter, a value is restored but not the current value, rather an old value, probably the value when the page was first opened.

For number with dimension, except problem 2, it seems to work well.

For problem 1, as it was working well before, I assume there is a link with the change you did in core framework. The problem is in the JavaScript as I see command sent to the server in the event log.

I will try to solve these 2 issues.

mherwege:
The java script code was not touched in https://github.com/openhab/openhab-webui/commit/3d6aa48e09180fe6e994952985262a17f861fa63, the commit adapting to the change in core. So both problems may already have been there with the other PR, or caused by the change in core.

lolodomo:
> The java script code was not touched in [3d6aa48](https://github.com/openhab/openhab-webui/commit/3d6aa48e09180fe6e994952985262a17f861fa63), the commit adapting to the change in core. So both problems may already have been there with the other PR, or caused by the change in core.

But in fact, I did not test recently to fill a value.
The problem should be in parseNumber method, I assume.

lolodomo:
> The problem should be in parseNumber method, I assume.

But this method works for number with dimension so I am not sure. That is strange.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1606,25 +1679,65 @@
...
+				}
+			}
 			_t.input.value = newValue;
 			_t.input.nextElementSibling.innerHTML = undefValue;
 
 			_t.input.parentNode.MaterialTextfield.change();
 			_t.input.parentNode.MaterialTextfield.checkValidity();
-
 			lastValue = value;
 			lastItemState = itemState;
```
lolodomo, line 1733: I am not sure if lastItemState should be updated when itemState is undefined (not passed to thge call) ?

mherwege: I am now sure it matters. But I notice itemState is not being set at all, and it should be set somewhere. That may explain the reverting to an old state.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1546,46 +1546,119 @@
...
+				var unitValue = newValue.substring(numberValue.length).trim();
+				newValue = numberValue.replace(/^\+/, "");
+				// when sending updates, only uppercase E for exponent is accepted, don't change when only parsing for visualisation
+				newValue = keepExponentChar ? newValue : newValue.replace("e", "E");
+				if (commaSeparatorPattern.test(newValue)) {
+					newValue = newValue.replace(/\./g, "").replace(",", ".");
+				}
+				if (unitValue.length > 0) {
+					newValue = newValue + " " + unitValue;
+				} else if (unit !== undefined) {
```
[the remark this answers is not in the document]
mherwege: Makes sense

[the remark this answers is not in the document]
lolodomo: Bingo. Number without dimension are working again after this change.

lolodomo:
I am not able to reproduce problem 1 (bad value restored) now !

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: I believe lastItemState should set only if itemState is defined. That means it will be updated only when a new item state is received, not when you trigger a reset value to previous one because a wrong value was filled. 
Do you agree?

mherwege: I agree.

mherwege:
Make sure to first change it to another good value and only then try clearing the field. If there is an issue, it may be after a first successful update.

lolodomo:
I just made more tests for numbers whatever the hint and that looks good.
I discovered that when input hint is number, you can use your mouse scroll to increase/decrease the value and it works.
I just found one issue when input hint is number: using the "+" like for example +12 is not possible, the value is rejected. I propose that this issue is fixed later in another PR.

I did not test again date/time hints but I made intensive tests and feedback few months ago on that part.

lolodomo (approved):
LGTM, thank you

mherwege:
Thank you for all your review and adjustment effort.

lolodomo:
Thank you to you first :)

I opened an issue for the remaining (minor) issue. I will try to fix it next week. I should check if this is specific to Firefox or not.

PR #1925, opened 2023-06-13
`oh-sipclient`: Fix audio playback stays active when foreground is left

florian-h05:
I noticed this especially on my iOS when I listened to music and opened the SIP client. After leaving the iOS app, openHAB kept active in the audio playback control and I was unable to continue listening to my music without closing the openHAB app.


florian-h05:
@ghys Can you please merge this fix? Would be cool

ghys (approved):
I'll trust you on this SIP widget as I cannot test it easily :)

florian-h05:
I "battle-test" it as my whole door intercom is build on it, so if there is any problem I will notice very soon!

PR #1932, opened 2023-06-18
Improvements to the add-on page

J-N-K:
- hide author line if author not present
- add correct provisioning information for jar file add-on service
- properly handle description in add-on details page


florian-h05 (approved):
LGTM, thanks!

Just a few things to think about:

- I'd also hide the Documentation link if none is available.
- The Content-Type column is still there, but empty for jar provisioned add-ons. I would either hide it, or even better: If we can tell the content-type based on the source, I'd do this.

J-N-K:
The empty content-type will be fixed by https://github.com/openhab/openhab-core/pull/3664.

florian-h05 (approved):

Issue #1926, opened 2023-06-13
Number:ElectricConductivity missing in mainUI config dialog


NorbertHD:
## The problem
in the openHAB docs [Units Of Measurement | openHAB](https://next.openhab.org/docs/concepts/units-of-measurement.html#list-of-units) at currently
supported units, there is ElectricConductivity listed, but in the item config dialog Number:ElectricConductivity is missing.

![MissingConductivity](https://github.com/openhab/openhab-webui/assets/8472266/9013e3bc-8d37-4a7c-a9dc-71420c881779)

## Expected behavior
Show the Number:ElectricConductivity in the list.

## Steps to reproduce

1. 

## Your environment
Hardware: raspberry pi 4
OS: openhabian 64 bit
Java Runtime Environment: 17
openHAB version: 4 Build #3493

```yaml

```

## Browser console

```txt

```

florian-h05:
See also https://github.com/openhab/openhab-distro/issues/1545.

juantxorena:
> See also [openhab/openhab-distro#1545](https://github.com/openhab/openhab-distro/issues/1545).

Seen. So is this fixed? I only see bugs referencing other bugs that reference other bugs, or merged PR, but I still cannot see ElectricConductivity on the frontend.

florian-h05:
No, this is not fixed. I just linked the issue to be aware of it.

PR #1936, opened 2023-06-24
Items: Fix `unit` metadata not available for UoM groups

### LINKED ISSUE #1933: Missing "Unit" selection for the metadata of a Group Item
I have a Group Item that has Number:Dimensionless.  I would like to set the Unit metadata to "%", but it isn't selectable in the MainUI.  I can use the custom metadata, but I should be able to select it directly as is possible with non-Group Items.

## Expected behavior

Should be able to select Unit metadata for Group Items.

## Steps to reproduce

1. Create a Group Item
2. Try to add Unit metadata - can't see that choice in the MainUI

## Your environment
OH 4 Build 3502

```yaml

```

## Browser console

```txt

```

florian-h05:
Fixes #1933.
Follow-up for #1901.

This enables the unit metadata for UoM groups (i.e. groups with a groupType of `Number:`).

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-menu.vue
@@ -38,7 +38,7 @@ export default {
     }
   },
   beforeMount () {
-    if (this.item.type.indexOf('Number:') < 0) this.metadataNamespaces = this.metadataNamespaces.filter(n => n.name !== 'unit')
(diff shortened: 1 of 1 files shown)


ghys (approved):
LGTM, thanks for keeping up on the UoM stuff

florian-h05:
FYI, there is actually still something missing: On Item creation, there should be a unit field that is pre-filled with the system default unit for the chosen dimension. This way the unit metadata is set for new Items and it is ensured, that even when changing the measurement system, your persisted data won't get corrupted.

https://github.com/openhab/openhab-core/pull/3611 added an REST endpoint for UoM information (all available dimensions with their system default unit), I plan to introduce a Vuex store for this data. I will also use this store for the Item type dialog, we then don't need to keep up with new dimensions as in #1926.

PR #1947, opened 2023-06-29
Developer sidebar: Support search for rule tags

mherwege:
It is possible to filter the rules on tags in the search bar of the rule list, as they are visible on screen. As items can be searched by tag in the developer toolbar search, I think it is still useful to extend the rule search in the developer toolbar to tags as well. I tend to give my rules tags, often corresponding to tags on items as well. This allows finding all of these in one place.


florian-h05 (approved):
LGTM, also thanks for fixing the comment.

PR #1948, opened 2023-07-02
context block typing, extended Qty block var type support

### LINKED ISSUE #1942: [blockly] contextual info new state of item cannot be interpreted as QuantityType
## The problem

Edit: problem is somewhere else, title adjusted and problem description is https://github.com/openhab/openhab-webui/issues/1942#issuecomment-1611078320.

The current quantity computation block only allows 2 arguments with the same dimension.

## Your suggestion

Ideally, for division and multiplication, it should be possible to have a decimal value as a second argument. It is a common use case to multiply/divide a quantity with a constant or calculated value.

## Additional information

As a bonus (but not critical), it would be nice to align the symbols for multiplication and division with the standard ones in the math calculation block.
The quantity computation block uses * and /. The standard block uses × and ÷.

stefan-hoehn:
Fixes #1942.
Requires version 4.5.0 of openhab-js, therefore awaiting https://github.com/openhab/openhab-addons/pull/15219.

- Adds context block typings.
- Adds supports for Item state input in addition to Item object only in case Qty Block has a var as input.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
@@ -173,6 +173,7 @@ export default function defineOHBlocks_Scripts (f7, isGraalJs, scripts) {
           ['triggered event', 'event']
         ]),
         'contextInfo')
+      this.contextInfo = this.getFieldValue('contextInfo')
       this.setInputsInline(true)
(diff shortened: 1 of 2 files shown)


florian-h05:
I will review this once I have released the new required version and it is included in the addon.

florian-h05 (approved):
LGTM, let‘s wait for the merge of https://github.com/openhab/openhab-addons/pull/15219.

PR #1955, opened 2023-07-11
Persistence configuration page: Minor enhancements & fixes

florian-h05:
Follow-up for #1917.

- Display enabled filters in config footer.
- Display filter configuration in filter footer.
- Time filter: Define options for time unit.
- Add a button to open the documentation.
- Fixes an issue, where Filter creation fails due to missing filter type arrays.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
@@ -21,6 +21,16 @@
     <f7-tabs>
       <!-- Design Tab -->
       <f7-tab id="design" @tab:show="() => this.currentTab = 'design'" :tab-active="currentTab === 'design'">
+        <f7-block class="block-narrow">
+          <f7-col>
+            <f7-block-footer>
+              Persistence stores data over time, which can be retrieved at a later time, e.g. to restore Item states after startup, or to display graphs in the UI.
+              <f7-link external color="blue" target="_blank" :href="`https://${$store.state.runtimeInfo.buildString === 'Release Build' ? 'www' : 'next'}.openhab.org/link/persistence`">
+                Learn more about persistence.
+              </f7-link>
+            </f7-block-footer>
+          </f7-col>
+        </f7-block>
         <!-- Skeletons for not ready -->
         <f7-block v-if="!ready" class="block-narrow">
           <f7-col class="modules">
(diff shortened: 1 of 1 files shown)


florian-h05:
@ghys Can you please review and merge this for openHAB 4? Probably to late for the feature freeze, but this PR contains a critical bug fix so IMO should be merged anyway.

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
```diff
@@ -162,9 +162,12 @@
               </div>
             </div>
           </f7-col>
-          <f7-col v-if="isEditable && !newPersistence">
+          <f7-col>
             <f7-list>
-              <f7-list-button color="red" @click="deletePersistence">
+              <f7-list-button color="blue" :href="`https://${$store.state.runtimeInfo.buildString === 'Release Build' ? 'www' : 'next'}.openhab.org/docs/configuration/persistence.html`" external target="_blank">
+                Open documentation
+              </f7-list-button>
```
ghys, line 169: A list button is not the best for linking to the docs IMHO, those are better suited for internal actions. For those concepts that are a little obscure and need a link to the docs a f7-link labeled "Learn more [about X]" or "About X" would be better, see the profile configuration for example (the blurb before the link can be optional):

https://github.com/openhab/openhab-webui/blob/9be42930d957c22e55c7b5e7931a9f24123e2c82/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue#L49-L53

or iOS settings (note how the list buttons does something while the documentation links are simply blue hyperlinks):

image

Also, as shown above for profiles (the www/next switch is missing there, sadly!) it would be better to create a `openhab.org/link/persistence` alias link or similar, that way if the docs are reorganized it wouldn't require a change in the code but only an update to the redirect. This happens here: https://github.com/openhab/website/blob/main/.vuepress/_redirects

ghys (changes requested):
Two minor remarks on the docs link, the rest LGTM.

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
florian-h05: I've added a short paragraph with the link to the top and created a PR for the redirect, thanks for your feedback, looks much better now!

See https://github.com/openhab/website/pull/413.

ghys (approved):
Thank you!

PR #1961, opened 2023-07-12
Add cacheSource repeater property

### LINKED ISSUE #1956: Repeater static values vs dynamic refresh
I'm moving [the conversation on the forms](https://community.openhab.org/t/cant-refresh-oh-repeater-widget-using-key/147667) here after further testing:

Quick recap:
Issue #1804 pointed out that oh-repeaters using a calculated value for the array source could be a significant drain on processing in larger systems due to constant api calls when the component is redrawn.

PR #1904 addressed this issue by calculating source once on component creation and caching that array for all future redraws of the repeater and its children.

Issue #1918 (and related forum topics) pointed out that this broke the repeater for several user who relied on the ability to refresh repeater information without a full page refresh.  This issue included an example where the `in` array of the repeater was dynamically defined but the repeater contents would not change even when the input array changed.

PR #1919 addressed this issue by creating a carve out in the caching method for the repeater's `in` array definition.  This did not fix the non-refresh issue for most users, especially users as discussed in the above linked forum topic.

The refresh issue can be addressed by users by, instead of refreshing the repeater directly, refreshing one of the repeater's parents.  This solution, however, is suboptimal because that destroys the repeater and redraws it from scratch which results in a significant visual change in the UI when elements shift with the repeater's removal and shift back when it is redrawn.

Now:
After further testing I think there is an argument for going back to the previous computed source method that refreshes the repeater in response to object and css changes, but to add a new repeater property (e.g., `static`) that caches the array and then prevents further updates.

As I was looking at the changes in PR 1919 that allowed for the `in` property to override the cached source array I realized that there a many more reasonable situations parallel to the demo widget used in 1918 that apply as well.  And it's likely that broad adoption of OH4 with this change will break a lot of widgets.

For example, let's think about a common enough simple widget that uses an array to list the equipment in a location.  Easy, that's just a `itemsInGroup` source with an `itemGroup` property.  But, now let's say the user wants three buttons which change that group between Kitchen, Dining room, and Front Hall.  That's also easy, `itemGroup` can be set to an expression with the variable value that's determined by the buttons.  The problem is, with this new system, that repeater won't refresh when any of the button are pushed.  A change to the variable will cause the repeater to be re-evaluated, but the cached array will be used and there will be no change in the repeater output.  So, now we have to create a second carve out in the cache method for the `itemsInGroup` kind of arrays.

The problem is we can do this for each type of source.  It's trivial to change the example above to a widget that lists the all Speaker equipment (an `itemsWithTags` source type) but with buttons that select other equipment types (again, a very reasonable kind of widget).  So, now we need to consider a carve out for `itemsWithTags`.

Based on this thought process, I've come to the conclusion that it is more sensible to return the repeater closer to it's original form while allowing users to opt in to caching the source array.
1) This is less of a breaking change
2) Because it uses a property to define whether to cache the source it can even be dynamically defined so that users have real-time control over whether to use caching or not.
3) The modification is very light and I have already tested what I believe to be a fully working version.

Before submitting a PR with the change, however, I wanted to open a direct discussion about it, because it does fully revert the changes in 1904 and 1919 by returning to the `asynComputed` method for `source`

Edit:
As discussed at the end of the forum topic, I did some testing as well with binding the key property properly to the repeater.  This has the same undesirable visual impact as refreshing a repeater parent element because the repeater is destroyed and redrawn.

In contrast, when the repeater source array is just recalculated as the result of an object change, the repeater's elements are maintained and only the children change resulting in a smoother visual transition.

JustinGeorgi:
fixes #1956 

Adds a boolean property to the repeater (and relevant docs), `cacheSource` which allows the user to determine whether the repeater should refresh the source array with css and object changes or cache the source array for performance improvements.

Signed-off-by : Justin Georgi justin.georgi@gmail.com

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/repeater.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/repeater.js
@@ -28,5 +28,5 @@ export default () => [
   pt('containerClasses', 'Classes of the container', 'Add these CSS classes to the container'),
   pt('containerStyle', 'Styles of the container', 'Add these CSS styles to the container'),
   pb('fragment', 'No container (fragment)', 'Render all children directly under the repeater\'s parent, without any container'),
-  pb('cacheSource', 'Suppress source refresh', 'The source array will be cached and not refreshed on page updates')
+  pb('cacheSource', 'Suppress source refresh', 'For loaded sources (e.g. with Items or rules), the source array will be cached and not refreshed on page updates')
 ]
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/repeater.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/repeater.js
@@ -27,5 +27,6 @@ export default () => [
   pb('listContainer', 'List container', 'The child components will be wrapped in a <code>ul</code> HTML elements instead of a <code>div</code>'),
   pt('containerClasses', 'Classes of the container', 'Add these CSS classes to the container'),
   pt('containerStyle', 'Styles of the container', 'Add these CSS styles to the container'),
-  pb('fragment', 'No container (fragment)', 'Render all children directly under the repeater\'s parent, without any container')
+  pb('fragment', 'No container (fragment)', 'Render all children directly under the repeater\'s parent, without any container'),
+  pb('cacheSource', 'Suppress source refresh', 'The source array will be cached and not refreshed on page updates')
 ]
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-repeater.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-repeater.vue
@@ -75,6 +75,7 @@ export default {
   },
   asyncComputed: {
     source () {
+      if (this.config.cacheSource && this.sourceCache) return this.sourceCache
       let sourceResult
       if (this.config.sourceType === 'range') {
         const start = this.config.rangeStart || 0
@@ -82,24 +83,23 @@ export default {
         const step = this.config.rangeStep || 1
         sourceResult = Promise.resolve(Array(Math.ceil((stop + 1 - start) / step)).fill(start).map((x, y) => x + y * step))
       } else if (this.config.sourceType === 'itemsWithTags' && this.config.itemTags) {
-        if (this.config.cacheSource && this.sourceCache) return this.sourceCache
         sourceResult = this.$oh.api.get('/rest/items?metadata=' + this.config.fetchMetadata + '&tags=' + this.config.itemTags).then((d) => Promise.resolve(d.sort(compareItems)))
+        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
       } else if (this.config.sourceType === 'itemsInGroup') {
-        if (this.config.cacheSource && this.sourceCache) return this.sourceCache
         sourceResult = this.$oh.api.get('/rest/items/' + this.config.groupItem + '?metadata=' + this.config.fetchMetadata + '&tags=' + this.config.itemTags).then((i) => Promise.resolve(i.members.sort(compareItems)))
+        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
       } else if (this.config.sourceType === 'itemStateOptions') {
-        if (this.config.cacheSource && this.sourceCache) return this.sourceCache
         sourceResult = this.$oh.api.get('/rest/items/' + this.config.itemOptions).then((i) => Promise.resolve((i.stateDescription) ? i.stateDescription.options : []))
+        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
       } else if (this.config.sourceType === 'itemCommandOptions') {
-        if (this.config.cacheSource && this.sourceCache) return this.sourceCache
         sourceResult = this.$oh.api.get('/rest/items/' + this.config.itemOptions).then((i) => Promise.resolve((i.commandDescription) ? i.commandDescription.commandOptions : []))
+        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
       } else if (this.config.sourceType === 'rulesWithTags' && this.config.ruleTags) {
-        if (this.config.cacheSource && this.sourceCache) return this.sourceCache
         sourceResult = this.$oh.api.get('/rest/rules?summary=true' + '&tags=' + this.config.ruleTags).then((r) => Promise.resolve(r.sort(compareRules)))
+        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
       } else {
         sourceResult = Promise.resolve(this.config.in)
       }
-      this.sourceCache = (this.config.cacheSource) ? sourceResult : null
       return sourceResult
     }
   }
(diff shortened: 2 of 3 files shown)


bundles/org.openhab.ui/web/src/components/widgets/system/oh-repeater.vue
```diff
@@ -22,15 +22,6 @@ export default {
     Fragment
   },
   widget: OhRepeaterDefinition,
-  data () {
-    return {
-      ready: false,
```
florian-h05, line 27: You need to add `sourceCache` to data, otherwise Vue will complain during runtime.

florian-h05 (commented):
Code looks good, but I would prefer that even when `cacheSource` is enabled, only loaded sources are cached. This could be accomplished by just moving the this.sourceCache line into the if blocks for loaded sources.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/repeater.js
```diff
@@ -27,5 +27,6 @@ export default () => [
   pb('listContainer', 'List container', 'The child components will be wrapped in a <code>ul</code> HTML elements instead of a <code>div</code>'),
   pt('containerClasses', 'Classes of the container', 'Add these CSS classes to the container'),
   pt('containerStyle', 'Styles of the container', 'Add these CSS styles to the container'),
-  pb('fragment', 'No container (fragment)', 'Render all children directly under the repeater\'s parent, without any container')
+  pb('fragment', 'No container (fragment)', 'Render all children directly under the repeater\'s parent, without any container'),
+  pb('cacheSource', 'Suppress source refresh', 'The source array will be cached and not refreshed on page updates')
```
florian-h05, line 31: After applying my suggestion to not cache the not loaded sources:

```suggestion
  pb('cacheSource', 'Suppress source refresh', 'For loaded sources (e.g. with Items or rules), the source array will be cached and not refreshed on page updates')
```

WDYT?

JustinGeorgi:
> I would prefer that even when `cacheSource` is enabled, only loaded sources are cached

That makes good sense now that you point it out.  That should all be taken care of.

florian-h05:
I would prefer to early return when cached, and to store the data in the cache inside the if-clauses instead of returning from there.

I would commit, but I cannot because your main branch is protected, can you please commit that code:

```js
<template>
  <ul v-if="config.listContainer" :class="config.containerClasses" :style="config.containerStyle">
    <generic-widget-component :context="ctx" v-for="(ctx, idx) in childrenContexts" :key="'repeater-' + idx" @command="onCommand" />
  </ul>
  <fragment v-else-if="config.fragment">
    <generic-widget-component :context="ctx" v-for="(ctx, idx) in childrenContexts" :key="'repeater-' + idx" @command="onCommand" />
  </fragment>
  <div v-else :class="config.containerClasses" :style="config.containerStyle">
    <generic-widget-component :context="ctx" v-for="(ctx, idx) in childrenContexts" :key="'repeater-' + idx" @command="onCommand" />
  </div>
</template>

<script>
import mixin from '../widget-mixin'
import { OhRepeaterDefinition } from '@/assets/definitions/widgets/system'
import { compareItems, compareRules } from '@/components/widgets/widget-order'
import { Fragment } from 'vue-fragment'

export default {
  mixins: [mixin],
  components: {
    Fragment
  },
  widget: OhRepeaterDefinition,
  data () {
    return {
      sourceCache: null
    }
  },
  computed: {
    childrenContexts () {
      const iterationContext = (ctx, el, idx, source) => {
        // takes the context with the added variables
        const loopVars = {}
        if (ctx.loop) {
          for (const loopKey in this.context.loop) {
            this.$set(loopVars, loopKey, this.context.loop[loopKey])
          }
        }
        loopVars[this.config.for] = el
        loopVars[this.config.for + '_idx'] = idx
        loopVars[this.config.for + '_source'] = source

        this.$set(ctx, 'loop', loopVars)

        return ctx
      }

      let source = this.source
      if (!Array.isArray(source)) return []

      if (this.config.filter) {
        source = source.filter((el, idx, source) =>
          this.evaluateExpression('filterExpr', '=' + this.config.filter,
            iterationContext(this.childContext(this.context.component), el, idx, source)))
      }
      if (this.config.map) {
        source = source.map((el, idx, source) =>
          this.evaluateExpression('mapExpr', '=' + this.config.map,
            iterationContext(this.childContext(this.context.component), el, idx, source)))
      }

      let contexts = []
      let idx = 0
      for (let i of source) {
        contexts.push(...this.context.component.slots.default.map((c) => {
          return iterationContext(this.childContext(c), i, idx, source)
        }))

        idx++
      }

      return contexts
    }
  },
  asyncComputed: {
    source () {
      if (this.config.cacheSource && this.sourceCache) return this.sourceCache
      let sourceResult
      if (this.config.sourceType === 'range') {
        const start = this.config.rangeStart || 0
        const stop = this.config.rangeStop || 10
        const step = this.config.rangeStep || 1
        sourceResult = Promise.resolve(Array(Math.ceil((stop + 1 - start) / step)).fill(start).map((x, y) => x + y * step))
      } else if (this.config.sourceType === 'itemsWithTags' && this.config.itemTags) {
        sourceResult = this.$oh.api.get('/rest/items?metadata=' + this.config.fetchMetadata + '&tags=' + this.config.itemTags).then((d) => Promise.resolve(d.sort(compareItems)))
        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
      } else if (this.config.sourceType === 'itemsInGroup') {
        sourceResult = this.$oh.api.get('/rest/items/' + this.config.groupItem + '?metadata=' + this.config.fetchMetadata + '&tags=' + this.config.itemTags).then((i) => Promise.resolve(i.members.sort(compareItems)))
        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
      } else if (this.config.sourceType === 'itemStateOptions') {
        sourceResult = this.$oh.api.get('/rest/items/' + this.config.itemOptions).then((i) => Promise.resolve((i.stateDescription) ? i.stateDescription.options : []))
        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
      } else if (this.config.sourceType === 'itemCommandOptions') {
        sourceResult = this.$oh.api.get('/rest/items/' + this.config.itemOptions).then((i) => Promise.resolve((i.commandDescription) ? i.commandDescription.commandOptions : []))
        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
      } else if (this.config.sourceType === 'rulesWithTags' && this.config.ruleTags) {
        sourceResult = this.$oh.api.get('/rest/rules?summary=true' + '&tags=' + this.config.ruleTags).then((r) => Promise.resolve(r.sort(compareRules)))
        this.sourceCache = (this.config.cacheSource) ? sourceResult : null
      } else {
        sourceResult = Promise.resolve(this.config.in)
      }
      return sourceResult
    }
  }
}
</script>

```

JustinGeorgi:
> I would prefer to early return when cached, and to store the data in the cache inside the if-clauses instead of returning from there.

Sorry, I misintrepreted your previous request :rofl:.  It seemed strange to me that you were asking to move the early return into the if blocks, but figured you knew something I didn't.  This should put it back right with the caching in the if blocks.

florian-h05:
> Sorry, I misintrepreted your previous request 🤣. 

Oops, sorry 🙈 

> Also I've invited you the pr repo if you need to change anything else. I'm about to get on a transoceanic flight.

Thanks, I think the problem is that you created the PR from your main and not a feature branch. I usually don't have any problems writing to PR branches. I will take care of the linting failure and then merge this.

florian-h05:
@JustinGeorgi 
Even as a collaborator I cannot commit because you enabled branch protection for your main branch.
Due to lack of time, I'll open a new PR to the openHAB repo from my branch and merge this one.

florian-h05:
Superseded by #1963.

PR #1963, opened 2023-07-15
`oh-repeater`: Make caching configurable

### LINKED ISSUE #1956: Repeater static values vs dynamic refresh

florian-h05:
Fixes #1956.

Adds a boolean property to the repeater (and relevant docs), `cacheSource` which allows the user to determine whether the repeater should refresh the source array with css and object changes or cache the source array for performance improvements.

Signed-off-by : Justin Georgi <justin.georgi@gmail.com>


florian-h05:
Supersedes #1961, where lint was failing and the contributor was not available to fix it and I was not able to push to his branch. 
As I just opened this PR and did not author it, I’ll self-merge.

PR #1969, opened 2023-07-17
Upgrade SAT and Spotless

wborn:
This upgrades SAT and Spotless so they can correctly check and format Java 17 code.


kaikreuzer (approved):
Thanks!

Issue #1965, opened 2023-07-06
Cant add colorpicker widget to Group


proggprod:
## Expected Behavior

I have a group that contains all 'Color' items of my lights. I want to be able to send a color to all items by using a colorpicker widget on the group. But when I choose colorpicker as the default standalone widget, I see nothing

![image](https://github.com/openhab/openhab-core/assets/81530084/de511a1f-7201-40a5-81a7-2b3029e34434)

florian-h05:
You have to set the group type to color and ensure that at least one of the members has a state. If the Item state of the Item for the color picker is no color (e.g. because it is `NULL`), the color picker won't show up.

I however will add an option to the color picker that allows the user to define a default value that will be used in that case.

PR #1971, opened 2023-07-18
oh-colorpicker: Add defaultColor option

florian-h05:
Related to #1965 and IMHO closes it.


florian-h05:
@ghys This is the best idea I had with regard to handling the linked issue.

If you have a different idea, please let me know.

ghys (approved):
I'm okay with the principle, as long as it doesn't have side effects (there were problems of unwanted commands sent reacting to value change event firing with this color control in the past)

PR #1982, opened 2023-07-23
Persistence page: Fix config update on filter removal

florian-h05:
When removing a filter, the persistence editor checks which configs use that filter and removes the filter from those configs, otherwise the save request fails with an HTTP error.

However, the implementation was buggy and also removed other filters from configs because of a missing check.


kaikreuzer (approved):
Thanks!

PR #1984, opened 2023-07-24
Developer sidebar: Fix Thing pinning fails

florian-h05:
Regression from #1661.

Using the cacheable list for Things leads to a failure because the status field is not available then.
This reverts the change from #1661 to this single REST request.


florian-h05:
@ghys Sorry, you were right that staticDataOnly does not work for some of the requests made by the developer sidebar search. This fix should be backported to 4.x.

ghys (approved):

PR #1989, opened 2023-07-26
oh-colorpicker: Fix command not sent when state null & style not accepted & cosmetic error on missing defaultColor

### LINKED ISSUE #1713: `oh-colorpicker` component doesn‘t accept `style` configuration
## The problem

All `oh-` components accept CSS style configuration in the widget editor, but `oh-colorpicker` does apply the `style` configuration.

## Expected behavior

`oh-colorpicker` should apply `style` configuration.

## Steps to reproduce

Create a new widget and add this code to the slots.default section:

```yaml
    - component: oh-colorpicker
      config: 
        item: Florian_RGB
        style:
          position: absolute
          right: 5px
          top: 65px
          height: 60px
          width: 60px
```

## Your environment

```yaml
openHAB 3.4.1
```

## Browser console

Nothing related to this issue.

## Additional information

Wrapping `oh-colorpicker` inside a `div` works:

```yaml
    - component: div
      config: 
        style:
          margin: 0 0 0 0
          position: absolute
          right: 5px
          top: 65px
          height: 60px
          width: 60px
      slots:
        default:
          - component: oh-colorpicker
            config:
              item: Florian_RGB
              modules:
              - current-color
```

florian-h05:
Fixes #1713.

- Fix command not sent when current Item state is null.
- Fix style can not be set.
- Fix a (cosmetic) error thrown when defaultColor has no value.

@ghys Please test this so we don't run into infinite command loops.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-colorpicker.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-colorpicker.vue
@@ -1,11 +1,13 @@
 <template>
-  <div v-if="!config.openIn" ref="container" style="width: 100%" />
-  <div v-else ref="swatch" :class="config.swatchClasses || ['elevation-4', 'elevation-hover-8', 'elevation-pressed-1', 'elevation-transition']" :style="{
+  <div :style="config.style">
+    <div v-if="!config.openIn" ref="container" style="width: 100%" />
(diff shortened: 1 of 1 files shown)


florian-h05:
@ghys Can you please have a look at that?

ghys (approved):
Sorry for letting this slide; I can't really test myself as I've never been confronted to the situation but I remember reports when for example HSB components were rounded and it caused trouble like these infinite loops. I guess if there are side-effects we'll know soon enough :)

PR #1997, opened 2023-07-30
Redesign settings page & Add doc links

### LINKED ISSUE #1935: Redesign (add-on) configuration page
*In https://github.com/openhab/openhab-webui/pull/1891#issuecomment-1589767165 @ghys wrote:*

> In fact maybe having the add-on's (formerly only binding's) configuration in the add-on store was a bad idea after all, as we could just have add-ons add entries to the settings menu like on iOS. So you would have e.g. "Philips Hue" directly under Settings and users wouldn't have to look for it in the add-on store, which was in the end really meant for browsing, installing and uninstalling, nothing more.
>  As for the logging, in fact it could have been a single consolidated page where you would configure levels for all loggers at once.
>
>  The removal of "Other Services" however offers an opportunity to adjust the Settings menu to keep it balanced when presented on two columns, as we could layout it like this:
> 
>  - Left column
>    - 1st-tier Configuration: Things, Model, Items, Pages
>    -  2nd-tier Configuration: Persistence, Tags (?), Transformations no need to have a section header title for those, just separate them from the 1st tier. There would be room for more, for instance "Metadata Namespaces" (allow the user to define their own namespaces).
>    -  Automation: Rules, Scenes, Scripts, Schedule
>  -  Right column
>    -  Add-ons: Bindings, Automation, UI, Other
>    -  Services (formerly System Services)

*In https://github.com/openhab/openhab-webui/pull/1891#issuecomment-1590818187 @florian-h05 wrote:*

> > 2nd-tier Configuration: Persistence, ...

> I'd probably keep persistence out of the main settings and only have it under the add-on settings in the right column.

For the addon setting, I can imagine having the list of installed addon with its icon and instead of just opening the addon settings from the config description, open the addon settings page as in the addon store.

florian-h05:
Closes #1935.

- Moves transformations to a new list for visual separaration, as it is only second class configuration.
- Renames System Services to System Settings.
- Add Add-on Settings, which opens the same settings page as the add-on store, i.e. provides log, service config and persistence config (if persistence service). Add-ons that only provide log settings are considered advanced and hidden by default.
- Consolidate the add-on store and the newly added add-on settings into a single place (refactoring included).
- Add doc links to Things, Items, pages, rules, scenes & scripts list pages.
- Make some system settings advanced.
- Use a third column for large screens.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -33,7 +33,7 @@
         <f7-block-title v-if="$store.getters.isAdmin" v-t="'sidebar.administration'" />
         <f7-list class="admin-links" v-if="$store.getters.isAdmin">
           <f7-list-item link="/settings/" :title="$t('sidebar.settings')" view=".view-main" panel-close :animate="false"
-                        :class="{ currentsection: currentUrl === '/settings/' || currentUrl.indexOf('/settings/addons/') >= 0 || currentUrl.indexOf('/settings/services/') >= 0 }">
+                        :class="{ currentsection: currentUrl === '/settings/' || currentUrl.indexOf('/settings/services/') >= 0 }">
             <f7-icon slot="media" ios="f7:gear_alt_fill" aurora="f7:gear_alt_fill" md="material:settings" color="gray" />
           </f7-list-item>
           <li v-if="showSettingsSubmenu">
@@ -72,6 +72,11 @@
               </f7-list-item>
             </ul>
           </li>
+
+          <f7-list-item link="/settings/addons/" :title="$t('sidebar.addOnStore')" panel-close :animate="false"
+                        :class="{ currentsection: currentUrl.indexOf('/settings/addons/') >= 0 }">
+            <f7-icon slot="media" ios="f7:bag" aurora="f7:bag" md="material:shopping_bag" color="gray" />
+          </f7-list-item>
 
           <f7-list-item link="/developer/" :title="$t('sidebar.developerTools')" panel-close :animate="false"
                         :class="{ currentsection: currentUrl.indexOf('/developer/') >= 0 && currentUrl.indexOf('/developer/widgets') < 0 &&
@@ -749,7 +754,7 @@ export default {
 
       this.$f7.on('pageBeforeIn', (page) => {
         if (page.route && page.route.url) {
-          this.showSettingsSubmenu = page.route.url.indexOf('/settings/') === 0
+          this.showSettingsSubmenu = page.route.url.indexOf('/settings/') === 0 && page.route.url.indexOf('addons') === -1
           this.showDeveloperSubmenu = page.route.url.indexOf('/developer/') === 0
           this.currentUrl = page.route.url
         }
--- a/bundles/org.openhab.ui/web/src/assets/addon-store.js
+++ b/bundles/org.openhab.ui/web/src/assets/addon-store.js
@@ -29,33 +29,6 @@ export const Formats = {
   'karaf': 'Karaf'
 }
 
-export const AddonStoreTabShortcuts = [
-  {
-    id: 'bindings',
-    label: 'Bindings',
-    icon: 'circle_grid_hex',
-    subtitle: 'Connect and control hardware and online services'
-  },
-  {
-    id: 'automation',
-    label: 'Automation',
-    icon: 'sparkles',
-    subtitle: 'Scripting languages, templates and module types'
-  },
-  {
-    id: 'ui',
-    label: 'User Interfaces',
-    icon: 'play_rectangle',
-    subtitle: 'Community widgets & alternative frontends'
-  },
-  {
-    id: 'other',
-    label: 'Other Add-ons',
-    icon: 'ellipsis',
-    subtitle: 'System integrations, persistence, voice & more'
-  }
-]
-
 export function compareAddons (a1, a2) {
   if (a1.installed && !a2.installed) return -1
   if (a2.installed && !a1.installed) return 1
--- a/bundles/org.openhab.ui/web/src/assets/i18n/common/en.json
+++ b/bundles/org.openhab.ui/web/src/assets/i18n/common/en.json
@@ -34,6 +34,7 @@
   "sidebar.noPages": "No pages",
   "sidebar.administration": "Administration",
   "sidebar.settings": "Settings",
+  "sidebar.addOnStore": "Add-on Store",
   "sidebar.developerTools": "Developer Tools",
   "sidebar.helpAbout": "Help & About",
   "sidebar.unlockAdmin": "Unlock Administration",
--- a/bundles/org.openhab.ui/web/src/assets/i18n/empty-states/en.json
+++ b/bundles/org.openhab.ui/web/src/assets/i18n/empty-states/en.json
@@ -11,6 +11,9 @@
   "things.nobindings.title": "No bindings",
   "things.nobindings.text": "To add things to your system, you first need to install binding add-ons.",
 
+  "transformations.title": "No transformations yet",
+  "transformations.text": "Transformations are used to translate data from a cluttered or technical raw value to a processed human-readable representation.",
+
   "model.title": "Start modelling your home",
   "model.text": "Build a model from your items to organize them and relate them to each other semantically.<br><br>Begin with a hierarchy of locations: buildings, outside areas, floors and rooms, as needed. Then, insert equipment and points from your things (or manually).",
 
--- a/bundles/org.openhab.ui/web/src/js/routes.js
+++ b/bundles/org.openhab.ui/web/src/js/routes.js
@@ -9,7 +9,7 @@ import AnalyzerPopup from '../pages/analyzer/analyzer-popup.vue'
 const AboutPage = () => import(/* webpackChunkName: "about-page" */ '../pages/about.vue')
 const UserProfilePage = () => import(/* webpackChunkName: "profile-page" */ '../pages/profile.vue')
 
-const SettingsMenuPage = () => import(/* webpackChunkName: "admin-base" */ '../pages/settings/settings-menu.vue')
+const SettingsMenuPage = () => import(/* webpackChunkName: "admin-base" */ '../pages/settings/menu/settings-menu.vue')
 const ServiceSettingsPage = () => import(/* webpackChunkName: "admin-base" */ '../pages/settings/services/service-settings.vue')
 const AddonsListPage = () => import(/* webpackChunkName: "admin-base" */ '../pages/settings/addons/addons-list.vue')
 const AddonsAddPage = () => import(/* webpackChunkName: "admin-base" */ '../pages/settings/addons/addons-add.vue')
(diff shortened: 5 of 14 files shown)


florian-h05:
image

florian-h05:
@ghys Please let me know what you think.

ghys:
Many thanks @florian-h05 for giving this a shot! It's pretty much what I imagined. I very much like the reintroduction of direct links to the add-on settings which makes more sense from a UX perspective than having to find them again in the add-on store and then finding and clicking on that little gears icon. Seeing that screenshot lead to more thoughts and ideas:

- the 2-tier configuration section is great, but maybe it's premature when the only entry in the 2nd part is "Transformations" which is therefore kind of lonely in there atm - not necessarily a huge problem though
- about this "Transformations" entries, the icon I chose for these in the add-on store and its predecessor was `function` from the f7 icons (the _"f(x)"_ one) which is maybe more representative than the one here? Also the description - "Manage transformations" - is a bit dull and could use some more creativity with a short sentence in the same spirit as the ones above it. Not offering suggestions at this stage, just saying.
- There are more and more system settings added over time and also your screenshot only has 4 add-on settings entries, which I'd say would be rather below average. The right column would inevitably grow much taller than the left one which doesn't change much.
- Maybe it's time to introduce a "show advanced" checkbox on this screen as on config sheets and some other places. For instance "Transformations" would be considered advanced and shown only when that box is checked; some system settings would also qualify for sure (at least "JSON Storage", "3rd-party Json add-on service", "Network Settings", and maybe "Charts" would be considered advanced given their current options). Eventually individual add-on settings would be considered advanced if they wouldn't offer any configuration properties nor persistence configuration but only log level settings (probably hard to figure out in a performant way).

I don't think we have to rush this, we can discuss this and schedule it for the next milestone or the one after. I've learned from other experiences that too many big changes in critical parts of an UI as we adjust and make up our minds can lead to confusion and frustration.

florian-h05:
I had a few thoughts about your comment:

> the 2-tier configuration section is great, but maybe it's premature when the only entry in the 2nd part is "Transformations" which is therefore kind of lonely in there atm - not necessarily a huge problem though

I understand what you mean, but when openHAB 4.1 is released hopefully transformations aren't that lonely anymore, because we have a page to manage the semantic tags in place.

Regarding transformations: 
I like the function icon more, I've updated it together with the description. Not so easy to come up with something short, but still meaningful.

> There are more and more system settings added over time and also your screenshot only has 4 add-on settings entries, which I'd say would be rather below average. The right column would inevitably grow much taller than the left one which doesn't change much.

For system settings, it is difficult to decide whether they are advanced or not, and I don't want to hardcode a list - I'd prefer to show them all.
For addons, I found a easy way to determine whether the add-on provides settings, just check the list of installed add-ons against the list of non-system services. This works by converting the add-on uid `automation-jsscripting` to configDescriptionURI `automation:jsscripting`, just a simple string replacement. If the add-on is a persistence add-on, it also is not considered advanced.


Now looks like this:

image

Note that I chose to add a list button "Show Advanced" (and also "Hide Advanced") because IMHO this is more intuitive than placing the "Show advanced" checkbox in the upper right corner.

ghys:
Thanks! I think it looks great, I have a couple little more nitpicks obviously:

- I would replace the "Show Advanced" label for add-ons to "Show All" and actually remove the button to collapse again. "Show Advanced" is not quite right because it could be interpreted as a hierarchy between the add-ons themselves, some being less important than others. It's a pattern you can come across on iOS where you have a filtered list (for instance, recently accessed) and have a way to show it unfiltered if you need, and it will render as collapsed again when the page is recreated anyway. Bonus points if the list is sorted so that when you hit "Show All" the entries are not reordered, but the new ones are just appended to the list.
- I'm not too sure the descriptions added with the last commit are really necessary, especially when the list is empty, you'll have both that and the "empty state placeholder" which will not fit well on a phone screen (I specifically made sure of that) and offer redundant information.

   

   Maybe a button to the documentation could be added to those instead, like here?

ghys:
> It's a pattern you can come across on iOS where you have a filtered list (for instance, recently accessed) and have a way to show it unfiltered if you need

Incidentally this is something which should be beneficial in other places, for instances the sessions list in the "profile" page, as illustrated on https://demo.openhab.org/profile/
Here the list takes ages to render when maybe the 20/50/100 most recent sessions could be displayed by default with a "Show All" button to show the rest.

lsiepel:
> ddons, I found a easy way to determine whether the add-on provides settings, just check the list of installed add-ons against the list of non-system services. (That works by checking if there is any non-system service whose configDescriptionURI's second part (the servcie name) equals the add-on id). If the add-on is a persistence add-on, it also is not considered advanced.

The current two column layout is ideal for 1024px, but has some limitations. Hiding visual items behind a 'show all' tab would be great and necessaire for the smaller viewports. Many of us have screens with a much higher resolution.
I would suggest make it more repsonseive by adding a third column (for the addons) for viewports with min 1440px and only show the 'Show all' for the smaller viewports that have two column layout.

A quick copy/paste paint mock:
image

ghys:
That's a fair remark, however the settings screen has been carefully designed with a 1280px screen width in mind (kind of the middle point), also considering you can use the developer sidebar at this resolution and above, so it had to fit just right:

image

image

Adding columns could be done but the developer sidebar reduces the available width without adjusting the viewport size, so all of this has to be taken into account.

> Hiding visual items behind a 'show all' tab would be great and necessaire for the smaller viewports.

This is also influenced by the fact that not all add-ons have settings to configure, some would just be log level settings, that's why they're behind a "show all" filter, as they don't have anything specific to configure.

Also in your mockup I would move the add-on store items to the middle column. This drives me to another completely different idea: we could have a consolidated "add-ons" section presenting the installed add-on settings, then a single "add-on store" link to add more.

florian-h05:
> I would replace the "Show Advanced" label for add-ons to "Show All" and actually remove the button to collapse again. "Show Advanced" is not quite right because it could be interpreted as a hierarchy between the add-ons themselves, some being less important than others.

Agreed, and already commited.

> I'm not too sure the descriptions added with the last commit are really necessary, especially when the list is empty, you'll have both that and the "empty state placeholder" which will not fit well on a phone screen (I specifically made sure of that) and offer redundant information.

Good point. I have added documentation buttons to the "empty state placeholders", but I wanted to keep a simple way to open the docs right from the UI, so I extended those "x Items" lines to "x Items - Open documentation".

Regarding the number of columns and ordering, I will have a look once I have the time.

jimtng:
> This drives me to another completely different idea: we could have a consolidated "add-ons" section presenting the installed add-on settings, then a single "add-on store" link to add more.

I love this idea, similar to how browser extensions are managed. It also gives us a place to see all the addons that are currently installed, and that it would load much more quickly.

florian-h05:
I have consolidated the add-on store and the add-on settings into a single block (even refactored them to their own component), introduced advanced system settings (hard-coded inside the UI) and added a third column for large screens.

On a smaller screen:

image

On a larger screen (note that I have chosen a good brakepoint to use the developer sidebar the same time):

image

ghys:
This last design might satisfy everybody's needs/be a good compromise! :) I'll test it ASAP.

 Wondering - for later, not blocking - if we could have icons for the add-on entries, if it doesn't look good then never mind.

Also it would be worth noting that the list of add-ons doesn't reflect all the add-ons that are installed, only Java ones (for instance widgets/rule templates/block libraries added from the store won't appear there, I think it's implied but still).

Another option, maybe, and it's just me playing with ideas, would be to have the add-on store out of the settings section altogether, and have its own entry in the left menu perhaps between Settings and Developer Tools, like this (only a mockup):

image

So the Settings would be for settings only and the add-on store link before the add-ons could actually be removed.

florian-h05:
> This last design might satisfy everybody's needs/be a good compromise! :) I'll test it ASAP.

FYI, I have just fixed a style issues, that was caused by the use of CSS visbility instead of v-if (took me a while to find the right v-if condition).

> Another option, maybe, and it's just me playing with ideas, would be to have the add-on store out of the settings section altogether, and have its own entry in the left menu perhaps between Settings and Developer Tools, like this (only a mockup):

Good idea, I like that and so I implemented it.

lsiepel:
> > This last design might satisfy everybody's needs/be a good compromise! :) I'll test it ASAP.
> 
> FYI, I have just fixed a style issues, that was caused by the use of CSS visbility instead of v-if (took me a while to find the right v-if condition).
> 
> > Another option, maybe, and it's just me playing with ideas, would be to have the add-on store out of the settings section altogether, and have its own entry in the left menu perhaps between Settings and Developer Tools, like this (only a mockup):
> 
> Good idea, I like that and so I implemented it.

Just a suggestion, maybe ‘copy’ instead of move it. I like both place to ‘enter the store’

ghys:
I think it looks pretty neat now :heart_eyes:!

image

image

> Just a suggestion, maybe ‘copy’ instead of move it. I like both place to ‘enter the store’

Fair point, might be a simple button like "Install More Add-ons" instead of the prominent item in the attempt above, but I actually like this current design and the separation of concerns of sorts. This separation is all more important if, in the future, when some sort of RBAC is implemented and/or OAuth2 scopes are implemented, we might allow certain users access to the settings, but not to the store.

One other technicality would be to move the add-on store pages to simply `/addons`  instead of `/settings/addons` (as the developer tools are simply under `/developer`) to further segregate the 3 administration sections. Maybe the 4 tabs of the store (Bindings, Automation, UI, Other) could have sublinks on the left-side bar when the add-on store is the current section, like the other sections do. I do however think we can merge this as it is and eventually work on that later.

Thanks very much @florian-h05!

ghys (approved):

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -73,6 +73,11 @@
             </ul>
           </li>
 
+          <f7-list-item link="/settings/addons/" :title="$t('sidebar.addOnStore')" panel-close :animate="false"
+                        :class="{ currentsection: currentUrl.indexOf('/settings/addons/') >= 0 }">
+            <f7-icon slot="media" ios="f7:bag" aurora="f7:bag" md="material:shopping_bag" color="gray" />
```
ghys, line 78: A minor detail I just noticed too late (it wouldn't have been blocking): in that left side menu the top-level entries' icons (Settings, Developer Tools, About) are "filled" and the sub-levels are "hollow", so in this case the icon should have been e.g. `f7:bag_fill` instead of `f7:bag`; again just a minor design quirk but it contributes to the consistency :sunglasses:

wborn:
You could also argue that this fixes a bug and it should be backported to 4.0.x. 😉

bundles/org.openhab.ui/web/src/components/app.vue
florian-h05: Agreed that it should be changed for visual consistency, so here it is: #2053.

florian-h05:
@wborn I have actually also though if this should be backported to 4.0.x, because it greatly improved UX and some users did not find the add-on settings in the store.

@ghys WDYT? I don‘t want to backport my own PR without your approval.

ghys:
Please do (when it all works), I think you can make these decisions on your own :)

PR #2000, opened 2023-08-02
[blockly] Add Quantity support to round block

stefan-hoehn:
Round block now also takes a Qty block as input.
Output type changes according to input.

image


florian-h05 (approved):
LGTM, thanks!

PR #2007, opened 2023-08-06
Widget editor: Add note that properties are not evaluated in the editor

florian-h05:


ghys (approved):
Thanks!
I think it's a "cop out" though. I'm sure there could be a way to support expressions for the preview in the widget editor - though I haven't looked into it yet.

florian-h05:
> I think it's a "cop out" though. I'm sure there could be a way to support expressions for the preview in the widget editor

Yes, I had a look into it but I haven’t found a quick and simple solution — didn’t want to spend too much time.

ghys:
Thinking back about my comment above, I don't want you to think you yourself "copped out"... it was a general remark about the problem and the ideal solution that should ultimately be provided, not about your contribution that you know I value.

florian-h05:
Everything fine on my side, no worries!

I'm not really satisfied myself either, that was the best thing to do quickly.

PR #1986, opened 2023-07-24
Semantic tags: Fix missing translations in case of missing label

florian-h05:
Fixes #1976.
Follow-up for #1882.

PR #2029, opened 2023-08-19
Fix missing labels for semantic cards

### LINKED ISSUE #2006: Openhab 4 translation fallback in main Ui
## The problem
Hello everyone,
I successfully migrate to Openhab 4 (from 3.4 / docker Version) and everything seems to work fine :).
The only thing I recognize is that the translation in the main UI sometimes switches between german and english (seems that it fallback to english).
Going to the setting menu and open the regional settings and saving it without changing anything, changes the language back to german. After some hours or days it switches back to english.
The “problem” occurs in the tab “Geräte” or “Eigenschaften” as you can see in the screenshot
image

## Expected behavior
The tab “Geräte” or “Eigenschaften” is translated and don't fallback after some time to english.

## Steps to reproduce
I don't really know, in my case it happens after some time.

1. 

## Your environment
openhab 4.0.1 // docker version 

```yaml

```

## Browser console

```txt

```

florian-h05:
Fixes #2006.
Probably fixes the issue described in https://github.com/openhab/openhab-webui/pull/1986#issuecomment-1656746939.

Reported on the community: https://community.openhab.org/t/openhab-4-0-release-discussion/147957/86.

Sometimes, the equipment and property cards did not use the translations.
I found out, that in some cases the equipment and property pages (Main UI in general) initialized before the translations loaded, this hopefully solves it.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -136,7 +136,7 @@
       <developer-sidebar />
     </f7-panel>
 
-    <f7-view main v-show="ready" class="safe-areas" url="/" :master-detail-breakpoint="960" :animate="themeOptions.pageTransitionAnimation !== 'disabled'" />
+    <f7-view main v-if="ready" class="safe-areas" url="/" :master-detail-breakpoint="960" :animate="themeOptions.pageTransitionAnimation !== 'disabled'" />
 
   <!-- <f7-login-screen id="my-login-screen" :opened="loginScreenOpened">
     <f7-view name="login" v-if="$device.cordova">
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -136,7 +136,7 @@
       <developer-sidebar />
     </f7-panel>
 
-    <f7-view main v-show="ready" class="safe-areas" url="/" :master-detail-breakpoint="960" :animate="themeOptions.pageTransitionAnimation !== 'disabled'" />
+    <f7-view main v-if="ready" class="safe-areas" url="/" :master-detail-breakpoint="960" :animate="themeOptions.pageTransitionAnimation !== 'disabled'" />
```
florian-h05, line 139: @ghys We should only initialize the f7-view after all data is loaded (i.e. ready is true). I have checked the app.vue code and found no reference to the f7-view, so I think this is fine.

ghys: Alright, normally I never use `v-show` if I could use `v-if` instead, so there could be a reason why I did here in 2020 (https://github.com/openhab/openhab-webui/commit/6d12fe94fe2756b8966a908ace38d69b5dd85347) and just can't remember why atm. Let me sit on this one for a while...

ghys: Let's just keep this last remark in mind and see how it goes.

ghys (approved):
Thanks!

florian-h05:
If it goes well for two weeks I would say let’s backport it then!

PR #2030, opened 2023-08-19
Item details: Fix metadata error thrown when opened

florian-h05:
When the Item was a group Item and groupType was not defined, an error was thrown. This fixes it.


ghys (approved):
thanks, lgtm!

PR #2034, opened 2023-08-21
Fix add-ons missing from About Details

wborn:
It is still using /bindings to get the data resulting in a 404 Not Found. The other pages were updated in #1468.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/about.vue
+++ b/bundles/org.openhab.ui/web/src/pages/about.vue
@@ -129,7 +129,7 @@ export default {
         runtimeInfo: this.$store.state.runtimeInfo,
         locale: this.$store.getters.locale,
         systemInfo: this.systemInfo,
-        bindings: this.bindings,
+        addons: this.addons,
         clientInfo: {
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!

wborn:
A backport would also be nice. Then the addons will also be there when this info is copy/pasted into an issue.

florian-h05:
I am currently heading towards vacation and the Wi-Fi on the ICE is currently really slow, but once it gets better I will try to do the backport from my phone (Working Copy is a great Git client for iOS).

florian-h05:
I am currently heading towards vacation and the Wi-Fi on the ICE is currently really slow, but once it gets better I will try to do the backport from my phone (Working Copy is a great Git client for iOS).

florian-h05:
I’ve just cherry picked it, however GitHub says https://github.com/openhab/openhab-webui/commit/a7b436f06fdbafadcb473eabf93eb7c244df1365 „florian-h05 committed“ and not „wborn authored and florian-h05 authored“ as usual. Seems that Working Copy is doing something different here.

PR #2037, opened 2023-08-23
[blockly] Allow `item xxx` to be used for `get xxx of item`

stefan-hoehn:
The block can now take both “get item item xxx” or (new) “item xxx” directly. I have also added some hint in the help tooltip that clarifies how to work with the block when using it through variables because Blockly has to make an assumption due to the fact that it cannot detect what is provided in the variable.

image

see https://community.openhab.org/t/less-then-greater-then-comparison-return-value-seems-to-be-inverted/148910/9


florian-h05 (approved):
LGTM, thanks!

PR #2040, opened 2023-08-23
[blockly] Fix ZDT comparison between block on GraalJS

### LINKED ISSUE #2038: [Blockly] Adding a DateComparison block always adds the Nashorn JS code
## The problem

Adding a "Date comparison Between" block in Blockly will always add the Nashorn zdt comparison code to the generated script.
image

## Expected behavior

The expectation is that it should add the GraalVM zdt comparison code, if GrallVM is selected as script type.
image

## Steps to reproduce

1. Add in the DateComparison Between block to an empty Blockly script
2. Save the script
3. Check the Code page, it will show the NasHorn zdt code.
4. Running the script will then fail, as the zdt functions can't be found.

## Your environment

runtimeInfo:
  version: 4.0.2
  buildString: Release Build
locale: da-DK
systemInfo:
  configFolder: /etc/openhab
  logFolder: /var/log/openhab
  javaVersion: 17.0.7
  javaVendor: Raspbian
  osName: Linux
  osVersion: 6.1.21-v7+
  osArchitecture: arm
  availableProcessors: 4
  freeMemory: 101884896
  totalMemory: 324403200
  startLevel: 100
bindings: null
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    pixelRatio: 1
    prefersColorScheme: dark
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 8
    language: da-DK
    languages:
      - da-DK
      - da
      - en-US
      - en
      - de
    onLine: true
    platform: Linux x86_64
  screen:
    width: 1920
    height: 1080
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: dark
    filled: true
    bars: light
    homeNavbar: default
    homeBackground: default
  userAgent: Mozilla/5.0 (X11; CrOS x86_64 14541.0.0) AppleWebKit/537.36 (KHTML,
    like Gecko) Chrome/116.0.0.0 Safari/537.36
timestamp: 2023-08-23T15:27:15.278Z

```yaml

```

## Browser console

```txt

```

Dooleweerdt:
Fixes #2038

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-dateoffsets.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-dateoffsets.js
@@ -737,7 +737,7 @@ export default function (f7, isGraalJs) {
   * Code part
   */
   javascriptGenerator['oh_zdt_between'] = function (block) {
-    let zdtCompare = addDateComparisonSupportNashorn()
+    let zdtCompare = (isGraalJs) ? addDateComparisonSupportGraalVM() : addDateComparisonSupportNashorn()
     let zdtOne = javascriptGenerator.valueToCode(block, 'zdtOne', javascriptGenerator.ORDER_ATOMIC)
     let zdtTwo = javascriptGenerator.valueToCode(block, 'zdtTwo', javascriptGenerator.ORDER_ATOMIC)
     let zdtThree = javascriptGenerator.valueToCode(block, 'zdtThree', javascriptGenerator.ORDER_ATOMIC)


Dooleweerdt:
Another way of solving this could be to move the original line in 740 to 750 (into the else clause of isGraalJS):
`      let zdtCompare = addDateComparisonSupportNashorn()
`
zdtCompare is only used in this part of the function.

stefan-hoehn (approved):
Thx @Dooleweerdt to catch this and provide the fix.

@openhab/webui-maintainers This fix looks good to me as it is similar to how it is done in "oh_zdt_compare". Do you mind merging it?

florian-h05 (approved):
LGTM, thanks!

MarkusThur:
question...
How to get the fix, once it is merged, as i did run into that issue also :-)

Intermediate workarounds I found:

Variant a: set scripting language to ECMA - 262 Edition 5.1 (Nashorn)

Variant b: dont use the IS - AFTER block when between block was used => reformulate it as IS - BETWEEN - AND  block, which should be possible in most cases
AFTER equals between <timepoint> and <something which is definitely in the future, like testitem + 1 year>
BEFORE equals <something definitely far in the past like testitem - 1 year> and <timepoint>
If after or equal or before or equal are critical or the precision is critical, the one needs also proper mathematical transformations of that to be aplied to the testitem and the timepoint to test against.

florian-h05:
> How to get the fix, once it is merged, as i did run into that issue also :-)

Either install the openHAB 4.1.0 SNAPSHOTs (I do not recommend this unless you are one of the very technical openHAB users and willing to regularly take care of it) or wait for a 4.0.x patch release. I will backport this.

MarkusThur:
> > How to get the fix, once it is merged, as i did run into that issue also :-)
> 
> Either install the openHAB 4.1.0 SNAPSHOTs (I do not recommend this unless you are one of the very technical openHAB users and willing to regularly take care of it) or wait for a 4.0.x patch release. I will backport this.

Thank you, somehow I am normally the guy for snapshots, but as I didn't understand or didn't investigate deep enough how to setup a full working dev environment, and not fully in the picture on the release logic, I think I will wait for the back port, especially as i have a working solution, which is not even that unelegant. Where maybe the reverse workaround formulating between as after and before clause is even more powerful.
Just a short warning. Think i did find a bug with "copy of datetime" today also.. Have to try to reproduce it.. Wanted to create last midnight, 8 am, 4 pm and next midnight, by using "copy of block" with now and set 0 hours, 0 minutes, 0 seconds.. But the result was that the milliseconds still remained, so it was 00:00:00.xxx 08:00:00.xxx 16:00:00.xxx and next day 00:00:00.xxx where xxx is the miliseconds when now was evaluated first time.. I tried to set Millis, micros and Nanos also to zero.. No difference.. It stayed that my created times are always exactly the amount to late that now is after the full second on evaluation time.. But i have to test that further and produce a better descriprion

stefan-hoehn:
`Think i did find a bug with "copy of datetime" today also..`

Please, in this case open a new issue for that with an example on how to reproduce it. Thanks.

PR #2041, opened 2023-08-25
[blockly] Support Quantity in more math blocks

### LINKED ISSUE #2001: [blockly] Add common Math operations support for Quantity
## The problem

Associated with PR #2000. 

Common mathematical operations provided by the standard JavaScript `Math` class cannot be used with `Quantity` objects. 

## Your suggestion

It probably doesn't make sense to support all of the possible operations but the following list makes sense to me.

- `abs`
- `ceil`
- `floor`
- `max`
- `min`
- maybe `pow` (I can go either way on this since raising a Quantity to a power likely changes the unit which obviously wouldn't actually happen here by implementing this outside of the QuantityType library)
- maybe `fround` (I'm not sure I've ever seen this used in rules)
- maybe `trunc` (also I'm not sure I've seen this used in rules and the same result can be achieved using `ceil` and `floor` I think

I can't imagine users needing to do trig or log match in rules with units so all that stuff can be skipped.

stefan-hoehn:
fixes #2001

- Adds Quantity support for more math blocks.
- math_single had to be reimplemented
- math_minmax was added

image

Note that there is a special case on min/max if the inputs are not of equal type an error will be shown to user:

image

In the special case of variables Blockly does its best to detect the right code to be generated for the min/max block:
- both are variables -> then numerical input is expected
- one of inputs is a variable: then blockly uses the type of the other non-var-block to base the generation on (either number or quantity comparison).
- note that no type conversion of the inputs is done

Here are working examples that depict what is possible:

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-math.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-math.js
@@ -264,23 +264,37 @@ export default function (f7, isGraalJs) {
           case 'max': return 'Return the maximum of both inputs'
         }
       })
+      this.setOnChange(function (changeEvent) {
+        if (changeEvent.type === 'move') {
+          const typeInfo = computeMinMaxOutputType(this, false)
+          this.setOutput(true, typeInfo.leadType)
+        }
+      })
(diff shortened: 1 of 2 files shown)


florian-h05 (approved):
LGTM, thanks!

I have fixed two bugs:
- the lead type was not correctly determined due to a typo
- the lead type did not default to `Number` if both input were vars, which made the block have not set an output type

PR #2042, opened 2023-08-26
Use EEAs with maven-compiler-plugin

wborn:
Using this configuration the maven-compiler-plugin will also use the Eclipse External Annotations. This results in more similar build results when not using Eclipse like CI, on the CLI and when using other IDEs.


wborn:
We use the same compiler config in other repos, see https://github.com/openhab/openhab-addons/pull/15504.

kaikreuzer (approved):
Thanks!

Issue #2043, opened 2023-08-28
When importing sitemap code into UI sitemap // coments cause errors


ubeaut:
## The problem
I just tried pasting in my old sitemap and it worked after I took out all my comments I had in it.
The new UI version of sitemap doesn’t seem to like // as a comment.
It also didn’t like spaces after the = sign like this Frame label= “Power”

## Expected behavior
It should seamlessly import the sitemap.

## Steps to reproduce
Copy the file sitemap code and paste it into the code tab of the UI sitemap.
If the imported code has // comments in it then you will see errors.

## Your environment

```yaml

```

## Browser console

```txt

```

## Additional information
Discussion about this is here and there was a request to open an issue for this problem:
https://community.openhab.org/t/what-is-the-most-common-ui-you-are-using/149096/17

PR #2045, opened 2023-08-30
[MainUI] Sitemap parser parse and discard comments

mherwege:
Signed-off-by: Mark Herwege [mark.herwege@telenet.be](mailto:mark.herwege@telenet.be)

Fixes https://github.com/openhab/openhab-webui/issues/2043

The sitemap parser was not able to cope with comments in the sitemap text when pasted in from a sitemap file.
Comments will now be discarded (the REST API does not have fields for comments), but the sitemap will still be parsed and saved.

Similarly, blanks where accepted in a number of places when defined in text files, but not in the parser. This is now also corrected.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
+++ b/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
@@ -3,16 +3,16 @@
 
   let lexer = moo.compile({
     WS:               /[ \t]+/,
-    comment:          /\/\/.*?$/,
     sitemap:          'sitemap ',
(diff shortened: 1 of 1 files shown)


ghys (approved):
@mherwege you're becoming quite proficient at Nearley ;)

I didn't consider the multiline comments at first because I figured it was not worth the hassle (sometimes a grammar looks correct and looks like it parses the input, but performance takes a huge hit if there are ambiguities, which are very easy to introduce but can be hard to identify and solve).

Glad to see them implemented, thanks!

PR #2054, opened 2023-09-04
Reverts v-if/v-show change from #2029.

digitaldan:
Regression from #2029.

@florian-h05 i noticed when upgrading my system this AM that a few of my UI pages stopped loading properly.  Specifically there seemed to be a condition where when the page would draw its widgets before other parts of the page were loaded ( or something similar, some race conditional).  This would result in widgets loading with the wrong state, colors, icons, etc that are dependent on item state.  I tracked it down to this one very small change,  reverting back to a `v-show` from `v-if` for the main view ready state fixes this.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -141,7 +141,7 @@
       <developer-sidebar />
     </f7-panel>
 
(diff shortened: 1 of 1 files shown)


florian-h05:
Ah sh*t. 
So I have to find another way to fix #2006. 

Since you tracked it down to this little change, let’s merge this PR then.

florian-h05 (approved):

florian-h05:
@ghys FYI. Seems like there was a reason to use v-show …
Weird that I did not notice those problems, but TBH I haven’t tested this on my production, just on my limited dev setup.

PR #2055, opened 2023-09-04
Adds an UI command Item

digitaldan:
This allows the local running instance of the UI to listen to an item for commands.  Specifically it will understand the following commands. 

- navigate
- popup
- popover
- sheet
- close (closes all popups/sheets/popover)
- back
- reload
- notification

Commands with arguments use a colon delimiter, mirroring widget actions in the UI, for example
`navigate:/pages/page`
`popup:widget:customWidget`
`notification:Text:Title:Sub Title:Title Right:5000`

~I have this as a WIP~ as i play with this some more, but its very handy for tablets i have mounted around the house, i can easily pop up a camera, or a SIP call when the doorbell rings, trigger a reload when i have made UI changes or after a system  update/restart,  etc.....  

users can select the item in the about menu,  since we have other local options set there its seemed to be the best place.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -272,12 +272,13 @@ import DeveloperSidebar from './developer/developer-sidebar.vue'
 
 import auth from './auth-mixin.js'
 import i18n from './i18n-mixin.js'
+import sseEvents from './sse-events-mixin.js'
 
 import dayjs from 'dayjs'
 import dayjsLocales from 'dayjs/locale.json'
 
 export default {
-  mixins: [auth, i18n],
+  mixins: [auth, i18n, sseEvents],
   components: {
     PanelRight,
     DeveloperSidebar
@@ -622,52 +623,6 @@ export default {
         ev.preventDefault()
       }
     },
-    startEventSource () {
-      this.eventSource = this.$oh.sse.connect('/rest/events?topics=openhab/webaudio/playurl', null, (event) => {
-        const topicParts = event.topic.split('/')
-        switch (topicParts[2]) {
-          case 'playurl':
-            this.playAudioUrl(JSON.parse(event.payload))
-            break
-        }
-      })
-    },
-    stopEventSource () {
-      this.$oh.sse.close(this.eventSource)
-      this.eventSource = null
-    },
-    playAudioUrl (audioUrl) {
-      try {
-        if (!this.audioContext) {
-          window.AudioContext = window.AudioContext || window.webkitAudioContext
-          if (typeof (window.AudioContext) !== 'undefined') {
-            this.audioContext = new AudioContext()
-            unlockAudioContext(this.audioContext)
-          }
-        }
-        console.log('Playing audio URL: ' + audioUrl)
-        this.$oh.api.getPlain(audioUrl, '', '*/*', 'arraybuffer').then((data) => {
-          this.audioContext.decodeAudioData(data, (buffer) => {
-            let source = this.audioContext.createBufferSource()
-            source.buffer = buffer
-            source.connect(this.audioContext.destination)
-            source.start(0)
-          })
-        })
-      } catch (e) {
-        console.warn('Error while playing audio URL: ' + e.toString())
-      }
-      // Safari requires a touch event after the stream has started, hence this workaround
-      // Credit: https://www.mattmontag.com/web/unlock-web-audio-in-safari-for-ios-and-macos
-      function unlockAudioContext (audioContext) {
-        if (audioContext.state !== 'suspended') return
-        const b = document.body
-        const events = ['touchstart', 'touchend', 'mousedown', 'keydown']
-        events.forEach(e => b.addEventListener(e, unlock, false))
-        function unlock () { audioContext.resume().then(clean) }
-        function clean () { events.forEach(e => b.removeEventListener(e, unlock)) }
-      }
-    },
     /**
      * Creates and opens a toast message that indicates a failure, e.g. of SSE connection
      * @param {string} message message to show
@@ -835,10 +790,8 @@ export default {
       if (window) {
         window.addEventListener('keydown', this.keyDown)
       }
-
-      if (localStorage.getItem('openhab.ui:webaudio.enable') === 'enabled') {
-        this.startEventSource()
-      }
+      // sse-event mixin
+      this.startEventSource()
     })
   }
 }
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -703,6 +703,7 @@ export default {
       if (window) {
         window.addEventListener('keydown', this.keyDown)
       }
+
       this.startEventSource()
     })
   }
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -272,7 +272,7 @@ import DeveloperSidebar from './developer/developer-sidebar.vue'
 
 import auth from './auth-mixin.js'
 import i18n from './i18n-mixin.js'
-import sseEvents from './sse-events-mixin.js'
+import sseEvents from './sse-events-mixin'
 
 import dayjs from 'dayjs'
 import dayjsLocales from 'dayjs/locale.json'
@@ -298,8 +298,6 @@ export default {
     return {
       init: false,
       ready: false,
-      eventSource: null,
-      audioContext: null,
 
       // Framework7 Parameters
       f7params: {
@@ -790,7 +788,6 @@ export default {
       if (window) {
         window.addEventListener('keydown', this.keyDown)
       }
-      // sse-event mixin
       this.startEventSource()
     })
   }
--- a/bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
@@ -79,6 +79,8 @@ export default {
                 const payload = {
                   text: segments[0],
                   closeOnClick: true,
+                  closeButton: true,
+                  swipeToClose: true,
                   closeTimeout: 5000
                 }
                 if (segments.length > 1) {
--- a/bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
@@ -78,7 +78,6 @@ export default {
               case 'notification':
                 const payload = {
                   text: segments[0],
-                  closeOnClick: true,
                   closeButton: true,
                   swipeToClose: true,
                   closeTimeout: 5000
--- a/bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
@@ -1,0 +1,160 @@
+import OhPopup from './widgets/modals/oh-popup.vue'
+import OhSheet from './widgets/modals/oh-sheet.vue'
+import OhPopover from './widgets/modals/oh-popover.vue'
+
+export default {
+  data () {
+    return {
+      eventSource: null,
+      audioContext: null
+    }
+  },
+  methods: {
+    startEventSource () {
+      const topicAudio = 'openhab/webaudio/playurl'
+      const commandItem = localStorage.getItem('openhab.ui:commandItem')
+      const topicCommand = `openhab/items/${commandItem || ''}/command`
+      let topics = null
+      if (localStorage.getItem('openhab.ui:webaudio.enable') === 'enabled') {
+        topics = topicAudio
+      }
+      if (commandItem) {
+        topics = topics ? `${topics},${topicCommand}` : topicCommand
+      }
+      if (!topics) return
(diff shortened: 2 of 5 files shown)


ghys:
Thanks @digitaldan, this seems like a useful feature.

My main worry is the new SSE connection dedicated to it.

As you might know (https://news.ycombinator.com/item?id=30313844) there's a 4-connections limit _per domain_ (not per tab) on these so as the Main UI uses two for normal operation (one for state updates and one for relevant event updates dependent on the current page), then there's the webaudio one (https://github.com/openhab/openhab-webui/pull/1422) and now this would potentially introduce another one. So that's up to 4 connections just for one tab of the Main UI, and in that case, if you open another tab (even if it's Basic UI or HABPanel) then it'll start to fail. Won't happen if you dedicate a tablet to displaying the UI but still.

I would suggest trying to try to mitigate that by refactoring these two features (webaudio and UI commands) which rely on a permanent connection so they can share a single one (you can check which features are enabled in localCache and tweak the `topics` parameter according to it with comma-separated topics in case both are enabled; so it could be e.g.
`topics=openhab/webaudio/playurl,openhab/items/${itemName}/command`.
Of course all this would become a moot point when/if we eventually switch to a websocket.

This could be a good opportunity to all this logic to other file(s) (either imported or a mixin) dedicated to these permanent listeners as `app.vue` is becoming quite messy IMO.

ghys:
Update, I re-checked and apparently it's 6 concurrent EventSource connections, not 4 per domain like I always thought:

https://bugs.chromium.org/p/chromium/issues/detail?id=275955
https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#listening_for_custom_events

> Warning: When not used over HTTP/2, SSE suffers from a limitation to the maximum number of open connections, which can be especially painful when opening multiple tabs, as the limit is per browser and is set to a very low number (6).

digitaldan:
Yeah, the additional connection seems wasteful, i did not think about combining it with the audio sink one, thats a good idea until we can move to a websocket model.  Agreed also to look at moving code outside of app.vue.

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
```diff
@@ -0,0 +1,127 @@
...
+                this.$f7.views.main.router.navigate(modalRoute, modalProps)
+                break
+              case 'back':
+                this.$f7.views.main.router.back()
+                break
+              case 'reload':
+                window.location.reload()
+                break
+              case 'notification':
+                const toast = this.$f7.toast.create({
```
florian-h05, line 75: I love your idea of having a command Item to control the UI!!

Just one very minor comment after having a look at the idea (not really at the code, I will leave the review to Yannick because he started it): IMO we should not use toasts for user-sent notifications here, I would consider them as only-used by the UI itself. I would rather use Framework7 notifications for this, see https://v5.framework7.io/docs/notification.

digitaldan: Thanks @florian-h05 !  I did not think about f7 notifications, i agree, it would be a much better experience.

digitaldan:
I still need to test that audio continues to work, but this is ready to be looked at for code review.  I also will need to add documentation about these new features.   I updated my first post with all supported commands and more examples.

digitaldan:
Confirmed that the audio sink works as well

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -272,12 +272,13 @@ import DeveloperSidebar from './developer/developer-sidebar.vue'
 
 import auth from './auth-mixin.js'
 import i18n from './i18n-mixin.js'
+import sseEvents from './sse-events-mixin.js'
```
florian-h05, line 275: Let's drop the .js extension here, as for most JS imports in MainUI:

```suggestion
import sseEvents from './sse-events-mixin'
```

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -835,10 +790,8 @@ export default {
       if (window) {
         window.addEventListener('keydown', this.keyDown)
       }
-
-      if (localStorage.getItem('openhab.ui:webaudio.enable') === 'enabled') {
-        this.startEventSource()
-      }
+      // sse-event mixin
```
florian-h05, line 793: Can you please remove the comment from this line, but keep the line? 
IMO makes the code cleaner, and if you hover over the method call, your IDE can anyway tell you the method is from the mixin (at least JetBrains IDEs do that).

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
```diff
@@ -0,0 +1,154 @@
...
+                  payload.closeTimeout = parseInt(segments[4])
+                }
+                this.$f7.notification.create(payload).open()
+                break
+            }
+            break
+        }
+      })
+    },
+    stopEventSource () {
```
florian-h05, line 103: I have just noticed that this method is neither called inside the mixin nor in app.vue.

Should we keep it for completeness or remove it? I guess for the webpack bundle that does not matter due to webpack's optimization.

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
```diff
@@ -0,0 +1,154 @@
...
+              case 'back':
+                this.$f7.views.main.router.back()
+                break
+              case 'reload':
+                window.location.reload()
+                break
+              case 'notification':
+                const payload = {
+                  text: segments[0],
+                  closeOnClick: true,
```
florian-h05, line 81: I would suggest to have a close button (`closeButton`) and swipe to close (`swipeToClose`), because both on iOS and Android a click on a notification does open it, but not close it.

This brings me to the next idea: I can imagine having the user define a action to perform when clicking on the notification. I would support the same actions as for the command Item.
This is BTW not directly supported, but if such a on-click action is defined, we could enable `closeOnClick` and add an event listener for closed, like here: https://framework7.io/docs/notification#examples.

I am not sure how to fit such an action into the `notification:Text:Title:Subtitle:Title_Right_Text:Close_Timeout` scheme, without making that totally unreadable. Probably something like `notification:Text...Closed_Timeout_navigate:/pages/page` (`_` as a delimiter?). Another option would be to have this parameterized, e.g. `notification:text=Some Text,title=Some Title,action=navigate:/pages/page`.

bundles/org.openhab.ui/web/src/components/theme-switcher.vue
```diff
@@ -127,6 +136,10 @@ export default {
     setWebAudio (value) {
       localStorage.setItem('openhab.ui:webaudio.enable', (value) ? 'enabled' : 'default')
       location.reload()
+    },
+    setCommandItem (value) {
+      localStorage.setItem('openhab.ui:commandItem', value)
+      location.reload()
```
florian-h05, line 142: This does not work, I guess because of the Item picker's popup.
Luckily I already have a solution:

```suggestion
      setTimeout(() => { location.reload() }, 50) // Delay reload, otherwise it doesn't work
```

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
```diff
@@ -0,0 +1,154 @@
...
+            const combined = segments.join(':')
+            switch (command) {
+              case 'navigate':
+                this.$f7.views.main.router.navigate(combined)
+                break
+              case 'popup':
+              case 'popover':
+              case 'sheet':
+                if (combined.indexOf('page:') !== 0 && combined.indexOf('widget:') !== 0 && combined.indexOf('oh-') !== 0) {
+                  console.error('Action target is not of the format page:uid or widget:uid')
```
florian-h05, line 48: Can you please update that error message, because `oh-` components are obviously also supported?

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
```diff
@@ -0,0 +1,154 @@
+import OhPopup from './widgets/modals/oh-popup.vue'
+import OhSheet from './widgets/modals/oh-sheet.vue'
+import OhPopover from './widgets/modals/oh-popover.vue'
+
+export default {
+  data () {
+    return {
+      eventSource: null,
+      audioContext: null
```
florian-h05, line 9: As `eventSource` and `audioContetx` are inside the mixin, they can be removed from `app.vue`'s `data`.

florian-h05 (commented):
LGTM, and works fine! 
Really good idea, probably you got inspired by HABPanel's "Switch dashboard with item value"?

Please have a look at my comments.

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
digitaldan: I also noticed it was not used, but though maybe it was there also for completeness.  I don't have an opinion one way or another. 🤷

digitaldan:
> LGTM, and works fine! Really good idea, probably you got inspired by HABPanel's "Switch dashboard with item value"?

Actually, i use https://github.com/vbier/habpanelviewer (well, a slightly modified version), but for our Main UI only, on a bunch of tablets around the house and use it's built in command item functionality extensively for reloading and navigating to pages on events, but this would replace most of that with the benefit of allowing me to use iPads if i need to ( my android tablets are very sluggish with our main UI).

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
florian-h05: Let‘s keep it there.

bundles/org.openhab.ui/web/src/components/app.vue
```diff
@@ -750,10 +703,7 @@ export default {
       if (window) {
         window.addEventListener('keydown', this.keyDown)
       }
-
-      if (localStorage.getItem('openhab.ui:webaudio.enable') === 'enabled') {
-        this.startEventSource()
-      }
+      this.startEventSource()
```
florian-h05, line 706: Let‘s add an empty line for readability.

```suggestion

      this.startEventSource()
```

bundles/org.openhab.ui/web/src/components/sse-events-mixin.js
florian-h05: `closeOnClick` can be removed IMO.

```suggestion

```

florian-h05 (commented):
LGTM, just two minor comments.
I will test this ASAP and then merge it.

digitaldan:
Thank @florian-h05 for the review !

digitaldan:
Also, i probably need to write some documentation up on this,  I guess adding this to https://github.com/openhab/openhab-docs/blob/main/ui/index.md would be the best path forward?

florian-h05:
Given that the title is "User Interface Design Overview", I would not expect the command Item there, however if the title is changed to "User Interface Overview", it could fit in there. If the docs are long, I would say introduce a new page.

florian-h05 (approved):
LGTM, thanks! I will merge this when the milestone build finished.

digitaldan:
I think https://www.openhab.org/docs/settings/aboutpage.html would probably be a better location

florian-h05:
Yeah, there already is a section about Web Audio sink support.
However IMHO the command Item should be promoted somewhere else - would be sad if no one notices it.

bundles/org.openhab.ui/web/src/components/theme-switcher.vue
```diff
@@ -74,14 +74,23 @@
             <span v-t="'about.miscellaneous.webaudio.enable'" />
             <f7-toggle :checked="webAudio === 'enabled'" @toggle:change="setWebAudio" />
           </f7-list-item>
+          <f7-list-item>
+            <span v-t="'about.miscellaneous.commandItem.title'" />
+            <item-picker title="Item" :multiple="false" :value="commandItem" @input="setCommandItem" />
+          </f7-list-item>
```
ghys, line 80: Also it's not looking too good on narrow screens:

image

I think we could just have the item-picker component directly added to the list with an appropriate label like "Listen to UI commands from", instead of the item-picker inside a list item (the item-picker is a list item itself).

OmegaQuest:
I noticed that the Audio Sink dosent work after a "Navigate to Page" Command UI is done after a short period of time, unless you manually click the new page.  Its like the Audio Sink times out after a while when the Command UI is done (unless you click the page somewhere).  This time-out dosent happen with the normal Navigate to Page or after you click somewhere on the screen "After" the Command UI navigated to a new page.

Server Side:
OpenHab: 4.2.0 M1
Ubuntu 22.04.4 LTS

Client/Web Side:
Windows 10
Chrome Browser

florian-h05:
That's a limitation of modern browsers:

> Please note that due to limitations in Safari (and possibly Chrome as well), a user interaction is required after the first audio stream has been sent to actually play it. This means, that after opening Main UI, the first audio that should be played on the web audio sink is only played after the user interacts with Main UI in any way (i.e. touching the screen is enough). For subsequent audio playback the above is not required anymore and the audio is played immediately.

(https://next.openhab.org/docs/mainui/about.html#web-audio-sink)

OmegaQuest:
But it also does it with Edge

florian-h05:
I guess it is the same in all modern browsers.
And Edge is Chromium based, if it doesn’t work in Chrome, it likely also won’t work in Edge.

PR #2061, opened 2023-09-11
Profile page: Only show latest 10 sessions

florian-h05:
The openHAB demo's profile page takes ages to render because of the huge sessions list. This should fix that problem.


florian-h05:
@ghys I remember a conversation where you said that it would be great for the openHAB demo's profile page if the number of shown sessions is limited, because currently it takes ages to render. Here is your PR!

ghys (approved):
Nice :star_struck: - thanks a lot!

PR #2066, opened 2023-09-14
Ensure semantic model pages are built after tags are loaded

### LINKED ISSUE #2006: Openhab 4 translation fallback in main Ui

florian-h05:
Fixes #2006.

With the attempt in #2029 failing due to #2054, this finally fixes an issue where the semantic model pages (Equipment, Properties) were missing translations.
This was due to a race condition, where the semantic model pages were built before the tags were loaded.

This PR fixes this by watching for semantic model updates (i.e. subscripting to the underlying Vuex mutation) and loading/building the model pages after such an update.

Futher improvements:
- Initial loading of semantic tags is put in parallel with other API requests performed on app init.
- Vuex semantics: A mutation is used for setting the semantic tags from the action.
- Debug logging is added to both the semantic tag and the semantic model page loading.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/home.vue
+++ b/bundles/org.openhab.ui/web/src/pages/home.vue
@@ -175,6 +175,12 @@ export default {
       this.$store.dispatch('stopTrackingStates')
     },
     onPageInit () {
+      this.$store.subscribe((mutation, state) => {
+        if (mutation.type === 'setSemantics') {
+          this.loadModel()
+        }
+      })
+
       if (window.OHApp) {
         if (window.OHApp.pinToHome) this.showPinToHome = true
         if (window.OHApp.exitToApp) this.showExitToApp = true
--- a/bundles/org.openhab.ui/web/src/pages/home.vue
+++ b/bundles/org.openhab.ui/web/src/pages/home.vue
@@ -156,7 +156,6 @@ export default {
   watch: {
     ready (val, oldVal) {
       if (val && !oldVal) {
-        this.loadModel()
         this.$store.dispatch('startTrackingStates')
       }
     }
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.ui/web/src/pages/home.vue
```diff
@@ -175,6 +174,12 @@ export default {
       this.$store.dispatch('stopTrackingStates')
     },
     onPageInit () {
+      this.$store.subscribe((mutation, state) => {
```
florian-h05, line 177: @ghys This could also be placed in `homecards-mixin.js` but it is only used by `home.vue` and `home-edit.vue` and I don't think we need this subscription for the edit page. I don`t think a user would want to update the semantic model tags and have the edit page open at the same time.

Do you agree?

ghys: Sure - makes sense.

ghys (approved):
Thanks for the investigation and the fix, nice to have some more debug logs too!

florian-h05:
One thing that came to my mind just a few seconds ago and that we have to keep an eye on, is that in case the home page is destroyed and re-created, the model won‘t load. I made it to only load when the tags are saved, because otherwise the model would be loaded two times on app init.

florian-h05:
I have thought about this again, and this cannot happen, since the model is loaded `onPageAfterIn`, so I will backport this then.

PR #2053, opened 2023-09-03
Fixes/Improvements for sidebar highlighting, settings page & add-on store

florian-h05:
Follow-up for #1997.

- Addresses https://github.com/openhab/openhab-webui/pull/1997/files#r1313446378: Use filled icon for the add-on store.
- Fixes #2049: Fix sidebar highlighting for add-on settings and add-on store pages.
- Fixes #2028: Fix tab is resetted to bindings when reloading the add-on store (by using seperate routes for all tabs, therefore addresses https://github.com/openhab/openhab-webui/pull/1997#issuecomment-1703221977).
- Closes #2048: Expand system settings by default on large screens where three columns are used.
- Localize the "Show All" buttons.

PR #2073, opened 2023-09-16
Change setup wizard icon to choose add-ons to install

MyRaceData:
remove shopping cart icon from add-ons wizard
I am user Andrew_Rowe on the openHAB community forum


florian-h05:
Can you please explain why you want to remove that icon?

MyRaceData:
> Can you please explain why you want to remove that icon?
Yes, of course sorry....
I think the shopping cart icon, next to the link that says 'Select Add-ons to Install' may give the appearance that the Add-ons might have to be purchased, especially to a new user that knows nothing about openHAB. I think a different icon or no icon would be better.
Here is a link to a community thread where I asked about it

https://community.openhab.org/t/first-time-user-wizard-for-oh4/149670/4?u=andrew_rowe

stefan-hoehn:
my 2 cents: I wouldn't want to drop the icon completely as it has some nice appearance. I would ask you to come up with a better icon instead.

hmerk:
Why not use the basket like in the side menu ?

MyRaceData:
image
integration icon?

I like the basket in the side menu better then the shopping cart

hmerk:
> I like the basket in the side menu better then the shopping cart and that would provide consistency with the side menu.

So why not update your PR and remove icon deletion in favour of replacing it with the shopping basket ?

MyRaceData:
well, if we are going to get into it, why are there two links that appear to go to the same thing?
image

florian-h05:
I agree that the shopping cart is not the best icon, it also reminds me of paying money for something.

When searching for a better icon, keep in mind that we both need a Material icon (https://jossef.github.io/material-design-icons-iconfont/) and a Framework 7 icon (https://framework7.io/icons/).

> well, if we are going to get into it, why are there two links that appear to go to the same thing?

They do not go to the same thing: The first link opens up https://www.openhab.org/addons/, where you can have a closer look at the add-ons, the second link opens a simple selection which add-ons to install.

Oggerschummer:
Personally I support the idea of removing the shopping cart. In any real-world or internet scenario this is commonly used to shop goods that are to be payed in some sort of a checkout process - which is not the case here.
An icon representing a simple "connection" would be fine, think connecting things is what new users are about first.
An icon with a plug and a socket or something like ["adding a link"](https://icons8.de/icon/UbGRP5y1uQu8/link-hinzuf%C3%BCgen) could fit. In general we are using the plus sign in MainUI to add/create something. So why not use it here as well? A "plus" is always obviously adding something.

And as we are talking about making things more understandable to Newbies:
Why no use a bit more simple speech - the two sentence with the "add-ons" assume that one knows what an "add-on" means (For us it is fine, for them a part of the undiscovered country). 

Yes, i know, there are ones for scripting, purely technical stuff etc.
But the first thing and main reason someone installs OH is that (s)he wants to connect stuff, to switch something - simple things in the beginning. Then why not name it like "Get started - choose the things you want to connect to from the supported add-ons".  Yes there is more, but for the beginning keep in mind what your first steps with OH were. I guess it was switching something or getting data from something else. And remember the smile on your face when it worked.

This is what we should try to achieve: Keep it simple and support initial quick wins for new users.
Nothing more frustrating that having to deal with terms a glossary before being able to do something.
Nevertheless link to the basic documentation of OH would be fine here as well - is could give the user a chance to directly look up what we are talking about when it comes to "Add-ons" - so (s)he would know what makes sense to choose.

Does that make sense to you?

ghys:
Just to try to clarify, when I made that screen and used that icon, the intent was that you would be "shopping" for add-ons that you would install as part of the setup process. I used `shopping-cart` not giving much thought about it, but following https://github.com/openhab/openhab-webui/pull/2053#issuecomment-1704392367 it might be better to use `shopping-bag` instead (and the equivalent MD icon), if only for consistency.

We can also wonder about this step, it's certainly convenient for users who are used to spinning up openHAB instances, but for real first-time users, it's an inferior experience than the add-on store.

ghys:
> In general we are using the plus sign in MainUI to add/create something. So why not use it here as well? A "plus" is always obviously adding something.

This is also a good point, so I would suggest this icon:

image

Denotes more the adding and less an unevitable payment as the supermarket cart suggests.

MyRaceData:
I added the bag_fill_badge_plus icon instead of the shopping cart

ghys (approved):
Let's make this change, thanks for the discussion.

Issue #866, opened 2021-01-30
Configure default page (Overview/Locations/Devices/Properties)


GumbyMan82:
## The problem
The location tab is one of the cooles features in OH. So many of the users would like to open it directly instead of manually have to switch to the location tab.
It is also discussed in the community:
https://community.openhab.org/t/oh3-use-oh-location-card-outside-of-overview-page/113021

## Make the defaulkt page configurable
Provide an additional configuration for selecting the default page (overview, locations, devices, properties):
Since it isn't possible to get a link to open the locations tab, i'd suggest to make the default tab configurable.
![image](https://user-images.githubusercontent.com/40233411/106359767-f5d7fa00-6314-11eb-9e87-0948e03e5cdf.png)

## Your environment
openHAB 3.0

```yaml

```

krocans:
It would be also nice to have possibility to change labels of "built in" pages which are displayed on bottom selector ant page top.
![image](https://user-images.githubusercontent.com/42302779/106384302-c8ea1c80-63d2-11eb-90b2-56d869cd878e.png)

bf8392:
+1 for configuring which sites are displayed....in most cases I don't need the overview, because the other tabs are so good! It would be really cool if one could simply turn it off :-)

SHU-red:
Hello,
i just can confirm!

I would want to have the possililty to set a tabbed-page (configured by my own) as default page

--> Opening the Android app and instantly seeing my favourite page (which is not overview)

Nic0205:
Are there any efforts here with this issue?

DerEsdas:
What is with this now? I will set my own page as default start page....

snemias:
+1

romanstyler:
+1

Daikon2022:
+1

PR #2074, opened 2023-09-17
Add routes for home page tabs

florian-h05:
Adds routes for `/overview`, `/locations`, `/equipment` and `/properties` so a home page reload keeps the tab and routing to individual tabs is possible.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/routes.js
+++ b/bundles/org.openhab.ui/web/src/js/routes.js
@@ -92,7 +92,45 @@ export default [
     // keepAlive: true,
     options: {
       transition: 'f7-dive'
-    }
+    },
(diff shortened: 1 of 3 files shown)


ghys (approved):
That's great! The much-requested feature I was referring to was https://github.com/openhab/openhab-webui/issues/866 - have a way to show the locations tab for those who don't configure an overview page.
While it doesn't address the issue completely it's a good step forward as they could just open `/locations` if they want.
If we choose not to have a setting to configure the default initial tab, changing the tab would set a `openhab.ui:home.lastHomePageTab` local storage setting so the last accessed tab would be restored if they hit the root URL - this PR has made it easier.
In any case it's a nice improvement as it is.

PR #2079, opened 2023-09-20
Fix missing jar add-ons on settings page

### LINKED ISSUE #2076: Missing Add-on Settings menu item for third party bundles
I've just updated to `4.1.0-SNAPSHOT - Build #3630` which includes #2053.

When I installed the official JRuby automation addon, the `JRuby settings` item would show up as an entry in the `Add-on Settings`.

However, when I am using a custom built jrubyscripting addon jar that I dropped into the `addons/` folder, the "Add-on Settings" would not show up the JRuby settings item.

I can still go to Add-on store -> Automation -> find the custom JRuby Scripting addon then click on the little blue cog wheel icon, in order to get to the settings page.

The same thing happened with jsscripting too.
This is when it's installed from the official distribution:
image

And this is when it's installed as a jar inside `addons/` folder:
image

Note that it's still not there even if I expanded / clicked "Show all".

jimtng:
Fix #2076

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/menu/settings-menu.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/menu/settings-menu.vue
@@ -215,7 +215,7 @@ export default {
       if (!this.apiEndpoints) return
 
       const servicesPromise = (this.$store.getters.apiEndpoint('services')) ? this.$oh.api.get('/rest/services') : Promise.resolve([])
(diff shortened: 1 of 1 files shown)


jimtng:
@florian-h05 this seems to fix it for me

florian-h05:
I will have a look when I am back home.

florian-h05 (approved):
LGTM, thanks!

PR #2084, opened 2023-09-23
Remove PNG icons from the classic icon set

### LINKED ISSUE #1768: Icon "player" from classic icon set broken
## The problem

The text "Power on" from the SVG version of the icon "player" is broken in the PNG version:
* https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/player.png
* https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/player.svg

## Expected behavior

Both should look the same.

## Steps to reproduce

1. Open the links above.

### LINKED ISSUE #2071: Proposal to remove PNG icons from the classic icon set
In order to simplify the maintenance of the classic icon set and in order to improve it more easily, I suggest to only provide SVG icons and so to remove all PNG files.
We could then also remove all the stuff in charge of converting SVG into PNG.

To be really clear, I do not propose to remove the support of the PNG format because this format could be used by certain users for their custom icons.
My proposal is only to simplify the classic icon set to have only one file per icon and not 2 as it is now.

This will fix #1768 for example.
And it will allow in a second step to simplify the update of our icons to make their color automatically adjustable to the selected theme ( #1769 ) and their color adjustable by the user (iconcolor field in sitemap syntax), something not possible if we have to keep the PNG version of the icon.

But we need first to know if certain UIs are still relying exclusively on the PNG format. Basic UI already uses by default the SVG format and do not need PNG icons. I am almost certain that Main UI and the Android app can also work without the PNG icons, I have only a doubt about the iOS app.
@openhab/webui-maintainers 
@openhab/ios-maintainers 
@openhab/android-maintainers 
Please let me know.

lolodomo:
The docker environment is (unfortunately) kept. It can be used to compress SVG files. Conversion from SVG into PNG format has been commented in the script .convert.sh.

Closes #2071
Closes #1768


lolodomo:
Before merging this change, I would prefer getting first answers from @stefan-hoehn regarding potential impact on our openHAB documentation.

bundles/org.openhab.ui.iconset.classic/README.sh
```diff
@@ -42,7 +42,7 @@ cat <<EOF >> "$READMEMD"
 {% for category in placesCategoryNames %}
   {% assign iconSrc = base | append: "/img/icon_no_category.png" %}
   {% if allIconsWithoutExtension contains category %}
-    {% assign iconSrc = "icons/" | append: category | append: ".png" %}
+    {% assign iconSrc = "icons/" | append: category | append: ".svg" %}
   {% endif %}
   <figure>
     <img src="{{iconSrc}}" alt="{{category}}" title="{{category}}">
```
mueller-ma, line 48: Maybe a size should be added here. Before the PNG icons had a fixed size, but SVG doesn't and it should be avoided to show them too big in the docs.

lolodomo: Yes, you're right.

stefan-hoehn:
@Confectrician I have no clue if it has an impact. Do you?

bundles/org.openhab.ui.iconset.classic/README.sh
lolodomo: What size should I set?

mueller-ma: The size of the png icons should be a good choice as the page doesn't change then. Currently the pngs are 64x64 pixel.

lolodomo: https://www.openhab.org/docs/configuration/iconsets/classic/

In case, this stuff is used to build this page, yes, the icons are currently displayed with 64x64 as size.

lolodomo:
@openhab/maintainers : is there any interest for this PR ? If not, I simply close it.

florian-h05:
From my POV nothing speaks against removing the PNG icons. AFAIK SVG is supported by all modern browsers.

kaikreuzer (approved):
Sure, let's work then.
Thanks for your work, @lolodomo!

PR #2090, opened 2023-09-24
Fix decimal formatting for oh-stepper

### LINKED ISSUE #2089: oh-stepper-item ignoring 2+ decimal places
The stepper widget only shows one decimal place at most.
You can define scale: 0.01 and it's honored the value changes but you have to click 10 times "+" before the shown value is changed.

The value should be shown with as many decimal places as the display pattern defined in item metadata.

JustinGeorgi:
Fixes #2089.
Possibly regression from #1875.

Stepper formatValue was calculating displayed number of decimals from the value of the step property, not it's length.

Signed-off-by: Justin Georgi <justin.georgi@gmail.com>

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
@@ -21,14 +21,14 @@ export default {
   },
   computed: {
     value () {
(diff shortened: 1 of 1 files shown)


florian-h05:
@JustinGeorgi FYI I have backported this to 4.0.x, so it will be published with the next patch release.

Issue #2082, opened 2023-09-21
Sitemap Label Colour not working


Dash:
## Which UI are you reporting an issue for?

- [X] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

Conditional Label colours no longer work on Sitemaps, though simply specifying a colour does work.  I'm not sure at which point this regressed.

This setup
![image](https://github.com/openhab/openhab-webui/assets/129838/62e690f4-cba0-4987-99c5-964489aa6abf)

Produces:
![image](https://github.com/openhab/openhab-webui/assets/129838/25a95ea5-921c-4437-a100-176fb7a92c0b)

The UI generates this code:

```javascript
sitemap page_c6c037636f label="New Sitemap" {
    Text item=KitchenLights labelcolor=[KitchenLights == ON =" red"] label="Lights are: [%s]"
}
```

I have checked in Firefox, Chrome, and through the Android App.

## Expected behavior

I'm expecting the above example to show the text "Kitchen Lights" in red on sitemap.

## Steps to reproduce

1. Build a new site map
2. Add a label to a switch
3. Add a conditional colour based on the switch status

## Your environment

```yaml
runtimeInfo:
  version: 4.1.0.M1
  buildString: Milestone Build
locale: en-GB
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 17.0.8
  javaVendor: N/A
  osName: Linux
  osVersion: 5.14.21-150400.24.81-default
  osArchitecture: amd64
  availableProcessors: 2
  freeMemory: 332692888
  totalMemory: 778043392
  uptime: 353815
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-astro
  - binding-exec
  - binding-gpstracker
  - binding-http
  - binding-icalendar
  - binding-mqtt
  - binding-network
  - binding-snmp
  - binding-systeminfo
  - binding-xmppclient
  - persistence-rrd4j
  - transformation-jsonpath
  - transformation-map
  - transformation-regex
  - ui-basic
  - ui-habpanel
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: true
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
[...]

PR #2091, opened 2023-09-25
Sitemap editor: Trim spaces from color value

mherwege:
When generating the sitemap DSL from a UI configured sitemap with color conditions, if the color in the condition has a leading space, it will be included in the color string.
While this does not have an impact on the sitemap functionality (it works), it is an issue if one would copy this sitemap to a sitemap file.
This PR strips the blanks from the color.

See https://github.com/openhab/openhab-webui/issues/2082 and the example provided there:
```
sitemap page_c6c037636f label="New Sitemap" {
    Text item=KitchenLights labelcolor=[KitchenLights == ON =" red"] label="Lights are: [%s]"
}
```


florian-h05 (approved):
LGTM thanks!

PR #2094, opened 2023-09-26
Fix `oh-` components cannot be configured in action modals

### LINKED ISSUE #927: Opening popover/popup/sheet doesn't work from oh-list-item (listButton)
## The problem

Triggering a popover / popup / sheet from `oh-list-item` with `listButton: true` doesn't work

openingSheet

## Expected behavior
I would expect the behavior similar to the one from the `oh-button` seen in the above *.gif

## Steps to reproduce
Create a widget with the following content and try to open the popover.
```
- component: f7-list
  slots:
    default:
        - component: oh-list-item
          config:
            listButton: true
            popoverOpen: .deleteItems
            
- component: f7-popover
  config:
    class: deleteItems
  slots:
    default:
        - component: Label
          config:
            text: Some sheet content
```

## Your environment
openHAB 3.1.0 (Build #2190) on RPi4b

florian-h05:
Fixes #927.

When attempting to open an `oh-` Vue component, e.g. `oh-colorpicker-card`, as action modal, it was not possible to configure it, even though `actionModalConfig` was set.
This was because `oh-` componens take their configuration from the computed `config` of the widget-mixin instead of the Vue component `props`. 
This PR fixes that by using the `actionModalConfig` for the computed `config` of `oh-` components in action modals.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/modals/modal-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/modals/modal-mixin.js
@@ -23,7 +23,8 @@ export default {
         root: component,
         store: this.$store.getters.trackedItems,
         props: this.modalConfig,
-        vars: this.vars
+        vars: this.vars,
+        modalConfig: this.modalConfig // For configuration of oh- components
       }
     },
     modalStyle () {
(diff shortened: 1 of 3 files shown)


florian-h05:
@ghys I have battle-tested this on my production system and I haven't noticed any regressions, so this should be safe to merge.

ghys (approved):
Can't say I have battle-tested this myself but upon reading the code it makes sense!

LGTM

PR #2087, opened 2023-09-23
Items tag: added tags to list, unification of items tag edit

Davek145:
Closes #2086

- Adds Non-semantic tags to Items list (e.g. /settings/items/) consistently with similar list of Rules / Scripts / Pages.
- Extends `getItemTypeAndMetaLabel` with relatesTo part of Semantic classification.
- Fixes incorrect calculation of vue virtual box height on Items list.
- Replaces custom tag input with accordion tag input (inspired by https://github.com/openhab/openhab-webui/pull/2078) and show number of tags (inspired by https://github.com/openhab/openhab-webui/pull/2083).
- Moves the custom tag input to `item-form.vue`, so it is available also in Model view, when creating Item from Thing etc.

At minimum for the calculation of vue virtual box height it would be great to fully test this PR for variety of browsers. Currently I do not have complete BrowserStack access avaible, but anyway I have tried to test and adjust height calculation for following browsers using trial and my own devices (This PR is tested with BrowserStack):
Windows – Chrome, Edge, Firefox
MacOS – Chrome
Linux – Chrome, Firefox
iOS – Safari, Chrome
Android - Chrome
I kindly ask maintainers to test this PR using BrowserStack for other browsers. If I get full access from BrowserStack in the meantime, I will complete the testing with other browsers as well.

PR #2095, opened 2023-09-26
Semantic model: Remove duplicate non-semantic tags in Item component

Davek145:
Changes introduced into #2087 in combination with #2093 create duplicity in showing non-semantic tags in the model page (component/model/item-details.vue). 
Non-semantic tags are displayed both in the Item and when edit mode is enabled, in the accordion tag input in item-form.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/model/item-details.vue
+++ b/bundles/org.openhab.ui/web/src/components/model/item-details.vue
@@ -3,7 +3,7 @@
     <f7-card-content>
       <f7-list media-list accordion-list>
         <ul>
-          <item v-if="!createMode" :item="model.item" :link="'/settings/items/' + model.item.name" :context="context" />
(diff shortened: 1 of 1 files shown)


florian-h05:
Can you please provide a screenshot where you see those duplicate tags?

Davek145:
It is in the Model when you select an Item and expand it by Edit:

After your change in item.vue:
image

With my minor fix it looks like this:
image

florian-h05:
I am a bit hesitate here, because on one hand they are duplicate, but only if edit mode is active. If it is not active, you won't see the tags.

Davek145:
It just looked weird to me. But you are right that it is like this only when in edit mode.
What about having them hidden only in the edit mode? `:no-tags="editMode || createMode"`

florian-h05:
Agreed! `:no-tags="editMode"` is enough, the component is only rendered if not in create mode (`v-if="!createMode"`).

florian-h05 (approved):
LGTM, thanks!

PR #2105, opened 2023-10-02
Add a "Add from template" page for the semantic model

JustinGeorgi:
This is the second of [my suggested additions](https://community.openhab.org/t/first-time-user-wizard-for-oh4/149670/36) for helping new users.

This adds a button and new sub page to the semantic model.  On that sub-page the user can select from a  few different basic templates for semantic locations which can then be automatically created and added to the model.

There are a few options for customization:
1) The user can select a custom prefix that will be applied to all the items to fit or establish a desired naming scheme (or just differentiate between two different semantic dwellings if this is a multi dwelling OH instance)
2) The user can optionally deselect extra pieces of the model that are not appropriate to their needs.

image

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
@@ -61,7 +61,15 @@
     <f7-block v-else class="semantic-tree-wrapper" :class="{ 'sheet-opened' : detailsOpened }">
       <f7-row>
         <f7-col width="100" medium="50">
-          <empty-state-placeholder v-if="empty" icon="list_bullet_indent" title="model.title" text="model.text" />
+          <f7-block v-if="empty">
+            <empty-state-placeholder icon="list_bullet_indent" title="model.title" text="model.text" />
+            <f7-row class="display-flex justify-content-center">
+              <f7-button color="blue" large raised fill @click="addLocationTemplate()">
+                Create Locations From Templates
+              </f7-button>
+            </f7-row>
+          </f7-block>
+
           <f7-block v-show="!empty" strong class="semantic-tree" no-gap @click.native="clearSelection">
             <!-- <empty-state-placeholder v-if="empty" icon="list_bullet_indent" title="model.title" text="model.text" /> -->
             <f7-treeview>
@@ -92,14 +100,6 @@
                   <f7-list-button color="blue" title="Add Equipment" @click="addSemanticItem('Equipment')" />
                   <f7-list-button color="blue" title="Add Point" @click="addSemanticItem('Point')" />
                   <f7-list-button color="blue" v-if="includeNonSemantic" title="Add Item" @click="addNonSemanticItem(false)" />
-                </f7-list>
-              </f7-card-content>
-            </f7-card>
-            <div><f7-block-title>Model Templates</f7-block-title></div>
-            <f7-card>
-              <f7-card-content>
-                <f7-list>
-                  <f7-list-button color="blue" title="Location Templates" @click="addLocationTemplate()" />
                 </f7-list>
               </f7-card-content>
             </f7-card>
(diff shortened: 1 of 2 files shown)


ghys:
Thank you for this!
It isn't what I had in mind (a multiple-step wizard with simple options) but it could do the job.
I think we should indeed offer the option only when the model is empty, so it could be added as a button to the empty state placeholder (under the "Start Modelling your Home" blurb) instead of an in-your-face dialog box ("There is no semantic model. Create model locations from template?").

In any case I'm good with adding this feature early even as a "beta" feature, even if we revisit it later, it is a net benefit (as it's meant to be used just once so there's no regressions to keep in mind).

JustinGeorgi:
> I think we should indeed offer the option only when the model is empty

That was my first intention as well, but on further thought I wondered whether 1) it's better for a user to still be able to find it after they've made a couple of their own test semantic elements 2) the available templates could be expanded in the future to be more than just this initial setup.

I'm not convinced either way is better, so I'm happy to go back just available when the model is blank.

JustinGeorgi:
Something like this?

image

ghys:
Sorry for not following up lately @JustinGeorgi.
I think the goal now would be to target it for 4.2 - and it's not all that bad, it allows us some time to get it just right.
This would be be an important feature, I think, so we shouldn't rush it.

ghys:
> Something like this?
> 
> image

I would simply name the button 'Get Started'. And then you would have something to.... well, get started, like a wizard.

JustinGeorgi:
> Sorry for not following up lately

No worries at all.  It's a busy time for everyone.

> I would simply name the button 'Get Started'.

Done.

> This is also interesting, but is it worth the drawback though?

It should be easy enough to go back and expand it to be available more generally if it winds up being a frequently used/popular option, so leaving it just as a start for now is fine by me.

florian-h05:
@ghys Can you please review this PR, since it seems you have a vision how it could look which I don't have?

ghys:
My point really was that sometimes you have to design screens for people who know their way around and others for people who don't. I felt it was a case of the former when it should have been the latter.

JustinGeorgi:
> sorry for not following up.

No worries, @ghys! Honestly, I've got so many other things going on right now, I had forgotten about this one too.

> My "vision" of sorts was that the first-time model setup could be a wizard (like the one you get when starting the instance for the first time) with several steps asking you very practical (as opposed to technical) questions:

That's fine too.  I'll admit, however, that I personally have no instinct at all for what that series of "practical questions" might be other than "Do you have a kitchen?", etc.  At that point, that's no different than the series of checkboxes, so clearly you've got something in that's a completely different approach.

> My point really was that sometimes you have to design screens for people who know their way around and others for people who don't.

If the "Item Name Prefix" section is too intimidating (and I can see that), that's easy to hide behind an advanced options toggle like so many of the other dialogs.

In the end though, we can shelve this if you don't see a way to salvage it.

ghys:
> In the end though, we can shelve this if you don't see a way to salvage it.

I wouldn't dare shelve it as it's a clear UX improvement, and I recognize the good work that you made.

In fact it could coexist with the "wizard" that I eventually would like to see. When both are available we could have some sort of menu which asks you whether you prefer this "create locations from template" page (which assumes you know what the model and locations are) or the full wizard.
In the meantime we can clearly add your page.

bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
```diff
@@ -376,6 +385,20 @@ export default {
...
+              url: 'add-template',
+              route: {
+                component: AddFromTemplate,
+                path: 'add-template',
+                props: {
+                }
+              }
+            })
+          })
+        }
```
ghys, line 401: I would rather not have this, to be honest.
The button in the empty state placeholder is enough, IMO.

JustinGeorgi: > I would rather not have this, to be honest.

Done. PR should be up to date now as well.

florian-h05 (approved):
LGTM, thanks.
Really nice addition!!

I have done a few minor changes:
- Fix styling of the add from template page for narrow screens (fix padding).
- Minor code improvements, including disabling the erroring ESLint rule for the affected line.
- Rename the button from "Get Started" to "Add Locations from Template".
- Set category/icon for the template Items, just looks nicer afterwards.

PR #2109, opened 2023-10-04
Fix `oh-stepper` not working if step is an integer

florian-h05:
Regression from #2090.

Fixes `Uncaught TypeError: Cannot read properties of undefined (reading 'length')` being thrown if the step property is an integer value, e.g. 1 or 1.0.


florian-h05:
@openhab/webui-maintainers Can this one-liner please be merged ASAP? 
The stepper is currently broken for all users who define a integer step, e.g. 1, 5, or 1.0, 5.0 etc.

kaikreuzer (approved):
Thanks!

PR #2112, opened 2023-10-06
[BasicUI] Hide header row if label not explicitly set

### LINKED ISSUE #2065: [Basic UI] Improve layout of header and buttons
This refers to https://github.com/openhab/openhab-webui/pull/2010 and the discussion at [the forum](https://community.openhab.org/t/openhab-4-1-milestone-discussion/149502/14):

After seeing the new feature on my production env, I feel that it could possibly be improved. What was a neat, slim list of charts before has now become bloated and imho unpleasantly arranged:

Bildschirmfoto 2023-09-13 um 08 35 39

Things that might be improved:
- Showing a text/icon although none has been specified feels like a regression and it should imho be avoided. Setting `label=""` and `icon="none` does not feel right, it is counter-intuitive (we usually define what we want and not what we do not want) and looks rather like a workaround.
- The new header wastes a lot of space, especially, if no label+icon is desired. One idea would be to show the buttons as an overlay when hovering over a chart. Imho this would be quite elegant, although I understand that it might not be simple to implement.

Looking forward to your thoughts on this, @lolodomo.

lolodomo:
Applied to Chart/Image/Viodeo/Mapview/Webview elements.

Chart and Image elements are also now clickable.
Clicking allows toggling between header row shown and header row hidden.

Fix #2065

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
@@ -127,6 +127,12 @@ public class ChartRenderer extends AbstractWidgetRenderer {
             }
 
             String snippet = getSnippet("chart");
+
+            boolean showHeaderRow = chart.getLabel() != null;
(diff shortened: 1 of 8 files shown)


kaikreuzer (approved):
Thanks!

PR #2119, opened 2023-10-08
Adapt motion icon to bright and dark themes

lolodomo:
Use currentColor as color.

Related to #1769


lolodomo:
This change depends on #2084 as only the SVG icon is updated.

lolodomo:
With this change:
image
image

Before this change:
image
image

The difference is subtle but at least now the user will be able to control the color of the icon.

lolodomo:
One advantage of this change is that users can now control the color of the icon.

```
sitemap test16 label="Test motion icon"
{
	Frame {
		Default item=TestString staticIcon="motion"
		Default item=TestString staticIcon="motion-open"
		Default item=TestString staticIcon="motion-closed"
		Default item=TestString staticIcon="motion" iconcolor=["blue"]
	}
}
```

image

lolodomo:
As Main UI does not adjust icon color depending on theme when "currentColor" is used in the SVG, I will update my change to not use "currentColor" and choose a "hardcoded" color that is fine in bright and dark themes.

lolodomo:
I am closing this change, the current motion icons are fine in bright and dark themes. The rendering in MainUI:
image
image

PR #2122, opened 2023-10-09
Rules list: Add filtering by (multiple) tags

### LINKED ISSUE #2123: Rules list: Grouping rules/scripts by semantic and by all tags and Filtering by Multiple tags
Features:
1. grouping by semantics
2. grouping by tags
3. filtering by tags

PR with description is here #2122 

## Group by semantic 
image
image

## Group by tags (all tags)
image

### without filtering
image

### filter by multiple tags
image

d51x:
Closes #2123.

This adds filtering by tags to the rules list, the filter is inside an accordion.
In case you select multiple tags, the filter behaves like OR, which means all rules which have at least one of the selected tags are displayed.
In case you use the searchbar, the rules list will be narrowed down to the rules which match the search query and have the selected tags.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
@@ -417,7 +417,7 @@ export default {
           searchbar.clear()
           searchbar.search(filterQuery)
         }
-        /*if (groupBy === 'alphabetical')*/ this.$refs.listIndex.update()
+        this.$refs.listIndex.update()
       })
     },
     toggleSearchTag (e, item) {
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks! Nice contribution.

To save me some time, I decided to implent my review:

- I removed sort by semantics (as I want to generally remove semantics for rules) and sort by tags (as this is type of duplicated functionality of the tag filter).
- For the tag filter, I cleaned-up and refactored the code a bit, and removed the capitalization of the tags.
- I added a "Reset filter" functionality for convenience. I don't want to unselect all tags by hand ;-)
- I wrapped the tag filter inside an accordion, so the filter does not need much space by default, when you need it, just open the accordion.

BTW I can imagine having this capability implemented for pages and Items as well. In that case, the filtering should be refactored into a mixin and Vue component to avoid code duplication.

florian-h05:
image

This is how it looks now.

PR #2125, opened 2023-10-09
Unwrap command Item picker from f7-list-item

florian-h05:
Addresses https://github.com/openhab/openhab-webui/pull/2055#discussion_r1350710682.


florian-h05:
@ghys I'm fast! Here is the updated command Item picker.

ghys (approved):
You are fast and so am I.

PR #2126, opened 2023-10-09
Set unit metadata & state description pattern when creating UoM Item

### LINKED ISSUE #2108: [MainUI] Add option to set unit to Model and "Add equipment to model"
## The problem

With the changes made to UoM handling in OH 4 the `unit` Item metadata has become very important. More Number:Something Items than not will require the `unit` to be set, particularly with units like energy, power, and especially dimensionless. 

Given this importance, it would ease the burden on users creating new Items if the `unit` could be set when creating Items through the Model view and from the "add equipment to model" without first creating the Item and then needing to click through to the newly created Item(s) to set the metadata. 

State Description could have a similar positive benefit.

## Your suggestion

I can see several approaches. 

1. Just add a `unit` text field where the user types in the desired unit. 
    a. Bonus if it's pre-populated from the unit supplied by the Channel, though I don't know whether that information is even available on the Channel.

2. Add `unit` and State Description options

4. Add an option to add any Meta data similar to how it's supported on the Item's page.

florian-h05:
Related to https://github.com/openhab/openhab-core/pull/3481.
Closes #2108.

- Ensures that the internal unit is set when an Item is created, so that in case the system unit changes (i.e. measurement system changes) persisted data does not get corrupted.
By default, the unit metadata is set to the system default unit, however the user can easily change that on Item creation.
  
  The unit can also be changed later as it is just normal metadata, however this can corrupt persisted data.
- Adds the ability to set state description pattern when creating a UoM Item.
- Shows the group type for groups, e.g. `Group (Number:Temperature)` instead of just the Item type.

This applies to both the Items and the model page.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
@@ -163,17 +163,36 @@ export default {
       if (this.item.groupType === 'None') delete this.item.groupType
       if (this.item.function === 'None') delete this.item.groupType
 
+      const unit = this.item.unit
+      delete this.item.unit
+      const stateDescriptionPattern = this.item.stateDescriptionPattern
+      delete this.item.stateDescriptionPattern
+
       // TODO: Add support for saving metadata
       this.$oh.api.put('/rest/items/' + this.item.name, this.item).then(() => {
         let unitPromise = Promise.resolve()
-        if (this.createMode && (this.item.type.startsWith('Number:') || this.item.groupType?.startsWith('Number:')) && this.item.unit) {
+        if (this.createMode && (this.item.type.startsWith('Number:') || this.item.groupType?.startsWith('Number:')) && unit) {
           const metadata = {
-            value: this.item.unit,
+            value: unit,
             config: {}
           }
           unitPromise = this.$oh.api.put('/rest/items/' + this.item.name + '/metadata/unit', metadata)
         }
         return unitPromise
+      }).then(() => {
+        let stateDescriptionPromise = Promise.resolve()
+        if (this.createMode && (this.item.type.startsWith('Number:') || this.item.groupType?.startsWith('Number:')) && stateDescriptionPattern) {
+          if (stateDescriptionPattern !== `%.0f ${unit}`) {
+            const metadata = {
+              value: ' ',
+              config: {
+                pattern: stateDescriptionPattern
+              }
+            }
+            stateDescriptionPromise = this.$oh.api.put('/rest/items/' + this.item.name + '/metadata/stateDescription', metadata)
+          }
+        }
+        return stateDescriptionPromise
       }).then(() => {
         if (this.createMode) {
           this.$f7.toast.create({
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
@@ -35,7 +35,7 @@
           </f7-col>
           <f7-col v-if="item && item.type === 'Group'">
             <f7-block-title>Group Settings</f7-block-title>
-            <group-form :item="item" />
+            <group-form :item="item" :createMode="createMode" />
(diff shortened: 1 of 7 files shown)


florian-h05:
This was actually discussed somewhere, but I cannot find this discussion now.
Maybe @J-N-K finds that discussion?

J-N-K:
This has been discussed several times. I'm not sure if fixing the unit is a good idea, while I agree that fixing the dimension is. The difference is: if you create an item, you have an idea what you want to use it for. This fixes the dimension, but not necessarily the unit. 

IMO the unit should be editable, but there should be a warning that this might be a bad idea when persistence is involved.

florian-h05:
> IMO the unit should be editable, but there should be a warning that this might be a bad idea when persistence is involved.

The unit metadata is still editable through metadata edit, where a big orange warning is shown that this can corrupt persisted data. This PR only sets the unit metadata to the system default when creating an Item, so that there is unit metadata.

rkoshak:
Does it make sense to merge this with #2108? Those are other places where Items are created and giving the users the option to set the unit before the Item is created right there in those respective forms (i.e. Model and add equipment to model) seems like more of the same as what's being implemented here.

ghys:
> This is how it looks when creating an Item:

> 

I'm sorry to say it, but this description is absolutely preposterous; it's 4 lines long which is probably 2 too many and seriously:

> All processed values are internally normalized to the specified unit. The normalized value is used to propagate the value to external integrations (e.g. persistence, REST API, WebSocket) so these values will always have the specified unit and scale.

You couldn't be as cryptic and obscure if you tried.

florian-h05:
@ghys I have updated the description, I am open for suggestions!

@rkoshak I have tried to address #2108, please have a look at the PR description.

spacemanspiff2007:
How about:

"Used internally, for persistence and external systems. It is independent from the state visualization in the GUI which is defined through the state description".

rkoshak:
I'm not good at translating this code to what it will look like but based on "This applies to both the Items and the model page" it's not clear this also impacts "create equipment from thing" and "add points to model".

image

 If I missed that I'd say it looks good. If It didn't then that's an important place to make suer we can set the unit too.

martingrassl:
> I'm not good at translating this code to what it will look like but based on "This applies to both the Items and the model page" it's not clear this also impacts "create equipment from thing" and "add points to model".
> image
> 
> If I missed that I'd say it looks good. If It didn't then that's an important place to make suer we can set the unit too.

+1 for this - if it is possible to set the unit in this screen that would be perfectly fine!

florian-h05:
> I'm not good at translating this code to what it will look like but based on "This applies to both the Items and the model page" it's not clear this also impacts "create equipment from thing" and "add points to model".

Those fields were displayed on both pages, but not pre-filled. I have fixed this.

> "Used internally, for persistence and external systems. It is independent from the state visualization in the GUI which is defined through the state description".

Thanks for the suggestion!

florian-h05:
@J-N-K @spacemanspiff2007 @rkoshak:
Should this PR be merged in the current state or is there anything left to be changed?

florian-h05:
@ghys Can you please review?

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
```diff
@@ -159,12 +160,10 @@ export default {
       if (this.currentTab === 'code') {
         if (!this.fromYaml()) return Promise.reject()
       }
-      if (!this.item.name) return // user cannot change name
+      if (!this.item.name) return this.$f7.dialog.alert('Please give Item a valid name').open() // user cannot change name
```
ghys, line 163: (not blocking, for later)
```suggestion
      if (!this.item.name) return this.$f7.dialog.alert('Please give the Item a valid name').open() // user cannot change name
```

ghys (approved):
Yep sorry I had this one on the backburner for some time.
Let's say I'm not happy with the added complexity but I'm happy with the solution and implementation.

A big thank you for the work involved!

PR #2132, opened 2023-10-10
Add copy rule functionality

d51x:
Sometimes it is necessary to make copies of the rule for later editing.

Added ability to copy a rule.
After "copy rule" new rule is opened
~~If original rule is disabled, then copied rule also disabled~~

#### Original rule

image

#### Copied rule
Opened as New rule with pre-filled values

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/routes.js
+++ b/bundles/org.openhab.ui/web/src/js/routes.js
@@ -297,7 +297,10 @@ export default [
             path: ':ruleId',
             beforeEnter: [enforceAdminForRoute],
             beforeLeave: [checkDirtyBeforeLeave],
-            async: loadAsync(RuleEditPage, (routeTo) => (routeTo.params.ruleId === 'add') ? { createMode: true } : {}),
+            async: loadAsync(RuleEditPage, (routeTo) =>
+              (routeTo.params.ruleId === 'add')
+                ? { createMode: true } : (routeTo.params.ruleId === 'copy')
+                  ? { copyMode: true } : {}),
             routes: [
               {
                 path: 'script/:moduleId',
--- a/bundles/org.openhab.ui/web/src/js/routes.js
+++ b/bundles/org.openhab.ui/web/src/js/routes.js
@@ -298,9 +298,9 @@ export default [
             beforeEnter: [enforceAdminForRoute],
             beforeLeave: [checkDirtyBeforeLeave],
             async: loadAsync(RuleEditPage, (routeTo) =>
-              (routeTo.params.ruleId === 'add')
-                ? { createMode: true } : (routeTo.params.ruleId === 'copy')
-                  ? { copyMode: true } : {}),
+              routeTo.params.ruleId === 'add' ? { createMode: true }
+                : routeTo.params.ruleId === 'copy' ? { copyMode: true }
+                  : {}),
             routes: [
               {
                 path: 'script/:moduleId',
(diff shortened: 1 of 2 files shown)


rkoshak:
I wonder if the new rule should be created initially disabled regardless of the state of the old rule. For example, if the rule is triggered by a system runlevel/system started trigger, the new rule will immediately run as soon as it's created as a copy which could be undesirable. Simply making a copy shouldn't cause "stuff" to happen.

Also, since the rule's UID cannot be changed once it's created, it would be desireable if the user could choose the UID of the newly created rule. Because one needs the UID of a rule to call it from another rule, and some error messages use the rule UID in the error, having the ability to choose our own UID on creation/copy is kind of important. 

Could this be used perhaps as a way to create a copy of an unmanaged rule (i.e. rule defined in a text file instead of through the UI)? I would expect not but if there is enough data there to recreate the rule as a managed rule that could be a cool way to import text file rules into the UI.

d51x:
> I wonder if the new rule should be created initially disabled regardless of the state of the old rule. For example, if the rule is triggered by a system runlevel/system started trigger, the new rule will immediately run as soon as it's created as a copy which could be undesirable. Simply making a copy shouldn't cause "stuff" to happen.
> 
> Also, since the rule's UID cannot be changed once it's created, it would be desireable if the user could choose the UID of the newly created rule. Because one needs the UID of a rule to call it from another rule, and some error messages use the rule UID in the error, having the ability to choose our own UID on creation/copy is kind of important. 
> 
> Could this be used perhaps as a way to create a copy of an unmanaged rule (i.e. rule defined in a text file instead of through the UI)? I would expect not but if there is enough data there to recreate the rule as a managed rule that could be a cool way to import text file rules into the UI. 

UID of copied rule is new (unique)

rkoshak:
> UID of copied rule is new (unique)

Yes but the end user should be allowed to change it before the new rule is created.

d51x:
> > UID of copied rule is new (unique)
> 
> Yes but the end user should be allowed to change it before the new rule is created.

don't understand....  "copy rule" feature makes uniq id for new rule automatically

JustinGeorgi:
> don't understand.... "copy rule" feature makes uniq id for new rule automatically. why change it?

Many users do not use the automatically generated random UIDs but prefer to give rules more descriptive identifiable names.  If this feature is going to be useful for all users then it needs to not create the rule as a saved configuration already but open the rule in the rule creation wizard so the user can, if they want to change the UID.

rkoshak:
I listed some examples above for why a user would want to change the UID. In brief:

- makes it easier to identify rules when wanting to call a rule from another rule
- the rule UID sometimes is the only identifier in an error message, using meaningful UIDs makes it easier to map an error to a specific rule

It's a lot easier to work with a rule with a UID of `door-open-detection` than with a UID `51f0c47a62`.

d51x:
the implementation of the copy rule has been changed
the form opens as for a new rule
user can change rule ID
the rule is not saved until the user clicks the Save button

florian-h05 (commented):
LGTM, thanks!
This has been on my "I want to implement it list" for a long time, thanks for doing that!

I just fixed some places you forgot to replace `createMode` with `isNewRule`.

Wrt to disabling a copied rule by default: Since a newly created rule is enabled by default, I would do the same for a copied rule. However technically it would be possible to create the new rule right after creating it, that would not solve the problem with start level trigger triggering. Seems like it is not possible to create the rule already disabled.

florian-h05 (approved):

PR #2136, opened 2023-10-14
Fix color picker not displayed correctly

### LINKED ISSUE #2134: Color Widget lacks default module setting
When creating an item that has color state which leads to the default of a oh_colorpicker_card the widget is created in a wrong way

image

The root cause is the missing module setting

image

The setting should therefore as follows by default

```
value: ""
config:
  modules:
    - wheel
```

as such 
image

which then reveals a working color widget:

image

florian-h05:
Fixes #2134.
Regression from #1989.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-colorpicker.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-colorpicker.vue
@@ -1,5 +1,8 @@
 <template>
-  <div :style="config.style">
+  <div :style="{
+    width: '100%',
+    ...config.style
+  }">
     <div v-if="!config.openIn" ref="container" style="width: 100%" />
     <div v-else ref="swatch" :class="config.swatchClasses || ['elevation-4', 'elevation-hover-8', 'elevation-pressed-1', 'elevation-transition']" :style="{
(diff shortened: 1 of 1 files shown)


Mark-VG:
Hi.
Any chance of a jar for this to try before merge. Hoping it fixes the issue I am seeing.

florian-h05:
Sure: https://we.tl/t-safSYPMgG8.

Mark-VG:
> Sure: https://we.tl/t-safSYPMgG8.

Thank you.

Same instructions as firvtge other one?

Deploy it to your addons folder, and log into the Karaf console:

bundle:list org.openhab.ui 

Stop the milestone bundle, restart the other one.

bundle:stop ID and bundle:restart ID.

florian-h05:
Exactly.
You can also stop the milestone bundle before dropping the JAR into the addons folder.

Mark-VG:
Thank you so much. Seems to have fixed my issue too :-) Also the reload doe snot happen any more when setting the "H".

image

florian-h05:
Self-merging this small PR to finally fix the broken color picker. 

Please note that the added `width: 100%` is already applied to the color picker‘s div, it is just missing for the wrapping div.

PR #2138, opened 2023-10-15
Developer sidebar: Support searching (in) sitemaps

florian-h05:
Extends the developer sidebar search to sitemaps, where uid, label and slots are searched for the query.
Case is ignored.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
+++ b/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
@@ -542,7 +542,7 @@ export default {
       return false
     },
     /**
-     * Search for the query string inside a single page.
+     * Search for the query string inside a single page or sitemap.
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
```diff
@@ -577,13 +577,15 @@ export default {
           Promise.resolve(this.cachedObjects[1]),
           Promise.resolve(this.cachedObjects[2]),
           Promise.resolve(this.cachedObjects[3]),
-          Promise.resolve(this.cachedObjects[4])
+          Promise.resolve(this.cachedObjects[4]),
+          Promise.resolve(this.cachedObjects[5])
```
ghys, line 581: This is becoming quite embarrassing, but okay ;)

ghys (approved):
I won't ever use it but I don't see why not!

PR #2141, opened 2023-10-18
Sitemap editor: fix staticIcon parameter

mherwege:
One missing blank in the lexer made the functionality fail, but the test still succeed.
Corrected the blank, and then also had to correct the DSL generator.


mherwege:
Included in https://github.com/openhab/openhab-webui/pull/2145

PR #2144, opened 2023-10-20
Configurable visibility of semantic model cards

### LINKED ISSUE #2143: Configurable visibility of semantic model cards (with visible / visibleTo)
## The problem

Currently it's not possible to show/hide the model cards (the ones generated from the semantic model) for roles or users.

With widget components, this can be achieved by defining the config options `visible` and `visibleTo`.

## Your suggestion

`model-card.vue` already uses the `widget-mixin.js` mixin, which provides the `visible` computed property based on those config options, but this computed property is not used in the template.
The template should be modified accordingly.

## Your environment

openHAB 4.1.0
Build #3680

## Additional information

See [this community thread](https://community.openhab.org/t/restricting-access-to-oh3-overview-and-pages-to-selected-users-other-than-visibleto/134874) for further examples

wuschi:
Only render model-card if `visible` computed property is `true`.
Hide divs in model-tabs which do not have a visible model-card.
Closes #2143

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/cards/model-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/model-card.vue
@@ -1,5 +1,5 @@
 <template>
-  <f7-card expandable ref="card" class="model-card" :class="type + '-card'" :animate="$f7.data.themeOptions.expandableCardAnimation !== 'disabled'" card-tablet-fullscreen @card:opened="cardOpening" @card:closed="cardClosed">
+  <f7-card v-if="visible" expandable ref="card" class="model-card" :class="type + '-card'" :animate="$f7.data.themeOptions.expandableCardAnimation !== 'disabled'" card-tablet-fullscreen @card:opened="cardOpening" @card:closed="cardClosed">
     <f7-card-content :padding="false">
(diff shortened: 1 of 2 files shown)


florian-h05:
> Hide divs in model-tabs which do not have a visible model-card.

What do you mean by this?

Just a sidenote: Have you noticed that you cannot propertly configure a model card? The drop down seems to have some CSS problems:

image

florian-h05 (approved):
LGTM, thanks! Congrats to your first contribution!

wuschi:
Thanks for merging!

> 
> Just a sidenote: Have you noticed that you cannot propertly configure a model card? The drop down seems to have some CSS problems:
> 
> image

Didn't notice that as I'm mostly doing configuration by code.  I just checked - looks normal on my system:


Bildschirmfoto 2023-10-29 um 16 39 47

florian-h05:
Can you open the drop down and it works? It’s the icon besides the orange checkbox.

wuschi:
> Can you open the drop down and it works? It’s the icon besides the orange checkbox.

Now i see what you mean.
Took me a while to figure out - the problem only seems to occur for the last item in the list.
For all other items the dropdown seems to work.

florian-h05:
Exactly. But I don’t find the cause of this issue, do you have an idea what’s going on?

PR #2146, opened 2023-10-21
Fix oh-clock context in clock card

JustinGeorgi:
This PR fixes an error in the context definition of the `oh-clock` used  in the `oh-clock` card that was causing the clock cards to fail to render in the updated configuration definition.

See the forum discussion:

https://community.openhab.org/t/oh4-oh-clock-card-empty-since-oh4-1-m2/150518


florian-h05 (approved):
LGTM, thanks!

PR #2157, opened 2023-10-29
SSE event mixin: Replace RegEx lookbehind assertion with split by string

### LINKED ISSUE #2153: Main UI shows only blank screen
## The problem

<!--
After updating to SNAPSHOT #3686 (issue might have been introduced earlier), Main UI only shows a blank screen at iPadOS 15.8, so I can no longer use my older iPads.
No issue at iOS 17.1 as far as I could test.

## Expected behavior

I would expect the same behavior like using iOS 17.1

## Steps to reproduce

Nothing special, just use the above named iOS/iPadOS version with any browser or openHAB‘s app.

## Your environment

I would post this information If I could see anything in MainUI

## Browser console

not available at Apple mobile devices AFAIK

## Browser network traffic

See above

## Additional information

None

florian-h05:
Fixes #2153.
Regression from #2055.

MainUI was broken on older iOS versions, e.g. 15.8, because RegEx lookbehind assertions were introduced with Safari 16.4.
This uses a split by string instead and adds an ESLint rule to forbid the usage of RegEx lookbehind assertions.

/cc @digitaldan

### DIFF
--- a/bundles/org.openhab.ui/web/.eslintrc.js
+++ b/bundles/org.openhab.ui/web/.eslintrc.js
@@ -24,7 +24,8 @@ module.exports = {
     '@babel',
     'import',
     'cypress',
-    'vue'
+    'vue',
+    'es'
   ],
 
   globals: {
@@ -57,6 +58,7 @@ module.exports = {
     'no-case-declarations': 'off',
     'no-console': 'off',
     'no-debugger': 'off',
+    'es/no-regexp-lookbehind-assertions': 'error', // Supported in Safari  >= 16.4, which breaks iOS 15.x.
     'no-trailing-spaces': 'error',
     'no-unsafe-optional-chaining': 'error',
     'no-whitespace-before-property': 'error',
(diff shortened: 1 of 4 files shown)


florian-h05:
I have tested this, I did not notice any difference in behaviour, and I also wrote the following script to check whether the behaviour of the different splits is the same:

```javascript
const tests = [
  'notification:Text:Title:Sub Title:Title Right:5000',
  'navigate:/pages/page',
  'popup:widget:customWidget',
  'reload',
  'back',
  'popover:oh-slider-card'
]


let cmdSame = true
let segSame = true

for (const i in tests) {
  const t = tests[i]
  const [command1, ...segments1] = t.trim().split(':') // NOT use a RegEx lookbehind assertions here, because they are unsupported on Safari < 16.4, i.e. iOS 15.x
  const [command2, ...segments2] = t.trim().split(/(?<!\\):/)
  if (!(command1 === command2)) cmdSame = false
  if (!(JSON.stringify(segments1) === JSON.stringify(segments2))) segSame = false
}

console.log('commands same: ' + cmdSame)
console.log('segments same: ' + segSame)
```

This script verified that the behaviour is the same.

florian-h05:
@digitaldan Once you approve this PR, I will self-merge this to get MainUI working again for iOS 15.x devices.

digitaldan (approved):
LGTM

hmerk:
Many Thanks @florian-h05 !!!!
It is working again 👍

PR #2160, opened 2023-11-03
Fix invalid chart period "3M" of sitemaps

stefanroellin:
"3M" for three months is an invalid chart period for a sitemap. Replace it with "4M" for four months, which is missing.

See https://next.openhab.org/docs/ui/sitemaps.html#element-type-chart


florian-h05 (approved):
LGTM, thanks!

PR #2162, opened 2023-11-04
[BasicUI] Fix updating of the chart period parameter in URL

lolodomo:
The period parameter syntax was extended.
Depends on openhab/openhab-core#3863

Also add few new default periods in the selection of chart period


mherwege:
LGTM I will look into adapting the UI sitemap definition checks.

lolodomo:
@kaikreuzer : a small change that is a consequence of the change in core.

kaikreuzer (approved):
Thanks!

PR #2163, opened 2023-11-07
[blockly] Cosmetic changes on cancel and reschedule timer

jimtng:
Make the blocks say "Cancel [private|shared] timer XXX" and "after xx reschedule [private|shared] timer XXXX"

Also put the reschedule block right after cancel.


florian-h05 (approved):
LGTM, thanks!

PR #2175, opened 2023-11-18
[blockly] Add parse number block

### LINKED ISSUE #2139: [Blockly] Add parseFloat/parseInt Blocks
## The problem

The other day I was helping someone with some Blockly code and I could not find any block to parse a string to a number. While this may not come up often, it will come up, particularly when dealing with data that is transformed in some way as the transform action always returns a String.

## Your suggestion

Add a parse block. The first place I looked was the Math category but I could see it belonging a number of places.

stefan-hoehn:
fixes #2139

I added a new toNumber block in the Math section that parses to float (I think we don't need both parseFloat and parseInt).

image

generates

````
console.info(parseFloat('3.14159'));
console.info((1 + parseFloat('2.14159')));
````

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-math.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-math.js
@@ -7,6 +7,24 @@ import { javascriptGenerator } from 'blockly/javascript'
 import { blockGetCheckedInputType } from './utils'
 
 export default function (f7, isGraalJs) {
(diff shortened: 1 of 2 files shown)


florian-h05 (approved):
LGTM, thanks!

BTW very good that you chose `parseFloat()` and not `Number()`. `Number()` is really picky about what it accepts, e.g. `Number(''20.5 °C)` returns `NaN`, `parseFloat` however returns the float. `Number` does not accept NaN chars inside the string.

PR #2180, opened 2023-11-19
Fix developer sidebar conceals "Other Apps" panel

### LINKED ISSUE #2177: Developer sidebar preventing access to other UIs
## The problem

Using snapshot 3725, Basic UI is no more displayed and accessible at right when the developer sidebar is visible:

image

It is hopefully visible when the developer sidebar is hidden:

image

florian-h05:
Fixes #2177.


ghys (approved):
Makes sense, thanks!

PR #2183, opened 2023-11-22
Un-expand system & addon settings when re-entering settings page

florian-h05:
@ghys When re-designing the settings page, our idea was that the expanded stuff is automatically de-expanded once you re-enter the page. That's however not always the case. This should fix this, however it would be better to de-expand `onPageAfterOut` or on `pageBeforeIn`. Unfortunately I wasn't able to get that working.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/menu/addon-section.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/menu/addon-section.vue
@@ -9,7 +9,7 @@
         :key="a.uid"
         :link="'addons/' + a.uid"
         :title="a.label" />
-      <f7-list-button v-if="!expanded && addonsInstalled.length > addonsSettings.length" color="blue" @click="expanded = true">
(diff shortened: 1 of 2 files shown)


ghys:
Thanks, with all due respect, I wonder if this is really useful; the framework retains the immediate former page in the DOM when you navigate forwards, it is very much on purpose, so that when you go back, you can have your former page as it was, with proper animations.

> the expanded stuff is automatically de-expanded once you re-enter the page. That's however not always the case.

That's because you're coming from 2 pages forward and had to go back twice. In fact  I think we need to keep the Settings menu in the DOM (https://v5.framework7.io/docs/routes#keep-alive) but I don't think I have managed to do it.

florian-h05:
When I created that PR, I wasn't quite sure if it is actually needed, so I'm fine with closing it.
This PR would probably also be annoying when you want to change settings of several "advanced" entries, because every time you come back to the settings menu, you have to re-expand.

PR #2190, opened 2023-11-26
Extend chart periods to 3Y, 5Y and 10Y

stefan-hoehn:
Allow further settings 3Y, 5Y, 10Y for dynamic period of chart.

image


florian-h05 (approved):
LGTM, thanks!

I guess especially the 10 year range might take a year to load for those who have a datapoint every minute in their persistence ;-)

stefan-hoehn:
I agree, I don't want anyone to complain if loading takes ages ;-)
However, if you only persist information once a day or so, then the setting makes sense.

PR #2194, opened 2023-12-01
Fix alexa item group type parameter support

jsetton:


florian-h05 (approved):
LGTM, thanks!

I haven't tested this since I don't use Alexa metadata, but I'll trust you as Alexa maintainer.

PR #2214, opened 2023-12-12
Support time in oh-input widget

mherwege:
Time only was not supported so far in the oh-input widget.

This was requested in the community forum: https://community.openhab.org/t/using-a-user-input/151884/9


florian-h05 (approved):
LGTM, thanks! 

I will create a follow-up PR to update the component param description and avoid a few errors being thrown ...

PR #2216, opened 2023-12-12
Addons store: Use swiper for addon cards on mobile & Reduce size of addon card on desktop

### LINKED ISSUE #1668: Why so big logo for addons and why different sizes ?
Why do we have different sizes of logo for the addons ?
Logo for "JavaScript Scripting is really too big and not at all the same size as the logo for "Jython Scipting".
The result on a mobile phone is even worse, one logo takes more than half the screen.
image

### LINKED ISSUE #2205: UI rendering/scaling of logos for suggested add-ons
## The problem

The UI looks like this:

image

## Expected behavior

Are the logos rendered this large intentionally? It looks a bit strange compared to the section below with smaller logos.

## Steps to reproduce

Go to the Add-on Store page.

## Your environment

```yaml
runtimeInfo:
  version: 4.1.0
  buildString: "Build #3760"
```

## Browser console

```txt

```

florian-h05:
Fixes #2205 and fixes #1668.

This slighly reduces the addon card size on desktop and uses a swiper (like in many app stores) instead of a flex on mobile.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
@@ -1,5 +1,5 @@
 <template>
-  <f7-link v-if="addon" class="addon-card padding-right-half" :href="`/addons/${addon.type}/${addon.uid}`">
+  <f7-link v-if="addon" class="addon-card" :href="`/addons/${addon.type}/${addon.uid}`">
     <div class="addon-card-inner card">
       <div class="addon-card-headline">
         <div>{{ headline || autoHeadline || "&nbsp;" }}</div>
@@ -32,19 +32,19 @@
 
 <style lang="stylus">
 .addon-card
-  padding 5px
-  width: 100%
+  width 100%
   position relative
 
   .addon-card-inner
     width 100%
     height 100%
-    margin: 0px
+    margin 0
     display flex
     flex-direction column
     scroll-snap-align center center
     padding 10px
     border-radius 5px
+    box-sizing border-box
 
     &:hover
       background var(--f7-list-link-hover-bg-color)
@@ -66,8 +66,6 @@
       text-overflow ellipsis
       overflow clip
       white-space nowrap
-      // width calc(100% - 5rem)
-      width 210px
       color var(--f7-text-color)
     .addon-card-title-after
       .preloader-inner .preloader-inner-left, .preloader-inner .preloader-inner-right, .preloader-inner .preloader-inner-line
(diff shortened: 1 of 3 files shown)


florian-h05:
<details>

<summary>This is how it looks (Click me to expand):</summary>

image
image
image
image
image


</details>

ghys:
Thanks, good attempt. Maybe a few remarks here:

image

- Could you just fix the horizontal gap between cards so that it matches the vertical one?

- My concerns were firstly that most titles would be truncated (word-wrapping didn't look well), in the above screenshot that's okay because you can guess the truncated word, but in other cases (in the UI tab for instance) it's more difficult. I'm not opposed to a 2-column view for bindings etc. and a single-column view for UI widgets in those cards (with a parameter on the component). But in any case, in the end, the title truncating, well we can live with it.

- Lastly I considered the fact that having 3 cards on 2 rows doesn't look too good (you have an empty space), but if we're not limiting the amount of "featured" add-ons to 3 in sections in the future then we can have what you find in "app store"-like UIs - a single line of horizontally-scrollable list of cards/ icons.

image

florian-h05:
> Could you just fix the horizontal gap between cards so that it matches the vertical one?

I tried to fix that, while doing that, I noticed that the vertical gap actually was not to large, but there was no horizontal gap left.

> My concerns were firstly that most titles would be truncated (word-wrapping didn't look well), in the above screenshot that's okay because you can guess the truncated word, but in other cases (in the UI tab for instance) it's more difficult.

Yes, that's not nice, but I don't have a better idea ...

> I'm not opposed to a 2-column view for bindings etc. and a single-column view for UI widgets in those cards (with a parameter on the component).

The problem is, that this is controlled by CSS, so it is not that easy to control it from JS.

> Another idea would be to just not display the "install/remove" button at least when space is too tight, as you can always go to the details and install from there.

I have had a look at this, it feels a bit weird to not have the install/remove button for those, but for the list.

> Lastly I considered the fact that having 3 cards on 2 rows doesn't look too good (you have an empty space), but if we're not limiting the amount of "featured" add-ons to 3 in sections in the future then we can have what you find in "app store"-like UIs - a single line of horizontally-scrollable list of cards/ icons.

That is a really nice proposal.

florian-h05:
@ghys 

> Could you just fix the horizontal gap between cards so that it matches the vertical one?

Done :+1: 

> Lastly I considered the fact that having 3 cards on 2 rows doesn't look too good (you have an empty space), but if we're not limiting the amount of "featured" add-ons to 3 in sections in the future then we can have what you find in "app store"-like UIs - a single line of horizontally-scrollable list of cards/ icons.


Implemented:

image

florian-h05:
Let’s give it a try 👍
We can always adjust later on — I will merge now.

PR #2229, opened 2023-12-20
Add a new settings menu for persistence

mherwege:
Persistence policies can be configured from the UI. However, it is difficult to find, and one needs to know where to look. Apart from a documentation issue, I think persistence is core enough to merrit its own entry in the settings, so it becomes more prominent.

See for example here: https://community.openhab.org/t/after-reboot-string-values-are-empty/152064/16

This adds a persistence entry below transformations to the settings menu.

The new persistence config page allows to set the default persistence, and links to defining the policies for each installed persistence add-on.
There also is a "add" list entry to install more persistence add-ons.
If none is installed, it will explain persistence and link to the docs and the add-on store.

The default persistence settings has been removed from the system settings, and the link to the persistence policy config from the add-on settings page has been converted to a button.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/i18n/empty-states/en.json
+++ b/bundles/org.openhab.ui/web/src/assets/i18n/empty-states/en.json
@@ -45,5 +45,8 @@
   "addons.text": "Add-ons add functionality to your openHAB system.<br><br>Install them with the button below.",
 
   "page.unavailable.title": "Page Unavailable",
-  "page.unavailable.text": "You are not allowed to view this page because of visibility restrictions."
+  "page.unavailable.text": "You are not allowed to view this page because of visibility restrictions.",
+
+  "persistence.title": "No persistence add-on installed",
+  "persistence.text": "With a persistence configuration you can automatically store item states for the future. To configure persistence, you need at least one persistence add-on to be installed.<br><br>Install one from the Add-On store following the button below."
 }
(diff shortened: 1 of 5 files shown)


hmerk:
> None of the previously defined links were removed, as it makes sense to be able to access it from multiple places.

While I second you PR, I disagree on not removing the existing persistence link. As both are shown, it is a little bit irritating, which one to use. Therefore I vote to remove the old link under system settings.

mherwege:
@hmerk Fine with removing it from the systems settings menu if nobody else objects. What about the link at the top of the persistence add-on?

hmerk:
If it is easier to access via your new menu, it could be removed as well.
Right now it just looks like s headline...

mherwege:
> If it is easier to access via your new menu, it could be removed as well. Right now it just looks like s headline...

I think it is easy to access from the new menu, but I also think it still has its place in the add-on configuration screen. Configuring persistence kind of falls in between configuring an add-on and a system wide configuration. And I rather have 2 places with the option, so a user finds it where they are looking for it. Therefore I changed the link to a button there, making it more obvious.
I did remove the persistence system setting menu.

florian-h05 (approved):
LGTM, thanks!
Good to give the persistence configuration more visibility.

I have changed a few things:
- Remove space between transformations and persistence setting entries. The space above the transformations is meant to separate 1st- from 2nd-level configuration.
- Update the empty state placeholder wording.
- Use a redirect link for the documentation button. This way we can easily adjust where this button links to without having to change the code.
- Fix dirty check (alert when unsaved changed on leave) not working.
- Minor code refactoring in the load method. From my POV there is no catch required as no 404 should occur because the `org.openhab.persistence` service should always be available.
- Align the style to general MainUI styling & Replace the plus button with an "add" list entry or a new button on the empty state placeholder:

image

image

mherwege:
Thanks.

PR #2231, opened 2023-12-22
[BasicUI] Fix icon none

mherwege:
As described in https://community.openhab.org/t/openhab-4-1-milestone-discussion/149502/222, none.png is not found for BasicUI.

This PR replaces it with non.svg, see discussion on forum.

@lolodomo @florian-h05 I was behind my screen when I saw the issue. I didn't test, but did the replacement based on info provided on the forum.

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
@@ -58,7 +58,7 @@ public class ChartRenderer extends AbstractWidgetRenderer {
 
     private final Logger logger = LoggerFactory.getLogger(ChartRenderer.class);
 
-    private static final String URL_NONE_ICON = "images/none.png";
+    private static final String URL_NONE_ICON = "/icon/none.svg";
 
     @Override
     public boolean canRender(Widget w) {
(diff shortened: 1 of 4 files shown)


florian-h05 (approved):
Thanks for the PR!

I just wonder whether `/images/none.png` works, because the resource folder contains that image. 
Anyway, I think this PR can be merged as a quick fix. I won‘t be at home in time so I cannot try out if there is a better solution.

florian-h05:
@kaikreuzer WDYT?

kaikreuzer:
I'm just looking at it - I would change as little as possible; as you say, the image exists, so it's rather a matter of fixing the icon. I'm trying to debug locally now.

kaikreuzer:
Argh, bad timing; my Internet connection just broken and I am also having trouble running the npm part of the Basic UI build locally -> I am failing to test it locally right now. 😢

florian-h05:
I‘m on the way home but it’ll definitely take till 1 pm … and unfortunately I also don’t have my notebook with me.

kaikreuzer (approved):
In general, I would also say the fix looks good.
So let's merge it and quickly test it on a new snapshot.

florian-h05:
And once the release is done we can have a look at fixing it in a cleaner way 👍

kaikreuzer:
I successfully tested the fix!

lolodomo:
Thank you guys.

Normally images/none.png should exist, it is one of the PNG file packaged with Basic UI. Here is the link:
https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.basic/src/main/resources/web/images/none.png

So you probably changed more things than expected.

mherwege:
I probably have. I was mislead by https://community.openhab.org/t/openhab-4-1-milestone-discussion/149502/226 stating images/none.png not existing, but it was using the wrong path, as the URL would have to be http://openHAB:8080/basicui/images/none.png to test.

PR #2232, opened 2023-12-22
{BasicUI] icon.png minimal fix

mherwege:
See comment https://github.com/openhab/openhab-webui/pull/2231#issuecomment-1867570889

This is the minimal fix to the issue.

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
@@ -58,7 +58,7 @@ public class ChartRenderer extends AbstractWidgetRenderer {
 
     private final Logger logger = LoggerFactory.getLogger(ChartRenderer.class);
 
-    private static final String URL_NONE_ICON = "/icon/none.svg";
+    private static final String URL_NONE_ICON = "images/none.png";
 
     @Override
     public boolean canRender(Widget w) {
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Yes, but I actually like the other fix. As we have changed all icons to svg, why should we keep the png file in use here?

lolodomo (approved):
I like this one more.

lolodomo:
@kaikreuzer : is it still time to merge this ?

lolodomo:
> why should we keep the png file in use here?

You could change it later, the idea here is rather to avoid requesting the server.

kaikreuzer:
No, release build is already running.
Why do you like it more?

lolodomo:
> Why do you like it more?

Because the icon servlet can be requested even for stuff not related to icons ! Like image or chart elements rendering.

lolodomo:
I am even not sure that none.png in images has the same size as the icon.

lolodomo:
But it is not dramatic if not changed in 4.1.

lolodomo:
I am not sure of the rendering (size) of the image when the image element will use icon none.svg.

lolodomo:
I am going to merge this fix because after analysing the impact of the changes done by #2231 very quickly at the last minute, I realized that one of the changes broke something !

This revert fix is necessary for both reasons:

1. It makes no sense to request the icon servlet when the purpose is to render an empty image but with absolutely no link with icon stuff. This was the reason of the static image.
2. For real icon management, the broken case by #2231 is if the user wants to do dynamic icon with icon "none". The icon will no more be updated with new state arriving. I agree that the probability that one user defines custom icons with none as prefix is low but there is no valid reason to not cover that case.

lolodomo:
Now, yes an improvement is still possible, consisting in replacing the static none.png by a static none.svg.

lolodomo:
I am going to cherry-pick this PR in branch 4.1.x.

PR #2243, opened 2023-12-29
[BasicUI] Move most of app settings to per-device (browser) settings

### LINKED ISSUE #1832: Consolidate per device MainUI and global BasicUI settings
My understanding of the web interfaces in openHAB 3.4 so far is:

*  For MainUI any unauthenticated user can visit `/about` and setup the “Mode preferences”: Auto (System preference), Light, Dark; and the “Enable Web Audio sink support”. These settings are per device (per browser, actually)
 * The above settings are however irrelevant for BasicUI. To adjust the latter one has to visit `/settings/services/org.openhab.basicui`. Then the settings are not per device, but for all BasicUI instances, whereever they are loaded. Moreover to alter that settings the user has to authenticate towards openHAB.

For me these two-fold settings are contradicting each other. I suggest removing from the URL `/settings/services/org.openhab.basicui` the settings, which are duplicated from `/about` (like color mode and Web Audio sink preferences). Then use on the device where BasicUI is loaded the preferences set at `/about`. This will simplify things.

Moreover it would be good to be able set the default WebAudio (ON) for all devices.  So that it is not necessary to enable it everywhere manually.

### LINKED ISSUE #1930: [Basic UI] On narrower screens Chart takes more place
Below are two screen shots from browser.  In the second one the width of the browser is smaller, but the chart is bigger.  This is against every logic.

I suggest either using always the larger variant of the chart, or using for the chart one column, and in the other column inserting several “normal” items, so that no vertical space is left unused.

image

image

### LINKED ISSUE #2342: [BasicUI] New device setting to enable a bigger font size

- [x] Basic UI

## The problem

Basic UI uses the default font size defined by Material Design Lite which is 14px for texts like sitemap element label and sitemap element value in particular.
On a phone with high resolution screen, texts could appear a little small. Users (not all of course) would appreciate a bigger font size. We need a setting to choose between the default font size and a bigger font size.

## Your suggestion

Now that I have implemented local device settings for Basic UI (PR in review), my proposal is to add a new device setting to enable or disable a bigger font size. The default value will be "disabled" meaning the rendering will be unchanged (backward compatibility).

lolodomo:
Closes #1832
Closes #1930
Closes #2342

A new page is available to show and update all per-device settings. This page is accessible through the sitemap list page or the home page of any sitemap.
Basic UI theme now matches the dark mode of Main UI. It can be updated either in Main UI or in the new page of Baisc UI.
Web Audio can also be enabled either in Main UI or Basic UI.

The iconify setting is now enabled by default. The script to handle iconify icons is now loaded in sitemap pages only if the iconify setting is enabled.

New settings are also available to control the size of the chart and the size of content (text, line) in the chart.

A new setting is also available to enable a bigger font size.


lolodomo:
This PR is finally ready for review.

The home icon at top left has been replaced by a setting icon.

image

When clicking on it, you open the page with all settings:

image

In case you open the settings page in a separate tab or window, your changes are immediately applied to your other tab/window opened on Basic UI.

lolodomo:
I have of course tested all these settings.
Regarding the "web audio" setting, first you have to know that the web audio is possible only when you are on a sitemap page, not when you are on the settings page or the list of sitemaps page. Nothing new here. The enabling/disabling of the "web audio" is working well in Chrome and Edge. In firefox on Windows 10, you have in fact to restart Firefox after enabling/disabling it to have it working, maybe a bug in Firefox.

Theme setting and web audio setting are now common to Main UI and Basic UI.

lolodomo:
Only 2 settings remain on server side:
1. the default sitemap to use if none is provided in WEB URL
2. the "inline SVG" feature because it is something that you should enable or disable globally for all devices, depending only on whether you decided to add or not custom SVG icons with CurrentColor.

lolodomo:
Without bigger font size enabled:

image

With bigger font size enabled:

image

kaikreuzer (approved):
lgtm, didn't test it though Thanks for this enhancement!

lolodomo:
That's cool, thank you for the review.
I am waiting for your feedback after you test it.

kaikreuzer:
Seems to work well and it looks pretty nice - thanks again!

bundles/org.openhab.ui.basic/src/main/resources/OH-INF/i18n/basic.properties
```diff
@@ -1,41 +1,47 @@
...
+ui.config.basic.chartSize.option.840x420 = 840 x 420 (desktop)
+ui.config.basic.chartSize.option.960x480 = 960 x 480 (desktop)
 ui.config.basic.condensedLayout.label = Condensed Layout
 ui.config.basic.condensedLayout.description = When enabled, changes the layout so that more widgets can fit on the screen.
-ui.config.basic.condensedLayout.option.true = Enable
-ui.config.basic.condensedLayout.option.false = Disable
 ui.config.basic.defaultSitemap.label = Default Sitemap
 ui.config.basic.defaultSitemap.description = The sitemap to show if no parameter is passed.
 ui.config.basic.enableIconify.label = Enable Iconify Icons
 ui.config.basic.enableIconify.description = If enabled, the UI will render iconify icons directly by downloading them from the Internet and caching them in the WEB browser. If disabled, the UI will use the standard method by requesting the server icon proxy and assuming iconify icons are installed on the server. Default is disabled.
```
mueller-ma, line 27: @lolodomo You wrote in the PR description: "The iconify setting is now enabled by default". So I guess you need to update this string?

lolodomo: Fixed in PR #2660

PR #2088, opened 2023-09-24
Intergrated Help Features

JustinGeorgi:
Closes #2058.

This converts the developer sidebar into a multipurpose dock. 
It keeps the previous developer sidebar as the `Tools` panel and adds a `Help` panel.

The new `Help` panel contains 4 different sections:
1) A context dependent section that will show basic help for the administrative page currently viewed.  These pages give basic definitions, some simple tips and links to the relevant full documentation.
2) An add-on section that lists all the currently installed add-ons and provides direct links to their documentation pages.
3) A FAQ section for simple answers to common questions with internal links to help guide the users, some external links to relevant resources, and links to documentation pages providing more details about the question.
4) A Quick-Start section that guides a new user with click-by-click instructions for everything from binding installation to adding a widget to a MainUI page with links to the detailed Getting Started Tutorial for each step.  This section is opened by default when an administrator opens a new MainUI instance with no configured overview page (i.e., most likely a new user).

All of the main setting list pages and the overview page now have a help icon in the navbar that toggles the help dock.

FAQ and Quick-Start data are stored in separate JSON formatted files for ease of addition and maintenance.

Signed-off-by: Justin Georgi <justin.georgi@gmail.com>

PR #2253, opened 2024-01-03
Help sidebar: Move docs to docs repo & Integrate them into the UI

florian-h05:
Follow-up for #2088.
Depends on https://github.com/openhab/openhab-docs/pull/2199.

Also add the help sidebar button to the settings and list pages where it is missing.


stefan-hoehn:
I think with the latest change you pushed, which we discussed offline, covering the non-availability of the internet, it is looking pretty good. I also looked through the code and I don't see any obvious issue. So, I think we are good to go.

This PR, based on the related restructuring of the docs sidebar, is a great step forward for context sensitive in the UI and a good integration into the online docs.
Thanks, Florian (not only for the implementation but also for the contribution of the related docs)!

stefan-hoehn (approved):
The code looks good to me. I don't see any obvious issue.

PR #2260, opened 2024-01-08
[BasicUI] Fix Buttongrid alignment on desktop

### LINKED ISSUE #2259: [BasicUI] Buttongrid alignment issue on the web browser

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

When we have a Buttongrid after some other widgets, it seems to not have a left margin, and as a result, it is not aligned the same way as the rest of the widgets on the page

image

## Expected behavior

It should be aligned like this:

image

## Steps to reproduce

1. Create a sitemap with some normal widgets, followed by a buttongrid
```
sitemap test {
  Frame {
    Default item=TestDimmer
    Default item=TestDimmer
    Buttongrid item=Testicon buttons=[1:1:cmd=Cmd]
    Default item=TestDimmer
  }
}
```

It is not necessary to have other widgets, nor to enclose it in a Frame. Even with a single lone buttongrid it's happening, but we can't see the problem easily because we have no "frame of reference".

## Your environment

MacOS 13.6.3 on Chrome Version 120.0.6099.129 (Official Build) (arm64)

## Browser console

```txt
GET http://192.168.1.10:8484/icon/buttongrid?iconset=classic&format=svg&anyFormat=true 404 (Not Found)
```
I did not specify any icons so this request for `buttongrid` icon is weird.

jimtng:
Fixes #2259

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/resources/snippets/buttongrid.html
+++ b/bundles/org.openhab.ui.basic/src/main/resources/snippets/buttongrid.html
@@ -6,7 +6,7 @@
 		%label%
 	</span>
 </div>
-<div class="mdl-form__row-buttongrid mdl-form__row--height-auto mdl-cell mdl-cell--12-col %visibility_class%">
+<div class="mdl-form__row mdl-cell mdl-cell--12-col mdl-form__row--height-auto %visibility_class%">
 	<div
 		class="mdl-form__control mdl-form__buttongrid mdl-grid"
 		data-control-type="buttons"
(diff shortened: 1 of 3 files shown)


lolodomo (approved):
Your code is better than mine, thank you.

lolodomo:
Tested on PC with different browsers and on my Android phone and works perfectly.

lolodomo:
Cherry-picked in 4.1.x branch

PR #2263, opened 2024-01-08
Cleanup Java code

wborn:
This cleanup includes:

* Fix deprecations
* Fix JavaDocs
* Remove redundant toString calls
* Remove redundant semicolons
* Simplify boolean expressions
* Use diamond operator
* Use enhanced for loops
* Use instanceof pattern matching
* Use isEmpty instead of 0 comparisons
* Use lambdas
* Use static inner classes
* Use StandardCharsets

Also adds the SA_LOCAL_SELF_COMPARISON suppression similar as used in other repositories for https://github.com/spotbugs/sonar-findbugs/issues/876.


bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/PageRenderer.java
```diff
@@ -136,12 +136,12 @@ public StringBuilder processPage(String id, String sitemap, String label, EList<
...
-    private void processChildren(StringBuilder sb_pre, StringBuilder sb_post, EList<Widget> children, String sitemap)
+    private void processChildren(StringBuilder sbPre, StringBuilder sbPost, EList<Widget> children, String sitemap)
             throws RenderException {
         // put a single frame around all children widgets, if there are no explicit frames
         if (!children.isEmpty()) {
-            EObject firstChild = children.get(0);
-            EObject parent = itemUIRegistry.getParent((Widget) firstChild);
+            Widget firstChild = children.get(0);
+            EObject parent = itemUIRegistry.getParent(firstChild);
             if (!(firstChild instanceof Frame || parent instanceof Frame || parent instanceof Sitemap)) {
```
lolodomo, line 145: We have to be sure that this change of type does not break the test `firstChild instanceof Frame`

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
```diff
@@ -423,9 +418,8 @@ protected String processColor(Widget w, String originalSnippet) {
     }
 
     protected @Nullable String getRGBHexCodeFromItemState(@Nullable State itemState) {
-        if (itemState instanceof HSBType) {
-            HSBType hsbState = (HSBType) itemState;
-            return "#" + Integer.toHexString(hsbState.getRGB()).substring(2);
+        if (itemState instanceof HSBType hsbState) {
+            return "#" + Integer.toHexString(ColorUtil.hsbTosRgb(hsbState)).substring(2);
```
lolodomo, line 422: I agree that your change will change nothing but I am asking myself since when hsbState.getRGB() is returning a sRGB rather a RGB ? Was it changed recently ? I am not sure it does not break the color picker in Basic UI.

lolodomo (commented):
Review of changes in Basic UI

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
lolodomo: Change was done in April 2023 but I think it is ok and the result is the same as before.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/PageRenderer.java
lolodomo: Frame => LinkableWidget => Widget => EObject
So it is OK.

lolodomo (approved):
OK for Basic UI.
Not reviewed for other UIs.

ghys (approved):
As far as HABot and HABPanel are concerned, these seem okay to me (but mostly I trust @wborn :wink:)

wborn:
IntelliJ cleaned up most code which I trust. :wink:

kaikreuzer (approved):
Thanks!

PR #2270, opened 2024-01-14
oh-colorpicker: Replace defaultColor option with an actual default color

florian-h05:
Remove the defaultColor option.
Instead use RGB(0,0,0) black as default color if the Item state holds no valid color, so that the color picker is accessible.


stefan-hoehn:
I agree with this approach as I think the default color is nothing really important and it makes the code simpler.

I have pulled the code and tested it and it works fine.
Thanks, Florian, for fixing this as this was really a pain for me ;-)

PR #2063, opened 2023-09-12
Update RelativeCI agent action

vio:
Hi! I noticed that you are still using the RelativeCI CLI agent with the inline API key. 

This PR replaces the CLI with the action agent running on workflow_run event and in the context of the default branch (secrets are not leaked to forks). To get a new API key, go to your organization settings 'API keys' page and generate a new one.

More resources:
- Example pull request from a forked repo: [Pull Request](https://github.com/relative-ci/openhab-webui/pull/2), [GitHub Actions](https://github.com/relative-ci/openhab-webui/actions/workflows/relative-ci.yml), [RelativeCI Job](https://app.relative-ci.com/projects/wDdf9e7l57l5Gn2wlV8Y/jobs/4-HZEwzCXmUeiL1LSvqGDa)
- [RelativeCI agent action setup guide](https://relative-ci.com/documentation/setup/agent/github-action/#workflow_run-event)
- [Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Feel free to reach out if you have any questions,
Vio

PR #1158, opened 2021-09-07
Replace Add-on Management UI with Add-on Store

ghys:
Depends on https://github.com/openhab/openhab-core/pull/2405.

This replaces the current section of the UI for managing
add-ons with a completely new "store-like" interface.
External data, like logos and detailed descriptions for
distribution add-ons are fetched from well-known locations
on the website or GitHub to provide a "store" look & feel.

The add-on store is orgnanized around 4 categories: bindings,
automation, UI, and others, with a separate search tab.
Each category is splitted into subsections. The availability
of these sections depend on the add-on services that are
registered. The sections currently supports add-ons provided
by the "karaf" & "marketplace" services, as provided by
https://github.com/openhab/openhab-core/pull/2405.

Signed-off-by: Yannick Schaus <github@schaus.net>

PR #2273, opened 2024-01-14
Update descriptions in the settings menu

ghys:
I thought these labels could be improved and those describe the relevant sections better IMO, getting more to the point. Native English speakers, please confirm and review!


ghys:
Btw, we have to check why the bundle metrics are like:

  | CurrentJob #1349 | BaselineJob #1158
-- | -- | --
Initial JS | 1.9MiB(+13.82%) | 1.67MiB

I think the baseline hasn't changed since #2063.
@vio could you confirm and tell us how to update it with every PR?

In any case 300kB more initial JS is a lot in 4 months; we'll have to investigate why it happened.

vio:
@ghys the baseline is the latest job in the default branch at the time of the build. Looks like your latest job in the default branch (`main`) was processed on `16/09/2023 16:28:44`: https://app.relative-ci.com/projects/ZNG5hy4VeSJQVQcq1Kvu/tree/main.  

The `Main UI` workflow runs only on `pull-request` and on `workflow_dispatch`, so the push to `main` is ignored atm. To run the workflow and send the latest from `main`, you need to trigger the workflow manually or update the workflow to build and send the data on `push` for the `main` branch

```suggestion
+  push:
+    branches: [ main ]
+    paths: bundles/org.openhab.ui/web/**
+
   pull_request:
-    branches: [ main ]
     paths: bundles/org.openhab.ui/web/**

  workflow_dispatch:
```

florian-h05:
@vio Thx for pointing out what to do.
I have now triggered a manual run, then triggered a re-run of the Main UI action on this PR and the bundle stats look fine again. I will update the workflow file in #2278.

@ghys I have tried to find out what caused the initial CS bundle-size increase, and it is not easy to find out the cause from checking the RelativeCI history, but it seems that #2088 at least lead to some increase. #2253 will help with that.

florian-h05 (approved):
LGTM, thanks!
But I am no native English speaker, so let's ping one @rkoshak.

JustinGeorgi:
The english all looks just fine.

PR #2276, opened 2024-01-17
Rule/Script edit: Fix not editable not properly handled & Always show tags

florian-h05:
Fixes an issue where the configuration of a script action of a file-based rule (e.g. those created using JSRule or Rule Builder of openhab-js) was shown as editable, even though the rule is not editable because it is provisioned from file.

image

Always show the tag input, even if there are no rule tags. This is more consistent and avoids confusion, as you can now clearly see that a rule has no tags.


stefan-hoehn:
Just to explain a bit more what the issue was:

Created a rule with JSRule, which is then still displayed as a file-based rule in the UI.

When opening that rule, only the code of the rule is displayed, but things like as UID, Label, Description, Tags, Script Language can still be edited

For file based rules all this should not editable, but the UI has allowed editing of all fields except the UID.
Also, the tag input was completely hidden if there were no tags instead of just being empty.

Even though it was editable, saving it did not have a effect (as it should be). 
In a nutshell, like any other file based rule, the information should be shown but not be editable via the UI, which was fixed here.

stefan-hoehn (approved):
The changes are straight forward and easy to understand. 
LGTM. Thanks @florian-h05  you fixed this.

Issue #1757, opened 2023-03-05
[BasicUI] Live update is broken in snapshot 3344 ! 


lolodomo:
- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

It looks like BasicUI is no more updating the states in its current page.

## Additional information

I don't think this is related to the other major recent changes (Karaf version, Jetty version, ...) because it was still working as expected for me yesterday in the morning. 
I suddenly encountered the problem yesterday afternoon.

It might be something I broke in one of my changes yesterday. I am sorry for that and I will try to find where is the problem.

lolodomo:
I am starting to have a serious doubt that it is due to my changes because I have undone all my changed in smarthome.js and the problem persists.
I am going to reinstall the snapshot 3340 (before all my UI changes).

lolodomo:
With snapshot 3340 or snapshot 3340 + updated jupp bundle (2.7.0), the live update is apparently working.
This confirms that the problem is in changes of yesterday.
In addition of my own changes in BasicUI, I also see this change: https://github.com/openhab/openhab-distro/pull/1460
Is there any chance that this change to impact communication between OH server and BaiscUI ?
@wborn @J-N-K  ?

J-N-K:
You can easily check, shutdown, edit the jetty.xml and start again. If it works, it‘s related, if it doesn’t, then it‘s another issue.

If it does work, then please try to add an `exclude` similar to the existing `include` for `/rest/events` and see if that improves the situation.

lolodomo:
Interesting, if I install the last BasicUI version in snapshot 3340, the live update is still working.

J-N-K:
So if you install BasicUI from 3340 in a recent snapshot it works, if you use the BasicUI coming with the distribution it does not work? Then it's most likely not related to the GZIP change.

lolodomo:
If I rmeove your change in jetty.xml with a fresh snapshot 3344, it looks like live update is back.

lolodomo:
Closed by openhab/openhab-distro#1476

lolodomo:
I reopen this issue because it happens again in 4.1.1 !
@J-N-K : was there another change ? Or was your fix not fixing the root cause of the problem ?

J-N-K:
With http or https connections? Safari doesn't support SSE on http connections since a long time, maybe other browser catch up there.

lolodomo:
[...]

PR #2280, opened 2024-01-20
[BasicUI] Fix SSE reconnection to the current page

lolodomo:
After an error in the SSE connection, the event source is now reconnected to the right sitemap page.

Fix one of both problems explained in #1757

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -2955,9 +2955,7 @@
 				responseJSON,
 				subscribeLocation,
 				subscribeLocationArray,
-				sitemap,
-				subscriptionId,
-				page;
+				subscriptionId;
 
 			try {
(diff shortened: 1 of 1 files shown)


lolodomo:
@kaikreuzer : this bug was there probably more or less since the first day of Basic UI but because the SSE connection is working very well, I never noticed it because SSE reconnection was never required. I discovered it today because the SSE connection is now no more reliable when using Firefox and now the reconnection code is used.

kaikreuzer (approved):
Thanks, I also never really noticed it, because usually it is rock stable (and I am not using Firefox).

lolodomo:
@kaikreuzer if you are ok, I cherry pick this fix in 4.1.x branch?

kaikreuzer:
@lolodomo Yes, sure, feel free!

lolodomo:
> @lolodomo Yes, sure, feel free!

Ok, done.

PR #2288, opened 2024-01-26
Change the default navbar style from orange to light for light theme

florian-h05:
This is a proposal, please discuss it.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -572,7 +572,7 @@ export default {
     },
     updateThemeOptions () {
       this.themeOptions.dark = localStorage.getItem('openhab.ui:theme.dark') || ((window.OHApp && window.OHApp.preferDarkMode) ? window.OHApp.preferDarkMode().toString() : (this.$f7.darkTheme ? 'dark' : 'light'))
-      this.themeOptions.bars = localStorage.getItem('openhab.ui:theme.bars') || ((this.$theme.ios || this.$f7.darkTheme || this.themeOptions.dark === 'dark') ? 'light' : 'filled')
+      this.themeOptions.bars = localStorage.getItem('openhab.ui:theme.bars') || 'light'
       this.themeOptions.homeNavbar = localStorage.getItem('openhab.ui:theme.home.navbar') || 'default'
       this.themeOptions.homeBackground = localStorage.getItem('openhab.ui:theme.home.background') || 'default'
       this.themeOptions.expandableCardAnimation = localStorage.getItem('openhab.ui:theme.home.cardanimation') || 'default'
--- a/bundles/org.openhab.ui/web/src/components/theme-switcher.vue
+++ b/bundles/org.openhab.ui/web/src/components/theme-switcher.vue
@@ -36,13 +36,13 @@
     </f7-row>
     <f7-block-title v-t="'about.navigationBarsStyle'" />
     <f7-row>
-      <f7-col width="50" class="nav-bars-picker nav-bars-picker-fill" @click="setBarsStyle('filled')">
-        <div class="demo-navbar" />
-        <f7-checkbox checked disabled v-if="barsStyle === 'filled'" />
-      </f7-col>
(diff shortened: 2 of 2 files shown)


florian-h05:
FTR https://community.openhab.org/t/permastore-ui-look-options/152052/7?u=florian-h05.

ghys:
before you merge: why?

florian-h05:
I put this as a proposal to discuss it, so I'm happy that you ask.

Whilst the orange navbar shows openHAB's color, the orange can be quite "pushy", so @mstormi asked me to make the normal navbar the default.

WDYT?

lolodomo:
Can we see a screenshot?
My general feeling is that we should keep OH main color.

mstormi:
You can already switch navbar to that mode (via "about" menu), it's just about making that the default.

ghys:
> I put this as a proposal to discuss it, so I'm happy that you ask.

Sorry I reacted urgently as I was unsure if you would merge this directly, and I'm admitting I didn't see  [https://community.openhab.org/t/permastore-ui-look-options/152052/7](https://community.openhab.org/t/permastore-ui-look-options/152052/7), so thanks for the link.

For full disclosure, I remember my initial UI project and it had the iOS and desktop themes set to light/dark and only the Android theme had the orange navbar (in light theme only), as Android apps tended to have such colored navbars at the time.
In fact it's still there: https://github.com/ghys/openhab-ui-f7/blob/47d5bf7a0c08defc902a92b079dbe93cf61a6e50/src/components/app.vue#L258
compared to now where the desktop theme gets the 'filled' bars unless in dark mode:
https://github.com/openhab/openhab-webui/blob/89766f32bb9c85c96ecb64810660189a351b3b14/bundles/org.openhab.ui/web/src/components/app.vue#L575

I seem to remember a conversation with @kaikreuzer (but I do not remember if it was on GitHub or on Slack) convincing me to put the filled bars on the desktop mode, as it was the 'brand color' - again, IIRC. I just didn't want those filled bars on iOS as I thought it was very un-iOSy to have filled bars on iPhones so I agreed.

So there you have it, I don't offer an opinion, just the context.

florian-h05:
Thanks for sharing the context.
I think that Main UI should look like the OS you are using it on, so if having these coloured bars is still common on Android, I’d keep them.

I haven’t used an Android phone for a while now, is that still the common Android design to have coloured bars?

florian-h05:
@openhab/android-maintainers Can you please tell me whether Android still often users coloured nav bars?

florian-h05:
Okay thanks. 

WDYT matches better to Material You? The orange navbar or the light/dark navbar?

maniac103:
> WDYT matches better to Material You? The orange navbar or the light/dark navbar?

I'd say the light one (is this orange-on-white?), as the app itself doesn't have a colored top bar anymore and thus the phone status bar has a white background as well:

foo

Ideally we could pass an accent color into the UI as well, but I imagine that would be larger effort...

florian-h05:
Your screenshot shows filled (this is orange on white), which is the current default on Android.

So we should keep this default?

maniac103:
> So we should keep this default?

My screenshot was intended to show this default should *not* be kept ;-) With older versions of the app, the status bar background was orange as well, as was the whole app title bar. Nowadays, the app title bar is white(-ish, might change slightly with dynamic theme):

Screenshot_20240223-080954

... thus I think the default should be changed, since the orange bar in combination with white status bar looks a bit out of place now. (I wasn't aware of this being an option, or I would have raised an issue before)

florian-h05:
Thanks for the clarification, so we should use the light (non-orange) bar as default for Android.

I will update this PR to only keep the orange bars as default for the desktop theme.

mstormi:
> I will update this PR to only keep the orange bars as default for the desktop theme.

Why? The orange is pushy, no matter if Android or desktop.

florian-h05:
True, but I'm hesitant because it still is the "brand-color" and no one else has complained yet.
But I guess since on desktop the openHAB logo in the side-menu is likely to be visible, this is not a big problem, so let's change it.
And the openHAB color is at least the accent color in the UI.

florian-h05:
Let me just switch the order of the navigation bar style pickers, so the new default is still on the left.

PR #2293, opened 2024-01-28
[blockly] Use Scratch-3 block style

jimtng:
Changes:
- Use the `Zelos` renderer which looks like MIT's Scratch blocks
- Upgraded Blockly to v10. This is needed to fix a bug that affected Zelos when using a shadow with multiple drop downs. It is still not 100% bug-free. The datetime shadow attached to the persistence block still behaves strangely - the "ago" drop down cannot be changed. But as soon as the numberic input field is changed, Blockly will create an unshadowed datetime input block and it becomes fully functional. So overall this doesn't affect usability.
- Generally, blocks are "inline" instead of multiple lines
- Changed the "now +/- X hours" block to read as "X hours ago/from now"
- Changed the "item xxxx" block to "<icon> xxx". The icon matches the "Items" menu icon in MainUI. The same goes for Thing.
- Reduced the verbosity of the wording, in hope that it reads/flows better when reading the blocks. Generally I'm aiming to change blocks that read as `Get state of item <item Item_Name>` or `Get state of item <get item Item_Name>` => `Get state of <Item_Name>` or `Get state of <get item Item_Name>`
- Changing the persistence method no longer disconnects the datetime input unless the method requires a boolean
- Fixed other minor bugs and typos

image

image

image
image
image


stefan-hoehn:
I am not completely against it as I agree that it is appealing and I honestly weren't aware of them. So thanks.

There is one caveat though, it uses a complete new renderer that needs to be tested. Also a read that it uses smaller font (which I have not yet verified) which would maybe have side effects (not necessarily only bad ones though ;-) ). Also I read that the new renderer is mainly optimized for touch interfaces. Note that also may have have positive effect - it is just that I am cautious.
There is even another renderer "thrasos" that we should look into and compare the three. 

So, give me a bit of time to compare and test them.

jimtng:
There is a bit of an issue with persistence blocks, hence why I marked it as draft.

stefan-hoehn:
Ok, does it mean you will be putting some more investigation into it? 👍

rkoshak:
I do like the looks. 

Given there are at least three renderers, would it be out of the question to let the user choose? The trouble with some of the blocks makes me think the answer is no but it doesn't hurt to ask. ;-)

jimtng:
The other two renderers look very similar. However, I'm sure there will be people who will prefer the current look and don't want to change it. Letting them switch back will make them happy. But imo, it's easier to support just one

jimtng:
I started troubleshooting this and ended up making a ton of changes/refactoring in the existing blocks, primarily for cosmetic reasons. I've also upgraded to the latest Blockly version, adjusting for the breaking changes in the Blockly library. I hope you'll like the changes.

rkoshak:
I think two of your added blocks already exist under the Text category.

image


I'm not sure what the difference is between "join" and the last block beyond the fact that the last block can join any number of strings, not just two. 

Or maybe I'm missing what these do? OR was that not intended to be a list of the blocks you changed?

jimtng:
> I'm not sure what the difference is between "join" and the last block beyond the fact that the last block can join any number of strings, not just two.

Yes it's one of the two blocks I've added so far. The "join" is much simpler and quicker to use. I left the other one for a more complex join involving multiple inputs.

I want to add a few more but perhaps in a separate PR.

The other block I added is in the last screenshot: "key is defined in xxx cache" to complement the "key is undefined in xxxx cache".

rkoshak:
I think adding new blocks is definitely better left for a separate PR. I know that there is a high bar to be met for adding new blocks in addition to the style and the code they produce. I'd expect much discussion that probably won't be relevant to the style changes proposed in this PR and possibly hold things up.

jimtng:
> I think adding new blocks is definitely better left for a separate PR.

Agreed. But I left the bug fixes in this PR.

stefan-hoehn:
@jimtng Just to set expectations: This will take _quite_ some time until I have reviewed this due to the amount of changes you did and the implication it has. I browsed through the comments and I would ask you to only to changes that are _necessary_ to run the new renderer.

Note that I am hesitant to change block naming (I noticed you did) as people are used to it...but I will for sure look into it.
We had long discussion about "item xxx" and the verbosity of those. I would therefore not agree to change these. 

And as Rich said, don't add anything know with this PR (I guess you already dropped it though).

And even though "Changing the persistence method no longer disconnects the datetime input unless the method requires a boolean" seems like a bug which should be fixed, I would rather have that as a separate fix.

"Fixed other minor bugs and typos" -> please keep that seperate.

Even though upgrading blockly is appreciated I have some other blocks to finish first before I will finally do the review. And I have about 100 block tests that I have to manually check against the new blockly version as I am very cautious not to break anything, I hope you understand.

jimtng:
@stefan-hoehn thanks for the heads up. 
Part of the change in style is that (almost) all blocks are inline (a single line) so the inputs spread towards the right, not one input each line going down.

In order to do that, I'm trying to minimise the horizontal space taken.

Would you reject how the blocks below are formatted in terms of wording and insist on keeping the original, much longer text? If that's the case i don't think the scratch block style would work well as it would look very long especially when blocks are inserted into another block as input.

I'm guessing that you've seen and are familiar with https://scratch.mit.edu?

image

stefan-hoehn:
"Part of the change in style is that (almost) all blocks are inline (a single line) so the inputs spread towards the right, not one input each line going down."

I actually had comprehensive discussion with @ghys on that. While I often use single line comments, he is more on the "external inputs" side. 

The following example also depicts why in some cases this make more sense:

image

So in general I would probably not accept it because the defaults were carefully taken (even though I am always open for feedback but again I would rather keep that out from the original intent of the renderer.

I met one of the Blockly Developers at Fosdem yesterday and the reason was indeed to add it to allow Scratch use the latest blockly version while keeping the looks. To be honest, now looking at the renderer >I< personally like the original a bit better, so I would definitely think I we should allow the user to chose switching it (as I would stay with the old one). 

Btw, thanks for adding images to the top description of the PR. This always makes the PR much more descriptive!

jimtng:
In that case I would close this pr. Maybe in the future if a renderer switcher is implemented, this can be revisited.

jimtng:
I've split up the fixes into individual PRs to keep them small and easier to review. Also created a PR just for the blockly version upgrade.

jimtng:
Superseded by #2315

PR #2294, opened 2024-01-28
Lazy load the developer dock

florian-h05:


florian-h05:
@ghys Lazy loading the developer dock/sidebar makes sense IMO as only admins can use it, and it saves us some initial JS.

ghys:
It's certainly increasingly needed when the developer dock adds more and more features.
But I don't think it's taking a lot of initial JS, as there are no big library dependencies.

I think it's just this, bundle 13:

image

image

compared to the initial JS (in purple), that's nice to have but not critical.
Another idea would be to put it in the "admin" bundle? That would be bundle 16 here, in red on the top right:

image

But I'm fine with your proposal to create a new one :)

florian-h05:
> But I don't think it's taking a lot of initial JS, as there are no big library dependencies.

According to the RelativeCI bot it is around 80 KiB, which sounds quite nice to me.

> Another idea would be to put it in the "admin" bundle? That would be bundle 16 here, in red on the top right:

I haven’t considered that, thanks for the suggestion! I’ll change that.

ghys:
Apparently it's now in the one with the add-on store:

image

florian-h05:
Given the fact that the add-on store is using `admin-base` as chunk name as well, that looks good to me.

ghys (approved):
So be it!

I'd strive in general to limit the chunk fragmentation as we have plenty already (100+ chunks), when we can combine some which be not in the initial JS but related, the better, IMO.

PR #2300, opened 2024-02-02
Fix regressions caused by wrong YAML library usage

florian-h05:
Regression from #2267.

In #2267 the YAML library was upgraded to a new version, but the usage of the library was not adjusted to breaking changes.
This PR fixes all issues discouvered in the browser console.


florian-h05:
FYI @GiviMAD

PR #2301, opened 2024-02-02
Widget expressions: (Partly) Fix handling of Item name being `undefined`

florian-h05:
When using the `@` and `@@` shortcuts for Item displayState and state in widget expressions, properly handle cases where the Item name is `undefined`.
Unfortunately I did not find a way to improve that for `items[itemName]` expressions.

Requesting the Item state from the server when the Item name is undefined leads to warnings in the log (which might cause some confusion for the users):
```
[WARN ] [se.internal.SseItemStatesEventBuilder] - Attempting to send a state update of an item which doesn't exist: undefined
```


florian-h05:
FYI @rkoshak IIRC you suggested to reduce the level of that message to info. I have now partially fixed the root cause.

PR #2306, opened 2024-02-06
[blockly] Fix persistence block ZDT disconnection

jimtng:
Fix so oh_get_persistvalue block does not unnecessarily disconnect the ZDT input every time a different methodname is selected.


florian-h05 (approved):
LGTM, thanks!

PR #2309, opened 2024-02-06
[blockly] Fix console warnings when switching to Math

jimtng:
Fix this to avoid a whole bunch of these warnings were logged when switching to math category

```
blocks-math.js:302  CodeGenerator init was not called before blockToCode was called.
blocks-math.js:303  CodeGenerator init was not called before blockToCode was called.
```

This is a quick and dirty solution without doing a bigger refactoring


florian-h05 (approved):
LGTM, thanks!

PR #2310, opened 2024-02-06
[blockly] Upgrade Blockly from v9.2.0 to v10.4.2

jimtng:


jimtng:
Updated to v10.4.1 and rebased to the current main branch

stefan-hoehn:
FYI, I'll start testing this... 🤞🏻

stefan-hoehn:
I noticed one weird behaviour that sometimes the new blockly version doesn't compute the cursor position correctly when trying to add a block into an input. In this case it won't find the input area and doesn't add it. The fact that there is an offset x/y wise to the tip of the block can be perceived because accidently some other input block a bit off the designated input is actually highlighted. Letting go the block, selecting a different one, and then trying it again with the original one sometimes then works fine. 
I am 100% sure though that this is not a bug in your PR but a glitch in that blockly version. We should keep that in mind in case openHAB users start noticing this.

jimtng:
Maybe that's what got fixed here https://github.com/google/blockly/releases/tag/blockly-v10.4.2

stefan-hoehn:
And yes, I can confirm it fixed it. I was actually be able to reproduce it pretty well before .2 and now I cannot anymore. Thanks for finding that while I was still testing more cases.

(btw, you tricked me because I was looking at the title and thought "well, we are already on .2 only, that's strange, when I then saw the force push 😂)

stefan-hoehn:
All in all after testing probably a hundred different rules I can say it looks good to me. So, I will merge that one.
Btw, that gets us into a good shape for a long lasting evil bug in the multiline text block which I told the cpcallen whom I met at the Fosdem and he promised me to fix it soon and he worked on it. It is not yet in this version and finally merged but I guess it will be soon.

stefan-hoehn (approved):
I checked the code migration changes and they look good to me. The amount of tests I did should make us confident enough to migrate to 10.4.2. Let's knock on wood that we won't introduce something breaking.

Thanks for pushing this, Jim 🥳 and sorry that it took some time to review as I needed a good time slot to test this in thorough way.

stefan-hoehn:
We need to wait until @florian-h05 has merged it.

florian-h05 (approved):

stefan-hoehn:
Note to myself: when migrating to 11 check all deprecated methods first, e.g. MutatorIcon.reconnect

PR #2311, opened 2024-02-06
[blockly] Set output type of oh_check_undefined_value to boolean

jimtng:
On the default renderer, it isn't obvious that this is wrong, but on Zelos where generic output (rounded) vs Boolean output (angled) have different looks, it's glaringly obvious.


stefan-hoehn:
@jimtng In general it would be great if in the future you add visuals on blockly topics, so one can see the issue and the fix. thx, 🙏🏻
Thanks for catching this.

Issue #117, opened 2019-09-10
[Q] How to add a rfc2217://.. port in the UI (PaperUI, Habmin)


curlyel:
I'm trying to use rfc2217://...  remote serial ports with the Zigbee ~~and Zwave~~ binding.
When opening the serial port configuration for the dongles, I get a drop down list containing all local serial ports available in the underlaying Debian operation system (physical as well as socat-provided virtual serial ports).

But how to configure (add) a rfc2217 port? The list is not editable (same for PaperUI and Habmin):
![image](https://user-images.githubusercontent.com/11849305/64619422-5a25b100-d3e2-11e9-8eb6-8671f1491a33.png)

This is the platform:
```
Release   = Debian GNU/Linux 9 (stretch)
Kernel    = Linux 4.9.0-8-amd64
Platform  = VMware Virtual Platform
openHAB   = openHAB 2.5.0~M2-1 (Milestone Build)
Java      = Zulu 8.40.0.25-CA-linux64
```

PR #2312, opened 2024-02-06
Provide meaningful list of units in item form

mherwege:
Closes https://github.com/openhab/openhab-core/issues/4082

This PR adds:

1. A curated list of units to show as a drop-down list when creating a Number item with dimension.
2. The possibility to use a different default unit on item creation than the system default unit.
3. The ability to change unit and state description for items.

By default, the system default unit (for the configured measurement system) will be shown when editing or creating an item. [`units.js`](https://github.com/openhab/openhab-webui/blob/7256bc1a3c38d38f16279d7c763d64556ca6e973/bundles/org.openhab.ui/web/src/assets/units.js) contains a number of frequently used units by dimension and measurement system. These will be available in a autosuggest dropdown list. It is still possible to not select from the list and use any other string as a unit. [`units.js`](https://github.com/openhab/openhab-webui/blob/7256bc1a3c38d38f16279d7c763d64556ca6e973/bundles/org.openhab.ui/web/src/assets/units.js) now only contains a relatively small list, but can easily be extended with other frequently used units.
All units for the dimension in `units.js` will be in the dropdown list, but they will be sorted by measurement system. If the measurement system is set to US, imperial units will appear higher in the list. Units that have not explicitely been listed as SI or US will always appear higher.
When typing a unit that is not in the curated list, a longer list will be used for autocompletion that considers allowed prefixes to base units and constructs all combinations.

[`units.js`](https://github.com/openhab/openhab-webui/blob/7256bc1a3c38d38f16279d7c763d64556ca6e973/bundles/org.openhab.ui/web/src/assets/units.js) also contains a field to set a different default unit on item creation than the system default unit.

With https://github.com/openhab/openhab-core/pull/4079, the REST API of channel types would provide a unit hint if defined in the binding channel types.

If such information is available, this PR adds support for this in the UI. When creating an item from a thing channel, it will suggest to use the unit from the unit hint if available. If not available, it will suggest the webui defined default from `units.js`, if not available, the system default unit for the dimension.

For dimensions with different units in SI or US measurement systems, it is possible to provide 2 units separated by comma in binding channel types. The first one will be SI, the second US. The suggested one will depend on the measurement system setting.

See also:

- https://github.com/openhab/openhab-core/issues/4082
- https://github.com/openhab/openhab-core/issues/4077
- https://github.com/openhab/openhab-core/issues/3854
- https://github.com/openhab/openhab-core/pull/4079
- https://github.com/openhab/openhab-core/pull/4078

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/units.js
+++ b/bundles/org.openhab.ui/web/src/assets/units.js
@@ -160,5 +160,5 @@ export const Units = [{
   baseUnitsMetric: ['m³/s', 'm³/min', 'm³/h', 'm³/d']
 }]
 
-export const prefixesMetric = ['y', 'z', 'a', 'f', 'p', 'n', 'µ', 'm', 'c', 'd', 'da', 'h', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
-export const prefixesBinary = ['ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi']
+export const MetricPrefixes = ['y', 'z', 'a', 'f', 'p', 'n', 'µ', 'm', 'c', 'd', 'da', 'h', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
+export const BinaryPrefixes = ['ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi']
--- a/bundles/org.openhab.ui/web/src/assets/units.js
+++ b/bundles/org.openhab.ui/web/src/assets/units.js
@@ -1,67 +1,164 @@
+// Units defines the possible units for UI unit selection. Each unit is defined by one or more fields:
+//   dimension: required field, unit dimension
+//   units: units used in a curated shortlist of units, not specific to SI or Imperial measurement system
+//   unitsSI: units used in a curated shortlist of units, specific to the SI measurement system
+//   unitsUS: units used in a curated shortlist of units, specific to the Imperial measurement system
+//   default: default unit, to be set to override core default unit, not specific to SI or Imperial measurement system
+//   defaultSI: default unit in SI measurement system, to be set to override OH core default SI unit
+//   defaultUS: default unit in Imperial measurement system, to be set to override OH core default Imperial unit
+//   baseUnits: all supported base units that don't allow metric or binary prefixes
+//   baseUnitsMetric: metric base units, the full list of units will include all of these with all metric prefixes
+//   baseUnitsBinary: binary base units, the full list of units will include all of these with all binary prefixes
+// If nothing is defined for an allowed dimension, the UI will fall back on the OH core default unit in the configured
+// measurement system.
+// For dimensions defined, any of the fields can be ommitted. Logical defaults will be used.
+// Units from curated units lists will always be added to the full unit list constructed from baseUnits and prefixes. So it
+// is not necessary to explicitely add what is already in the curated units for the full units list. However, no prefixes
+// will be applied to these. If you want prefixes to be applied, you should add them in the respective baseUnits Array as well.
+
 export const Units = [{
+  dimension: 'Acceleration',
+  baseUnits: ['gₙ'],
+  baseUnitsSI: ['m/s²']
+}, {
+  dimension: 'AmountOfSubstance',
+  baseUnits: ['°dH'],
+  baseUnitsSI: ['mol']
+}, {
   dimension: 'Angle',
-  units: ['°', 'rad']
+  units: ['°', '\'', '"', 'rad']
 }, {
   dimension: 'Area',
   unitsSI: ['m²', 'km²', 'ha'],
-  unitsUS: ['ft²', 'mi²']
+  unitsUS: ['ft²', 'mi²'],
+  baseUnits: ['ca', 'a', 'in²', 'ac'],
+  baseUnitsMetric: ['m²']
 }, {
   dimension: 'DataAmount',
-  units: ['bit', 'B', 'KB', 'KiB', 'MB', 'MiB', 'GB', 'GiB', 'TB', 'TiB']
+  units: ['bit', 'B', 'kB', 'kiB', 'MB', 'MiB', 'GB', 'GiB', 'TB', 'TiB'],
+  baseUnitsMetric: ['bit', 'B', 'o'],
+  baseUnitsBinary: ['bit', 'B', 'o']
 }, {
   dimension: 'DataTransferRate',
-  units: ['bit/s', 'Kbit/s', 'Mbit/s', 'GBit/s']
+  units: ['bit/s', 'kbit/s', 'Mbit/s', 'GBit/s'],
+  baseUnitsMetric: ['bit/s'],
+  baseUnitsBinary: ['bit/s']
 }, {
   dimension: 'Density',
-  units: ['g/l', 'g/m³', 'kg/m³']
+  units: ['g/l', 'g/m³', 'kg/m³'],
+  baseUnits: ['lb/in³'],
+  baseUnitsMetric: ['g/m³', 'g/mm³', 'g/cm³', 'g/dm³', 'g/ml', 'g/cl', 'g/dl', 'g/l']
 }, {
   dimension: 'Dimensionless',
-  units: ['one', '%', 'dB', 'ppm'],
+  units: ['one', '%', 'dB', 'ppm', 'ppb'],
   default: '%'
 }, {
+  dimension: 'ElectricCapcitance',
+  baseUnitsMetric: ['F']
+}, {
   dimension: 'ElectricCharge',
-  units: ['Ah', 'C']
+  units: ['Ah', 'C'],
+  baseUnitsMetric: ['C']
+}, {
+  dimension: 'ElectricConductance',
+  baseUnitsMetric: ['S']
+}, {
+  dimension: 'ElectricConductivity',
+  baseUnitsMetric: ['S/m']
+}, {
+  dimension: 'ElectricCurrent',
+  baseUnitsMetric: ['A']
+}, {
+  dimension: 'ElectricInductance',
+  baseUnitsMetric: ['H']
+}, {
+  dimension: 'ElectricPotential',
+  baseUnitsMetric: ['V']
+}, {
+  dimension: 'ElectricResistance',
+  baseUnitsMetric: ['Ω']
 }, {
   dimension: 'Energy',
-  units: ['kWh', 'Wh', 'VAh', 'varh', 'J', 'kJ', 'cal', 'kcal']
+  units: ['kWh', 'Wh', 'VAh', 'varh', 'J', 'kJ', 'cal', 'kcal'],
+  baseUnitsMetric: ['Ws', 'Wh', 'VAs', 'VAh', 'vars', 'varh', 'J', 'cal']
 }, {
   dimension: 'Force',
-  units: ['N', 'kN']
+  units: ['N', 'kN'],
+  baseUnitsMetric: ['N']
 }, {
   dimension: 'Frequency',
-  units: ['Hz', 'kHz', 'MHz', 'GHz', 'rpm']
+  units: ['Hz', 'kHz', 'MHz', 'GHz', 'rpm'],
+  baseUnitsMetric: ['Hz']
 }, {
   dimension: 'Intensity',
-  units: ['W/m²', 'µW/cm²']
+  units: ['W/m²', 'µW/cm²'],
+  baseUnitsMetric: ['W/mm²', 'W/cm²', 'W/dm²', 'W/m²']
 }, {
   dimension: 'Length',
   unitsSI: ['mm', 'cm', 'dm', 'm', 'km'],
-  unitsUS: ['in', 'ft', 'mi']
+  unitsUS: ['in', 'ft', 'mi'],
+  baseUnits: ['yd', 'ch', 'fur', 'lea']
+}, {
+  dimension: 'LuminousFlux',
+  baseUnitsMetric: ['lm']
+}, {
+  dimension: 'LuminousIntensity',
+  baseUnitsMetric: ['cd']
+}, {
+  dimension: 'MagneticFlux',
+  baseUnitsMetric: ['T']
 }, {
   dimension: 'Mass',
-  unitsSI: ['mg', 'g', 'kg', 't']
+  unitsSI: ['mg', 'g', 'kg', 't'],
+  baseUnits: ['lb', 'oz', 'st'],
+  baseUnitsMetric: ['g', 't']
 }, {
   dimension: 'Power',
-  units: ['W', 'kW', 'VA', 'kVA', 'var', 'kvar', 'dBm']
+  units: ['W', 'kW', 'VA', 'kVA', 'var', 'kvar', 'dBm'],
+  baseUnits: ['hp', 'kgf', 'lbf'],
+  baseUnitsMetric: ['W', 'VA', 'var']
 }, {
   dimension: 'Pressure',
   unitsSI: ['Pa', 'hPa', 'bar', 'mbar', 'mmHg'],
-  unitsUS: ['inHg', 'psi']
+  unitsUS: ['inHg', 'psi'],
+  baseUnits: ['atm'],
+  baseUnitsMetric: ['Pa', 'bar']
+}, {
+  dimension: 'RadiationAbsorbedDose',
+  baseUnitsMetric: ['Gy']
+}, {
+  dimension: 'RadiationEffectiveDose',
+  baseUnitsMetric: ['Sv']
+}, {
+  dimension: 'Radioactivity',
+  unitsSI: ['Bq', 'Ci'],
+  baseUnitsMetric: ['Ci']
 }, {
   dimension: 'Speed',
   unitsSI: ['km/h', 'm/s'],
-  unitsUS: ['mph', 'in/h']
+  unitsUS: ['mph', 'in/h'],
+  baseUnits: ['kn'],
+  baseUnitsMetric: ['m/s', 'm/h']
 }, {
   dimension: 'Temperature',
   unitsSI: ['°C', 'K'],
-  unitsUS: ['°F', 'K']
+  unitsUS: ['°F', 'K'],
+  baseUnits: ['mired']
 }, {
   dimension: 'Time',
   units: ['s', 'min', 'h', 'd', 'wk', 'mo', 'y']
 }, {
   dimension: 'Volume',
-  unitsSI: ['ml', 'cl', 'l', 'm³']
+  unitsSI: ['ml', 'cl', 'l', 'm³'],
+  unitsUS: ['gal'],
+  baseUnits: ['in³', 'ft³'],
+  baseUnitsMetric: ['l', 'm³']
 }, {
   dimension: 'VolumetricFlowRate',
-  unitsSI: ['l/min', 'm³/s', 'm³/min', 'm³/h', 'm³/d']
+  unitsSI: ['l/s', 'l/min', 'm³/s', 'm³/min', 'm³/h', 'm³/d'],
+  unitsUS: ['gal/min'],
+  baseUnitsMetric: ['m³/s', 'm³/min', 'm³/h', 'm³/d']
 }]
+
+export const prefixesMetric = ['y', 'z', 'a', 'f', 'p', 'n', 'µ', 'm', 'c', 'd', 'da', 'h', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
+export const prefixesBinary = ['ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi']
--- a/bundles/org.openhab.ui/web/src/assets/units.js
+++ b/bundles/org.openhab.ui/web/src/assets/units.js
@@ -1,21 +1,28 @@
-// Units defines the possible units for UI unit selection. Each unit is defined by one or more fields:
-//   dimension: required field, unit dimension
-//   units: units used in a curated shortlist of units, not specific to SI or Imperial measurement system
-//   unitsSI: units used in a curated shortlist of units, specific to the SI measurement system
-//   unitsUS: units used in a curated shortlist of units, specific to the Imperial measurement system
-//   default: default unit, to be set to override core default unit, not specific to SI or Imperial measurement system
-//   defaultSI: default unit in SI measurement system, to be set to override OH core default SI unit
-//   defaultUS: default unit in Imperial measurement system, to be set to override OH core default Imperial unit
-//   baseUnits: all supported base units that don't allow metric or binary prefixes
-//   baseUnitsMetric: metric base units, the full list of units will include all of these with all metric prefixes
-//   baseUnitsBinary: binary base units, the full list of units will include all of these with all binary prefixes
-// If nothing is defined for an allowed dimension, the UI will fall back on the OH core default unit in the configured
-// measurement system.
-// For dimensions defined, any of the fields can be ommitted. Logical defaults will be used.
-// Units from curated units lists will always be added to the full unit list constructed from baseUnits and prefixes. So it
-// is not necessary to explicitely add what is already in the curated units for the full units list. However, no prefixes
-// will be applied to these. If you want prefixes to be applied, you should add them in the respective baseUnits Array as well.
+// Units defines the possible units for UI unit selection.
+// If nothing is defined for an allowed dimension, the UI will fall back on the OH core default unit in the configured measurement system.
+// For dimensions defined, any of the fields can be omitted. Logical defaults will be used.
+// Units from curated units lists will always be added to the full unit list constructed from baseUnits and prefixes.
+// So it is not necessary to explicitly add what is already in the curated units for the full units list.
+// However, no prefixes will be applied to these. If you want prefixes to be applied, you should add them in the respective baseUnits Array as well.
 
+/**
+ * @typedef Unit
+ * @property {string} dimension unit dimension (required)
+ * @property {string[]} [units] units used in a curated shortlist of units, not specific to SI or Imperial measurement system
+ * @property {string[]} [unitsSI] units used in a curated shortlist of units, specific to the SI measurement system
+ * @property {string[]} [unitsUS] units used in a curated shortlist of units, specific to the Imperial measurement system
+ * @property {string} [default] default unit, to be set to override core default unit, not specific to SI or Imperial measurement system
+ * @property {string} [defaultsSI] default unit in SI measurement system, to be set to override OH core default SI unit
+ * @property {string} [defaultUS] default unit in Imperial measurement system, to be set to override OH core default Imperial unit
+ * @property {string[]} [baseUnits] all supported base units that don't allow metric or binary prefixes
+ * @property {string[]} [baseUnitsMetric] metric base units, the full list of units will include all of these with all metric prefixes
+ * @property {string[]} [baseUnitsBinary] binary base units, the full list of units will include all of these with all binary prefixes
+ */
+
+/**
+ * Defines the possible units for UI unit selection.
+ * @type {Unit[]}
+ */
 export const Units = [{
   dimension: 'Acceleration',
   baseUnits: ['gₙ'],
@@ -160,5 +167,14 @@ export const Units = [{
   baseUnitsMetric: ['m³/s', 'm³/min', 'm³/h', 'm³/d']
 }]
 
+/**
+ * Metric prefixes for metric base units.
+ * @type {string[]}
+ */
 export const MetricPrefixes = ['y', 'z', 'a', 'f', 'p', 'n', 'µ', 'm', 'c', 'd', 'da', 'h', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
+
+/**
+ * Binary prefixes for binary base units.
+ * @type {string[]}
+ */
 export const BinaryPrefixes = ['ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi']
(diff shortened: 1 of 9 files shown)


andrewfg:
@mherwege I can’t really comment on the code; but I will make a build and test how it looks on screen.

andrewfg:
> I will make a build and test how it looks on screen

Ok. I did a full build and found three problems..

1. Item Creation: Initially when I first tried to create an item attached to a channel with unitHint, the 'Unit' and 'State Description Pattern' fields could not be selected, and thus could not be edited i.e. they were read only. But after some time playing around with creating and deleting items, when I tried to create an item again, the fields became editable once more. I was using the Chrome browser so I wonder if it may have been related to the recently posted browser cache issue -- but perhaps not. I tried it also in Edge and did not see the same issue.

image

2. Item Editing: When editing the (already created) item the respective input page does not display the 'Unit' or 'State Description Pattern' fields. So to change an already existing item, the only solution is to delete and recreate the item.

image

3. For the respective channel item shown in the screenshots above, I had set unitHint="dB" in the respective binding, however the input screen in the browser showed 'one'..

andrewfg:
^
Apropos point 3. above, I can see the "unitHint=dB" in the REST API JSON provided by OH core.
So in other words, the issue seems to be in openhab-webui rather than openhab-core..

```
  {
    "parameters": [],
    "parameterGroups": [],
    "description": "Quality of Service: percentage of configured devices currently connected to the RF mesh network",
    "label": "Mesh Network QoS",
    "itemType": "Number:Dimensionless",
    "unitHint": "dB",
    "kind": "STATE",
    "stateDescription": {
      "pattern": "%.0f %%",
      "readOnly": true,
      "options": []
    },
    "tags": [],
    "UID": "neohub:meshNetworkQoS",
    "advanced": false
  },
```

mherwege:
@andrewfg Thanks for testing. Issues are noted and I will look into them.

I am currently working on integrating a currated list of units in this PR as well. Let's come back on this and test again when I have done that.

mherwege:
@andrewfg @jimtng I have updated the scope of this PR and changed the original description.

mherwege:
@andrewfg If you want to have another go at testing this, that would be very much appreciated.

jimtng:
For Time, is it `wk` or `week`, and `mo` is not listed on https://www.openhab.org/docs/concepts/units-of-measurement.html

For Temperature, can someone in SI still pick `°F`?
An idea for consideration: instead of excluding the "other" units, perhaps put them _after_ the preferred units, so for people using SI, they'll get SI units first, followed by US units, and vice versa...?

mherwege:
> For Time, is it `wk` or `week`, and `mo` is not listed on https://www.openhab.org/docs/concepts/units-of-measurement.html

I based it on: https://github.com/openhab/openhab-core/blob/6be00bd5855d42d6a83b7d3e09cc19811fb0f2d7/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java#L318

I don't know if `week` is supported, it probably is. `wk` definitely is.

mherwege:
> For Temperature, can someone in SI still pick `°F`?

It won't be in the suggestion list, but it is always possible to put it in manually.

mherwege:
> An idea for consideration: instead of excluding the "other" units, perhaps put them _after_ the preferred units, so for people using SI, they'll get SI units first, followed by US units, and vice versa...?

Possible, but from a perspective of a European, I don't want to see the Imperial units. I don't deal with them at all. That perspective could be different for users in the US though.

mherwege:
> From usability perspective, when it's not there, _some_ people might think it's not possible.

Agree, but I don't see a good way to overcome it. The issue is there are too many combinations possible for units and their multipliers. Showing them all does not make sense. And if you don't show all possible combinations, the proposed list should be relevant with the possiblity to put something else if needed. That's what I try to achieve.

andrewfg:
> If you want to have another go at testing this, that would be very much appreciated.

Ok. I just started building it now..

andrewfg:
> They might need to use metric sometimes, and at that time, having the hints would probably be quite helpful. I'm happy either way

Please just remember the poor old UK .. 

.. unlike the US which uses 'Imperial' (I always find it odd that the US refers to its units by reference its one-time imperial over-lords) .. the UK has mostly adopted SI and dropped 'Imperial' (Brexit notwithstanding) .. EXCEPT in a couple of peculiar special cases -- namely as follows..
1. On the road: where distance and speed are measured in miles, fuel is sold in litres, and consumption is measured in miles-per-gallon (sic) .. and note that UK 'Imperial' gallons are different from US 'Imperial' gallons .. go figure ..
2. In the pub: where beer is sold in pints..

=> Ergo for UK users, presenting both units, would probably make sense..

mherwege:
> 1. UK 'Imperial' gallons are different from US 'Imperial' gallons

Does that mean standard conversion to a metric unit does not work, as different for US and UK?

mherwege:
> Please just remember the poor old UK ..

So, what does a UK user set as measurement system, metric or imperial?

mherwege:
OK, so it sounds like it may make sense to show both, but sequence switched depending on measurement system.

andrewfg:
> Does that mean standard conversion to a metric unit does not work, as different for US and UK?

Ummm. It gets complicated. See below. .. Please, please, please don't ask me why

- The British Imperial fluid ounce is equal to 28.413 milliliters, while the US Customary fluid ounce is 29.573 ml.
- The British Imperial pint is 568.261 ml (20 fluid ounces), while the US Customary pint is 473.176 ml (16 fl oz).
- The British Imperial quart is 1.13 liters (40 fl oz), while the US Customary quart is 0.94 L (32 fl oz).
- The British Imperial gallon is 4.54 L (160 fl oz), while the US Customary gallon is 3.78 L (128 fl oz).

mherwege:
> Ummm. It gets complicated. See below. .. Please, please, please don't ask me why
> 
> * The British Imperial fluid ounce is equal to 28.413 milliliters, while the US Customary fluid ounce is 29.573 ml.
> * The British Imperial pint is 568.261 ml (20 fluid ounces), while the US Customary pint is 473.176 ml (16 fl oz).
> * The British Imperial quart is 1.13 liters (40 fl oz), while the US Customary quart is 0.94 L (32 fl oz).
> * The British Imperial gallon is 4.54 L (160 fl oz), while the US Customary gallon is 3.78 L (128 fl oz).

I won't ask. I was just wondering what conversion OH is using if you convert between the units using QuantityTypes. This can get messy fast. What if you put in volume liters for how much you put in your car tank (so an item with unit liter), have the distance that your car travelled in miles (an item with unit miles) and a rule that calculats consumption (distance/volume) and show it in miles/gallon. What conversion will OH use for this? US? British? If it is US, the system doesn't work for UK, or? And to me that would be an argument **not** to show units from the other measurement system.

rkoshak:
> That perspective could be different for users in the US though.

People in the US *do* use metric almost day to day. It's really just that the government decided it wasn't worth the cost to force it's use. So we buy our milk in gallons but buy our Cokes in two liter bottles. Our cars travel miles but the bolts that hold them together are often in mm. It often depends on context which  type of unit is used. But the units are kind of unofficially standardized. You won't see milk sold in liters nor Cokes sold in gallons. So we are not very good at switching between the two for the same thing. Forecast that it's going to be -3 °C and people will be confused. 

Anyway, that's a long way of saying showing both for US users wouldn't be a problem and in some rare cases the metric will be preferred over the Imperial unit. We're not that far off from the UK in that regard.

Anyway, back to why I as tagged...

> From usability perspective, when it's not there, some people might think it's not possible.

I understand both sides here. There are *far* too many units to present in one list. OTH, it may not be clear that in cases where the unit isn't in the list that the user can just type it in. I've seen people become confused by this in Item Triggers in Rules. They see this

image


And become confused about why there's that ON/OFF toggle, can an Item only trigger a rule with ON and OFF?

So in the description we need to make it absolutely clear that the field is editable and to edit it when the desired unit isn't in the list. If there are hints in the description of the field as well as some UI hints (e.g. a blinking cursor in the field when it's clicked on perhaps) that may be as good as we can go it. 

This is a little bit out there, but would it make sense to create a "unit builder" akin to the cron builder? Then instead of a giant list, the user can "build" the unit. They could select the Dimension, then select from the list of those units the Dimension supports (that list at least will be short) and finally select a scale modifier, if supported. 

So to get to `kWh` for example the user would select Energy for the Dimension, `Wh` for the unit from the list of only those units that Energy supports (J, Ws, Wm, Wh, VAh, varh, and cal) and then `kilo` for the scale modifier. For density and speed type units there can be an optional "per" field (e.g. Bq/m³, mph, etc.). It's just a half thought out idea. I was just thinking that if we could split the list so the user doesn't have to sort through everything all at once it might be more workable to present most of the units or at least a bigger set of the units. It's probably significantly more work though.

jimtng:
If understand correctly, that's the whole point of this PR.
When "Length" dimension is selected, users will be given the list of _length_ units only. So this may not be as long as you may think. Plus this PR is about _curating_ only the common units, so it's even shorter than the list of _all_ possible units.

mherwege:
> Anyway, back to why I as tagged...

I tagged a specific post, but actually am happy you comment on the overall discussion. Thank you.


> This is a little bit out there, but would it make sense to create a "unit builder" akin to the cron builder?

Yes, definitely a lot more work, and not it makes selecting a unit a longer process as well. There suddenly is another screen with multiple parameters to set. I am not prepared to try to do that with this PR.

> If understand correctly, that's the whole point of this PR.
> When "Length" dimension is selected, users will be given the list of _length_ units only. So this may not be as long as you may think. Plus this PR is about _curating_ only the common units, so it's even shorter than the list of _all_ possible units.

I am afraid it is longer than you think. If you look at @rkoshak example, you have a set of base units (J, Ws, Wm, Wh, VAh, varh, and cal), but these are not even exhaustive. Notice Ws, Wh, VAh, ... are already multiplied units. In theory, other combinations would be possible. And then there are all the possible modifiers in front.
The idea of the PR is to give a reasonable list, not a full list. Notice I have not even attempted to provide units for everything.

> So in the description we need to make it absolutely clear that the field is editable and to edit it when the desired unit isn't in the list. If there are hints in the description of the field as well as some UI hints (e.g. a blinking cursor in the field when it's clicked on perhaps) that may be as good as we can go it.

We can't put a `Any` placeholder, because it will always be filled with a default already. But I can extend the info text to try to make it clearer.

rkoshak:
> Yes, definitely a lot more work, and note it makes selecting a unit a longer process as well. There suddenly is another screen with multiple parameters to set.

But not longer for those who just know what to type in. 

Anyway, it was just an idea.

> are already multiplied units. In theory, other combinations would be possible. And then there are all the possible modifiers in front.

I wasn't trying to suggest that the builder be exhaustive to support every possible combination of every possible unit. There can be some really complicated units for sure (Bq/m³ would probably be among them and was probably a bad example to list above).

I was just thinking that `kilo` is `k` whether you are talking about `b` or `Wh` or `l`. If you made setting that part separate you wouldn't have to include them in a list. Complicated stuff like `Ws` vs. `Wm`, well those are listed separately in the docs already so just treat them as separate units in the tool.

But again, it was just an idea, not even a request.

andrewfg:
^
The current behaviour is that OH offers _for the respective dimension_ **just one single choice of unit only** -- namely the 'default' unit. But yes you can override that one single choice by manually typing in something else.

The purpose of this PR is to offer _for the respective dimension_ a curated **selection** of units. And yes you can choose something else by manually typing it in.

andrewfg:
> If you want to have another go at testing this, that would be very much appreciated.

@mherwege I had problems with git (or something) today so my build failed. I will wait until tomorrow and resynch and try again.

ghys:
Enjoying the debate, just chiming in on this:

> I've seen people become confused by this in Item Triggers in Rules. They see this
> image
> And become confused about why there's that ON/OFF toggle, can an Item only trigger a rule with ON and OFF?

I'll admit it wasn't my best design, the intent was to present the actual input control, and then below it quick & easy suggestions for the value. So the suggestions are just that and don't prevent you from inputting whatever you want in the textbox.

> This is a little bit out there, but would it make sense to create a "unit builder" akin to the cron builder? Then instead of a giant list, the user can "build" the unit. [...]
So to get to kWh for example the user would select Energy for the Dimension, Wh for the unit from the list of only those units that Energy supports (J, Ws, Wm, Wh, VAh, varh, and cal) and then kilo for the scale modifier.

Certainly interesting, as long as it's not a pain to use. If you want your energy measurement in Wh or kWh you should just have to type "Wh" or "kWh" into a textbox and be done with it without dealing with a "unit builder" UI. The cron builder allows that so I would expect an unit builder would too.

mherwege:
I don't feel comfortable creating a real unit builder.

But an alternative could be the following:
- When creating an item, the unit field will default as in the current PR state (unitHint, default from UI default, system default).
- Dropdown (autosuggest) on the field shows the currated list.
- Start typing, and you get all possible combinations for completion. E.g. if you are in an Energy dimension item, start typing `k`, you will get as suggestion `kWs`, `kWh`, `kVA`, `kvar`, `kJ`, `kcal` (and possibly others). This would not be restricted to the currated list.

What is needed for the last part, is code to create a list of valid options (beyond the currated one), and as complete as possible. Would there be any js library that could be included for that? Anyway, if someone wants to build a unit builder, something similar would be required.

andrewfg:
@mherwege two things..

### 1. Test Results

TLDR summary: LGTM when creating a new item for a channel; but not working when editing that same (already created) item.

**A. Create New Item for Channel**

- Type => Ok
- Dimension => Ok
- Unit / Hint / when xml 'unitHint' exists => Ok
- Unit / Hint / when xml 'unitHint' missing i.e. takes default unit => Ok
- Unit / Curated drop down => Ok
- State Description Pattern => Ok

**B. Edit Already Created Item (for Channel)**

- Type => Ok
- Dimension => Ok
- Unit => Accepts only manual input. **i.e. neither prior value, unitHint, default unit, nor drop down**
- State Description Pattern => Accepts only manual input **i.e. no prior value**

image

### 2. Thoughts Re: "Unit Builder"

@mherwege personally I think your **four** level functionality (unit hint + default unit + curated dropdown + manual entry) is fully sufficient (so long as it is also applied when editing an already created item). I also support your wish not to (try to) devise a "unit builder". I also think that a _filter-as-you-type_ would be incredibly difficult to implement.

=> I suggest to start with the current four level functionality (which is already a **three level improvement** versus the current approach). And if someone wishes it, they can open a future Issue for an eventual further refinement, in a second step.

mherwege:
> I also think that a _filter-as-you-type_ would be incredibly difficult to implement.

It actually isn't all that difficult, working on it.

Thanks for the bug report. I will look into this.

andrewfg:
> code to create a list of valid options (beyond the currated one)

@mherwege did you look at [this](https://github.com/unitsofmeasurement/indriya/blob/67e5e6974ee543d0665dd82456ba37f073adac55/src/main/java/tech/units/indriya/AbstractSystemOfUnits.java#L86)

rkoshak:
>  the intent was to present the actual input control,

Yes and that intent was clear to me but it's tripped up some users. I only brought it up here to make sure we take that into consideration for units.

> If you want your energy measurement in Wh or kWh you should just have to type "Wh" or "kWh" into a textbox and be done with it without dealing with a "unit builder" UI. 

That was part of the idea from the start. No matter how this gets implemented the user needs to be able to just type in what they want and they have to be able to type in units not already known. The UoM capability allows one to combine units in cool and interesting ways that cannot always be anticipated but OH should support.

mherwege:
> @mherwege did you look at [this](https://github.com/unitsofmeasurement/indriya/blob/67e5e6974ee543d0665dd82456ba37f073adac55/src/main/java/tech/units/indriya/AbstractSystemOfUnits.java#L86)

Yes, I did, but it doesn't help much. It is partial and not available from the REST API.

bundles/org.openhab.ui/web/src/assets/units.js
```diff
@@ -0,0 +1,164 @@
...
+  baseUnitsMetric: ['H']
+}, {
+  dimension: 'ElectricPotential',
+  baseUnitsMetric: ['V']
+}, {
+  dimension: 'ElectricResistance',
+  baseUnitsMetric: ['Ω']
+}, {
+  dimension: 'Energy',
+  units: ['kWh', 'Wh', 'VAh', 'varh', 'J', 'kJ', 'cal', 'kcal'],
```
andrewfg, line 82: IMHO these curated entries should be `kVAh` and `kvarh` -- with `k` prefix rather than without; by analogy to line #117 below..

bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -23,11 +23,11 @@
         </select>
       </f7-list-item>
       <!-- Use v-show instead of v-if, because otherwise the autocomplete for category would take over the unit -->
-      <f7-list-input v-show="!hideType && item.type && item.type.startsWith('Number:') && createMode" label="Unit" type="text" :value="item.unit"
-                     info="Used internally, for persistence and external systems. It is independent from the state visualization in the UI, which is defined through the state description."
+      <f7-list-input v-show="!hideType && item.type && item.type.startsWith('Number:')" ref="unit" label="Unit" autocomplete="off" type="text" :value="item.unit"
+                     :info="(createMode) ? 'Type any valid unit or select from one of the proposed (non-exhaustive list of) units. Used internally, for persistence and external systems. It is independent from the state visualization in the UI, which is defined through the state description.' : ''"
```
andrewfg, line 27: Any reason why this applies only to createMode and not to editMode?

bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -23,11 +23,11 @@
...
                      @input="item.unit = $event.target.value" :disabled="!editable" :clear-button="editable" />
-      <f7-list-input v-show="!hideType && item.type && item.type.startsWith('Number:') && createMode" label="State Description Pattern" type="text" :value="item.stateDescriptionPattern"
-                     info="Pattern or transformation applied to the state for display purposes. Only saved if you change the pre-filled default value."
+      <f7-list-input v-show="!hideType && item.type && item.type.startsWith('Number:')" label="State Description Pattern" type="text" :value="item.stateDescriptionPattern"
+                     :info="(createMode) ? 'Pattern or transformation applied to the state for display purposes. Only saved if you change the pre-filled default value.' : ''"
```
andrewfg, line 30: Ditto above.

andrewfg:
@mherwege / @jimtng for info: please note that the two PRs #2312 and #2326 are working on some of the same forms, and probably the overlap is more than can easily resolved via a regular Git merge

bundles/org.openhab.ui/web/src/components/item/item-form.vue
mherwege: We are still looking for a way to have these info messages hidden behind a question mar link when not in create mode, see also @jimtng PR. It makes the screen less cluttered and is mostly relevant when creating.

bundles/org.openhab.ui/web/src/assets/units.js
mherwege: fixed

mherwege:
@andrewfg The issues you discovered should have been fixed now.
I also slightly changed the autocomplete list: as long as what you type has a match with the curated unit list (including unitHint), show those matching units. When not, show the matching units from the full list.

andrewfg:
> The issues you discovered should have been fixed now.

@mherwege Ok, I will do another build and test this afternoon.

andrewfg:
@mherwege I did some tests. Generally it looks great. Many thanks! However I did find a couple of errors..

1. Create Mode: On clean build system start-up, initially the `Unit` and `State Description Pattern` fields are read only (i.e. they cannot be edited). Editing the `Type` or `Dimension` field does NOT change the fields to become editable. However editing the `Category` field DOES make the fields become editable. Once the form has become editable, it remains editable even when creating yet another new Item. (i.e. I get the feeling that you probably only experience this issue once, on the first time the form is shown, on a clean new OH installation..)

image

2. Create / Edit Mode: the `Unit` field always accepts a manually typed a unit that is incompatible with the selected dimension => I wonder if it is possible to indicate such a mismatch by (say) changing the text color of the unit?

image

mherwege:
> 2\. I wonder if it is possible to indicate such a mismatch by (say) changing the text color of the unit?

Even the full autosuggest list may not be exhaustive. Giving something that is not in the full list another color is probably possible, but it is not always an error. I don’t have a mechanism to validate a unit is valid.

> 1. Create Mode: On clean build system start-up, initially the `Unit` and `State Description Pattern` fields are read only

I need to see if I can reproduce this in a development environment. I have little time at the moment, so may take a few days.

mherwege:
Rebased, but not retested yet. As there were many overlapping changes with https://github.com/openhab/openhab-webui/pull/2326, this needs careful retesting still.

florian-h05:
@mherwege Please let me know once you are finished with testing and I should start review.
Does this PR depend on https://github.com/openhab/openhab-core/pull/4079?

andrewfg:
> Does this PR depend on https://github.com/openhab/openhab-core/pull/4079?

Yes.

andrewfg:
> Rebased, but not retested yet. As there were many overlapping changes with https://github.com/openhab/openhab-webui/pull/2326, this needs careful retesting still.

Ok @mherwege I am building it now.

mherwege:
> Ok @mherwege I am building it now.

I am afraid there are quite a few issues now. I will need more time to resolve.

andrewfg:
@mherwege just for the avoidance of doubt, I would test the Item form in the following test cases..

- Create new Item from an existing Thing Channel
- Edit Item that is linked to existing Thing Channel
- Create new non-linked Item
- Edit non-linked Item
- .. do above for Item type Number / any other

mherwege:
@jimtng Maybe you want to have a look at this as well. Also make sure I didn't destroy any of your work with this PR by accident.

andrewfg:
> @andrewfg This should be ready for a test again

Ok. I should have some time tomorrow..

andrewfg:
Ok. Here are some initial tests..

1. On initial start up in a clean built and clean installed version of OH -- in the Chrome browser -- the Unit field is READ ONLY. However in Edge browser the unit field can be edited. And once an item has been created, the Unit field can once more be edited.

image

2. If you start to type something, then the drop down selection does show the typed text in bold (e.g. 'pp'). However if you change your mind and click on something else (e.g. '%') then the newly clicked value '%' does not overwrite the already typed value 'pp'. => You have to backspace to delete the already types 'pp' before you can select a new element from the dropdown.

image

3. Once the dropdown has been dropped down, it is impossible to close it. Pressing neither 'tab' nor 'return' has any effect..

image

Unfortunately because of the above item 3. I could not test anything else..

mherwege:
@andrewfg Indeed, rebased and tried on a new install and I now have the same issues. I had none of these issues in my development environment. This has been a moving target, and everytime I do something, it looks like there are other changes that may or may not impact this.
I need to find some time to figure this out.

mherwege:
@andrewfg There was an error in the rebase. I changed a few things and I cannot reproduce the problems you saw anymore.

@florian-h05 I do need your help.
On the item form, the first time using the unit dropdown, and the first time after changing the dimension, I get an error in the console:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'find')
    at eval (autocomplete-class.js:100:24)
    at Autocomplete.source (item-form.vue:221:1)
    at HTMLInputElement.onInputChange (autocomplete-class.js:70:24)
    at HTMLInputElement.handleEvent (dom7.module.js:333:14)
    at Dom7.trigger (dom7.module.js:442:10)
    at Autocomplete.onOpen (autocomplete-class.js:648:19)
    at Autocomplete.openDropdown (autocomplete-class.js:825:8)
    at Autocomplete.open (autocomplete-class.js:836:19)
    at HTMLInputElement.onInputFocus (autocomplete-class.js:187:10)
    at HTMLInputElement.handleEvent (dom7.module.js:333:14)
```
It looks like everything is working, but there might still be something. I can't figure out what the problem is though.
This is the line in `autocomplete-class.js` that is causing the issue:
```
ac.$dropdownEl.find('ul').html(itemsHTML);
```
That seems to suggest the dropdown does not exist yet, but it does open. So I am at a loss.
I have been testing this in Chrome.

andrewfg:
> cannot reproduce the problems you saw anymore.

Ok, I will test again. And thank you for your great efforts.

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
```diff
@@ -197,9 +199,14 @@ export default {
         category: (this.channelType) ? this.channelType.category : '',
         groupNames: [],
         type: this.channel.itemType || 'Switch',
+        unit: this.linkUnit(),
         tags: (defaultTags.find((t) => this.$store.getters.semanticClasses.Points.indexOf(t) >= 0)) ? defaultTags : [...defaultTags, 'Point']
       })
     },
+    linkUnit () {
+      const dimension = this.channel.itemType.startsWith('Number:') ? this.dimensions.find(d => d.name === this.channel.itemType.split(':')[1]) : ''
```
florian-h05, line 207: Is it really required to search the dimension in the list if we only need the dimension name?

```suggestion
      const dimension = this.channel.itemType.startsWith('Number:') ? this.channel.itemType.split(':')[1] : ''
```

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
```diff
@@ -170,6 +170,25 @@ export default {
       if (this.validateItemName(this.item.name) !== '') return this.$f7.dialog.alert('Please give the Item a valid name: ' + this.validateItemName(this.item.name)).open()
       if (!this.item.type || !this.types.ItemTypes.includes(this.item.type.split(':')[0])) return this.$f7.dialog.alert('Please give Item a valid type').open()
 
+      const dimensionChange = this.$refs.itemForm.dimensionChanged()
+      const unitChange = this.$refs.itemForm.unitChanged()
+      if (dimensionChange || unitChange) {
+        const title = 'WARNING: ' + (dimensionChange ? 'Dimension' : 'Unit') + ' Changed'
+        const text = dimensionChange ? 'Existing links to channels with dimension may no longer be valid!' : 'Changing the internal unit can corrupt your persisted data!'
```
florian-h05, line 177:
```suggestion
        const text = dimensionChange ? 'Existing links to channels with dimension may no longer be valid!' : 'Changing the internal unit can corrupt your persisted data and affect rules!'
```

Do you agree?

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
```diff
@@ -170,6 +170,25 @@ export default {
...
+            { text: 'Cancel', color: 'gray', close: true, onClick: () => this.$refs.itemForm.revertDimensionChange() },
+            { text: 'Save Anyway', color: 'red', close: true, onClick: () => this.saveConfirmed() }
+          ],
+          destroyOnClose: true
+        }).open()
+      } else {
+        this.saveConfirmed()
+      }
+    },
+    saveConfirmed () {
```
florian-h05, line 191:
```suggestion
    doSave () {
```

bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -162,7 +245,7 @@ export default {
     },
     onNameInput (event) {
       this.item.name = event.target.value
-      this.$set(this, 'nameErrorMessage', this.validateItemName(this.item.name))
+      this.thisnameErrorMessage = this.validateItemName(this.item.name)
```
florian-h05, line 248: Typo:

```suggestion
      this.nameErrorMessage = this.validateItemName(this.item.name)
```

bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -174,14 +257,27 @@ export default {
   },
   mounted () {
     if (!this.item) return
-    const categoryControl = this.$refs.category
-    if (!categoryControl || !categoryControl.$el) return
-    const inputElement = this.$$(categoryControl.$el).find('input')
-    this.initializeAutocomplete(inputElement)
+    if (!this.item.category) this.$set(this.item, 'category', '')
```
florian-h05, line 260: Please do not do create these props here, this causes issues with dirty handling.

Either make the form work if those are not available (I think it does work without them) or set them during load time in the components using the form.
However before setting something in these components, rebase. IIRC I already pushed a fix for that, see #2408.

florian-h05 (changes requested):
So, I'm finished with the first review round, please have a look at my comments.
I will see if I can help you with the auto-complete error.

florian-h05:
@mherwege I have created https://github.com/mherwege/openhab-webui/pull/2 which addresses some of my comments (I will mark them as resolved) and fixes a few issues.
Unfortunately I haven't been able to fix the autocomplete log level.

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
mherwege: Agreed

mherwege:
> @mherwege I have created [mherwege#2](https://github.com/mherwege/openhab-webui/pull/2) which addresses some of my comments (I will mark them as resolved) and fixes a few issues.

I merged your PR in my branch on github, but couldn't get it to my local anymore afterwards. So I reapplied your changes and force pushed. All should be in there.

I went through your other comments, and I believe I have tackled all of them.

florian-h05:
> So I reapplied your changes and force pushed. All should be in there.

I have checked the commits, looks good :+1: 

> I went through your other comments, and I believe I have tackled all of them.

I am currently re-reviewing, I already found a few minor issue, I think I will fix them myself (just push a new commit to your branch).

florian-h05 (approved):
LGTM, thanks!!

I have applied a few fixes and JSDoc improvements, you can have a look at my recent commits (they are all divided into separate commits).

As this PR is a major improvement to the configuration of UoM Items even without the unit hint being not yet available , I will merge this PR.
Regarding the units.js asset, I am fine with having this file in the UI. In the past, we also had the semantic tags hard-coded and translated inside the UI, which are now loaded from REST.

florian-h05:
And with regards to the error message of the unit autocompletion: As the autocompletion works anyway, I would propose to merge this PR and then try to fix that error message later. No need to make this PR even bigger ;-)

mherwege:
Thank you for the review and merge.

PR #2313, opened 2024-02-06
Config sheet: Show advanced params if not default or not empty

florian-h05:
Advanced parameters should be shown to the user if they have:

- a default value defined and the value is not the default value
- no default value defined and they have a value set

This is more transparent to the user and improves the UX, because the user does not need to enable "show advanced" just to check whether he has set up advanced parameters.


stefan-hoehn:
As explained to Florian 

"This is more of a behavioral change than a visual one ... The point is that if, for example, you use the API security configuration and change an advanced setting there, it should then always be displayed without having to activate "Show Advanced". So as soon as an "advanced" setting is not at its default value, it is always displayed, even if "Show Advanced" is not activated."

stefan-hoehn (approved):
LGTM

PR #2316, opened 2024-02-07
[blockly] Fix warning when show code button is clicked twice

jimtng:
Avoid this warning popup when clicking the show code button when already in show code mode.

image


florian-h05 (approved):
LGTM, thanks!

PR #2318, opened 2024-02-07
Fix fields should be editable for a new script

jimtng:
Regression from #2276 which caused a new rule unable to be created because the rule name/description couldn't be entered (fields are disabled)


florian-h05 (commented):
LGTM, thanks!!

I wonder how no one noticed that regression before. 
I have created https://github.com/jimtng/openhab-webui/pull/1 which fixes a regression in the script general settings, which has the same PR as cause.

florian-h05 (approved):

PR #2320, opened 2024-02-07
Model picker popup: Add "Show name" and "Show tags" options

jimtng:
This makes it easier to find items e.g. when building rules


florian-h05 (approved):
LGTM, thanks!

I am not very happy with the styling though, in the popup the styling issue IMO is more obvious than on the model page, but that is not related to your PR.

PR #2321, opened 2024-02-07
Fix code editor's Ruby detection

jimtng:
This fixes the syntax highlighting for Ruby because it's now recognised correctly.


florian-h05 (approved):
LGTM, thanks!

PR #2325, opened 2024-02-08
Item edit: Fix semantic class cleared when property set to None

jimtng:
On the Item Editor:

Current behaviour, when we have:
Semantic Class: Control
Property: Light

When changing Property to None, it would clear Semantic Class "Control" too.

The Fix is not to clear the Semantic Class.


florian-h05 (approved):
LGTM, thanks!

PR #2328, opened 2024-02-09
Item details: Fix group member selection allows to select unmanaged Items

florian-h05:
This lead to an error being thrown because openHAB core does not allow to modify group membership of unmanaged Items and returns a 405.


florian-h05:
To reproduce this error, create a few Items by file, e.g.:

```
Switch test_switch
Number:Temperature test_temperature
```

Create a group from the UI or by file, then change direct group members and select `test_switch` and `test_temperature`. Apply this change by confirming the following dialogs and you will receiver an error.

stefan-hoehn (approved):
straight forward and clear. LGTM.

PR #2332, opened 2024-02-09
Settings page: Improve loading behaviour

### LINKED ISSUE #2323: Settings page is slow to load
When clicking the settings page, it sits as an empty page for quite a while, enough to make me think that something is wrong and click away.

Expected behaviour:
- It loads static contents that it can show immediately (Configuration, Automation, System Settings)
- It would load the further details asynchronously afterwards 
  - fill in the number of things, items, etc. Showing `...` during loading instead of the count.
  - Populate the "Addon Settings" menu
Perhaps showing a message `Loading addon settings...`

florian-h05:
Fixes #2323.


digitaldan:
Thank you for fixing this !  Was driving me nuts !

florian-h05:
Yeah, that probably was one of the most annoying things in the UI.

PR #2334, opened 2024-02-10
Change add item route to avoid clashing with item name

### LINKED ISSUE #2333: Unable to show the details of an item named `add`
Because the URL of mainui for showing an item is `/settings/items/<Item_name>` and for adding a new item is `/settings/items/add`, if you happen to have an item named `add`, you can't show its configuration because MainUI will show the Add new item page instead! :)

jimtng:
Fix #2333

Change the route from `/settings/items/add` -> `/settings/items/add-item`

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
+++ b/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
@@ -272,7 +272,7 @@
               Inbox
             </f7-list-button>
             <f7-list-item divider title="Items" />
-            <f7-list-button href="/settings/items/add" color="blue" :animate="false">
+            <f7-list-button href="/settings/items/add-item" color="blue" :animate="false">
               Create Item
             </f7-list-button>
             <f7-list-button href="/settings/items/add-from-textual-definition" color="blue" :animate="false">
--- a/bundles/org.openhab.ui/web/src/js/routes.js
+++ b/bundles/org.openhab.ui/web/src/js/routes.js
@@ -176,7 +176,7 @@ export default [
         async: loadAsync(ItemsListPage),
         routes: [
           {
-            path: 'add',
+            path: 'add-item',
             beforeEnter: [enforceAdminForRoute],
             async: loadAsync(ItemEditPage, { createMode: true })
           },
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
@@ -114,7 +114,7 @@
       <f7-icon ios="f7:plus" md="material:add" aurora="f7:plus" />
       <f7-icon ios="f7:multiply" md="material:close" aurora="f7:multiply" />
       <f7-fab-buttons position="top">
-        <f7-fab-button fab-close label="Add Item" href="add">
+        <f7-fab-button fab-close label="Add Item" href="add-item">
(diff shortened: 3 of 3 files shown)


JustinGeorgi:
Doesn't this just move the conflict?  It just means that you can't view an item with the name of  `add-item`.  Now, I don't see either option as being very likely, but what do I know?

Since the `\add` route is consistent with a few others as well, maybe it makes more sense to just include `"add"` (and maybe a few other keywords) in the validation check for the item name.

jimtng:
add-item is not a valid name (dash is not allowed for item name)

Preventing "add" in UI is pointless because add is a valid item name and it can be created outside of the UI eg .items file

JustinGeorgi:
> add-item is not a valid name (dash is not allowed for item name)

D'oh.  Forgot about the dash, that's a solid point.

> Preventing "add" in UI is pointless because add is a valid item name and it can be created outside of the UI eg .items file

I'm not convinced that this is pointless (yet, though I am persuadable) .  If the item is created in a file then editing (and changing the name) can be done whether the UI page can be accessed anyway, so it's not nearly the same issue as if the item were named `add` during the UI creation process.

It just feels to me that changing the established route away from the pattern shared by other routes is a big step for this very small edge case and other options should be considered as well.

jimtng:
I agree that this makes it inconsistent with other add routes (e.g.things etc) and I'm unhappy about that.

I know this is an edge case, so perhaps leaving it be is a better option.

I'm open to other ideas.

JustinGeorgi:
You could check for the presence of an `add` item when opening the add new item wizard.  If the item exists then offer a dialog with the option to go to that item instead (this have have to include a work around to open the `add` item details page through some secondary name which is then converted back into `add` during the details page load).

That would have the added benefit of, after three or four times, annoying user enough to cause them to change the item to a better name :wink:

> It's inconsistent vs buggy.

Inconsistent could, of course, be addressed by just changing all the equivalent routes.  But that's even more work (plus updating the links in the integrated docs; there are some direct links to add pages there).

Or maybe this is better handled in the core repo with a log warning about item names with potential UI conflicts?

jimtng:
> You could check for the presence of an `add` item when opening the add new item wizard. If the item exists then offer a dialog with the option to go to that item instead (this have have to include a work around to open the `add` item details page through some secondary name which is then converted back into `add` during the details page load).

That's a clever idea. I will play with this idea and see if I could come up with an implementation

jimtng:
Brainstorming: how about this:

| path                            | desc                     |
| ------------------------------- | ------------------------ |
| /settings/items/                | list the items           |
| /settings/items/view/itemname | View the item itemname |
| /settings/items/edit/itemname | edit                     |
| /settings/items/add             | add a new item           |

So basically, instead of changing the route for add, we change the route for viewing the item, and by extension also changing the route for editing

But this means changing the routes for other things too (things, channels, transformation, etc).

@ghys @florian-h05 are these paths used in any documentation? What would be the impact of changing routes?

The other option is to do nothing and leave it as is, of course, which is the easiest option.

rkoshak:
> I'm open to other ideas.

Maybe add to the docs and some validation to the Name field that "add" is a reserved word and cannot be used as an Item name would be sufficient? 

I think the likelihood that there are a lot of users using "add" as an Item name or wanting to do so is relatively low, just based on what Items represent. So maybe just a tiny bit of documentation and checking is enough. 

All these other proposals seem to becoming larger and more involved than the original problem warrants.

jimtng:
Thanks for your feedback. Closing this.

PR #2338, opened 2024-02-10
Fix sidebar search scene and script integration

JustinGeorgi:
Fixes a few minor error with scripts and scene in the developer toolbar search function:
1) Scenes and scripts could be pinned but not unpinned
2) Scenes and scripts were not enumerated with the other types for determining existence of search results


florian-h05 (approved):
LGTM, thanks!

PR #2340, opened 2024-02-10
Allow custom page icons

JustinGeorgi:
This adds the ability to set custom icons for pages.  The custom icons can be any icons available to the the oh-icon component.  They are visible in the side bar and in the page settings list:
image

image

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -27,7 +27,7 @@
                         :class="{ currentsection: currentUrl === '/page/' + page.uid || currentUrl.indexOf('/page/' + page.uid + '/') === 0 }"
                         :link="'/page/' + page.uid"
                         :title="page.config.label" view=".view-main" panel-close>
-            <f7-icon slot="media" :f7="pageIcon(page)" />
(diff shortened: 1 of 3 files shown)


florian-h05 (approved):
LGTM, thanks!

ghys:
This will be a people-pleaser for sure!
By the way, @JustinGeorgi, do you track your beer brewing with openHAB? (just assuming from your first screenshot)

I'm asking because I attended a talk last weekend about that:
https://fosdem.org/2024/schedule/event/fosdem-2024-2213-brewing-free-beer-with-esphome-and-home-assistant/
and an openHAB alternative would be most welcome :)

JustinGeorgi:
@ghys Nothing complicated.  I'm on the very less serious side of home brewers.  I don't have a fermentation fridge so I use an OH page to track the temp in the room where my fermentation kettle is (and get an alert if it runs out of range) The same page keeps a quick record of what stage the batch is in and when I need to next to do something and a link to the recipe for reference.

I admit, though, the iSpindle post a little while back caught my attention:
https://community.openhab.org/t/ispindle-track-your-homebrews-with-an-opensource-mqtt-hydrometer-temperature-sensor/151009

That might be the hook that starts me on the slippery slope to full OH integration with my brews...

PR #2341, opened 2024-02-10
[BasicUI] Show buttons on several lines when not enough place

lolodomo:


lolodomo:
With this change:

image


Before this change:

image

jimtng:
The android app solved this more elegantly

lolodomo:
Question of taste. I prefer to have all buttons (and variable button width) and only one click for user.
I don't say Android app choice is wrong. By the way, I don't think I can implement it that way easily and responsively with HTML/CSS.

lolodomo:
I propose to merge my solution as a fast fix of one of the most annoying issues in Basic UI.

Later if someone proposes a PR with a better solution, it can be of course reviewed.

jimtng:
> I propose to merge my solution as a fast fix of one of the most annoying issues in Basic UI.
> 
> Later if someone proposes a PR with a more elegant solution, it can be of course reviewed.

That sounds like a plan

kaikreuzer (approved):
Fine for me, thanks.

PR #2347, opened 2024-02-13
Remove padding in list grouping toolbar

jimtng:
A minor cosmetic change

Before:
image

After:
image


ghys:
These were added on purpose to avoid them touching the edge of the screen on mobiles:
Screen Shot 2024-02-13 at 18 17 50

jimtng:
> These were added on purpose to avoid them touching the edge of the screen on mobiles:

Would adding a @media condition be an acceptable solution?

florian-h05 (approved):
LGTM, thanks!

PR #2348, opened 2024-02-13
Accept double click to pick a model in model picker popup

jimtng:
Allow double clicking to select an item here, without having to click "Pick" at the top right corner

image


florian-h05 (commented):
In general LGTM, however I wonder if we should check that the clicked Item is the same one.

jimtng:
> I wonder if we should check that the clicked Item is the same one.

I've added this check

florian-h05 (approved):
LGTM, thanks!

PR #2349, opened 2024-02-13
Remember checkbox settings in model picker popup

jimtng:
Remember these checkboxes so they don't need to be re-checked every time

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/model/model-picker-popup.vue
+++ b/bundles/org.openhab.ui/web/src/components/model/model-picker-popup.vue
@@ -79,13 +79,14 @@ export default {
     ModelTreeview
   },
   data () {
+    if (!this.$f7.data.modelPicker) this.$f7.data.modelPicker = {}
     return {
       ready: false,
       loading: false,
       initSearchbar: false,
-      includeNonSemantic: localStorage.getItem('openhab.ui:model-picker.includeNonSemantic') === 'true',
-      includeItemName: localStorage.getItem('openhab.ui:model-picker.includeItemName') === 'true',
-      includeItemTags: localStorage.getItem('openhab.ui:model-picker.includeItemTags') === 'true',
+      includeNonSemantic: this.$f7.data.modelPicker.includeNonSemantic || false,
+      includeItemName: this.$f7.data.modelPicker.includeItemName || false,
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
Better to use `$f7.data` instead - I've already commited this.

Thanks!

florian-h05:
@jimtng Do you think this is useful for the model page as well?

jimtng:
For "Show name" and "Show tags", yes. 

For "Show non-semantic", I'm not so sure. The "Model" page is supposed to show the semantic stuff. Having it show non-semantic by default (because of the stored option) would feel weird, but perhaps if that's what the user wants, then so be it?

florian-h05:
Clearly agreed on show name and show tags. 

"Show non-semantic" tags is difficult, and the argument that it is supposed to show semantic stuff is a good point.
But ff the two other settings remember their state, "Show non-semantic" not doing that might be considered as bug by most users.

florian-h05:
Not sure if you've seen #2355. 
I will partly revert changes from #2355 and #2349 to make the "show non-semantic" checkbox non-persistent.

jimtng:
IMO the item pickers show non semantics and the other check boxes should be persistent. I wish such things can be done in such a way that it can be made persistent for those who want it so and not persistent for the rest if that's what they prefer.

I don't understand why one would want it to be non persistent.

I agree that show non semantics in the model view stays off by default. In fact it probably shouldn't even be there.

florian-h05:
FYI when using the item picker, the button on the left wirh the menu items icon opens the model picker and a click on the rest of the item picker opens the normal, non-model item picker.
IMO we don’t need to store the non-semantic option if the model picker, as for non-semantic items you should simply use the normal item picker.

jimtng:
> FYI when using the item picker, the button on the left wirh the menu items icon opens the model picker and a click on the rest of the item picker opens the normal, non-model item picker. 

I am aware of this, but the model picker is what pops up when you're adding a new rule module. The choice of the normal item picker is only available after that.

jimtng:
Also, blockly only uses the model picker to select items.

florian-h05:
True, but it is not really consistent to have different behavious between the model picker and the model page.

@ghys WDYT? Do you have an idea how to solve that?

PR #2352, opened 2024-02-13
Rule edit: Remove redundant module title prefix

jimtng:
See https://github.com/openhab/openhab-core/pull/4090

- When/If removed
- Added prefix "item", "thing", similar to "channel"
- play sound "xxx.mp3"
- thing changed -> thing status changed

Before:
image

After:
image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/module-description-suggestions.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/module-description-suggestions.js
@@ -12,80 +12,81 @@ export default {
         // triggers
         case 'timer.TimeOfDayTrigger':
           if (!config.time) return moduleType.label
-          return 'When the time is ' + config.time
+          return 'the time is ' + config.time
         case 'timer.GenericCronTrigger':
           if (!config.cronExpression) return moduleType.label
           try {
-            return cronstrue.toString(config.cronExpression, {
+            const cron = cronstrue.toString(config.cronExpression, {
               use24HourTimeFormat: true
             })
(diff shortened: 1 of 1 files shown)


jimtng:
I've updated the screenshots

- When/If removed
- Added prefix "item", "thing", similar to "channel"
- play sound "xxx.mp3"

bundles/org.openhab.ui/web/src/pages/settings/rules/module-description-suggestions.js
```diff
@@ -12,80 +12,80 @@ export default {
...
           if (!config.time) return moduleType.label
-          return 'When the time is ' + config.time
+          return 'the time is ' + config.time
         case 'timer.GenericCronTrigger':
           if (!config.cronExpression) return moduleType.label
           try {
             return cronstrue.toString(config.cronExpression, {
               use24HourTimeFormat: true
-            })
+            }).toLowerCase()
```
ghys, line 21: This won't work well with labels like "At 08:00, only on Monday and Thursday".
Just lowercase the first character.

ghys (changes requested):
Thanks, this was also bugging me. LGTM except for the lowercasing of the cronstrue expressions.

bundles/org.openhab.ui/web/src/pages/settings/rules/module-description-suggestions.js
jimtng: Sorry I force pushed - habit. But it's now only lowercasing the first letter.

florian-h05 (approved):
The last change LGTM - luckily GitHub allows to compare between force pushes.

PR #2356, opened 2024-02-15
Globally align smart-select value with other input values

jimtng:
Related to #2326 
@florian-h05 
This will make all of the smart-selects aligned the same way without needing to add special class to each element

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -239,7 +239,23 @@
   // --f7-list-item-media-margin 24px
   // --f7-list-item-padding-horizontal 32px
   // --f7-list-chevron-icon-color var(--f7-color-blue-tint) !important
+</style>
 
+<style lang="stylus">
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/components/app.vue
florian-h05, line 1: This should rather go to https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/css/app.styl.

florian-h05 (commented):
In general I’m hesitant whether we want to do that.
Before doing that, I’d like to check where and how the smart select is used in the UI, i.e. check these search results: https://github.com/search?q=repo%3Aopenhab%2Fopenhab-webui+Smart-select+language%3AVue&type=code&l=Vue

ghys:
> In general I’m hesitant whether we want to do that.

We don't. I'm shutting this down before it gets to a point where it's just "we accept it because of all of work".
The smart selects are fine as they are.

PR #2360, opened 2024-02-16
Add back the 'When' prefix for file-based script trigger titles

jimtng:
Fix regression from #2352 

image


florian-h05 (approved):
LGTM, thanks!
FYI, I have fixed the linter error.

jimtng:
oops thanks for fixing it!

PR #2367, opened 2024-02-17
Fix rule module editor popup for blockly

jimtng:
Currently for Blockly Script Action or Condition, clicking on "Edit module" the pencil icon would open up rule-module-popup, which allows user to edit the raw javascript directly without regard to the blockly's block code. 

image

This PR changed it by replacing the configuration sheet with a button that leads to the blockly editor

image


florian-h05 (approved):
LGTM and thanks for the fix!

PR #2369, opened 2024-02-17
Add a global component developer-dock-icon

jimtng:
Also add the icon to widget-list and script-edit pages.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
@@ -1,11 +1,14 @@
 <template>
   <f7-page @page:afterin="onPageAfterIn" @page:afterout="onPageAfterOut">
     <f7-navbar :title="isNewRule ? 'Create rule' : rule.name" back-link="Back" no-hairline>
-      <f7-nav-right v-if="isEditable">
-        <f7-link @click="save()" v-if="$theme.md" icon-md="material:save" icon-only />
-        <f7-link @click="save()" v-if="!$theme.md">
-          Save<span v-if="$device.desktop">&nbsp;(Ctrl-S)</span>
-        </f7-link>
+      <f7-nav-right>
+        <developer-dock-icon />
+        <template v-if="isEditable">
+          <f7-link @click="save()" v-if="$theme.md" icon-md="material:save" icon-only />
+          <f7-link @click="save()" v-if="!$theme.md">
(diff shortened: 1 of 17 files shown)


bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -1,11 +1,14 @@
 <template>
   <f7-page @page:afterin="onPageAfterIn" @page:afterout="onPageAfterOut">
     <f7-navbar :title="isNewRule ? 'Create rule' : rule.name" back-link="Back" no-hairline>
-      <f7-nav-right v-if="isEditable">
-        <f7-link @click="save()" v-if="$theme.md" icon-md="material:save" icon-only />
-        <f7-link @click="save()" v-if="!$theme.md">
-          Save<span v-if="$device.desktop">&nbsp;(Ctrl-S)</span>
-        </f7-link>
+      <f7-nav-right>
+        <developer-dock-icon />
```
florian-h05: As long as we don't have extra content for the rule edit page to show in the sidebar, I would not add the help icon there.

florian-h05 (approved):
Nice refactoring, thanks!

I've reformatted your code a bit, removed the icon from rule edit and added it to the widgets list.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
florian-h05: @jimtng WDYT?

jimtng: Perhaps we should add an extra (help) content?

It's still handy to a) lookup items, b) create a new item using the dev sidebar

florian-h05: > Perhaps we should add an extra (help) content?

In case you have an idea what content to show there, it would be great if you could add that content.
I can provide help regarding the conventions required by the UI to properly render that Markdown file.

> It's still handy to a) lookup items, b) create a new item using the dev sidebar

Agreed, so I will add it back to rule-edit and merge.

florian-h05:
@jimtng One thing I now have noticed with the developer dock icon being on non-admin pages: It need to be hidden if the user is not logged in as admin.

Can you please fix that?

See https://github.com/openhab/openhab-webui/blob/c703f4b23e977caabb3beeca48eec34b512bae65/bundles/org.openhab.ui/web/src/pages/home.vue#L15 for reference how to check for admin, and BTW it seems you forgot to use the new component there.

PR #2372, opened 2024-02-17
Developer sidebar: Add clipboard icons

jimtng:
image
image


florian-h05 (approved):
Nice, thanks!

FYI I have added some padding to the icons in the search results, they were IMO too close together.

ghys:
image

Hmm, this is clearly a regression as we don't get the statuses anymore? Or the play/pause buttons?

jimtng:
Both those screenshots are post-pr. Nothing has been removed. Only new icon added.

florian-h05:
> Hmm, this is clearly a regression as we don't get the statuses anymore? Or the play/pause buttons?

IIRC they have never been shown for the search results, only for the pinned stuff.

PR #2374, opened 2024-02-18
Script & Item edit: Add/Fix dirty handling

florian-h05:
- Add dirty handling for script-edit.
- Fix dirty handling of item-edit always saying dirty due to wrong initialization order.
- Code improvements to other dirty handlings.


florian-h05:
@jimtng Can you please test this PR?

jimtng:
This doesn't seem to check whether a rule module is dirty when the Back link is clicked. This is the screen I'm referring to:

Rules -> select a rule -> Select an existing module (trigger / action / condition) but other than a script. Make changes in this screen then click Back (not Save). It should prompt when dirty.

image

florian-h05:
> This doesn't seem to check whether a rule module is dirty when the Back link is clicked.

I haven't worked on this, glad that you reminded me that this also need dirty handling.

I only worked on script-edit and item-edit.

florian-h05:
@jimtng FYI I have edited my last comment, what you tested is out of scope of this PR as I only focused on the actual pages, not dirty handling for popups.

jimtng:
> @jimtng FYI I have edited my last comment, what you tested is out of scope of this PR as I only focused on the actual pages, not dirty handling for popups.

Fair enough. I wasn't aware that the other areas needed fixing (they of course did, evidenced by this PR). The popup was what was on my mind :)

florian-h05:
@jimtng It now became a bit bigger than expected, but now it works fine in my tests.
Can you please re-test the script edit dirty handling?

jimtng (commented):
What I found so far:

- Run Now will forcibly save current changes without prompting/warning
- After Run Now, edit the script again. This should make it dirty but it doesn't warn about the new changes when clicking Back out of the script editor. The changes made after clicking Run Now is lost!

florian-h05:
I cannot reproduce that issue, for me this works fine.

jimtng:
> What I found so far:
> 
> * Run Now will forcibly save current changes without prompting/warning

This is still the case. Shouldn't we prompt user whether they wanted to save first before running?

> * After Run Now, edit the script again. This should make it dirty but it doesn't warn about the new changes when clicking Back out of the script editor. The changes made after clicking Run Now is lost!

It seems that this is not the case now.

florian-h05:
> Shouldn't we prompt user whether they wanted to save first before running?

I would tent to say no, because I guess when you click "Run Now" you want to run what you have actually written, and not what is the saved state on the server. WDYT?

jimtng:
I would like to be be prompted before it saves my changes.

florian-h05:
I have added a dialog for that.
From my POV we are ready to merge, WDYT?

jimtng:
I don't like the current behaviour.

The prompt should say 

> Changes not saved yet
> Do you want to save the changes before running the script?
> [Cancel] [Save]

When cancel is clicked, don't run.

Right now it runs the old script prior to editing, which is not what I see on the screen, and that, to me, is confusing.

florian-h05:
Pushed :+1:

florian-h05:
> Does this happen to you?

Yes.

I will merge this PR as it is unrelated to it, but it should be taken care of that.

jimtng:
@florian-h05 I've just tried this:
- View an item
- Click edit, make no change
- Click Back - I got the dirty warning.

florian-h05:
I cannot reproduce this with the dev server on main (commit 69f487611f2f763f7889c36bbe4ebd812d17de14).

PR #2377, opened 2024-02-18
Rule edit: Generate suggested trigger title only when none is available

jimtng:
This lets the helper library provide a more accurate label for custom triggers

Related: https://github.com/openhab/openhab-jruby/pull/254

image


florian-h05 (approved):
LGTM, thanks!

PR #2378, opened 2024-02-19
Developer sidebar icon: Check for admin role

jimtng:
Fix https://github.com/openhab/openhab-webui/pull/2369#issuecomment-1951462800


florian-h05 (approved):
Thanks!

PR #2387, opened 2024-02-21
Model Treeview: Remember the expanded state

jimtng:
Applies to both the Settings -> Model page and the model picker popup


florian-h05 (approved):
Nice, thanks!

PR #2388, opened 2024-02-22
[BasicUI] Align and optimize available space for switch with mappings

jimtng:
- Should only affect switch with mappings, not player control
- Utilise the wasted space instead of wrapping the buttons into multiple rows, if possible.
- Right align the buttons. This makes it look _much_ neater.
- Instead of limiting the width of the buttons, reserve a minimum width for the label
- However if the label is shorter than 6 characters (including blank labels), reduce the label's minimum width to just what's actually taken up by the shorter label. This gives more space for the buttons with shorter labels.
- When buttons wrap to multiple rows, make sure that each row contains almost the same number of buttons, instead of having the first row filling up the horizontal space, move the buttons down. By doing this, there is more space for the label to occupy.
- Squeeze extra space for more buttons in "condensed layout" by reducing padding, inter-button gaps, min-width, etc.
- The reduced padding in condensed layout also affects buttons in buttongrid.

Before:
image

After:

image

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -233,7 +233,7 @@
 		}
 	}
 	&__label-multiline {
-		min-width: 5em;
+		min-width: 3em;
 	}
 	&__control {
 		padding-right: $form-row-desktop-padding;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -517,7 +517,7 @@
 		gap: 4px;
 		html.ui-layout-condensed & {
 			margin: 0;
-			gap: 2px;
+			gap: 3px;
 		}
 		padding-top: 2px;
 		padding-bottom: 2px;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -225,13 +225,15 @@
 		@include flex-shrink(0);
 		@include flex-grow(2);
 		@include flex-2011(2 2 auto);
-		min-width: 5em;
 		white-space: nowrap;
 		overflow: hidden;
 		text-overflow: ellipsis;
 		html.ui-bigger-font & {
 			font-size: 18px;
 		}
+	}
+	&__label-multiline {
+		min-width: 5em;
 	}
 	&__control {
 		padding-right: $form-row-desktop-padding;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -281,6 +281,7 @@
 			box-shadow: none;
 			-webkit-box-shadow: none;
 			text-transform: unset;
+			min-width: 54px;
 			html.ui-layout-condensed & {
 				min-width: 40px;
 				padding-left: 4px;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -514,6 +514,7 @@
 	}
 	&__buttons-multiline {
 		margin: 6px 0;
+		gap: 4px;
 		html.ui-layout-condensed & {
 			margin: 0;
 			gap: 2px;
@@ -523,7 +524,6 @@
 		display: flex;
 		flex-wrap: wrap;
 		justify-content: end;
-		gap: 4px;
 	}
 	&__buttongrid {
 		padding: 0;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -237,9 +237,13 @@
 		padding-left: 4px;
 		html.ui-layout-condensed & {
 			padding-left: 0;
+			padding-right: $form-row-desktop-padding-condensed;
 		}
 		@media screen and (max-width: $layout-tablet-size-threshold) {
 			padding-right: $form-row-mobile-padding;
+			html.ui-layout-condensed & {
+				padding-right: $form-row-mobile-padding-condensed;
+			}
 		}
 		font-weight: 700;
 		html.ui-capitalize-values & {
@@ -503,7 +507,8 @@
 		width: 100%;
 	}
 	&__buttons {
-		padding: 2px;
+		padding-top: 2px;
+		padding-bottom: 2px;
 	}
 	&__buttons-multiline {
 		margin: 6px 0;
@@ -511,7 +516,8 @@
 			margin: 0;
 			gap: 2px;
 		}
-		padding: 2px;
+		padding-top: 2px;
+		padding-bottom: 2px;
 		max-width: 70%;
 		display: flex;
 		flex-wrap: wrap;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -225,6 +225,7 @@
 		@include flex-shrink(0);
 		@include flex-grow(2);
 		@include flex-2011(2 2 auto);
+		min-width: 5em;
 		white-space: nowrap;
 		overflow: hidden;
 		text-overflow: ellipsis;
@@ -518,7 +519,6 @@
 		}
 		padding-top: 2px;
 		padding-bottom: 2px;
-		max-width: 70%;
 		display: flex;
 		flex-wrap: wrap;
 		justify-content: end;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -168,7 +168,7 @@
 		@media screen and (max-width: $layout-tablet-size-threshold) {
 			padding-left: $form-row-mobile-padding;
 			html.ui-layout-condensed & {
-				padding-left: $form-row-mobile-padding-condensed;
+				padding-left: 0; // there's already padding in .mdl-form__row
 			}
 		}
 		html:not(.ui-icons-enabled) & {
@@ -212,15 +212,15 @@
 		}
 	}
 	&__label {
+		padding-left: $form-row-desktop-padding;
+		html.ui-layout-condensed & {
+			padding-left: $form-row-desktop-padding-condensed;
+		}
 		@media screen and (max-width: $layout-tablet-size-threshold) {
 			padding-left: $form-row-mobile-padding;
 			html.ui-layout-condensed & {
 				padding-left: $form-row-mobile-padding-condensed;
 			}
-		}
-		padding-left: $form-row-desktop-padding;
-		html.ui-layout-condensed & {
-			padding-left: $form-row-desktop-padding-condensed;
 		}
 		@include flex-shrink(0);
 		@include flex-grow(2);
@@ -233,16 +233,13 @@
 		}
 	}
 	&__control {
-		@media screen and (max-width: $layout-tablet-size-threshold) {
-			padding-right: $form-row-mobile-padding;
-			html.ui-layout-condensed & {
-				padding-left: $form-row-mobile-padding-condensed;
-			}
-		}
 		padding-right: $form-row-desktop-padding;
 		padding-left: 4px;
 		html.ui-layout-condensed & {
-			padding-left: $form-row-desktop-padding-condensed;
+			padding-left: 0;
+		}
+		@media screen and (max-width: $layout-tablet-size-threshold) {
+			padding-right: $form-row-mobile-padding;
 		}
 		font-weight: 700;
 		html.ui-capitalize-values & {
@@ -270,7 +267,7 @@
 		.buttongrid-cell {
 			height: 36px;
 			.buttongrid-button {
-				min-width: 100%;
+				min-width: 100% !important;
 				text-transform: none;
 			}
 		}
@@ -279,6 +276,11 @@
 			box-shadow: none;
 			-webkit-box-shadow: none;
 			text-transform: unset;
+			html.ui-layout-condensed & {
+				min-width: 40px;
+				padding-left: 4px;
+				padding-right: 4px;
+			}
 		}
 		.mdl-button-text {
 			html.ui-bigger-font & {
@@ -501,16 +503,19 @@
 		width: 100%;
 	}
 	&__buttons {
-		padding-top: 2px;
+		padding: 2px;
 	}
 	&__buttons-multiline {
 		margin: 6px 0;
 		html.ui-layout-condensed & {
 			margin: 0;
+			gap: 2px;
 		}
-		max-width: 60%;
+		padding: 2px;
+		max-width: 70%;
 		display: flex;
 		flex-wrap: wrap;
+		justify-content: end;
 		gap: 4px;
 	}
 	&__buttongrid {
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1349,7 +1349,14 @@
 			_t.minimizeWidth = function() {
 				var labelLength = _t.label.textContent.trim().length;
 				_t.label.style.paddingLeft = labelLength === 0 ? 0 : null; // null reverts it to the css
-				_t.label.style.minWidth = labelLength < 6 ? "min-content" : null;
+
+				var labelWidth = _t.label.offsetWidth;
+				// Try and see if min-content could make the label narrower.
+				// If not, undo it so it uses the fixed min-width set in the css.
+				_t.label.style.minWidth = "min-content";
+				if (labelWidth <= _t.label.offsetWidth) {
+					_t.label.style.removeProperty("minWidth");
+				}
 
 				// Minimize the width taken by the buttons without adding extra rows.
 				// Start from the maximum width (limited by `mdl-form__buttons-multiline.max-width`),
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1347,13 +1347,14 @@
 
 		if (_t.buttons.length > 1 && _t.parentNode.classList.contains(o.buttonsMultilineClass)) {
 			_t.minimizeWidth = function() {
+				var labelLength = _t.label.textContent.trim().length;
+				_t.label.style.paddingLeft = labelLength === 0 ? 0 : null; // null reverts it to the css
+				_t.label.style.minWidth = labelLength < 6 ? "min-content" : null;
+
 				// Minimize the width taken by the buttons without adding extra rows.
 				// Start from the maximum width (limited by `mdl-form__buttons-multiline.max-width`),
 				// then shrink it down to the minimum without causing additional wrapping.
 				var buttons = _t.parentNode;
-				if (_t.label.textContent.trim() === "") {
-					buttons.style.maxWidth = "100%";
-				}
 				var width = buttons.parentElement.offsetWidth + 100; // start wider than max-width allows
 				buttons.style.width = width + "px";
 				var height = buttons.offsetHeight;
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1384,7 +1384,8 @@
 				buttons.style.width = (width+1) + "px";
 			};
 
-			_t.minimizeWidth();
+			// Wait until after all the icons are loaded before running minimizeWidth()
+			window.addEventListener("load", _t.minimizeWidth);
 			window.addEventListener("resize", _t.minimizeWidth);
 		} else {
 			_t.minimizeWidth = function() {};
@@ -1412,6 +1413,7 @@
 				}
 			});
 			componentHandler.downgradeElements(_t.buttons);
+			window.removeEventListener("load", _t.minimizeWidth);
 			window.removeEventListener("resize", _t.minimizeWidth);
 		};
 
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1403,6 +1403,7 @@
 				buttons.style.width = (width+1) + "px";
 			};
 
+			_t.minimizeWidth();
 			// Wait until after all the icons are loaded before running minimizeWidth()
 			window.addEventListener("load", _t.minimizeWidth);
 			window.addEventListener("resize", _t.minimizeWidth);
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1177,6 +1177,7 @@
 		Control.call(this, parentNode);
 
 		var
+			maxButtonWidth = 0,
 			_t = this;
 
 		_t.ignoreState = _t.parentNode.getAttribute("data-ignore-state") === "true";
@@ -1239,6 +1240,7 @@
 			) {
 				_t.valueMap[itemState].classList.add(o.buttonActiveClass);
 			}
+			_t.minimizeWidth();
 		};
 
 		_t.setValueColor = function(color) {
@@ -1337,7 +1339,35 @@
 				icon.addEventListener("load", _t.convertToInlineSVG);
 				icon.addEventListener("error", _t.replaceImageWithNone);
 			}
+
+			if (maxButtonWidth < button.offsetWidth) {
+				maxButtonWidth = button.offsetWidth;
+			}
 		});
+
+		if (_t.buttons.length > 1 && _t.parentNode.classList.contains(o.buttonsMultilineClass)) {
+			_t.minimizeWidth = function() {
+				// Minimize the width taken by the buttons without adding extra rows.
+				// Start from the maximum width (limited by `mdl-form__buttons-multiline.max-width`),
+				// then shrink it down to the minimum without causing additional wrapping.
+				var buttons = _t.parentNode;
+				if (_t.label.textContent.trim() === "") {
+					buttons.style.maxWidth = "100%";
+				}
+				var width = buttons.parentElement.offsetWidth + 100; // start wider than max-width allows
+				buttons.style.width = width + "px";
+				var height = buttons.offsetHeight;
+				while (buttons.offsetHeight === height && width >= maxButtonWidth) {
+					buttons.style.width = --width + "px";
+				}
+				buttons.style.width = (width+1) + "px";
+			};
+
+			_t.minimizeWidth();
+			window.addEventListener("resize", _t.minimizeWidth);
+		} else {
+			_t.minimizeWidth = function() {};
+		}
 
 		_t.destroy = function() {
 			_t.buttons.forEach(function(button) {
@@ -1361,6 +1391,7 @@
 				}
 			});
 			componentHandler.downgradeElements(_t.buttons);
+			window.removeEventListener("resize", _t.minimizeWidth);
 		};
 
 		_t.setValueColor(_t.valueColor);
@@ -3936,6 +3967,7 @@
 	buttonTextClass: "mdl-button-text",
 	buttonIconText: ".mdl-button-icon-text",
 	buttonIconTextClass: "mdl-button-icon-text",
+	buttonsMultilineClass: "mdl-form__buttons-multiline",
 	modal: ".mdl-modal",
 	modalContainer: ".mdl-modal__content",
 	selectionRows: ".mdl-form__selection-rows",
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1375,8 +1375,8 @@
 				// Start from the maximum width (limited by `mdl-form__buttons-multiline.max-width`),
 				// then shrink it down to the minimum without causing additional wrapping.
 				var buttons = _t.parentNode;
-				var width = buttons.parentElement.offsetWidth + 100; // start wider than max-width allows
-				buttons.style.width = width + "px";
+				buttons.style.width = "100%";
+				var width = buttons.offsetWidth;
 				var height = buttons.offsetHeight;
 				while (buttons.offsetHeight === height && width >= maxButtonWidth) {
 					buttons.style.width = --width + "px";
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -168,6 +168,12 @@
 		});
 	}
 
+	function getElementWidth(element) {
+		var style = getComputedStyle(element);
+		return element.offsetWidth + parseFloat(style.marginLeft) + parseFloat(style.marginRight) +
+			parseFloat(style.paddingLeft) + parseFloat(style.paddingRight);
+	}
+
 	function EventMapper() {
 		var
 			_t = this;
@@ -1346,6 +1352,7 @@
 		});
 
 		if (_t.buttons.length > 1 && _t.parentNode.classList.contains(o.buttonsMultilineClass)) {
+			var labelMinWidth = 0;
 			if (_t.label.textContent.trim().length === 0) {
 				_t.label.style.paddingLeft = 0;
 				_t.label.style.minWidth = 0;
@@ -1367,16 +1374,28 @@
 				var minContentWidth = _t.label.offsetWidth;
 				if (minContentWidth > defaultMinWidth) {
 					_t.label.style.removeProperty("min-width");
+					labelMinWidth = defaultMinWidth;
+				} else {
+					labelMinWidth = minContentWidth;
 				}
 			}
 
 			_t.minimizeWidth = function() {
 				// Minimize the width taken by the buttons without adding extra rows.
-				// Start from the maximum width (limited by `mdl-form__buttons-multiline.max-width`),
+				// Start from the maximum width the buttons can take,
 				// then shrink it down to the minimum without causing additional wrapping.
 				var buttons = _t.parentNode;
-				buttons.style.width = "100%";
-				var width = buttons.offsetWidth;
+				var buttonsStyle = getComputedStyle(buttons);
+				var labelStyle = getComputedStyle(_t.label);
+				// Calculate the maximum width the buttons can take
+				var width = buttons.parentElement.offsetWidth -
+					parseFloat(buttonsStyle.paddingLeft) - parseFloat(buttonsStyle.paddingRight) -
+					getElementWidth(_t.iconContainer) - getElementWidth(_t.value);
+				if (labelMinWidth) {
+					width -= labelMinWidth + parseFloat(labelStyle.paddingLeft) + parseFloat(labelStyle.paddingRight);
+				}
+				buttons.style.width = width + "px";
+				width = buttons.offsetWidth;
 				var height = buttons.offsetHeight;
 				while (buttons.offsetHeight === height && width >= maxButtonWidth) {
 					buttons.style.width = --width + "px";
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1346,18 +1346,31 @@
 		});
 
 		if (_t.buttons.length > 1 && _t.parentNode.classList.contains(o.buttonsMultilineClass)) {
-			_t.minimizeWidth = function() {
-				var labelLength = _t.label.textContent.trim().length;
-				_t.label.style.paddingLeft = labelLength === 0 ? 0 : null; // null reverts it to the css
+			if (_t.label.textContent.trim().length === 0) {
+				_t.label.style.paddingLeft = 0;
+				_t.label.style.minWidth = 0;
+			} else {
+				// Try to see if setting min-width: min-content would result in a narrower min-width
+				// than the one set in _layout.scss, e.g. when the label is short.
+				// If it does make it narrower, it frees up more space for the buttons.
+				// If it is not narrower, un-set it, so that the min-width from _layout.scss can take effect.
 
-				var labelWidth = _t.label.offsetWidth;
-				// Try and see if min-content could make the label narrower.
-				// If not, undo it so it uses the fixed min-width set in the css.
+				// To measure the min-width using offsetWidth,
+				// we need to make the neighbouring element (buttons) as wide as possible
+				// to force the label to shrink to its min-width
+				// Note that _t.parentNode.style.width will get readjusted inside minimizeWidth()
+				// so setting it to 100% here wouldn't affect the final layout.
+				_t.parentNode.style.width = "100%";
+
+				var defaultMinWidth = parseFloat(getComputedStyle(_t.label).minWidth);
 				_t.label.style.minWidth = "min-content";
-				if (labelWidth <= _t.label.offsetWidth) {
-					_t.label.style.removeProperty("minWidth");
+				var minContentWidth = _t.label.offsetWidth;
+				if (minContentWidth > defaultMinWidth) {
+					_t.label.style.removeProperty("min-width");
 				}
+			}
 
+			_t.minimizeWidth = function() {
 				// Minimize the width taken by the buttons without adding extra rows.
 				// Start from the maximum width (limited by `mdl-form__buttons-multiline.max-width`),
 				// then shrink it down to the minimum without causing additional wrapping.
--- a/bundles/org.openhab.ui.basic/snippets-src/buttons.html
+++ b/bundles/org.openhab.ui.basic/snippets-src/buttons.html
@@ -2,7 +2,7 @@
 	<span class="mdl-form__icon">
 		%icon_snippet%
 	</span>
-	<span class="mdl-form__label">
+	<span class="mdl-form__label %label_class%">
 		%label%
 	</span>
 	<span class="mdl-form__value mdl-form__value--group">
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SwitchRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SwitchRenderer.java
@@ -133,9 +133,17 @@ public class SwitchRenderer extends AbstractWidgetRenderer {
 
             snippet = preprocessSnippet(snippet, w);
 
-            snippet = snippet.replaceAll("%height_auto%", multiline ? "mdl-form__row--height-auto" : "");
-            snippet = snippet.replaceAll("%buttons_class%",
-                    multiline ? "mdl-form__buttons-multiline" : "mdl-form__buttons");
+            if (multiline) {
+                snippet = snippet //
+                        .replaceAll("%height_auto%", "mdl-form__row--height-auto")
+                        .replaceAll("%buttons_class%", "mdl-form__buttons-multiline")
+                        .replaceAll("%label_class%", "mdl-form__label-multiline");
+            } else {
+                snippet = snippet //
+                        .replaceAll("%height_auto%", "") //
+                        .replaceAll("%buttons_class%", "mdl-form__buttons") //
+                        .replaceAll("%label_class%", "");
+            }
 
             StringBuilder buttons = new StringBuilder();
             if (s.getMappings().isEmpty() && item != null) {


lolodomo:
Your commit of vscode settings is probably unexpected ?

jimtng:
> The idea was to keep a minimum of place for label and value.

I haven't tested this with various widths of labels.



> Your commit of vscode settings is probably unexpected ?

oops, removed now.

lolodomo:
Using 4.2 M1, I can see that I have to increase the zone for buttons to not have controls for a Player item cut on 2 lines on a phone.

And I also see that the buttons in the new settings page are not rendered perfectly on a phone when cut on 2 lines.

jimtng:
I haven't had a chance to look into this yet, and I'm generally clueless with css. If you have a fix, feel free post it and close this PR.

Ideally, the buttons can take up as much horizontal space as possible, especially when there are no labels or the label is short. This way user can set a blank label to maximise the space for the buttons. But when there's a label, then reserve some space for the label.

lolodomo:
Even with your change, I have strange result like this one:

image

If I increase just a little the window, it is then OK:

image

Maybe the problem occurs when the label is truncated and  "..." is injected ?

lolodomo:
I would like to find the proper CSS properties to at least have player controls always on one line when there is enough place for that.

lolodomo:
If I reduce the length of my label, this is then fine:

image

The problem seems to be when the label is truncated.

jimtng:
The solution would probably include changing the css of the labels and/or the container.
Could we perhaps turn it into two lines (dynamically) when there's not enough space? So the label on top, then controls below it, but if there's enough space, keep them both in one line.

jimtng:
I'm continuing on this effort.

max-width still removed, also padding removed and justify/align to the end

image

jimtng:
With max-width: 70% instead of 60%

image

jimtng:
Another improvement:

image

jimtng:
Original post updated with the latest result.

jimtng:
@lolodomo this still reserves some space for the label / value, but also tries to make the buttons take up as much space as possible, but if it must wrap into multiple lines, maximise the space for the label again.

jimtng:
For anyone interested in testing, you can put this jar file in your addons folder, then disable the built in org.openhab.ui.basicui bundle, and restart/enable this bundle.

Note: rename the file and remove the .txt extension.

[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16409278/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

GeVaSta:
New JAR is worse I am afraid. Even wider screen does not put them on one line.
Clipboard_07-29-2024_02

GeVaSta:
New JAR is worse I think. Even wider screen does not put the buttons on one line.
Clipboard_07-29-2024_02

jimtng:
> New JAR is worse I think. Even wider screen does not put the buttons on one line. Clipboard_07-29-2024_02

Make sure you do a force refresh (Cmd+Refresh on mac, shift+reload or something - depending on your browser).

GeVaSta:
I did a refresh (Shift+F5 on Windows).
Looks like it could be something with the number of colums setting of the browser. I have mine on one (1) column.

jimtng:
Works fine for me, both 1 column and 2 columns.



image

image

image

GeVaSta:
Clipboard_07-29-2024_03
Clipboard_07-29-2024_06
After playing with the number of colums it changed a bit but still plenty of room to add button to the right instead on two lines.

jimtng:
I suspect it hasn't fuly refreshed. What's your O/S, browser version? I'll try it on BrowserStack

jimtng:
This is on BrowserStack running on Windows 11, Edge 125:

image

GeVaSta:
The difference is, I think, I use a much smaller window. Thats for me the whole point. I like the window as small as possible to put it right side of my screen to keep it visible.

GeVaSta:
Can you show a screenshot of the smallest browser window where one button moved to a second line? And the smallest on one line please?

jimtng:
I see, so when the window is narrowed, it does wrap. Thanks! I'll make some adjustments to fix that.

image

GeVaSta:
Perhaps reducing the overall width of the buttons (also the white spaces left and right of the text) resolves my problems. A button-width setting would be perfect I think.

Below is a screenshot withoud label text. It shows my problem clearly.
Clipboard_07-29-2024_08

jimtng:
To make it work in your specific situation, a few more unrelated margins / padding / sizes need to be tweaked. I can make some changes to make it fit better for narrower displays, however, I think that should be done in a separate PR to keep things cleanly separate. As it is, this PR addressed the issue for wider screens.

I would suggest you to open an issue regarding this specific point, i.e. optimise screen real estate for narrow displays.

GeVaSta:
I don't understand why to open a new PR. Read the Title of this PR and it perfectly reflects my issue. And I think my last screenshot perfectly shows the issue, even without asking for smaller margings inside the buttons.

jimtng:
@lolodomo wdyt? - Should I add more to this PR to reduce margins / adjust for narrow screens? It would affect a lot more things globally, hence why I think it should be a separate PR.

GeVaSta:
So should it look in my opnion. This is from my iPhone UI using the openHAB app.

IMG_1280

GeVaSta:
When you could make the buttons use the space that is available when there is no or a short label, there would already fit much more on one line. So right align the buttons until there does not fit another label on the left side. Occupying all the space from the end of the label. 
And I also do not understand why the buttons have to be so wide. I can push a much smaller button with ease.
Hope you can make an easy fix for me.

jimtng:
@GeVaSta please try this jar, and also turn on `Condensed Layout` in your BasicUI preferences
image

[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16424894/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

GeVaSta:
Thanks Jim. It got al partially better but still a lot of space wasted her with item "Speakers".
With item "Afzuiging...." it looks nice like this.
Looks like not all space after the label is taken. Could it also be possible without much trouble, to reduce the space left and right from the text within the button?
Clipboard_07-30-2024_03
Clipboard_07-30-2024_04
Appreciate your effort ver much.

jimtng:
I'm going away for several days. Once I'm back I can work on this some more.

jimtng:
@GeVaSta I managed to make some changes before leaving
Please try this
[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16434703/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

GeVaSta:
It is a little better already. See the two screenshots (just before and after wrapping). Is it a problem to remove (almost) all whitespace after the label to make it available for the buttons?
Clipboard_07-31-2024_01
Clipboard_07-31-2024_02

jimtng:
I'm afk for a few days. Typing this on a phone. Maybe we can collapse the space for VALUE when there's none to display.

GeVaSta:
Oh, yes. That is perhaps the space there is lost in front of the buttons and not only the white space after the label. Good idea.
Enjoy your stay for the coming days. Hope to hear back from you.

jimtng:
I've thought about this a bit, but not sure about "squashing" / eliminating the space for the _value_. What if the value got changed and suddenly we need to show it? I look around for what can be squeezed more, but the best is 2 pixels here and 4 pixels there.

GeVaSta:
I understand your concern about squeezing but how much space should there be reserved? 1 character, 2,3,4, 10? 
Why not using the same technique used for the label? Show one or two dots when the space is not enough when the window gets smaller. 
But afterall I think when you gain a few more pixels it will be enough (at least for me). The way it is now is much better than it was before you changed it.

GeVaSta:
I noticed a strange behaviour. I have put the two screenshots next to each other. The left one shows the first mapping-row with 5 buttons next to each other. When I reduce the width a little (right screenshot), the last button of the first mapping moves to a second line, BUT also from the other mapping (Speakers), the right button moves to the next line however there is plenty of space left.
Clipboard_08-05-2024_03

GeVaSta:
Another thing would give me more space. Leave the Case for the text on the buttons as entered and do not Capitalize all characters. I have never understood why it is automatically Capitalized. When I want the text Capitalized I will enter it that way.

jimtng:
> Leave the Case for the text on the buttons as entered and do not Capitalize all characters. I have never understood why it is automatically Capitalized.

I looked around the basicui code and found that this uppercasing isn't done by basicui. It is inherited from the default material design style.

See https://m2.material.io/components/buttons#anatomy under "Text Label" and I quote here:
> By default Material uses capitalized button text labels (for languages that have capitalization). This is to distinguish the text label from surrounding text. If a text button does not use capitalization for button text, find another characteristic to distinguish it such as color, size, or placement.

Our buttons in BasicUI do have a distinct background color, so I think it should be fine to _not_ uppercase the button text, but I'm not an expert in material design nor in basicui. @lolodomo ?

If we changed this, I'm pretty sure there'd be people who wouldn't like it and prefer to keep the old behaviour too :)

BasicUI has the configuration setting for "Capitalize Values"
image

We could either use that or add yet another setting "Capitalize Buttons" and make it default to enabled.

In any case, such changes _definitely_ belong in a separate PR.

jimtng:
@GeVaSta please try this version. It contains the following improvements: 
- when the label is empty, the buttons will take up the space normally reserved for the label. 
- In condensed layout, the 4px padding-left of the button is removed since there's also a padding-right on the value.

[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16517766/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

lolodomo:
@jimtng : I will have a look to your PR next week. We have to be very careful in case you changed something that impacts several (or maybe even all) widgets.
And please also check the result with condensed layout enabled, and on different sizes of screen.

GeVaSta:
> when the label is empty, the buttons will take up the space normally reserved for the label.

This behaviour did not work. The wrapping occurs on exactly the same width, independant from the presence of a label. I tried both 'label="[]"' and label="". After installing new jar and refreshing browser, it stayed also the same. 

This is the smallest I get before wrapping.
Clipboard_08-07-2024_01


> If we changed this, I'm pretty sure there'd be people who wouldn't like it and prefer to keep the old behaviour too :)

In the IOS app the Capitalation is not present either. When someone likes Capitalized buttons, he can enter the text in the Sitemap Capitalized.

jimtng:
> We have to be very careful in case you changed something that impacts several (or maybe even all) widgets.

Is there a "torture test" sitemap to be used to test all types / combinations of widgets?

jimtng:
I presume you did a force refresh

@GeVaSta this is the narrowest I can get before wrapping:
image

jimtng:
> In the IOS app the Capitalation is not present either. When someone likes Capitalized buttons, he can enter the text in the Sitemap Capitalized.

Indeed, also on Android, it's not capitalized. I'm happy to submit a separate PR and it's up to the maintainers to approve or reject it.

jimtng:
@GeVaSta I've created a PR for making the button text not uppercased here: #2697

GeVaSta:
> I presume you did a force refresh

Sure I did. And also checked all kind of variations with the label. Also removed other mappings/buttons on the same page to prevent possible influence.

jimtng:
@GeVaSta I've just made an unrelated change (to fix buttongrids), but in case I uploaded an older jar, please try this one now

[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16523238/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

GeVaSta:
>  I've just made an unrelated change (to fix buttongrids), but in case I uploaded an older jar, please try this one now

Yes, much better now!
Together with the possibility of Lowercase buttontext, it is perfect for me.
Clipboard_08-07-2024_02
Thanks a lot!

One thing I noticed (not impportant to me however) when I use the condensed layout with Desktop 2 columns, the smallest I can get is like below in the screenshot. Looks like there, the not present label-space is not taken for the buttons. I little smaller windows and the layout swaps to one column instead of using the labelspace.
Clipboard_08-07-2024_04

GeVaSta:
I noticed an unintented negative sideffect.  Only one button gets cropped on the right side.
Clipboard_08-07-2024_05

jimtng:
> I noticed an unintented negative sideffect. Only one button gets cropped on the right side.

Please try this next jar
[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16534392/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

jimtng:
> One thing I noticed (not impportant to me however) when I use the condensed layout with Desktop 2 columns, the smallest I can get is like below in the screenshot. Looks like there, the not present label-space is not taken for the buttons. I little smaller windows and the layout swaps to one column instead of using the labelspace.

This is because the threshold for single/double column is reached first. Here's an example with more buttons to take up the space in 2 column mode.
image

GeVaSta:
> > I noticed an unintented negative sideffect. Only one button gets cropped on the right side.
> 
> Please try this next jar [org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16534392/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

I can confirm the cropping is fixed in this jar. Thanks.

> This is because the threshold for single/double column is reached first.

Ok, clear.

jimtng:
> I can confirm the cropping is fixed in this jar. Thanks.
> 
> >

Great! Thanks for testing @GeVaSta! Do you think I can mark this PR ready for review?

GeVaSta:
> Do you think I can mark this PR ready for review?

I think so.

One thing I noticed while testing on my IOS device (iPad), but this has probably nothing to do with this issue. I can’t change the settings using the gear icon (left-top), by just tapping a button.  I tried changing the number of colums, but this did not work. However, double tapping changed the setting.

jimtng:
@GeVaSta the issue with a single button being truncated got me thinking that it could also happen to multiple buttons where each / every button has very wide text that only one button can fit per row, and sure enough, the same problem occurred.

I've made some changes to address this.

Would you mind testing this jar with the fix just to be sure that everything else is still working?

[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16543235/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

jimtng:
> One thing I noticed while testing on my IOS device (iPad), but this has probably nothing to do with this issue. I can’t change the settings using the gear icon (left-top), by just tapping a button. I tried changing the number of colums, but this did not work. However, double tapping changed the setting.

I'm guessing that this is an unrelated issue. If you think it's a bug, please open an issue.

lolodomo:
> I'm guessing that this is an unrelated issue. If you think it's a bug, please open an issue.

An issue already exists I believe.

GeVaSta:
> Would you mind testing this jar with the fix just to be sure that everything else is still working?

I tested your latest jar and everything looks fine to me. Could not find any ommisions as far as I am concerned.

jimtng:
@lolodomo this is now ready for review

jimtng:
@GeVaSta I found an issue with mapping buttons not lining up with setpoint controls (and perhaps others). So I've moved the right padding into the controls container and not a special case for just multiline buttons. Here's the new jar with the padding update. It added a few extra pixels of padding to the right of the buttons.

[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16567872/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

GeVaSta:
Looks ok to me. One thing I don't understand is why in one situation with longer text on the buttons, there is much less space taken from the label before wrapping occurs.
I made two examples side to side. The first screenshot with small text buttons and very little space left befire wrapping.
The second with wider text buttons and quite a lot of space left (between icon and first button) before wrapping occurs.
Clipboard_08-10-2024_02

Clipboard_08-10-2024_01

After some more testing it showed up this behaviour occured due to the single Letter labeltext in de second case. Without a labeltext at all, there is much more space taken from the labelposition than with one (or probably more) letter-labeltext.

jimtng:
That is correct. There is a special case for when the label is empty. I haven't tried to optimise it with super short labels. However that could be something I'll try next

GeVaSta:
👍

jimtng:
@GeVaSta Please try this jar
[org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt](https://github.com/user-attachments/files/16571163/org.openhab.ui.basic-4.3.0-SNAPSHOT.jar.txt)

There are still 10 extra pixels left in the padding for the value field that can be taken away if the value is empty, and I haven't taken it away here, but I hope what we've got now is good enough.

Sitemap to demonstrate it:
```java
    Switch item=TestString1 icon="light" label="[]" mappings=["long"="1", "longer"="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="t[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="t1[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="t12[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="t123[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="t1234[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="t1234[]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="Not so long[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]
    Switch item=TestString1 icon="light" label="Test with a long label to see where it clips[%s]" mappings=[1="1", 2="2", 3="3", 4="4", 5="5", 6="6"]

    Switch item=TestString1 icon="light" label="[]" mappings=["Long"="This is a Long label", "Longer"="The quick brown fox jumps"]
    Switch item=TestString1 icon="light" label="[%s]" mappings=["Long"="This is a Long label", "Longer"="The quick brown fox jumps"]
    Switch item=TestString1 icon="light" label="test label [%s]" mappings=["Long"="This is a Long label", "Longer"="The quick brown fox jumps"]

    Default item=TestPlayer1 label="Test with a long label to see where it clips [%s]"

```

GeVaSta:
Yes, this is very nice. I looks a lot better (at least for the way I use the sitemap). Found no issues.

Your remark
> There are still 10 extra pixels left in the padding for the value field that can be taken away if the value is empty, and I haven't taken it away here, but I hope what we've got now is good enough.

wonders me a little. Why would there be any need of this reserved space if no value is configured in the sitemap.
It is or it isn't configured. In both cases the sitemap maker will adjust the window or sitemap to display the sitemap at will.
It can't be (I think) there suddenly popups a value for the item where reserved space is neccessary for. It is configured or not!

But  also without these extra 10 pixels it looks already nice for me. Thanks.

jimtng:
OK I'll leave the extra 10 pixels alone. The added code+effort to remove them may not be worth the extra benefits.

@lolodomo this is once again ready for review. Thanks @GeVaSta for refining this.

lolodomo:
> @lolodomo this is once again ready for review

I will review only in few days when I am back home. Can you please check that the screen captures in the initial message are still uptodate ?

jimtng:
the screenshots are still applicable, except for the uppercased button text

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1337,8 +1339,37 @@
...
+				// Start from the maximum width (limited by `mdl-form__buttons-multiline.max-width`),
+				// then shrink it down to the minimum without causing additional wrapping.
+				var buttons = _t.parentNode;
+				var width = buttons.parentElement.offsetWidth + 100; // start wider than max-width allows
+				buttons.style.width = width + "px";
+				var height = buttons.offsetHeight;
+				while (buttons.offsetHeight === height && width >= maxButtonWidth) {
+					buttons.style.width = --width + "px";
+				}
+				buttons.style.width = (width+1) + "px";
```
lolodomo, line 1364: Maybe we could have a variable and set "buttons.style.width" only at end ?

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1361,6 +1392,7 @@
 				}
 			});
 			componentHandler.downgradeElements(_t.buttons);
+			window.removeEventListener("resize", _t.minimizeWidth);
```
lolodomo, line 1395: Maybe remove it only if it was added first ?
That would mean adding the test on `_t.buttons.length > 1 && _t.parentNode.classList.contains(o.buttonsMultilineClass)`

bundles/org.openhab.ui.basic/web-src/smarthome.js
jimtng: No, that's not possible. We need to _actually_ set and decrease the width incrementally, and we want this to be applied to the actual element, incrementally, so that when it's too narrow, the actual offset height would change, which indicates that the buttons have now wrapped around a new row. That's why we then back off and re-add the width to prevent the wrapping. This can only be done like this that I know of to figure out the breakpoint / threshold.

bundles/org.openhab.ui.basic/web-src/smarthome.js
jimtng: Is there any harm in removing it like this? If _t.minimizeWidth is null, [removeEventListener()](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener) doesn't mind.

> Calling removeEventListener() with arguments that do not identify any currently registered [event listener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback) on the EventTarget has no effect.

lolodomo: Ok, thank for the link.

lolodomo:
Remains to test intensively on different kind of devices, with and without condensed layout. I see in particular that we could have impacted icon positioning on phone when in condensed mode ?

jimtng:
What do you mean by impacted? From memory, the paddings / margins got reduced to squeeze more space out, but nothing looked bad / wrong to me.

bundles/org.openhab.ui.basic/web-src/_layout.scss
```diff
@@ -212,19 +212,20 @@
...
 			}
 		}
-		padding-left: $form-row-desktop-padding;
-		html.ui-layout-condensed & {
-			padding-left: $form-row-desktop-padding-condensed;
-		}
 		@include flex-shrink(0);
 		@include flex-grow(2);
 		@include flex-2011(2 2 auto);
+		min-width: 5em;
```
lolodomo, line 228: What is the impact of this added line on all other sitemap elements ?

lolodomo:
> What do you mean by impacted? From memory, the paddings / margins got reduced to squeeze more space out, but nothing looked bad / wrong to me.

I am referring to your change at line 171.

jimtng:
> I am referring to your change at line 171.

IIRC (as mentioned in the comment), there's already sufficient padding from the enclosing element.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1337,8 +1339,37 @@
 				icon.addEventListener("load", _t.convertToInlineSVG);
 				icon.addEventListener("error", _t.replaceImageWithNone);
 			}
+
+			if (maxButtonWidth < button.offsetWidth) {
+				maxButtonWidth = button.offsetWidth;
+			}
 		});
 
+		if (_t.buttons.length > 1 && _t.parentNode.classList.contains(o.buttonsMultilineClass)) {
```
lolodomo, line 1348: isn't your minimize code also expected when there is only one button ?
As I understood, this will provide more place for label (and value).

bundles/org.openhab.ui.basic/web-src/_layout.scss
jimtng: tbh, I'm not sure. From my limited testing with my own sitemaps, nothing looked weird.

Ideally we need to have some form of "unit test", e.g. a standard sitemap which has a rigorous set of combinations of elements so that when we try to render it, we could determine, at a glance, that everything "still looks ok" / not broken. That would have greatly aided in answering questions such as this.

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: Maybe not because the height will never change if we have only one button ?

jimtng: hmm it has been a while since I worked on this, but from my vague memory, there _was_ a reason why it was only applied to > 1 buttons. Oh yes, it came to me just now... 1 button cannot "wrap" around, so it's not possible to detect the breakpoint in the manner it's done in the loop above.

jimtng: Perhaps I should add a comment to remind myself (and explain to others) about this in the future?

lolodomo: But buttons.style.width will already be optimized in that case, no ?
I will include in my tests the case of only one button to be sure that the label is not smaller than it should be.

jimtng: > But buttons.style.width will already be optimized in that case, no ?

Not through js, but through the various tweaks (statically) made on the scss.

bundles/org.openhab.ui.basic/web-src/_layout.scss
lolodomo: Do you remember what is the default value if not set ?
PS: I can check that.

jimtng: I believe it would've been 0 if not set specifically

lolodomo:
Here is a first feedback after first tests with my usual sitemap tested on my phone and my desktop PC.
First, I would like to say that you improved the rendering, no contest.

I encountered only one real bug, that is the rendering of the Player item which is the only element that is forced to one unique line. We are back to the previous bug that the last button is partially outside its expected position and can override the next column. It is the case on phone in non condensed layout. It was the reason of my initial change. Maybe we should restore a fixed place for the Player item ?
image

IMHO the space between each button is ultra small in condensed layout, especially on hi-res phone. You used 2 pixels instead of 4 in non condensed ? Maybe we could make a try with 3 pixels. WDYT ?

You reduced the minimum size of buttons in condensed layout, 40 pixels I believe. We have to take care that it is enough to select such button with a finger on a phone. That's currently fine for the player buttons or a button with label "Bas", I will have to test for example with a label "i" or "1". That being said, your idea to reduce the minimum size is good and I would like to apply it to the non condensed layout, but not so hardly, just something that is a little smaller that the current MDL default. This will allow more buttons on the same line. WDYT ?

I have cases when using bigger font option where I have now more lines than before on my phone but it is in favour in a non truncated label. It does not annoy me, I just hope it is not a bug.

I have also a case with two different items with the same buttons but a different label. The one with longer label has buttons on 2 lines while the other has all buttons on one unique line. I have the feeling that there is enough place for the same rendering on one unique line. I will have to provide a capture from my phone later.

Finally, on my desktop with Firefox and Chrome in full screen and same settings, I see the label for the element at right truncated in Firefox while not truncated in Chrome. Any idea why ? It is in not condensed layout.

image

image

jimtng:
> I encountered only one real bug, that is the rendering of the Player item which is the only element that is forced to one unique line. We are back to the previous bug that the last button is partially outside its expected position and can override the next column. It is the case on phone in non condensed layout. It was the reason of my initial change. Maybe we should restore a fixed place for the Player item ?

AFAIK, I didn't make changes (outside the global reduction in paddings) that would affect playeritem. So if you could provide a sitemap for me to test the issue, I'd like to take a look.



> IMHO the space between each button is ultra small in condensed layout, especially on hi-res phone. You used 2 pixels instead of 4 in non condensed ? Maybe we could make a try with 3 pixels. WDYT ?

Can you tell me how it's viewed on the phone? Is it via a browser, or the app (android / ios)? 

To clarify, you're suggesting to increase the gap for the condensed version from 2 to 3, and leave the gap for non-condensed at 4?



> You reduced the minimum size of buttons in condensed layout, 40 pixels I believe. We have to take care that it is enough to select such button with a finger on a phone. That's currently fine for the player buttons or a button with label "Bas", I will have to test for example with a label "i" or "1".

OK, please let me know what you think after you've tested it.



> That being said, your idea to reduce the minimum size is good idea and I would like to apply it to the non condensed layout, but not so hardly, just something that is a little smaller that the current MDL default. This will allow more buttons on the same line. WDYT ?

Not a bad idea. I'll push this change shortly.



> I have also a case with two different items with the same buttons but a different label. The one with longer label has buttons on 2 lines while the other has all buttons on one unique line. I have the feeling that there is enough place for the same rendering on one unique line. It is in condensed layout with bigger font. WDYT ?

I think there's some space reserved for the value, and also the label width is slightly larger than it propably could be. See line 1352 in smarthome.js, when the label length is < 6 characters, it would optimise the width. If it's longer, it won't do it. It's a tradeoff between that and getting it too wide.

It's not an ideal strategy but I couldn't come up with a better solution.




> Finally, on my desktop with Firefox and Chrome in full screen and same settings, I see the label for the element at right truncated in Firefox while not truncated in Chrome. Any idea why ? It is in not condensed layout.
> 
> image
> 
> image

I'm not sure.

lolodomo:
> AFAIK, I didn't make changes (outside the global reduction in paddings) that would affect playeritem. So if you could provide a sitemap for me to test the issue, I'd like to take a look.

It is simply because you removed the 60% of the size dedicated to buttons.
You just need a Player item to reproduce it with relatively long label.

lolodomo:
> Can you tell me how it's viewed on the phone? Is it via a browser, or the app (android / ios)?

Browser of course as we are discussing about Basic UI. Your question looks strange, I hope I correctly understood it 

> To clarify, you're suggesting to increase the gap for the condensed version from 2 to 3, and leave the gap for non-condensed at 4?

Yes, exactly.

lolodomo:
> I'll push this change shortly.

Did you check what is MDL default?

lolodomo:
> I think there's some space reserved for the value, and also the label width is slightly larger than it probably could be.

Yes, it is probably that if we also consider paddings.

> See line 1352 in smarthome.js, when the label length is < 6 characters, it would optimise the width. If it's longer, it won't do it. It's a tradeoff between that and getting it too wide.
> 
> It's not an ideal strategy but I couldn't come up with a better solution.

I don't think that's the "problem" here. But I could check by increasing your threshold to 8 for example..
That's being said, I am not sure to understand why you apply your optimization only to short labels?

lolodomo:
Shouldn't you retrieve the height before the first update of the width with +109 ?

jimtng:
> > AFAIK, I didn't make changes (outside the global reduction in paddings) that would affect playeritem. So if you could provide a sitemap for me to test the issue, I'd like to take a look.
> 
> It is simply because you removed the 60% of the size dedicated to buttons. You just need a Player item to reproduce it with relatively long label.

I'm unable to reproduce this. Can you please paste the sitemap that can demonstrate the problem?



> Shouldn't you retrieve the height before the first update of the width with +100 ? I mean moving line 1360 before line 1358.

No.

We start from the widest possible button widths, get the height for that. If you moved the lines as you suggested, you'd be getting the height of already wrapped / messed up layout, and the rest of the code would not do what it's meant to do.




> Did you check what is MDL default?

The default seems to be 64px, I wrote it in the commit message.




Regarding the gaps - IMO, bigger gaps are more for visual effects. I don't think an extra 1px will significantly change the usability / margin of errors in clicking adjacent buttons. The width of the individual buttons would have a better impact on that, because people aren't going to tap the edge, they're going to tap the middle of the button. 

Increasing the gap however, increases the chance of the buttons wrapping, but I'm not too fussed on this and happy to increase to 3px or even back to 4px if necessary. I must say it does "look" better with 4px.

But comparing this to the android app for example, the buttons don't have any gaps inbetween, just thin vertical lines separating them.

GeVaSta:
Follow your discussion from the sideline. I hope the outcoming does not revert too many things because I am now very happy with the result for my sitemaps.
For instance, why would I need 3 pixels between the buttons instead of 2. As long as one can distinguish the buttons, I would say, thats enough and oaky with me. In the IOS app the buttons are joined together with only a black line between them and thats okay also.

lolodomo:
@jimtng : I can no more compile your branch as it is too old and the method `getFrequency` was removed from core during the weekend. Can you please rebase your branch ?

jimtng:
rebased

lolodomo:
Regarding the new min width for buttons, that is fine for me. 40 pixels in condensed layout leads to relative small buttons but acceptable for a selection with a finger on a phone. 54 pixels instead of 64 pixels in non condensed layout is fine.

lolodomo:
> I have also a case with two different items with the same buttons but a different label. The one with longer label has buttons on 2 lines while the other has all buttons on one unique line. I have the feeling that there is enough place for the same rendering on one unique line. It is in condensed layout with bigger font. WDYT ?
> 
> IMG_20240908_112425

I also tried with label "MPEG N" and the buttons are still on 2 lines and this is even more obvious with that new example that they could be on one line.
I understand what it's happening.
In the first line (label is "NR"), you set min-width to min-content. In the second line, min-width is set to 5em. 5em = 70 pixels (5*14); 90 pixels (5*18) when large font option is enabled.
When my label is "MPEG N", I estimated the real width of this text is larger font is around 75 pixels while 90 is considered as minimum. That is the reason why buttons are rendered on 2 lines. In this case, using min-content would resolve the problem.
But I understand that you want to set min-content only to short labels (and you use the number of characters to determine that) and I understand the reason (we want to cut a big label in favour of buttons).

I assume you are not able to retrieve the real width of the text (instead of its number of characters) ? If you could, then you will apply min-content if this width is < 70 pixels.

As a possible other solution, we could reduce min-width from 5em to 4em, making the hypothesis that 70 pixels is generally ok for a standard text of 5 characters ?
I will text with 4em but I am sure it will be ok in my case.
WDYT ?

jimtng:
> I assume you are not able to retrieve the real width of the text (instead of its number of characters) ? If you could, then you will apply min-content if this width is < 5em.

I can try this with javascript.

jimtng:
@lolodomo I've made a minor tweak for the min-width. It's still set to 5em in the css, but now that's the only parameter to tweak.

lolodomo:
Maybe you should not remove the property but rather restore it to 5em ?

jimtng:
It's a bit more complicated than I originally thought, but I hope it works now.

jimtng:
The force pushes were just for editing the comments to clarify the purpose of the min-content code.

lolodomo:
> done

No, you kept 2px

jimtng:
https://github.com/openhab/openhab-webui/pull/2388/commits/ff4a916ca8934a3b14f5bcca6d0029639ec80523

lolodomo:
> Finally, on my desktop with Firefox and Chrome in full screen and same settings, I see the label for the element at right truncated in Firefox while not truncated in Chrome. Any idea why ? It is in not condensed layout.
> 
> image
> 
> image

Apparently (using dev tools), the size for the widget is a little bigger in case of Chrome: 507.86 vs 506.667.
 When I look at global page, width is 1603.62 in Chrome vs 1600 in Firefox.
So it explains the result.

lolodomo:
I take a look on the possible impact of your changes regarding mdl-form__row for a Buttongrid and a Chart.
No rendering change noticed in non condensed layout.
In condensed layout, I can see that the grid/chart has been moved to the left, that means the border at left is now smaller than at right. Before your change, it was not yet perfect but the reverse, bigger border at right than at left. In both cases, one of the border is too big, compared to non condensed layout. With your change, this is the border at right that is now too big.
In non condensed layout, the centering looks very good.
@jimtng : Even if this is not dramatic (and not new), would you like to try to fix that ?

lolodomo:
My tests with my usual sitemap are finished and the only bug that remains is the re-introduction of the bug with the Player item (display of button,s forced on one line) that can lead to buttons moved too much to the right. I will provide a sitemap to to reproduce but a simple Payer item should be sufficient to reproduce it. Setting bigger font option shows even more the problem.

bundles/org.openhab.ui.basic/web-src/_layout.scss
lolodomo: We have to see what happens with a small label for other widgets having a big control block. I tried with a slider with value displayed at left and that already looks good.

lolodomo:
The additional tests I would like to execute are when the value is displayed at left of the buttons and when icons are used inside buttons.

jimtng:
> In condensed layout, I can see that the grid/chart has been moved to the left, that means the border at left is now smaller than at right. Before your change, it was not yet perfect but the reverse, bigger border at right than at left. In both cases, one of the border is too big, compared to non condensed layout. With your change, this is the border at right that is now too big.
> In non condensed layout, the centering looks very good.
> @jimtng : Even if this is not dramatic (and not new), would you like to try to fix that ? If not, could be the object of a next PR.

Could you please provide a sitemap for me to test?



> My tests with my usual sitemap are finished and the only bug that remains is the re-introduction of the bug with the Player item (display of button,s forced on one line) that can lead to buttons moved too much to the right and overriding the widget at right. I will provide a sitemap to reproduce it but a simple Payer item should be sufficient to reproduce it when in non condensed layout. Setting bigger font option shows even more the problem.

I also need a sitemap to show this issue.

lolodomo:
> My tests with my usual sitemap are finished and the only bug that remains is the re-introduction of the bug with the Player item (display of button,s forced on one line) that can lead to buttons moved too much to the right and overriding the widget at right.

I found what is your change that re-introduced the bug, it is a direct consequence to set min-width to 5em for the =label. We are in the case that this minimum is too much and there is then no enough place to show the 4 player buttons.
If I reduce this min size to 1em for example, then it works well.

It is exactly what I said in my comment about this adding a min width, we have to be sure that it is without any consequence on any widget. It has on the Player item.

Can you tell me if there is a real need to add such minimum width for the label ?

lolodomo:
> Can you tell me if there is a real need to add such minimum width for the label ?

In case it is usefull for multiline buttons, why not rather setting it only in JavaScript code and only for this widget when we are in multilines mode ?
It would have the advantage that we are certain that change will not impact other widgets.
WDYT ?

jimtng:
min-width for label is absolutely needed for the multiline buttons, because we removed the max-width on the buttons.

Would reducing the min-width work for the player? I would prefer that than setting it in javascript (or css) only for multiline buttons.

I don't care what it's reduced to, as long as it's set to something that's not too small that would cause the labels to be unreadable.

lolodomo:
> min-width for label is absolutely needed for the multiline buttons, because we removed the max-width on the buttons.

Ok

> Would reducing the min-width work for the player? I would prefer that than setting it in javascript (or css) only for multiline buttons.

Not sure to understand why you prefer this option.

> I don't care what it's reduced to, as long as it's set to something that's not too small that would cause the labels to be unreadable.

I should then experiment what could be a reasonable value.

jimtng:
@lolodomo in the latest change, I only set the min-width for multiline buttons. Could you please test it against the player item?

lolodomo:
> @lolodomo in the latest change, I only set the min-width for multiline buttons. Could you please test it against the player item?

Very good option. I will tell you if it works but I am already sure it will.

lolodomo:
```
[INFO] [15:50:42] Starting 'eslint'...
[INFO] [15:50:42] Starting 'snippetsAll'...
[INFO] Error in plugin "sass"
[INFO] Message:
[INFO]     web-src\_layout.scss
[INFO] Error: expected ':' after $--label-multiline in assignment statement
[INFO]         on line 235 of web-src/_layout.scss
[INFO]         from line 9 of web-src/smarthome.scss
[INFO] >>       $__label-multiline {
[INFO]
[INFO]    -^
[INFO]
```

lolodomo:
Typo at line 235: you used $ intsead of &

jimtng:
> Typo at line 235: you used $ intsead of &

Thanks! I was scratching my head!

lolodomo:
> Thanks! I was scratching my head!

What is strange is that the build was not failing despite this error. This is something we have to fix (in another PR).

lolodomo:
The bug is fixed, the rendering of the player item is fine on my phone either in portrait or paysage (2 columns) modes whatever the values of "condensed layout" and "bigger font size" parameters.

To be precise, it is still possible to reproduce the bug but it is not the consequence of any of your changes, it was already possible before. On a desktop screen with 3 columns, if you reduce the window width, at a certain time there is no more place to display the label at all and the buttons start to shift to the right to be fully rendered. In fact, the solution would be to use multiline buttons once there is no more place for the label. But this problem will be encountered in very rare cases so I propose to ignore it at this time.  Another PR could be added later to switch from mono-line to multi-line in the JavaScript when the label size is 0.

bundles/org.openhab.ui.basic/web-src/_layout.scss
lolodomo: This line has now been removed.

lolodomo:
When using icons in buttons, I have a cut in two lines when it could be displayed on one line (was before your change).

image

Before:

image

Sitemap:
```
frame {
Switch item=TestCommand mappings=[0=NULL,1=UNDEF,2="Set 1",3="Set 2"]
Switch item=TestCommand mappings=[0=NULL=f7:house,1=UNDEF="material:favorite",2="Set 1"=if:lucide:activity-square,3="Set 2"=chart]
Switch item=TestCommand mappings=[0=NULL=f7:house,1=UNDEF,2="Set 1"=if:lucide:activity-square,3="Set 2"]
}
```

This is in non condensed layout with standard font size.

lolodomo:
Another annoying case: look at the widget on laster line in the middle. Its size is bigger as expected at left and right. On my phone, I can see that it can also happen even without the display of the value, that is for the widget just before (last line at left). This is in non condensed layout with bigger font size.

image

Before:

image

Sitemap:
```
    Frame {
        Switch item=DemoString mappings=[ VALUE1="Val 1", VALUE2="Value 2"]
        Switch item=DemoString label="Demo String [%s]" mappings=[ VALUE1="Val 1", VALUE2="Value 2"]
        Switch item=DemoString mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3"]
        Switch item=DemoString label="Demo String [%s]" mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3"]
        Switch item=DemoString mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3", VALUE4="Value 4"]
        Switch item=DemoString label="Demo String [%s]" mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3", VALUE4="Value 4"]
        Switch item=DemoString mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3", VALUE4="Value 4", VALUE5="Val 5", VALUE6="Long value 6", VALUE7="An ultra mega long value 7"]
        Switch item=DemoString label="Demo String [%s]" mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3", VALUE4="Value 4", VALUE5="Val 5", VALUE6="Long value 6", VALUE7="An ultra mega long value 7"]
        Default item=DemoString label="Demo String [%s]"
    }
```

jimtng:
> When using icons in buttons, I have a cut in two lines when it could be displayed on one line

I'm investigating this


> Another annoying case: look at the widget on last line in the middle.

It seems that the "after" has a min-width that's wider than the "before". The advantage is we can read more of the label, at the expense of a narrower space for the buttons.

If you prefer to sacrifice the space for the label, then we can simply reduce the min-width.

> Its size is bigger than expected at left and right

Can you please clarify this?

lolodomo:
> Can you please clarify this?

Look the alignment of icons at left and the buttons at right. They are no more aligned.

jimtng:
> > When using icons in buttons, I have a cut in two lines when it could be displayed on one line
> 
> I'm investigating this

OK this one is solved and will be in the next commit. When using (some types of) icons, the icons weren't loaded yet when the minimizeWidth() was initially executed, resulting in an incorrect calculation.





> Another annoying case: look at the widget on last line in the middle. Its size is bigger than expected at left and right.

This is a result of the label's min-width. If you compare to "before" notice the label's size was smaller to compensate.

However, try make your button even wider, e.g. 
```
        Switch item=DemoString label="Demo String [%s]" mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3", VALUE4="Value 4", VALUE5="Val 5", VALUE6="Long value 6", VALUE7="An ultra mega long value 7 supercalifragilisticexpialidocious"]
```

Even the "before" will also suffer the same problem, although the "after" version is worse because buttons have no max-width to limit it.

I'm not sure of the best way to solve this because either way, something needs to either getting truncated, or pushed around. Perhaps truncation is better because it won't affect the neighbouring controls.

lolodomo:
> Perhaps truncation is better because it won't affect the neighbouring controls.

Yes I believe.

jimtng:
@lolodomo please test the latest changes

lolodomo:
@jimtng : the last version is worst I think because now labels are truncated even when there is place to not truncate them.

image

Maybe we should switch to your previous version which not so bad even if not perfect. WDYT ?

GeVaSta:
Guys, any idea when this will be finished? 
Today I installed the openHAB 4.3.0.M2 version and noticed immediately the progress you have made had disappeared.

jimtng:
I'm on holiday until mid November

lolodomo:
@jimtng : are you back ?

jimtng:
> @jimtng : the last version is worst I think because now labels are truncated even when there is place to not truncate them.
> 
> image

I'm back!

I had a look at this again, rebased my branch and recompiled without any changes, and I can't see any problems.

In your screenshot above, is it the "Espa..." that you think behaves the wrong way? IMO, it's correct because the last button wouldn't fit in one line. Yes, after it's wrapped, it "seems" that there is a lot of space but it's just not quite wide enough to accommodate that button.

lolodomo:
> In your screenshot above, is it the "Espa..." that you think behaves the wrong way?

Bot at all! The current problem is that now all my labels are truncated and only 3 or four characters are shown. This happens even when there is a lot of place to show the full label. Look at my screenshot, except for the label "NR", all others are truncated.

jimtng:
> The current problem is that now all my labels are truncated and only 3 or four characters are shown. This happens even when there is a lot of place to show the full label. Look at my screenshot, except for the label "NR", all others are truncated.

I tried but cannot reproduce this. Can you paste me a sitemap that I can use to test it?

GeVaSta:
I noticed something very strange. At every screen update the values of a setpoint label, move from left justified to right justified. 
see the two pictures of every state. 
IMG_2032
IMG_2031

jimtng:
> I noticed something very strange. At every screen update the values of a setpoint label, move from left justified to right justified.

did this happen on this branch's build but not on the main milestone / snapshot?

If so, can you please provide a sitemap for me to test, along with details about your basicui settings.

GeVaSta:
Wait! False Alarm. This problem has nothing to do with your adaptions in the Basic UI. 
The problem I have are in the IOS app! So why I bothered you stays unclear.
Will report it in IOS branch. Sorry.

lolodomo:
> I tried but cannot reproduce this. Can you paste me a sitemap that I can use to test it?

I'm terribly annoyed with myself, I can't reproduce it myself !!! I don't know what I did !
I will run again final tests.

lolodomo:
Ok, I found how to reproduce it. Just open the Basic UI preferences, change nothing, go bnack and finally come back to your page. Labels are then truncated.
If I do "Ctrl>+F5", labels are then correct again.

lolodomo:
There is even no need to open the preferences, just changing page and coming back shows the problem.

lolodomo:
Now I have non truncated labels only I do "Ctrl+F5" in any page.
It is not specific to browser, I reproduce it with Firefox and Chrome.

lolodomo:
Ok, here is how to reproduce the problem in IDE with the Demo sitemap.

Update the demo sistemap by adding two new elements:
```
sitemap demo label="Demo Sitemap" {
	Frame label="Demo Items" {
		Text item=DemoContact label="Contact [MAP(en.map):%s]"
		Text item=DemoString label="Message [%s]"
		Text item=DemoDateTime label="Date [%1$tA, %1$td.%1$tm.%1$tY]"
		Group item=DemoSwitchGroup icon=icon1
                Switch item=DemoString label="Switch demo long label" mappings=[1="Value 1", 2="Value 2", 3="Value 3"]
		Switch item=DemoString label="Switch demo long label" mappings=[1="Value 1", 2="Value 2", 3="Value 3", 4="Value 4", 5="Value 5", 6="Value 6"]
	}
	Frame label="Magic Test Items" {
	   Colorpicker item=MagicColor
	   Switch item=MagicOnOff
	   Slider item=MagicDimmer
	}
	Frame label="Location" {
		Mapview item=DemoLocation height=10
	}
```

Starts openHAB. Here is the initial rendering that is fine for buttons rendering:

image

Now click on DemoSwitchGroup to open tyhe sub page and go back to the main page. Here is the new rendering with truncated labels and different dispatching of buttons on the two lines !

image

Now click Ctrl+F5 and the initial and correct rendering is back.

jimtng:
It seems that onload event didn't fire when navigating to a subpage (Text element) and back. I've added an extra call to run minimizeWidth immediately.

GeVaSta:
Hmm. Yesterday I wrote an additional comment but one or the other way it disappeared. 
My issue about the Setpoint justification is ‘false alarm’. The issue I mentioned is only in the IOS app and not in the Web UI. My mistake.

lolodomo:
My tests with my current sitemap in production are all positive now.

Will run again tests with specific cases leading to your changes in september and start again a final review.

lolodomo:
> Will run again tests with specific cases leading to your changes in september

Done.
There is still one case on my phone with bigger font set and value displayed where buttons are shifted too much to the right.
But I think you have probably no choice, you defined a minimum size for the label and that is expected.
So for me, tests are considered as OK.
By the way, all this is better than before.

Remains to do a new quick review.

lolodomo (approved):
LGTM, thank you

lolodomo:
I will have to check if this does not impact the new buttons containing a square color (Colorpicker and Colortemperaturepicker widgets).

lolodomo:
Thank you @jimtng for this work and your patience.

jimtng:
> There is still one case on my phone with bigger font set and value displayed where buttons are shifted too much to the right.

Let me know how to reproduce that and I'll have a play with it. It'd be much easier (I hope) now that this is merged, thanks!

lolodomo:
> Let me know how to reproduce that and I'll have a play with it. It'd be much easier (I hope) now that this is merged, thanks!

With the following element:
```
Switch item=DemoString label="Demo String [%s]" mappings=[ VALUE1="Val 1", VALUE2="Value 2", VALUE3="Val 3", VALUE4="Value 4", VALUE5="Val 5", VALUE6="Long value 6", VALUE7="An ultra mega long value 7"]
```
Select the second button to have a value and on a a phone, set bigger font as preference.
If I correctly remember, you explained that there is a case while due to label and value, you cannot prevent from shifting the buttons.

jimtng:
This? 
image

lolodomo:
Yes, exactly

PR #2391, opened 2024-02-22
Persistence edit: Change empty-state-placeholder icon

mherwege:
Change the icon shown when no persistence is configured to be the same as the icon on the settings page.


florian-h05 (approved):
LGTM, thanks!

PR #2392, opened 2024-02-22
Item Details: left align Semantic Classification values

jimtng:
before
image

after
image


florian-h05:
I would tend not to align them left as long as there are no other left-aligned fields/items.

jimtng:
Right alignment just looks weird. Your eyes need to bounce back and forth from extreme left to right

ghys:
Never had the complaint before... it's just a standard ios-style label/value list that every Apple user will be familiar with.
Besides inset inputs have even been removed from later versions of Framework7 (along with the desktop theme, sadly).

image

jimtng:
> it's just a standard ios-style label/value list that every Apple user will be familiar with

This is a good point. It seems the same on android too. Thanks for pointing this out to me 😁

PR #2401, opened 2024-02-23
Fix issues with unchangeable page after cancelled dirty warning

florian-h05:
/cc @jimtng 
This seems to fix the issue, however now the browser history behaves different.


jimtng:
Was there an alternative solution?

florian-h05:
No, this was closed by accident.

florian-h05:
Closing as I haven't been able to reproduce the issue lately and this is likely a bug in F7 router, which will be fixed by migrating to Vue router instead.

PR #2402, opened 2024-02-23
Add support for SVG add-on logos

florian-h05:
Currently, Main UI attempts to load all add-on logos as PNG (the add-on logo url is generated using the add-on name and expecting a PNG).

This PR adds support for SVG add-on logos, which means: 
The UI will now attempt to load a SVG logo, 
if that fails, it will attempt to load a PNG, 
and if that fails, fall back to the add-on‘s default logo (which is dependent on the add-on type).


florian-h05:
@stefan-hoehn I have added a description. 
BTW I have taken this PR as a chance to refactor the add-on logo to its own component.

stefan-hoehn (approved):
I have tested it on final and it retrieves the svgs nicely. Styling looks good and I see nothing that breaks.

PR #2403, opened 2024-02-24
Material theme: Fix lock icon is visible on editable items

jimtng:
Regression from one of the previous PRs.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
@@ -6,7 +6,7 @@
         <f7-link @click="save()" v-if="editable && !$theme.md">
           Save<span v-if="$device.desktop">&nbsp;(Ctrl-S)</span>
         </f7-link>
-        <f7-link v-else icon-f7="lock_fill" icon-only tooltip="This Item is not editable through the UI" />
+        <f7-link v-if="!editable" icon-f7="lock_fill" icon-only tooltip="This Item is not editable through the UI" />
       </f7-nav-right>
     </f7-navbar>
     <f7-toolbar tabbar position="top">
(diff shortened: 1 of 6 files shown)


florian-h05 (commented):
I guess it should probably be: v-if, v-else-if, v-else?

jimtng:
> I guess it should probably be: v-if, v-else-if, v-else?

Good idea! I'll push the change shortly

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
```diff
@@ -2,11 +2,11 @@
   <f7-page @page:afterin="onPageAfterIn">
     <f7-navbar :title="createMode ? 'Create New Item': (editable ? 'Edit Item' : 'Item Details')" :back-link="editable ? 'Cancel': 'Back'">
       <f7-nav-right>
```
florian-h05, line 4: We probably should add a `v-if="ready"` here, because currently the lock is displayed for a short period of time.

jimtng: Good catch!

jimtng: oops I used v-show. Should I change it to v-if?

florian-h05 (approved):
LGTM, thanks!

florian-h05:
Hmm, I just read your comment about flickering, and it is not fixed now, but it is also really difficult to avoid as we would need a placeholder, but this placeholder would need a variable length.
Tip: Open the browser dev tools, network tab, limit bandwidth. This way you can easily check flickering caused by loading.

jimtng:
> Hmm, I just read your comment about flickering, and it is not fixed now, but it is also really difficult to avoid as we would need a placeholder, but this placeholder would need a variable length. Tip: Open the browser dev tools, network tab, limit bandwidth. This way you can easily check flickering caused by loading.

I see, thanks for the tip about that bandwidth limit. On mobile emulation there's also settings for "low end mobile" that does something similar.

Let me see if this can be avoided.

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
florian-h05, line 1: For persistence edit, it is still flickering, I will fix that :+1:

florian-h05 (approved):
LGTM, thanks!
One minor comment which I will take care of myself.

PR #2406, opened 2024-02-24
Add-on store: Title and searchbar focus tweaks

jimtng:
- Display specific title for each tab, but avoid becoming too long on smaller screens
- Don't autofocus on the search bar on mobile because it opens up the on-screen keyboard and covers up half the screen
~~- Add tooltip to the bottom tabbar on narrow screens~~

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/addon-store.js
+++ b/bundles/org.openhab.ui/web/src/assets/addon-store.js
@@ -8,7 +8,7 @@ export const AddonIcons = {
   voice: 'chat_bubble_2_fill'
 }
 
-export const AddonTabs = {
+export const AddonTitles = {
   binding: 'Bindings',
   automation: 'Automation',
   transformation: 'Transformations',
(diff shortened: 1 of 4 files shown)


florian-h05:
> Add tooltip to the bottom tabbar on narrow screens

I don't think this is a good idea, as having tooltips on clickable links is usually not working very well with touch devices.

jimtng:
> > Add tooltip to the bottom tabbar on narrow screens
> 
> I don't think this is a good idea, as having tooltips on clickable links is usually not working very well with touch devices.

BTW, the little buttons on the bottom of the blockly editor screen also have tooltips.

florian-h05:
> BTW, the little buttons on the bottom of the blockly editor screen also have tooltips.

Yes, but these buttons are also displayed on desktop, where the tooltip is usable. The toolbar of the store is only displayed on mobile, where it is not really usable.

florian-h05 (approved):
Thanks for the PR, however I think we can achieve the same without adding more data to addon-store.js. I will push a commit for that.

PR #2408, opened 2024-02-24
Item edit: Fix dirty warning for not-editable Items

florian-h05:
Follow-up for one of the previous PRs.

Fix Item dirty warning for:
- Item not editable, but warning still displayed
- editable group Item

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -174,8 +174,6 @@ export default {
   },
   mounted () {
     if (!this.item) return
-    if (!this.item.category) this.$set(this.item, 'category', '')
-    if (!this.item.groupNames) this.$set(this.item, 'groupNames', [])
     const categoryControl = this.$refs.category
     if (!categoryControl || !categoryControl.$el) return
     const inputElement = this.$$(categoryControl.$el).find('input')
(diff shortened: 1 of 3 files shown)


jimtng:
seems to have fixed it thanks

bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -174,8 +174,6 @@ export default {
   },
   mounted () {
     if (!this.item) return
-    if (!this.item.category) this.$set(this.item, 'category', '')
```
florian-h05, line 177: @jimtng This fixes what you reported in another PR.
I have removed these two lines you added in your Item field alignment PR and haven't seen any misfunctionality (note that I updated toCode() to have them in the code tab, no matter they are defined or not.)

florian-h05:
Apart from group items, I am currently working on fixing them as well.

bundles/org.openhab.ui/web/src/components/item/item-form.vue
jimtng: IIRC the item.category was already there, I added the second line. Removing them here is better anyway, because this file isn't responsible for this.item. The caller of this component should've fixed that up before passing `item` to it.

florian-h05: Missing category and groupName don't seem to be a problem to me, I see no warnings logged or UI misfunctions.

jimtng: I'm guessing that it might have only been a problem in the data() which you fixed by adding `|| []`

florian-h05: Yes, seems so.
I will merge this and we will see if any new problems arise.

PR #2409, opened 2024-02-25
Settings menu: Fix search bar to include system and addons settings

jimtng:


florian-h05 (approved):
LGTM, thanks for the fix!

PR #2411, opened 2024-02-25
[blockly] Provide HTTP request block

stefan-hoehn:
This is finally the missing HTTP Request block for Blockly.

In this case I tried to make the block as concise as possible. Therefore it contains to block mutation switches that allow hide/unhide timeout and headers if they are needed or not.

image
image
image


jimtng:
Can GET also set headers?

Another thing that would be handy is a query builder, ie given a dictionary of key value, append `key=Val&key=val` to the url

stefan-hoehn:
Yes, by clicking on the H button, you can unhide that input. The T button unhides the timout. This allows the block to be very small in case you don't need it.
image

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,184 @@
...
+   *
+   * @param block
+   * @returns {[string,number]}
+   */
+  javascriptGenerator.forBlock['oh_httprequest'] = function (block) {
+    const requestType = block.getFieldValue('requestType')
+    const contentType = block.getFieldValue('contentType')
+    const headers = javascriptGenerator.valueToCode(block, 'requestHeader', javascriptGenerator.ORDER_ATOMIC)
+
+    let timeout = javascriptGenerator.valueToCode(block, 'timeoutInput', javascriptGenerator.ORDER_ATOMIC)
```
florian-h05, line 159:
```suggestion
    const timeout = javascriptGenerator.valueToCode(block, 'timeoutInput', javascriptGenerator.ORDER_ATOMIC) || 3000
```

... and remove the next if statement.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,184 @@
...
+    if (!timeout) {
+      timeout = 3000
+    }
+
+    let url = javascriptGenerator.valueToCode(block, 'url', javascriptGenerator.ORDER_ATOMIC)
```
florian-h05, line 164: These both can be consts I think.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,184 @@
...
+    let payload = javascriptGenerator.valueToCode(block, 'payload', javascriptGenerator.ORDER_ATOMIC)
+
+    const hasContent = (requestType !== 'HttpGetRequest' && requestType !== 'HttpDeleteRequest')
+    let code = ''
+    if (hasContent) {
+      if (!headers) {
+        code = `actions.HTTP.send${requestType}(${url},'${contentType}',${payload},${timeout})`
```
florian-h05, line 171: Let's increase the readability of the generated code a bit:

```suggestion
        code = `actions.HTTP.send${requestType}(${url}, '${contentType}', ${payload}, ${timeout})`
```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,184 @@
...
+      } else {
+        code = `actions.HTTP.send${requestType}(${url},'${contentType}',${payload},${headers},${timeout})`
```
florian-h05, line 173:
```suggestion
        code = `actions.HTTP.send${requestType}(${url}, '${contentType}', ${payload}, ${headers}, ${timeout})`
```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,184 @@
...
+      }
+    } else {
+      if (!headers) {
+        code = `actions.HTTP.send${requestType}(${url},${timeout})`
```
florian-h05, line 177:
```suggestion
        code = `actions.HTTP.send${requestType}(${url}, ${timeout})`
```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,184 @@
...
+      } else {
+        code = `actions.HTTP.send${requestType}(${url},${headers},${timeout})`
```
florian-h05, line 179:
```suggestion
        code = `actions.HTTP.send${requestType}(${url}, ${headers}, ${timeout})`
```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  /**
+   * Implements GET, DELETE, POST and PUT requests
+   *
+   * @param block
+   * @returns {[string,number]}
+   */
+  javascriptGenerator.forBlock['oh_httprequest'] = function (block) {
```
florian-h05, line 154: Hmm, either the generator needs to handle `isGraalJs` and needs to generate different code when running on Nashorn or HTTP blocks need to be considered not available on Nashorn.
It should be relatively easy to execute the HTTP actions on Nashorn, you just need the Java import.

florian-h05 (changes requested):
In general LGTM, just a few comments, mostly wrt to code style.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
stefan-hoehn: I would have a hard time to actually backport and test it as I have no system available to test it. Therefore I have decided to exclude the block in case there is no graalJS

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
stefan-hoehn: Thanks: good hint - I didn't know that syntax 😍

florian-h05: JS has some very nice syntactic features, but probably too many to know them all.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
florian-h05: Fine :+1:

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-http.js
```diff
@@ -0,0 +1,186 @@
...
+
+  const unavailMsg = 'HTTP blocks aren\'t supported in "openHAB < 4.2"'
```
florian-h05, line 12:
```suggestion
  const unavailMsg = 'HTTP blocks aren\'t supported on "application/ecmascript;version=5.1"
```

The old message isn't really helpful, as no one before 4.2 will receive that block.
Instead it should be renamed to make clear that Nashorn is not supported.

florian-h05 (changes requested):

florian-h05 (approved):
LGTM, thanks!

mherwege:
> Another thing that would be handy is a query builder, ie given a dictionary of key value, append `key=Val&key=val` to the url

I am experimenting with this new functionality. I would be interested in this as well.

And another thing that could be useful in this context: a urlEncode function. Parameters in GET calls will have to be URL encoded. If extending the block with a query builder, the logic could include URL encoding the parameters.

mherwege:
> so we don’t have to write the logic ourselves ;-)

Sure, I think so as well.

To give an idea, here is what I came up with purely in Blockly:

image

I think this can be made nicer. The URL encode function is of course not complete, but covers the need in my case.

PR #2412, opened 2024-02-26
Rule module popup: add dirty checking

jimtng:
This PR adds dirty checking in the rule module popup.

Rule module popup comes up when you're editing a rule's module, e.g. trigger, action, condition. Previously when you opened the rule module and made changes, then clicked back, you wouldn't get a warning and the changes will simply be lost.


stefan-hoehn:
Would you mind adding some description of what the intent of the change is?

jimtng:
Updated description

florian-h05 (approved):
LGTM, thanks! Much needed improvement.

PR #2415, opened 2024-02-26
Blockly Editor: combine renderer selection and show item label icons

jimtng:
The current icons are too crowded for mobile devices. This PR combines the renderer choice and the show labels into one icon.

Before
image

After
image


jimtng:
The second commit capitalizes the renderer name. I think it looks nicer

image

florian-h05:
For mobile phones, this looks really good!
I however wonder if the options however shouldn’t stay separated on big enough screens for faster access.

jimtng:
> I however wonder if the options however shouldn’t stay separated on big enough screens for faster access.

I agree. Just need to add more code. I'll add that tomorrow.

jimtng:
It wasn't too hard, so I've made the changes.

On desktop it now looks like this (with an icon instead of the name)
image

ghys (approved):
It does look good, thank you!

PR #2419, opened 2024-02-27
[blockly] Add multi-select feature

### LINKED ISSUE #2417: Blockly mouse/trackpad navigation
@stefan-hoehn Currently in blockly:
- two finger drag = Zoom
- click and drag = pan

I would like to propose the following changes (I'll implement the changes)
- two finger drag = pan
- two finger drag + Ctrl held down = zoom
This is consistent with other block coding environments: [Scratch](https://scratch.mit.edu/), [MIT App Inventor](https://appinventor.mit.edu/), [Microsoft MakeCode](https://arcade.makecode.com/), which is also used by microbit block code, and possibly others.

This will also pave the way for another improvement that I'll submit shortly ;)

jimtng:
- Shift + Click to select multiple blocks
- Also shift + mouse drag to multi-select
- Currently multiselect doesn't work on mobile (AFAIK) but copy/pasting even across different scripts works
- Provides cross tab copy/paste, including copy/pasting multiple selected blocks.

The multi-select plugin conflicts with the cross-tab-copy-paste plugin and makes it redundant, so it is removed in this PR

Also in this PR, make two fingers do panning instead of zooming.

These have been split into three commits for granularity so please don't squash them.

Resolve #2417

### DIFF
--- a/bundles/org.openhab.ui/web/package-lock.json
+++ b/bundles/org.openhab.ui/web/package-lock.json
@@ -18,6 +18,7 @@
         "@jsep-plugin/object": "^1.2.1",
         "@jsep-plugin/regex": "^1.0.3",
         "@jsep-plugin/template": "^1.0.2",
+        "@mit-app-inventor/blockly-plugin-workspace-multiselect": "^0.1.11",
         "blockly": "^10.4.2",
         "cronstrue": "^1.100.0",
         "crypto-browserify": "^3.12.0",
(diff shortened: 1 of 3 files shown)


rkoshak:
Thank you!

This was really annoying when I was trying to do some testing of my block library.

stefan-hoehn:
I also need to check closely that it removes the cross-tab-plugin. 
Jimmy, please keep in mind that it is not only about adding new features but they also should be consistent what the users have used in the past. And it MUST be documented as well (sometimes that takes more time than the actual implementation).
So without having looked at it in detail yet, I am kind of worried because of the cross-rule-copy/paste.

stefan-hoehn:
So copy and paste across rules still work. That's good. 

But when you shift-click blocks and then do a copy of them and exception is thrown in the background and you can't go on except that you have the blocks clinging to your mouse:

image

See the attached 

https://github.com/openhab/openhab-webui/assets/5937600/8920b5e4-4250-4523-bac7-71f34fa544a6

jimtng:
> But when you shift-click blocks and then do a copy of them and exception is thrown in the background and you can't go on except that you have the blocks clinging to your mouse:

- A problem/bug does exist.
- This is not related to copy paste
- This is purely from trying to drag a multi-selected output (non connectable) blocks.
- No problems when selecting and moving connectable blocks. 

Perhaps hold off merging this until this issue is resolved. 

I will split the panning/zooming change into a separate PR because that should be fine and unrelated to multi-select I think.

jimtng:
@stefan-hoehn I've updated a patch although I'll open a PR on the multi-select repo to have it fixed there too. But this means we can merge this now instead of waiting.

jimtng:
Looks like it has already been [fixed](https://github.com/mit-cml/workspace-multiselect/pull/38) 9 hours ago.

stefan-hoehn:
Ok, I tested it again and from I see it looks good. 

Can you please not only provide this PR but also the PR that updates the docs accordingly, so the users are not confused as we had "cross rule copy" before and now the new behaviour. Also multiselect should be described. 

So in particular https://next.openhab.org/docs/configuration/blockly/rules-blockly-before-using.html#cross-rule-copy-paste should be updated at best with a new animated gif that explains how to use the new functionality.

florian-h05 (approved):
LGTM. I just had a look at it, and this is a really nice feature!

florian-h05:
@jimtng FYI I updated the commit messages a bit so these contain a reference to this PR when I rebase.

PR #2420, opened 2024-02-27
Add UI for health checks (broken links)

seime:
Starting point for a UI that shows issues with the users installation.

First out is broken links.

Depends on https://github.com/openhab/openhab-core/pull/4115

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -39,6 +39,10 @@
           </f7-list-item>
           <li v-if="showSettingsSubmenu">
             <ul class="menu-sublinks">
+              <f7-list-item v-if="$store.getters.apiEndpoint('links')" link="/settings/health/" title="Health checks" view=".view-main" panel-close :animate="false" no-chevron
+                            :class="{ currentsection: currentUrl.indexOf('/settings/health') === 0 }">
+                <f7-icon slot="media" f7="heart" color="gray" />
+              </f7-list-item>
               <f7-list-item v-if="$store.getters.apiEndpoint('things')" link="/settings/things/" title="Things" view=".view-main" panel-close :animate="false" no-chevron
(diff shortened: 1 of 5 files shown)


ghys:
Thanks @seime - I had the same thoughts a few years ago.
My conclusion was that as most issues can be identified on the server, they should then be collected and processed on the server, not the client.

In practice you'd have perhaps a `/rest/issues/scan` API call which will run code on the server that will identify issues like:
- orphan links (missing item or channel)
- items with invalid parent group names
- expressions in Pages or config props referencing unknown items
- etc.

then you could get some details about the issue (`/rest/issues/1234`), its resolution options ('Remove Group1 from Item1 definition', 'Dismiss this issue for now', 'Ignore/delete this issue') and execute a resolution (`/rest/issues/1234/resolve`).

It would be up to server code in a Java class to detect, report, store (in a registry), prevent duplicates, and eventually resolve a particular kind of issue. That is, every kind of issue would have its own class able to detect it, report it, offer options to solve it, and exercise one of these options.

And with OSGi we would be able to introduce and inject new types of issues - ie. an OH binding could run its own diagnostics and insert its own issues.

seime:
Yes, there is plenty to do in this space, and I was thinking a bit of the same. Actually started on a separate rest endpoint, but then reverted it. 

WDYT about starting small and improve and refactor over time - opposed to crafting a whole new set of backend features just to get started?

seime:
@ghys ☝️

florian-h05:
@ghys Can you please comment on

> WDYT about starting small and improve and refactor over time - opposed to crafting a whole new set of backend features just to get started?

and let me know if you are fine with this in general, so I can review it?

florian-h05:
Gentle ping @ghys

ghys:
Yes I'm sorry for the delay, I still would like an UI for "general problem solving" (like I described above) but let's do this organically. You have my support!

florian-h05 (approved):
Thanks, nice addition!

I have fixed a number of styling and functional issues, but it is now ready for merging - just in time for 4.2.0.

PR #2423, opened 2024-02-28
Help sidebar: Reduce list indentation

florian-h05:
Fixes https://github.com/openhab/openhab-docs/issues/2243.


ghys:
When you're adding style without scoping it you might be effectively propagating it to everywhere else.

I have never trusted the `scoped` addition to `<style>` so I would suggest that you add a CSS class and change the `ul` style only under that class.

PR #2198, opened 2023-12-05
Location cards: Swap status & measurement badges to avoid layout shift

ThunderboxEntertainment:
Swapped the location of measurement with the status badges to eliminate layout shifts in location cards.

This way If you have some rooms with equipment active, and some without, the temp and humidity readings will all be aligned (as opposed to offset).

PR #2425, opened 2024-02-28
Location card: Swap status & measurement badges to avoid layout shift

florian-h05:
The status badges are displayed based on the actual state of that status category. This causes the measurement badges, which are always shown to be either first or second line.

This swaps those, so that measurement badges don't move around anymore.


florian-h05:
Supersedes #2198.

PR #2426, opened 2024-02-28
oh-player-controls: Hide default tooltip

florian-h05:
Previously, when hovering over a `oh-player-control` component, a tooltip, which shouldn't be there, was shown.
This was caused by the config title being set as the divs title.
It is now fixed by manually setting title to an empy string.


florian-h05:
Supersedes #532.

PR #2428, opened 2024-02-28
Item picker: Disable picker & hide chevron if not yet ready

florian-h05:


florian-h05:
Supersedes #1066.
Please note that I have not picked up the type filter improvements (in general,  I like them) , because the type filter on the server requires exact match with the type, which does not allow to filter by all `Number[:...]` Items.

PR #2432, opened 2024-02-29
Remove Cordova from Main UI

florian-h05:
This removes Cordova from Main UI.

Cordova was used to build Main UI as an iOS or Android app (a webview with pre-packed assets), but the Cordova component of Main UI is currently unmaintained due to requiring lots of maintenance.
Both the iOS and the Android app have an integrated webview (without pre-packed assets though), so Cordova is unlikely to be required in the future.


stefan-hoehn:
Very nice. Again something tha reduces complexity. Thanks a lot, Florian 👏🏻

PR #2441, opened 2024-03-02
Widget expressions: Add `#` as a shorthand for item.numericState

jimtng:
See: https://github.com/openhab/openhab-core/pull/4123

Add `#` alongside the existing shorthands:
```
@'ItemName' => items.ItemName.state
@@'ItemName' => items.ItemName.displayState
#'ItemName' => items.ItemName.numericState
```

Yes, the docs will be updated if/after this is merged.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
@@ -20,12 +20,17 @@ expr.jsep.plugins.register(jsepRegex, jsepArrow, jsepObject, jsepTemplate)
 
 expr.addUnaryOp('@', (itemName) => {
   if (itemName === undefined) return undefined
-  const itemState = store.getters.trackedItems[itemName]
-  return itemState.displayState || itemState.state
+  const item = store.getters.trackedItems[itemName]
+  return item.displayState || item.state
 })
 expr.addUnaryOp('@@', (itemName) => {
   if (itemName === undefined) return undefined
   return store.getters.trackedItems[itemName].state
+})
+expr.addUnaryOp('#', (itemName) => {
+  if (itemName === undefined) return undefined
+  const item = store.getters.trackedItems[itemName]
+  return item.numericState || item.state
 })
 
 dayjs.extend(relativeTime)
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
@@ -30,7 +30,7 @@ expr.addUnaryOp('@@', (itemName) => {
 expr.addUnaryOp('#', (itemName) => {
   if (itemName === undefined) return undefined
   const item = store.getters.trackedItems[itemName]
-  return item.numericState || item.state
+  return item.numericState
 })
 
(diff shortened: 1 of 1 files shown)


florian-h05:
Would `##` be more consistent?

jimtng:
> Would `##` be more consistent?

It's a matter of opinion, I guess. I'd rather it be a single `#` to save typing. Two ## would indicate there's a version with one # too?

I picked # because it's often used as "number" e.g. "My #1 OS is Linux" whereas `##` doesn't sound as nice

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
```diff
@@ -20,13 +20,18 @@ expr.jsep.plugins.register(jsepRegex, jsepArrow, jsepObject, jsepTemplate)
...
+  return item.displayState || item.state
 })
 expr.addUnaryOp('@@', (itemName) => {
   if (itemName === undefined) return undefined
   return store.getters.trackedItems[itemName].state
 })
+expr.addUnaryOp('#', (itemName) => {
+  if (itemName === undefined) return undefined
+  const item = store.getters.trackedItems[itemName]
+  return item.numericState || item.state
```
florian-h05, line 33: IMO it doesn't make sense to fallback to the string state.

jimtng: agreed totally!

jimtng: how about `|| Number.parseFloat(item.state)` ?

florian-h05: As it is a shortcut for `numericState`, I think it should behave the same; so no fallback to parseFloat.

jimtng: OK in that case I'll leave the last commit as is (no fallback)

florian-h05: Very minor thing: Could you please remove the item const and make it a single liner?

jimtng: D'oh! Thanks for noticing that!

florian-h05 (approved):
LGTM, thanks :+1: 
Just in time for the first milestone of openHAB 4.2.0 :raised_hands:

bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js
```diff
@@ -20,13 +20,17 @@ expr.jsep.plugins.register(jsepRegex, jsepArrow, jsepObject, jsepTemplate)
 
 expr.addUnaryOp('@', (itemName) => {
   if (itemName === undefined) return undefined
-  const itemState = store.getters.trackedItems[itemName]
-  return itemState.displayState || itemState.state
+  const item = store.getters.trackedItems[itemName]
+  return item.displayState || item.state
```
ghys, line 24: I'm too late, but be careful with this syntax with the OR operator - here if `item.displayState === 0`  you do fallback to `item.state` which is not what you want.
Nice feature btw!

PR #2447, opened 2024-03-04
Only set API auth headers if requesting the REST API

florian-h05:
This should avoid leaking the user's access token to third-parties in case a UI component uses `this.$oh.api.` to request external resources.
This is currently not the case, but IMO we should add this check anyway as a component might make such a request with a future code change.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/api.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/api.js
@@ -12,7 +12,7 @@ function wrapPromise (f7promise) {
 Framework7.request.setup({
   xhrFields: { withCredentials: true },
   beforeSend (xhr) {
-    if (getAccessToken() && xhr.requestParameters.method !== 'HEAD') {
+    if (getAccessToken() && xhr.requestParameters.method !== 'HEAD' && xhr.requestParameters.url.startsWith('/rest/')) {
       if (getTokenInCustomHeader()) {
         xhr.setRequestHeader('X-OPENHAB-TOKEN', getAccessToken())
       } else {


florian-h05:
@ghys Can you please have a look at this PR? It does not have real-world benefits at the moment, but avoids token leak in case a PR changes the code so that `this.$oh.api.` is used to request something else then the REST API.

ghys:
> so that `this.$oh.api.` is used to request something else then the REST API.

The point of this interface is precisely to access OH's own REST API, so I'm curious what the intent or the potential use cases are?
In general I was reticent opening the can of worms of allowing the frontend to make arbitrary HTTP requests to 3rd parties (outside the occasional image), the "proper" way being to have the backend do it through HTTP binding/things or rules and putting the responses in items.

florian-h05:
Hmm, there is no potential use case, this was meant as guard rails in case someone uses `$oh.api.` accidentally for external requests, but since that would need pass our review, I guess this should not happen.

ghys:
Yes, this was my point, even if we don't exclude 3rd party API requests in the UI, it should be made with lower level interfaces (e.g. `Framework7.request`), not this `$oh.api` which is meant for the OH REST API only, so it shouldn't pass review.

In fact if anything we might want guard it from being used elsewhere altogether, by mistake or by malicious intents, so your PR kind of made sense, only it was not "restrictive" enough.

Note there was an alternative Cordova build at the time, where you had to specify the openHAB server you wanted to connect to, so setting a limit to URLs not beginning with `/rest/` wouldn't have worked in this case (a moot point because Cordova had its own implementation of these `this.$oh` interfaces).

PR #2448, opened 2024-03-04
Items List: Remove Add item from textual

jimtng:
Make the + Icon in the items list go straight into adding a new item. This makes the most common operation quicker.

Adding from textual definition can still be done through Developer Tool, as it's not a common task.


florian-h05 (approved):
Agreed and LGTM!

PR #2450, opened 2024-03-04
Upgrade echarts and vue-echarts & Add fallback for visualMap when using heatmap

florian-h05:
This upgrades `echarts` and `vue-echarts` to the latest versions.
`@vue/composition-api` was added as dependency because `vue-echarts` needs it on Vue 2.

See https://echarts.apache.org/handbook/en/basics/release-note/v5-upgrade-guide/ and https://github.com/ecomfe/vue-echarts#migration-to-v6 for upgrade guides.

This also adds a fallback default for `visualMap` because it is required to define a `visualMap` when using a heatmap, otherwise echarts fails to render the chart. 
Even though the docs and the chart-designer mentioned this until now, it worked without defining a visualmap, which would not be the case anymore (and therefore break charts - defining a visual map is easy though) without providing a fallback.


stefan-hoehn:
I tested it on my side. Even though openHAB charts seem to work, the network chart in the Z-Wave plugin does not. The latter doesn't even through an exception in the logs or anything obvious shows up in the console.

stefan-hoehn:
I can confirm that it works now as it should. Looks good to me overall. Hope I have not overseen something.

ghys:
To be thorough, the migration guide from ECharts 4.x to 5.x is here:
https://echarts.apache.org/en/tutorial.html#ECharts%205%20Upgrade%20Guide

and the migration advice of vue-echarts package to 6.x is here:
https://www.npmjs.com/package/vue-echarts#migration-to-v6

Appreciate the work, as always!

PR #2453, opened 2024-03-05
Item list & details: Hide group type if not set

jimtng:


florian-h05 (approved):
LGTM, thx for the fix.

I also fixed the Item mixin (this is used to get the type text in the Items list, where this bug also occured) and now use the Item mixin from item-details to get the type text.

PR #2455, opened 2024-03-05
Channel edit: Fix regression causing editable channels to be locked

jimtng:
Regression from #2403.

Channel doesn't have its own `editable` property.


florian-h05 (approved):
LGTM, thanks!

PR #2465, opened 2024-03-08
[blockly] Add undefined block

stefan-hoehn:
Adds the undefined block as discussed at https://community.openhab.org/t/blockly-proposal-to-change-contextual-info-blocks/154403 

image

creates 

```
var notDefinedVar;


if (notDefinedVar == undefined) {
  console.log('I am undefined');
}
```

Note that the logic block never creates "===" but always "=="


florian-h05 (approved):
LGTM, thanks!

Issue #2483, opened 2024-03-18
Saving a file in Eclipse suddenly triggers this systematic error


lolodomo:
When saving a file, this popup is shown:
![image](https://github.com/openhab/openhab-webui/assets/10186704/698d4bec-a1dc-4a0b-8d33-7e920493db38)

The result is that the file is not properly reformatted at saving.

@florian-h05 @wborn : any idea what is happening ?

If seems it happens in openhab-webui repo (java class of Basic UI for example) but not in openhab-addons repo.

lolodomo:
This morning, I got the same error, for the same jar file but the mentioned file was another, something like ringtone.mp3.

lolodomo:
I should mention that my Eclipse recently upgraded (automatically) to the most recent version:
```
Version: 2024-03 (4.31.0)
Build id: 20240307-1437
```

mherwege:
My Eclipse also updated automatically. I have a very similar thing:

<img width="491" alt="image" src="https://github.com/openhab/openhab-webui/assets/23501634/9cb19b71-c9c4-4135-bf71-b88c1f4b9d77">

This appeared when trying to look up the declaration of the `Item` type (but the same happens for any type). I don't have the `Item` type open in the projects, so it would look for it in a downloaded jar I suppose.

mherwege:
Could it have something to do with https://github.com/openhab/openhab-webui/pull/2267 ?

lolodomo:
I did not see it before but you're right, I have a doubled . in my popup.

PR #2484, opened 2024-03-19
Fix webpack asset filename

mherwege:
Resolves https://github.com/openhab/openhab-webui/issues/2483

The webpack [ext] template string already includes a leading . (https://webpack.js.org/configuration/output/#template-strings). This PR removes the the extra dot that was added.


florian-h05 (approved):
LGTM, thanks!

PR #2485, opened 2024-03-19
Save stateDescriptionPattern from channel type for UoM Items & Save UoM metadata when adding from model

mherwege:
When creating a new Number with Dimension item from a channel link, the stateDescriptionPattern does not propose the state description pattern from the channel type as the default.

This PR introduces this.

It only saves this state description pattern from the channel type to metadata of the item if it is not the default state description pattern for a Number type.

Subsequent edits of the item will present a blank state description if none has been saved explicitely. Reasons are:
- a channel stateDescription can contain more than a pattern. Saving only the pattern would override the rest of the stateDescription.
- an item can be linked to multiple channels with (possibly conflicting) stateDescription. The only solution would be for the user the set it on the item and override the channel stateDescriptions. If that is not done, nothing is shown and it is left to core to take one when presenting the formatted state.

See https://github.com/openhab/openhab-addons/pull/16531#discussion_r1530394678

This PR also now saves the unit (and state description pattern) when creating equipment or point from a thing.
See discussion https://community.openhab.org/t/units-are-not-propagated-when-creating-items/154748.
So far, changes were only saved when editing or creating items directly or from thing channels, not when creating from things in the model.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/group-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/group-form.vue
@@ -107,7 +107,7 @@ export default {
         const dimension = this.dimensions.find((d) => d.name === newDimension)
         this.$set(this.item, 'groupType', 'Number:' + dimension.name)
         this.groupUnit = this.getUnitHint(dimension.name)
-        this.$set(this.item, 'stateDescriptionPattern', this.getStateDescription())
+        this.$set(this.item, 'stateDescriptionPattern', this.stateDescriptionPattern)
       }
     },
     groupUnit: {
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -277,7 +277,7 @@ export default {
     if (this.createMode && this.stateDescription && (this.stateDescription !== this.item.stateDescriptionPattern)) {
       // If there is a state description from the channel type that is different from the default,
       // set it as the item state description
-      this.$set(this, 'item.stateDescriptionPattern', this.stateDescription)
+      this.item.stateDescriptionPattern = this.stateDescription
     }
   },
   beforeDestroy () {
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -274,6 +274,11 @@ export default {
     if (!this.item) return
     this.initializeAutocompleteCategory()
     if (this.dimensionsReady) this.initializeAutocompleteUnit()
+    if (this.createMode && this.stateDescription && (this.stateDescription !== this.item.stateDescriptionPattern)) {
+      // If there is a state description from the channel type that is different from the default,
+      // set it as the item state description
+      this.$set(this, 'item.stateDescriptionPattern', this.stateDescription)
+    }
   },
   beforeDestroy () {
     if (this.unitAutocomplete) {
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -40,11 +40,11 @@
         <f7-list-input v-show="itemDimension"
                        label="State Description Pattern"
                        type="text"
-                       :info="(createMode) ? 'Pattern or transformation applied to the state for display purposes. Only saved if you change the pre-filled default value.' : ''"
-                       :disabled="!editable"
+                       :info="(createMode) ? 'Pattern or transformation applied to the state for display purposes. Only saved if you change the pre-filled default value.' : 'Pattern can only be changed from the state description metadata page after Item creation!'"
+                       :disabled="!createMode"
                        :value="stateDescriptionPattern"
                        @input="stateDescriptionPattern = $event.target.value"
-                       :clear-button="editable" />
+                       :clear-button="createMode" />
 
         <!-- Group Item Form -->
         <group-form ref="groupForm" v-if="itemType === 'Group'" :item="item" :createMode="createMode" />
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -98,7 +98,7 @@ import uomMixin from '@/components/item/uom-mixin'
 
 export default {
   mixins: [ItemMixin, uomMixin],
-  props: ['item', 'items', 'createMode', 'hideCategory', 'hideType', 'hideSemantics', 'forceSemantics', 'unitHint'],
+  props: ['item', 'items', 'createMode', 'hideCategory', 'hideType', 'hideSemantics', 'forceSemantics', 'unitHint', 'stateDescription'],
   components: {
     SemanticsPicker,
     ItemPicker,
@@ -149,7 +149,6 @@ export default {
         const dimension = this.dimensions.find((d) => d.name === newDimension)
         this.$set(this.item, 'type', 'Number:' + dimension.name)
         this.itemUnit = (this.unitHint ? this.unitHint : this.getUnitHint(dimension.name))
-        this.$set(this.item, 'stateDescription', this.getStateDescription())
       }
     },
     itemUnit: {
@@ -171,7 +170,7 @@ export default {
     stateDescriptionPattern: {
       get () {
         if (this.item.stateDescriptionPattern) return this.item.stateDescriptionPattern
-        return this.item.metadata?.stateDescription?.config.pattern || '%.0f %unit%'
+        return this.item.metadata?.stateDescription?.config.pattern || this.stateDescription || (this.createMode ? '%.0f %unit%' : '')
       },
       set (newPattern) {
         this.$set(this.item, 'stateDescriptionPattern', newPattern)
--- a/bundles/org.openhab.ui/web/src/components/item/item-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/item/item-mixin.js
@@ -77,17 +77,15 @@ export default {
           return Promise.resolve()
         }
       }).then(() => {
-        // Save state description if Item is an UoM Item and if state description changed from the default value
+        // Save state description if Item is an UoM Item
         if ((item.type.startsWith('Number:') || item.groupType?.startsWith('Number:')) && stateDescriptionPattern) {
-          if (stateDescriptionPattern !== '%.0f %unit%') {
-            const metadata = {
-              value: ' ',
-              config: {
-                pattern: stateDescriptionPattern
-              }
+          const metadata = {
+            value: ' ',
+            config: {
+              pattern: stateDescriptionPattern
             }
-            return this.$oh.api.put('/rest/items/' + item.name + '/metadata/stateDescription', metadata)
           }
+          return this.$oh.api.put('/rest/items/' + item.name + '/metadata/stateDescription', metadata)
         } else {
           return Promise.resolve()
         }
--- a/bundles/org.openhab.ui/web/src/components/item/item-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/item/item-mixin.js
@@ -64,34 +64,42 @@ export default {
       const stateDescriptionPattern = item.stateDescriptionPattern
       delete item.stateDescriptionPattern
 
-      // TODO: Add support for saving metadata
       return this.$oh.api.put('/rest/items/' + item.name, item).then(() => {
-        // Save unit metadata if Item is an UoM Item
-        if ((item.type.startsWith('Number:') || item.groupType?.startsWith('Number:')) && unit) {
-          const metadata = {
-            value: unit,
-            config: {}
-          }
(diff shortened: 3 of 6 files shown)


mherwege:
From linked PR: https://github.com/openhab/openhab-addons/pull/16531

> I understand your point. But please tell me why you think the state pattern element in thing xml exists if not for the purposes of (helping) creating items? It seems to me that 455 binding developers may have been wasting their time writing state pattern elements if the whole concept has no purpose.

I think this whole concept is unfortunately messy, but difficult to change without breaking things.
Current behaviour, if you have one item connected to 2 channels with different channel type and state description pattern, no state description pattern set on the item (in metadata, a config file) or on a sitemap (for the sitemap based UI's), one of the channel type state descriptions will be used for formatting. Which one is entirely random.
If the item is only linked to one channel, it should use the state description pattern of the channel type if nothing else is set, even if not visible in the UI. That definitely is the most common case. Typically, very few items get linked to multiple channels.

This PR proposes to show the channel type state description pattern when originally creating an item linked to a channel of a thing. That gives users the possibility to set something else if that makes sense. It avoids changing item display state update behaviour.
An alternative could be to always set the channel type state description through the UI in metadata (so create a specific state description pattern). In that case, not just the state description pattern should be set, but also the other parameters from the state description should be copied over.
Of course, this alternative solution would not apply if you first create the item, and then link it to (one or more) channels. In that case, it would still be the behaviour as today.

rkoshak:
I'm sure you've already thought about this but one little wrinkle to keep in the back of your mind as you work this is sometimes the type of the destination Item may be dramatically different from the type the Channel (and therefore the state description pattern pushed by binding) may not even make sense. 

For example, if you have a timestamp profile on the link between a Number:Dimensionless Channel and a DateTime Item, `%.0f %%` as a State Description pattern is worse than nonsense, it generates an error. I think/hope that pattern pushed by the binding is ignored in this case but want to make sure we don't reintroduce errors in this particular use case. 

I don't think there's a problem here, just wanted to mention it.

mherwege:
@rkoshak Thank you for the insight. I didn’t consider that so far. But with the change as implemented so far, it shouldn’t change the current behavior. But I honestly don’t know how the system currently copes with it. Does it ignore channel type state descriptions when a profile changing to a different type is used?
The risk is higher when the code would take the state description from the channel type and force it on the item, store it in the item metadata.
As long as it doesn’t do that by default, behavior does not change.

rkoshak:
> Does it ignore channel type state descriptions when a profile changing to a different type is used?

Awhile back it was actually causing a bunch of problems because some bindings would push state description "options" which were silently applied to the Item but because the Item type changed due to a profile it messed up everything in the state description. This was a couple years ago now I think. 

I was under the impression that when that problem was sorted it also addressed the pattern. But I don't know how to test it. My searching has so far failed to show the original issue or PR.

I don't think it forces the metadata to be created though. I think it's only kept in memory somehow.

mherwege:
> My searching has so far failed to show the original issue or PR.

Probably this: https://github.com/openhab/openhab-core/issues/2283
But it doesn't look like that problem has been solved.

And here is another issue with this when using profiles: https://github.com/openhab/openhab-core/issues/4050

rkoshak:
Those both look like the same overall issue but neither is the one I'm thinking about. But I could be mistaken about my memory that it got fixed if these are open. And it could have been a forum thread I'm remembering. There's so much to keep track of.

andrewfg:
@mherwege I was wondering how to make progress on this topic. And how to resolve the prior as yet unresolved issues.

I suggest that the code must comply with the "functional specification" for state descriptions as written in the [state description developer document](https://www.openhab.org/docs/developer/bindings/thing-xml.html#state-description) 

This functional specification is either a) fully correct, b) fully wrong, or c) missing something. If the functional specification is fully correct, then this PR MUST follow that verbatim. Otherwise if the functional specification is wrong or missing something then we should first edit and agree upon the functional specification BEFORE we try to review the code in this PR. => WDYT?

mherwege:
@andrewfg This PR does not have the ambition to resolve all issues with state descriptions when applying profiles. It merely tries to capture the state description pattern from the channel type and apply it when creating a new item with dimension number and a type and make sure it is applied to the item. It avoids mismatches in the UI and unpredictability because of mismatches in channel type state description patterns. The UI naively supposed so far a %.f %unit% pattern when nothing was set in the UI. From my perspective, this PR still complies with the functional specification. All other parameters in the state description should not be touched by this PR. Where do you think I go wrong with this reasoning?
Second element in this PR is to make saving units and state description patterns work properly when creating items from the model.

andrewfg:
^
@mherwege I am not trying to drive a polemic here. But I do want to find a path forward. I don't know the webui language so for me it is easier to discuss the doc. If the doc and your PR code are mutually consistent then I think we should merge your PR immediately. But if there are inconsistencies then we should identify those, and map a path to resolve them. Such resolution could be either by changing the doc, and/or by changing the ui code (in this PR and/or in other PRs). But in all cases I would suggest that we agree on the wording of the doc BEFORE we try to interpret and merge your code..

mherwege:
> If the doc and your PR code are mutually consistent then I think we should merge your PR immediately. But if there are inconsistencies then we should identify those, and map a path to resolve them.

@andrewfg I still think the doc and this PR are mutually consistent. I spend some extra time testing.

This PR will only set the state description pattern (not the full state description) for a Number Dimension item when creating an item from a thing channel. It copies the state description pattern from the channel type and saves that, or saves a new value set by the user.
If the user links the same Number Dimension item to another channels (or links an existing item to a channel) it will not be set or overwritten.
State description patterns for any other item type will not be touched by this PR. They are not available in these forms and would need to be explicitely through metadata.
State descriptions are used to calculate the display state of items to be used in the UI. The state description used is a merge of the item state description and the state descriptions on the channels the item is linked to. The item state description takes precedence when the same state description field exists on the item and the channel. I tested with the following example:

Channel type for a channel on a binding (I used nikohomecontrol):
```
	<channel-type id="setpoint">
		<item-type>Number:Temperature</item-type>
		<label>@text/channelSetpointLabel</label>
		<description>@text/channelSetpointDescription</description>
		<category>Temperature</category>
		<tags>
			<tag>Setpoint</tag>
			<tag>Temperature</tag>
		</tags>
		<state min="0" max="100" step="0.5" pattern="%.1f %unit%"/>
	</channel-type>
```
I create on item linked to a channel on the binding and explicitely set a state description pattern of `%.5f %unit%`.
I then called the REST API (`/items/{itemname}` endpoint, with the following result:
```
{
  "link": "http://localhost:8080/rest/items/Thermostat_Setpoint",
  "state": "NULL",
  "stateDescription": {
    "minimum": 0,
    "maximum": 100,
    "step": 0.5,
    "pattern": "%.5f %unit%",
    "readOnly": false,
    "options": []
  },
  "unitSymbol": "°C",
  "metadata": {
    "unit": {
      "value": "°C"
    },
    "semantics": {
      "value": "Point_Setpoint",
      "config": {
        "relatesTo": "Property_Temperature"
      }
    },
    "stateDescription": {
      "value": " ",
      "config": {
        "pattern": "%.5f %unit%"
      }
    }
  },
  "editable": true,
  "type": "Number:Temperature",
  "name": "Thermostat_Setpoint",
  "label": "Setpoint",
  "category": "Temperature",
  "tags": [
    "Temperature",
    "Setpoint"
  ],
  "groupNames": []
}
```
You clearly see the effect of merging the state description.

So I see no inconsistency between documentation and what I try to do here. The only thing you could argue about is that no warning is given on the effect of merging state descriptions (which is already the case without this PR) on the resulting state description in the documentation.

If you feel differently, please let me know. I will be happy to change this PR. But at this time, I don't see how it would invalidate the documentation and current contract in the slightest way.

andrewfg:
@mherwege many thanks for the explanation above.

Based on that, I think the following..

1. The doc does not need to be changed.
2. This PR => LGTM

Further questions..

3. Do you have any idea if the 'min', 'max' and 'step' parameters are also properly taken over? And if so, where? Maybe there is a need for another PR to cover that?
4. As you PR here was partly triggered by my work on a binding with humidity channels having the unit "%rH" I wonder if you have any thoughts concerning my https://github.com/openhab/openhab-core/pull/4157 ??

andrewfg (approved):
LGTM

mherwege:
> 3\. Do you have any idea if the 'min', 'max' and 'step' parameters are also properly taken over? And if so, where? Maybe there is a need for another PR to cover that?

My understanding is it is part of the state description and will get merged with the rules I explained. There is no attempt in this PR to do anything with the state description beyond the state description pattern for Number Dimension items.
In the example above, you see the REST response returning min, max and step from the channel. They are not set on item metadata.
There are issues with it though, but they are outside of the scope of this PR. There is no unit set on min, max and step. So the behavior changes when changing the unit (same step in other unit). Solving the issue is out of scope for this PR. And it may also be tricky to solve (min, max and step are probably device dependent and may or may not always be in a specific unit).

lolodomo:
> I suggest that the code must comply with the "functional specification" for state descriptions as written in the [state description developer document](https://www.openhab.org/docs/developer/bindings/thing-xml.html#state-description)

A small fix in this documentation would be necessary.

> The pattern attribute can be used for Number and String items.

This is also applicable to DateTime items.

lolodomo:
> It only saves this state description pattern from the channel type to metadata of the item if it is not the default state description pattern for a Number type.

@mherwege : can you check you are aligned with my PR openhab/openhab-core#4175 for the default pattern.

As a reminder, for a number with dimension item, ChannelStateDescriptionProvider will set "%.0f" as default pattern if none is provided by the channel.
With my PR, ChannelStateDescriptionProvider will no more set any default pattern but DefaultStateDescriptionFragmentProvider will set "%.0f %unit%" as default pattern.

The current order when merging state descriptions from the different providers, from the highest priority to the lowest priority:

1. MetadataStateDescriptionFragmentProvider
2. GenericItemProvider (the .items file)
3. ChannelStateDescriptionProvider 
4. DefaultStateDescriptionFragmentProvider 

With my change, it will become:

1. MetadataStateDescriptionFragmentProvider
2. GenericItemProvider (the .items file)
3. ChannelStateDescriptionProvider 
4. DefaultStateDescriptionFragmentProvider

mherwege:
> @mherwege : can you check you are aligned with my PR [openhab/openhab-core#4175](https://github.com/openhab/openhab-core/pull/4175) for the default pattern.

I am aligned with that and I believe both complement each other well.

Note that this PR at its current state only does impact the state description pattern for items with dimension. With the changes in https://github.com/openhab/openhab-core/pull/4175 a next step would be to expose the state description pattern for all items and set it in he same way as for items with dimension.

This would solve the issue of potential conflicting channel state description patterns when linking items to two channels with different state description patterns (at least in the UI when creating items from a channel). It effectively promotes one of the state description patterns to the highest priority level by setting it explicitly in metadata.

lolodomo:
> I am aligned with that and I believe both complement each other well.

Normally with my PR, you don't need to set any default state pattern.

mherwege:
> Normally with my PR, you don't need to set a default state pattern if not provided by channel.

This PR only sets it when creating an item from a channel or thing  (model).

florian-h05 (approved):
LGTM, thanks!

Just one thing: IMO we should disable the state description field for existing Items, because changing the state description is a potentially dangerous operation. If you change the pattern, the rest of the metadata is overwritten, so you may loose your defined pattern etc. by mistake.
I will soon push a commit to disable this field for existing Items and show a message:

> Pattern can only be changed from the state description metadata page after Item creation!

florian-h05:
@mherwege Please let me know if you agree!

mherwege:
> @mherwege Please let me know if you agree!

Wouldn’t it be better to merge with the rest of the state description if one already exists on save to avoid this?
E.g. if you change the unit, you may also want to change the change description pattern. Not showing the second field makes it very confusing again.

florian-h05:
> Wouldn’t it be better to merge with the rest of the state description if one already exists on save to avoid this?

This would allow to change the state description pattern easily when changing the unit, but in case you have a min/max you would have to go the state description metadata page anyway to change them. And „hiding“ min/max when allowing to change the state description pattern could cause confusion.

> Not showing the second field makes it very confusing again.

It is not hidden, only disabled with a hint to go to the state description metadata page.


IMO it is safer to disable the state description pattern field in the item-form, because you then force the user to see and think about the whole state description when he wants to change the pattern afterwards.

mherwege:
> It is not hidden, only disabled with a hint to go to the state description metadata page.
> 
> IMO it is safer to disable the state description pattern field in the item-form, because you then force the user to see and think about the whole state description when he wants to change the pattern afterwards.

OK to disable.

while I started this for the state description pattern in the context of UOM, I still think there is further room for improvement. It is not only the state description pattern that can contain conflicting information when linked to multiple channels, but the whole of the state description. For the pattern, it is solved here in the UI by copying he pattern from the channel when creating the item, thereby giving a higher priority to the first one (with the ability to change). Something similar would probably be good to do for the rest of the state description. But I considered that beyond he scope of his PR.

florian-h05:
> I still think there is further room for improvement.

Agreed and your proposal for another PR sounds nice :+1:

andrewfg:
@mherwege many thanks for the contribution!

PR #2492, opened 2024-03-23
[BasicUI] All buttons forced on one line for a Player item

lolodomo:
Follow-up #2341
Partial revert to make two exceptions and restore old behaviour (buttons on a unique line) for:
* Control buttons of a Player item
* Buttons in the settings page


kaikreuzer (approved):
Thanks!

PR #2526, opened 2024-04-07
Thing details: Fix code tab not editable

### LINKED ISSUE #2523: Cannot update using the Code tab

- [ ] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

It is not possible to update things using the Code tab

## Expected behavior

Update using the Code tab was possible up to 4.1

## Steps to reproduce

1. Pick an existing thing
2. Select the Code tab
3. Try to update any line -> it will be refused 

## Your environment

OH 4.2.0.M1 (it also happens in snapshot 4027)
Windows 11
Zulu Java 17

## Additional information

This is a serious problem with high impact. It also occurs in QNAP / QTS

florian-h05:
Fixes #2523.

Regression from one of the previous PRs to the Thing details page.

This is a timing issue similar to #2375, where the code editor is initialized as read-only because the Thing is not loaded yet. 
After initialisation, the editor can not be set to read-write.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -151,7 +151,7 @@
 
       <f7-tab id="code" :tab-active="currentTab === 'code'">
         <f7-icon v-if="!editable" f7="lock" class="float-right margin" style="opacity:0.5; z-index: 4000; user-select: none;" size="50" color="gray" :tooltip="notEditableMsg" />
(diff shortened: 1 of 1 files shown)


kaikreuzer:
@florian-h05 Do I understand your comment on https://github.com/openhab/openhab-webui/issues/2523 correctly that this should better not be merged for M2 as it is untested?

florian-h05:
I expect it to fix the issue as it is similar to the other linked PR, but I can’t verify that it really fixes the issue as I can’t test.
However I don’t think there will be a regression from this PR, so it should be safe to merge. It might not solve the issue though.

kaikreuzer:
Ok, thanks. So if there is no risk through it, let's merge and see if it fixes the issue.

kaikreuzer (approved):
Thanks!

PR #2534, opened 2024-04-14
[BasicUI] Add special handling for "none" icon

lolodomo:
Same behaviour as in the Android app is now implemented for the "none" icon. This icon means "no icon" and we don't want to request the icon servlet. When this icon is associated to a sitemap element, the icon servlet is no more requested. Instead, a simple empty SVG tag is considered and inserted in the HTML page. There is also no dynamic behaviour (with item state) applied for this icon.

Additionally, in case the loading of any OH icon fails, the IMG tag is replaced by an empty SVG tag in the HTML page instead of updating the IMG source with the none icon URL (triggering a potential new icon servlet request).

Fix openhab/openhab-core#3977

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
@@ -63,9 +63,10 @@ public abstract class AbstractWidgetRenderer implements WidgetRenderer {
     private static final String ICON_SOURCE_MATERIAL = "material";
     private static final String ICON_SOURCE_FRAMEWORK7 = "f7";
     private static final String ICON_SET_OH_CLASSIC = "classic";
+    private static final String ICON_NAME_NONE = "none";
     private static final String DEFAULT_ICON_SOURCE = ICON_SOURCE_OH;
     private static final String DEFAULT_ICON_SET = ICON_SET_OH_CLASSIC;
-    private static final String DEFAULT_ICON_NAME = "none";
+    private static final String DEFAULT_ICON_NAME = ICON_NAME_NONE;
 
     public static final String ICON_TYPE = "svg";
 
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
```diff
@@ -166,7 +167,11 @@ protected String preprocessIcon(String originalSnippet, @Nullable String icon, b
         try {
             switch (iconSource.toLowerCase()) {
                 case ICON_SOURCE_OH:
-                    iconSnippet = getSnippet(ignoreState ? "icon_oh_no_state" : "icon_oh");
+                    if (ICON_SET_OH_CLASSIC.equals(iconSet) && ICON_NAME_NONE.equals(iconName)) {
```
kaikreuzer, line 170: Is there a special reason why we apply this logic only for the classic icon set?

mueller-ma: In the Android app `none` is treat as "no icon", but for `f7:none` a request is made: https://github.com/openhab/openhab-android/blob/9a2a12726e2eed8a7bb810e2b3afd6e1f9535cbf/mobile/src/main/java/org/openhab/habdroid/model/IconResource.kt#L158

Reason for this is simple: The handling of `none` has been implemented when only classic existed and hasn't been adapted later. 

If https://github.com/openhab/openhab-core/issues/4042 will be implemented and all icons without explicit icon source will be prefixed in the rest API, a user would need to define `oh:classic:none` to hide the icon. So I suggest to handle `none` as "no icon" for all icon sets, even when that disables all icons called `none` that may exist in other icon sets.

WDYT?

lolodomo: We are talking only about OH icon source or also other icon sources (material, framework7, ...) ?
I would suggest to keep that only for OH icon source but for any icon set of that source.

mueller-ma: Yes, makes sense. We have full control about all icon sets in the openHAB icon source.

kaikreuzer: @lolodomo If it should apply to any OH icon source, I assume you will do an update to this PR?

lolodomo: Yes, a small change is then needed to handle none for OH icon source whatever the icon set.

lolodomo: @kaikreuzer : this is now changed.

kaikreuzer (approved):
Thanks!

dilyanpalauzov:
Are now `icon=""` and `icon="none"` semantically identical?

mueller-ma:
At least for Android it's handled the same way

PR #2535, opened 2024-04-19
[BasicUI] Fix slider behaviour and implement two different modes

### LINKED ISSUE #2525: [BasicUI] Slider not working well, especially in Firefox

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

When clicking in the slider using a mouse, two problems can occur:
1. Basic UI emits an event to the bus but with the old value of the slider => the item state is not changed
2. Basic UI emits a correct event to the bus but the feedback SSE event is ignored and the value at the left of the slider is not updated

The first issue occurs almost all the time when using Firefox but not when using Chrome.
The difference between Firefox and Chrome is that Firefox receives first event `mouseup` and then `event` change, while Chrome receives events in the reverse order. Maybe the new value is not yet available until event `change` is received.

The second problem can occur in Firefox and in Chrome. It could be because resuming of listening to SSE events is resumed too late.

## Expected behavior

1. Correct event with the new value should be sent to the bus when using Firefox.
2. The value displayed at left of the slider should always be correctly refreshed.

## Your environment

Windows 10, with Firefox or Chrome browser.

lolodomo:
Fix #2525
Also related to openhab/openhab-core#4084 and openhab/openhab-core#3430

Two behaviour modes are now supported.
- If releaseOnly parameter is set, the new value is sent to the item only when the slider is released.
- If releaseOnly parameter is not set, new values are sent to the item while moving the slider. Events are sent at a certain frequency, this frequency is defined by the sendFrequency parameter if set or every 200 ms by default. Event is not sent when the value is is unchanged (when stopping the move but keeping the mouse pressed).

The brightness slider of the colorpicker widget is also updated to send regular new color commands when moving the slider.

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SliderRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SliderRenderer.java
@@ -83,6 +83,7 @@ public class SliderRenderer extends AbstractWidgetRenderer {
 
         snippet = preprocessSnippet(snippet, w);
         snippet = snippet.replace("%frequency%", frequency);
+        snippet = snippet.replace("%release_only%", s.isReleaseOnly() ? "true" : "flase");
         snippet = snippet.replace("%switch%", s.isSwitchEnabled() ? "1" : "0");
         snippet = snippet.replace("%unit%", unit == null ? "" : unit);
         snippet = snippet.replace("%minValue%", minValueOf(s));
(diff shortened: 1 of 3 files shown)


lolodomo:
Change tested with success using Firefox, Chrome and Edge on Windows 10  and using Chrome on an Android phone.

Note that the previous implementation tries to handle browsers not respecting the standard* but with problems with all browsers respecting this standard.
I can"t test with Safari browser but I read that Safari respects the standard so it should work well.

* Expected behaviour for a browser is to trigger "input" events when the slider is moving but not "change" events. "change" event should be triggered only when the slider is finally released. I read that 10 years ago, browsers like Internet Explorer were triggering "change" events while the slider was moving.

kaikreuzer (approved):
Thanks!

PR #2538, opened 2024-04-20
[BasicUI] Fix update of buttons (switch element)

### LINKED ISSUE #2537: [BasicUI] Switch with buttons: item state update can be ignored

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

It can happen that an item state update is ignored for a switch element rendered with buttons.
It happens after clicking twice on the same button. Then the next received item state update that should change the selected button is ignored.

## Expected behavior

When receiving any item state update, the buttons should be updated to show this last state whatever the previous actions done on this sitemap element.

## Steps to reproduce

1. Click a first time to a particular button. The command associated to this button is sent to the item.
2. Click a second time on a same button. The command associated to this button is sent to the item but this time there is no feedback state received because the item state is not changed (same state).
3. Now change the item state from another mean (another sitemap element for example). You can see that the sitemap element is not updated, the same button is still enabled while another should. 

## Additional information

This is due to the fact that when clicking a button, the state of all buttons is updated immediately before sending the command and then Basic UI is setting a flag to ignore the next event which generally is the feedback event to the command acting the change of the state. But if the command does not trigger any state update, Basic UI still is in a state that the next update will be ignored for this element.

I believe the idea was to show immediately the result of pushing a button without waiting for the server feedback. And then there was a kind of optimization to not reselect the button when receiving the feedback event as it is the most frequent case.

To cover all cases, this optimisation has to be removed.

lolodomo:
Fix #2537

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1156,7 +1156,6 @@
 		_t.ignoreState = _t.parentNode.getAttribute("data-ignore-state") === "true";
 		_t.hasValue = _t.parentNode.getAttribute("data-has-value") === "true";
 		_t.value = _t.parentNode.parentNode.querySelector(o.formValue);
-		_t.suppressUpdateButtons = false;
 
 		_t.reset = function() {
 			_t.buttons.forEach(function(button) {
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Let's hope that this does not cause any regression - I would imagine that this was there for a reason and not just for a theoretical technical optimization...

PR #2545, opened 2024-04-21
Charts: Add option to extend initial period to future & Refactor prop defs

florian-h05:
This adds a new option to charts which allows to extend the initial period to the future by default.
Until now, when a chart was opened its period always extended to the past.
This is especially useful with the new timeseries support for Items as it allows to display forecasts. 

Example:
When opening a temperature chart with period `D`, the chart opened up with the temperature for the last 24h.
When setting the new `future` prop to true, the chart will now open up with the temperature for the next 24h.

This also refactors the chart property definitions to use the helpers.

image


florian-h05:
@jlaur FYI, since you work much with forecasts in your energy data service binding, you might find this useful.

PR #2553, opened 2024-04-30
Sitemap editor: Add press release button support

mherwege:
Related to https://github.com/openhab/openhab-core/pull/4183
Requires https://github.com/openhab/openhab-core/pull/4204

This PR implements press/release button support.

To make this work, some refactoring had to be done to allow proper quoting of arguments in mappings and conditions. So far, quotes where stripped and the full mapping or condition strings where interpreted by splitting on special characters (=, : ...). This is limiting as it blocks these characters from being used in the arguments or makes working with blanks difficult. The sitemap syntax explicitely works with quoted strings for this, but that information was lost in the current implementation.
Quotes are now preserved, and therefore they need to be removed in core at usage (https://github.com/openhab/openhab-core/pull/4204).
Note that this introduced a backward compatibility constraint, as sitemaps saved in the UI, containing mappings or conditions with blanks, will require quotes to be manually added in the UI in order to be able to save changes to it. If not touched, the sitemaps should still work.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
+++ b/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
@@ -44,7 +44,7 @@
     comma:            ',',
     colon:            ':',
     hyphen:           '-',
-    number:           /-?[0-9]+(?:\.[0-9]*)?/,
+    number:           /[+-]?[0-9]+(?:\.[0-9]*)?/,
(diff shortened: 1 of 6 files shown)


lolodomo:
@mherwege : please rebase your branch. Build is failing due to a change in core framework that needed a fix in Basic UI.

mherwege:
@lolodomo done

florian-h05 (approved):
LGTM, thanks!

PR #2554, opened 2024-04-30
Sitemap editor: Fix buttongrid editor

mherwege:
Fields should use all space available. In a button grid definition the row and column fields where not visible.

I can't say what caused this, but this PR should fix this.


florian-h05 (approved):
LGTM, thanks!

PR #2562, opened 2024-05-04
[blockly] add more thing blocks

stefan-hoehn:
Add missing blocks to work with a thing or a group of things

- oh_getthing (one particular)
- oh_getthing_state (the state of the string - to be consistent with items)
- oh_things (get all things - can be used to iterate over all things)
- oh_getthing_attribute (retrieve any attribute of a thing)

image

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-things.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-things.js
@@ -6,6 +6,9 @@
 import Blockly from 'blockly'
 import { javascriptGenerator } from 'blockly/javascript.js'
 import { FieldThingPicker } from './fields/thing-field.js'
+import { blockGetCheckedInputType } from './utils.js'
+
+const unavailMsg = 'Advanced Thing blocks aren\'t supported in "application/javascript;version=ECMAScript-5.1"'
 
 export default function defineOHBlocks (f7, isGraalJs) {
   Blockly.Blocks['oh_thing'] = {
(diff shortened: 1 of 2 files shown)


rkoshak:
> Changing this would imply a complete reimplementation of the standard block which IMHO is not worth the effort.

I've noticed this too and agree. It's an unfortunate coincidence but I kind of blame OH for using the term "Item" the way it does in the first place instead of something less likely to conflict (Thing too). But since that choice was made long before even my time with OH, :shrug: .

stefan-hoehn:
re "Thing". I understand why Thing was used -> Internet of Things, so I am happy with that. Wording is always difficult, I suppose. I would rather have hoped that blockly had chosen "element" instead of item ;-)

florian-h05 (commented):
Overall looks good to me, thanks.

Just two things:
- I would rather name the get all Things block `oh_getthings` instead of `oh_things` because this is more inline with `oh_getthing`.
- Why do you have "complicated" code for the Thing attribute block? It also works with simpler code.

I will soon push my review changes for these two points, please have a look at them.

Boldfor:
@stefan-hoehn, thanks a lot for your work!

Quick question, maybe I misunderstand something or use things wrongly (I also couldn't answer it myself with the [docu](https://next.openhab.org/docs/configuration/blockly/rules-blockly-items-things.html#get-thing-status)): 
image
Thr first two logging-blocks work with the new thing-loop, the third one throws an error.
```
2024-07-11 21:57:33.523 [ERROR] [script.javascript.thing_status_check] - Failed to execute script: TypeError: things.getThing is not a function
        at <js>.:program(<eval>:10)
        at org.graalvm.polyglot.Context.eval(Context.java:399)
        at com.oracle.truffle.js.scriptengine.GraalJSScriptEngine.eval(GraalJSScriptEngine.java:458)
        at com.oracle.truffle.js.scriptengine.GraalJSScriptEngine.eval(GraalJSScriptEngine.java:426)
        at java.scripting/javax.script.AbstractScriptEngine.eval(AbstractScriptEngine.java:262)
        ... 75 more
2024-07-11 21:57:33.524 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'thing_status_check' failed: org.graalvm.polyglot.PolyglotException: TypeError: things.getThing is not a function
```
Am I using things wrongly? If yes, maybe others run into the same problem and this usage could be blocked in blockly?

rkoshak:
For 1 I think you need to file a new issue. Be sure to also post the full rule from the code tab so we can see the XML that makes up the blocks, the JS that the blocks compiled to, and the full rule context.

For 2 see https://community.openhab.org/t/thing-status-reporting-4-0-0-0-4-9-9-9/143180. Ultimately adding a new trigger would be an issue for core but this rule template can be used as is or as an example for how to work with the generic rule trigger in UI rules.

Boldfor:
> For 1 I think you need to file a new issue. Be sure to also post the full rule from the code tab so we can see the XML that makes up the blocks, the JS that the blocks compiled to, and the full rule context.

Thanks. I just did here: https://github.com/openhab/openhab-webui/issues/2670.
Let me know if I missed something that's needed as well.

PR #2564, opened 2024-05-06
[blockly] Adjust persistence blocks to breaking changes

florian-h05:
This adjusts the persistence blocks for the **breaking** changes of `org.openhab.core.persistence.extensions.PersistenceExtensions`.

For NashornJS, see https://github.com/openhab/openhab-core/pull/3736 for the changes.
For GraalJS, see https://next.openhab.org/addons/automation/jsscripting/#itempersistence for the current API docs and https://github.com/openhab/openhab-js/pull/331 for the changes.

In a follow-up PR, Blockly should be extended to make use of the new capabilities provided by the changed return type of variance, deviation, average, sum & delta.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
@@ -166,7 +166,7 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs, persistenceSe
         persistenceNameInput.setShadowDom(Blockly.utils.xml.textToDom('<shadow type="oh_persistence_dropdown" />'))
       }
       this.appendValueInput('dayInfo')
-        .appendField(new Blockly.FieldDropdown([['has changed since', 'changedSince'], ['has been updated since', 'updatedSince']]), 'methodName')
+        .appendField(new Blockly.FieldDropdown([['has changed since', 'changedSince'], ['will have changed until', 'changedUntil'], ['has been updated since', 'updatedSince'], ['will have been updated until', 'updatedUntil']]), 'methodName')
         .setAlign(Blockly.ALIGN_RIGHT)
         .setCheck(['ZonedDateTime'])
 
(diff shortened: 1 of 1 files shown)


stefan-hoehn:
Tested together in a session with @florian-h05 in an online session and looks good to us.
Can be merged.

**This is a BREAKING change that requires all Blockly rules that use persistence blocks to be reopened and saved again.**

stefan-hoehn (approved):

mherwege:
> In a follow-up PR, Blockly should be extended to make use of the new capabilities provided by the changed return type of variance, deviation, average, sum & delta.

@stefan-hoehn Do you have plans to do this? This will be another breaking change and I think we should try to do it before the release to have all changes done in one go. I will have to adapt my blocky rules working with persistence and quantity-types myself, but it will make it a lot cleaner afterwards. Currently I convert to text and add the unit atthe end again before updating an item. I just want to work with the values with the unit. Typed variables will also help for this.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
```diff
@@ -121,28 +138,53 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs, persistenceSe
...
+      case 'deltaSince':
+      case 'deltaUntil':
+      case 'deviationSince':
+      case 'deviationUntil':
+      case 'sumSince':
+      case 'sumUntil':
+      case 'varianceSince':
+      case 'varianceUntil':
+        dayInfo = javascriptGenerator.valueToCode(block, 'dayInfo', javascriptGenerator.ORDER_NONE)
+        code = (isGraalJs) ? `${itemCode}.persistence.${methodName}(${dayInfo}${persistenceExtension})?.numericState` : `parseFloat(${persistence}.${methodName}(${itemCode}, ${dayInfo}${persistenceExtension}).getState())`
```
mherwege, line 182: @stefan-hoehn This is inconsistent. For GraalJS, you focus on keeping backward compatibility by returning `numericState`. For Nashorn, it returns the full state.
Moreover, by returning `numericState` in GraalJS, you basically strip the unit and make it impossible to do straight quantity calculations on the result. I understand this would be a breaking change to remove this, but I think we should do it now rather than wait until after 4.2.
As an alternative, one could create extra methods (something like `deltaStateSince`). That would allow to keep backward compatibility for GraalJS at the expense of having different and less intuitive names in Blockly vs pure JS. I would prefer the breaking change here.
@florian-h05 What do you think?

florian-h05: This is not inconsistent, for Nashorn there is also a numeric state returned (note the parseFloat around the state).

I have done this PR this way because after the breaking changes were done to code and the helper library, we needed to update Blockly quickly before the new milestone so it is not effectively „broken“.
Adjusting Blockly to allow use the new functionality was postponed to a later PR.

See https://github.com/openhab/openhab-webui/issues/2595#issuecomment-2139088094 for my implementation proposal.

PR #2565, opened 2024-05-06
Setup wizard: Add primary IP config

mherwege:
Add-on finders scan the network for devices supported by OH add-ons to suggest suitable add-ons. These are presented in the setup-wizard (and add-ons store).

To limit network traffic, especially for ip broadcast and multicast, finders could limit the traffic to one subnet. This is especially relevant if the setup would be on servers with many network interfaces or when using Docker. The broadcast address can be controlled through network settings. However, these network settings so far are only accessible through the UI after initial setup.
If the broadcast address or primary IP address is not set, one will be used from the list randomly.

This PR adds one step in the setup process, setting up a primary IP address. This will automatically also default the broadcast address to the broadcast address of the primary IP subnet. Querying the suggestion finder service for add-on suggestions is delayed until after this step, and some delay is built into the process to allow suggestions finders to scan the network.

See discussion in https://github.com/openhab/openhab-core/pull/4036

image

image


florian-h05 (commented):
In general looks good to me, however there are a few points I will take care of:

- Seems like something went wrong when rebasing my changes from #2556 (I will clean-up the re-added code)
- Use parameter-options for primary network selection: This makes only one of the network interfaces selectable and improves consistency with the actual network settings.
- Refactor API call logic to use one big Promise.all (makes the code more readable and loading more predictable).
- Handle locale change for the add-on selection inside the addon-setup-wizard component (related to the code clean-up above).
- Fix an issue, where suggestions were not applied when an add-on tab was opened before the suggestions were loaded.

florian-h05 (approved):

lolodomo:
I just tested snapshot 4077 on a Windows PC. I do not see any step "Setup Primary Network" during the Setup wizard.

lolodomo:
Don't know if this is important but I passed/ignored the previous step (localisation).

florian-h05:
Skipping location doesn’t matter.
The primary network step only shows up when you have more that one interface so select from.

lolodomo:
> The primary network step only shows up when you have more that one interface so select from.

Ok. On Windows, it looks like the WiFi interface does not linked to an IP address when the Ethernet interface is up, even if WiFi is enabled.

PR #2568, opened 2024-05-07
Setup Wizard: improve suggestions refresh on network change

mherwege:
@florian-h05 I was working on an improvement in parallel with your improvements and merge. Thank you for your improvements.
When I was testing it, I still saw 2 issues I tried to resolve:

1. There would also be a delay when setting the network without a change. This is not required.
2. While in the wait cycle, going back to the network selection and setting the network again (with or without change) would lead to the progress bar alternating between short and long because 2 timer cycles would be running in parallel. I now cancel the existing timer if there is a real network change.

I hope these improvements still make sense.


florian-h05 (approved):
LGTM, thanks!

I just pushed a minor JSDoc improvement, other then this I changed nothing.

PR #2536, opened 2024-04-20
[BasicUI] Implement press & release button

lolodomo:
Related to openhab/openhab-core#3822

Depends on openhab/openhab-core#4183

Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #2572, opened 2024-05-09
[BasicUI] Remove a call to window.console

lolodomo:
Was added when implementing #2536 for test purpose and I forgot to remove it.


kaikreuzer (approved):
Thanks!

PR #2575, opened 2024-05-12
[blockly] Add timer context

stefan-hoehn:
Adds the possibility to provide any information to the timer's statement block that can be used when this block is eventually executed. This has become possible with the awesome addition in the jsscripting lib that @florian-h05 has provided to the createTimer-function.

Here are some examples how this can be used:

image
 
pinging @rkoshak as you might be interested to notice that.


florian-h05 (approved):
LGTM thanks!
I have to admit the code looks weird if there is no context, but it works and I couldn't find a quick solution.

PR #2576, opened 2024-05-14
[CI] Add Java 21 compiling to J17 and native J21

holgerfriedrich:
As we have implemented all changes that allow for Java 21 builds, I'd like to introduce it in the CI as well.
For core and addons, the same approach was already implemented quite a while ago.

Refs: openhab/openhab-distro#1590

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ '17' ]
+        java: [ '17', '21', 'profile-j21' ]
         maven: [ '3.9.6' ]
         python: [ '3.12' ]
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):

florian-h05:
@holgerfriedrich Seems like there is an issue with SAT annotation upload … I guess only one of the three build jobs can upload its SAG annotations, not all three.

holgerfriedrich:
> @holgerfriedrich Seems like there is an issue with SAT annotation upload … I guess only one of the three build jobs can upload its SAG annotations, not all three.

Yes, exactly. Hopefully this is already fixed, lets wait for the CI to complete....

florian-h05 (approved):
Looks good now, thanks!

PR #2578, opened 2024-05-18
[BasicUI] Remove a useless call to function reloadIcon (slider widget)

lolodomo:
Follow up #2535

The function reloadIcon is already called by the function setValue at line 636 just before calling setValuePrivate.

The icon continues being refreshed while moving the slider except if releaseOnly parameter is set.


kaikreuzer (approved):
Thanks!

PR #2580, opened 2024-05-18
[BasicUI] Switch between several Button elements at a place in the grid

lolodomo:
Several Button elements are accepted at the same place in the grid only when they all have visibility conditions.
In this case, the user must define proper conditions so that only one button is visible at any time.


lolodomo:
Here is an example with two buttons switching depending on the item state:
```
            Button item=DemoSwitch row=1 column=1 stateless click=ON staticIcon=switch-off visibility=[==OFF]
            Button item=DemoSwitch row=1 column=1 stateless click=OFF staticIcon=switch-on visibility=[==ON]
```

kaikreuzer:
@lolodomo Do you wish this to be included into the next milestone?

lolodomo:
Yes Kai, please, it finishes the feature.

kaikreuzer (approved):
Thanks!

Issue #2050, opened 2023-09-02
Improve binding selection by filtering (including usage of binding metadata)


jlaur:
Seeing the nice results of #1997 I thought I'd finally suggest to use the new [add-on metadata](https://www.openhab.org/docs/developer/addons/addon.html#xml-structure-for-add-on-definitions) introduced in 4.0 to provide either filtering or additional search options.

I don't know if this is technically feasible, i.e. if the UI has access to these metadata when bindings are not yet installed. If not, please feel free to close this issue.

## The problem

There is currently more than 350 bindings available which are shown in a long list per default:

![image](https://github.com/openhab/openhab-webui/assets/19519842/3a164604-4be7-4075-b5f2-ac4ea0d9e354)

It's possible to search by name, so it's not hard to find a specific binding.

## Your suggestion

I'm not UI expert so I'm not sure exactly what would be the best suggestion here. But I will throw a few ideas anyway. 🙂

First, I'm wondering if the search could be changed to work the same way as when looking for Things, so that it would be more like a filter:

![image](https://github.com/openhab/openhab-webui/assets/19519842/96528eaf-2a75-4078-ad22-d1dd1c9e9705)

Currently you have to click the Search icon to go to a separate view that initially is empty. Then you can perform the search (which btw seems to have very big icons?):

![image](https://github.com/openhab/openhab-webui/assets/19519842/dc8439f9-8098-47dd-addb-515108dd3b4e)

I think it would simplify the UI being able to filter directly on the first page without having to navigate to a separate search page. Also it would make the UI more consistent if it behaved the same way as for example when looking for Things and Items.

Additionally I'm wondering if the metadata **connection** and **country** could be taken into consideration. Per default we could exclude bindings having **countries** declared where none of those countries match the country configured in regional settings. So for example, when in Denmark:

![image](https://github.com/openhab/openhab-webui/assets/19519842/2d520385-ab44-4d38-b8d1-aa65932f793f)

I would not see any of these:
```shell
$ grep -R "<countries>" | grep -v "dk"
bundles/org.openhab.binding.dsmr/src/main/resources/OH-INF/addon/addon.xml:     <countries>nl,be,lu,at</countries>
bundles/org.openhab.binding.hccrubbishcollection/src/main/resources/OH-INF/addon/addon.xml:     <countries>nz</countries>
bundles/org.openhab.binding.kvv/src/main/resources/OH-INF/addon/addon.xml:      <countries>de</countries>
[...]

PR #2582, opened 2024-05-20
Add-on store: Add connection and country filter

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/2050

This implements filtering for connection type and country (related to region/country setting) in the addon store.

By default, it will filter for all connection types and filter out addons not supported for a specific country if info on countries is available.

Here is how this looks like on an Android mobile phone (Galaxy S22).

Filters dropdown:
bs_realdroid_Mobile_Samsung Galaxy S22-12 0-1080x2340

Change connection type to 'No cloud allowed':
bs_realdroid_Mobile_Samsung Galaxy S22-12 0-1080x2340 (1)

Search for 'amazon' (only one found):
bs_realdroid_Mobile_Samsung Galaxy S22-12 0-1080x2340 (2)

Switch to 'Cloud allowed' again, search result updated to 2 add-ons:
bs_realdroid_Mobile_Samsung Galaxy S22-12 0-1080x2340 (3)

Clear search and filter on country 'Specific for country':
bs_realdroid_Mobile_Samsung Galaxy S22-12 0-1080x2340 (4)

Select bindings tab, only one binding shown:
bs_realdroid_Mobile_Samsung Galaxy S22-12 0-1080x2340 (5)

Detail of binding shows connection and country detail when available:
bs_realdroid_Mobile_Samsung Galaxy S22-12 0-1080x2340 (6)

I tried this same procedure on Chrome and Firefox (Windows), iPhone 14 (iOS 16) and iPad 9th Gen (iOS 15) and it looks OK on all these.


bundles/org.openhab.ui/web/src/assets/addon-store.js
```diff
@@ -41,6 +41,27 @@ export const AddonSuggestionLabels = {
   }
 }
 
+export const AddonConnectionTypes = {
+  cloud: {
+    label: 'Cloud allowed',
+    values: ['none', 'local', 'hybrid', 'cloud']
+  },
+  cloudConfig: {
+    label: 'Cloud configuration only',
```
jlaur, line 50: This text is slightly misleading, since according to the current [definition](https://www.openhab.org/docs/developer/addons/addon.html#xml-structure-for-add-on-definitions) hybrid means "add-ons that interact locally without internet access, but can optionally use a cloud connection for extended functionality (such as discovery)". This is not specific to configuration.

mherwege: Would ‘cloud discovery only’ fit better? I need to keep it short to fit the UI on a mobile.

ghys: I would also reconsider "Exclude not for country" as it's quite puzzling IMO.

Suggestion:
- All applicable to you
- Only your country
- All

jlaur: > Would ‘cloud discovery only’ fit better? I need to keep it short to fit the UI on a mobile.

I don't think so, since discovery is just an example. You could have a binding where 20% of the (optional) functionality is provided through a cloud service. Unfortunately I don't have a good suggestion for a short way to put this. Maybe something like "Optional cloud functionality"?

florian-h05 (commented):
Codewise LGTM, thanks!
Just two comments about the type of the smart select and the wording, I really like @ghys and @jlaur suggestions/proposals.

jlaur:
Just for completeness, I can think of two queries that will currently not be possible. You probably already thought about it and decided to keep it simple, because they are not common use-cases, but nevertheless:
- I cannot (out of curiosity) see which add-ons are available specifically for Germany (I'm in Denmark).
- I cannot (again, out of curiosity) see which add-ons runs purely local, i.e. not even requiring LAN access.

bundles/org.openhab.ui/web/src/components/addons/addon-info-table.vue
```diff
@@ -78,6 +78,22 @@ export default {
...
+          title: 'Connection Type',
+          value: this.addon.connection
+        })
+      }
+
+      if (this.addon.countries && this.addon.countries.length > 0) {
+        info.push({
+          id: 'countries',
+          title: 'Regions/Countries',
+          value: this.addon.countries.join(',').toUpperCase()
```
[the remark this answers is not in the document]
mherwege: Fair enough. I kept the space out because it may limit what you see if there are many countries listed. But I will add it.

bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
```diff
@@ -20,6 +19,28 @@
         :disable-button="!$theme.aurora"
         @searchbar:search="search"
         @searchbar:clear="clearSearch" />
+      <f7-list accordion-list style="margin-top: 0px; margin-bottom: 0px">
+        <f7-list-item accordion-item title="Filters">
+          <f7-accordion-content>
+            <f7-list>
+              <f7-list-item smart-select title="Connection Type" :smart-select-params="{ closeOnSelect: true }">
```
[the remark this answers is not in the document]
mherwege: OK

mherwege:
> * I cannot (out of curiosity) see which add-ons are available specifically for Germany (I'm in Denmark).

I think ideally that should be something for the website, rather than in the software. It would make filtering in the software more complex. I thought about it, but decided against it.

> * I cannot (again, out of curiosity) see which add-ons runs purely local, i.e. not even requiring LAN access.

Again, thought about it but considered add-ons that require no network access typically of limited use. LAN access would be accepted. However, this is easy enough to add and doesn't make the filtering more complex (4 instead of 3 options for the filter). So I added it.

mherwege:
@ghys @florian-h05 @jlaur Thank you all for the feedback. I incorporated it in my last commit.

florian-h05 (approved):
LGTM now, thanks!

I agree that any "advanced" filtering should be implemented on the website.

PR #2585, opened 2024-05-25
Provide on/off icons for soundvolume_mute

jlaur:
Related to openhab/openhab-core#4245


jlaur:
I found this script also which I'm not sure about - is it still in use?

https://github.com/openhab/openhab-webui/blob/d23f407b2a169eb938ceac306428648baf293dc8/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/.copy.sh#L40

lolodomo:
> I found this script also which I'm not sure about - is it still in use?

It is better to maintain it I believe.

jlaur:
> > I found this script also which I'm not sure about - is it still in use?
> 
> It is better to maintain it I believe.

I'm not sure how, can you guide me a bit? The script doesn't copy `soundvolume_mute.svg` currently, and it copies only `soundvolume-100.svg` into `soundvolume.svg`. What is the logic behind this?

lolodomo:
I would add these lines:
```
cp -f soundvolume-100.svg   soundvolume_mute-off.svg
cp -f soundvolume_mute-on.svg   soundvolume_mute.svg
```

jlaur:
> cp -f soundvolume-100.svg   soundvolume_mute-off.svg
> cp -f soundvolume_mute-on.svg   soundvolume_mute.svg

Done. Do you know what the script is used for? Since we have all icons (even duplicates) in the icons folder, the script could copy 1:1?
```shell
cp -f soundvolume_mute.svg      soundvolume_mute.svg
cp -f soundvolume_mute-off.svg  soundvolume_mute-off.svg
cp -f soundvolume_mute-on.svg   soundvolume_mute-on.svg
```

lolodomo (approved):
LGTM, thank you

lolodomo:
The logic of this script was to create all the expected icons from a minimal list of not duplicated icons and to be sure that icons that should be identical are really the same.
By running it you are sure for example that soundvolume.svg is not different than soundvolume-100.svg.
If we don't use it, we have to be sure for example in your case that you really copy/paste the existing files and you did not create a new icon.

I have just checked the appearance of the new icons, I assumed that you just copied the existing icons. I am sure at 99.999% this is what you did ;)

jlaur:
> I have just checked the appearance of the new icons, I assumed that you just copied the existing icons. I am sure at 99.999% this is what you did ;)

I copied them according to your suggestion in the core PR. 🙂

PR #2588, opened 2024-05-26
[blockly] Fix Thing "is enabled" code generation

stefan-hoehn:
blockly thing isEnabled produces undefined

using thing.isEnabled instead of thing.enabled


florian-h05 (approved):
LGTM, thanks for the fix!

PR #2600, opened 2024-06-02
Add-on store: Fix region is null error

mherwege:
Fixes https://github.com/openhab/openhab-core/issues/4266

Error was introduced by https://github.com/openhab/openhab-webui/pull/2582 and manifests itself when no country was set in the setup wizard or the locale settings.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
+++ b/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
@@ -310,8 +310,10 @@ export default {
     load () {
       this.stopEventSource()
       this.$oh.api.get('/rest/services/org.openhab.i18n/config').then((data) => {
-        this.region = data.region ? data.region : ''
-        this.regionReady = true
+        if (data.region) {
+          this.region = data.region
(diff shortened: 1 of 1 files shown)


florian-h05:
What do you think about not setting regionReady in that case and hiding the then effectively useless region filter?

mherwege:
> What do you think about not setting regionReady in that case and hiding the then effectively useless region filter?

I am not too sure about that. Should I then show all? Right now the default is to not show the ones that have a region defined that does not match the region parameter. If I set it to an empty value, these will effectively be filtered, but with the option to still see all. The filter to show the specific ones will show empty.
If I remove the filter, I have to decide what to show all, or all with countries set filtered, with no option to change. Maybe there should be some kind of indication that there is no region set for the system.

mherwege:
@florian-h05 Thinking about it again, hiding the filter and showing all in that case is probably the best thing to do.

florian-h05:
I agree, I think with no region set, the filter behaviour could also be a bit confusing.

florian-h05 (approved):
LGTM, thanks for the fix!

Issue #2605, opened 2024-06-11
sitemaps with 'AND' in visibilty can not be saved after editing in code tab


nolantern:
## The problem
When I create a sitemap with the mainUI and use the `AND` operator in the visibility attribute, I can not save the sitemap after I edited it in the code tab. The "Sitemap updated" toast does not appear. If I reload the page, the version before my edits shows up.

## Expected behavior
I can save sitemaps that contain `AND` operators and have been edited in the code view of the mainUI.

## Steps to reproduce

1. Create a new sitemap
2. Add a Text widget
3. Add a visibility attribute with an `AND` operator
4. Go to the Code tab
5. Change anything using the Code tab (for example the name of a widget)
6. Try to save the sitemap

## Your environment

```yaml
runtimeInfo:
  version: 4.1.2
  buildString: Release Build
locale: de-DE
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 17.0.11
  javaVendor: Azul Systems, Inc.
  javaVendorVersion: Zulu17.50+19-CA
  osName: Linux
  osVersion: 5.15.0-1053-raspi
  osArchitecture: aarch64
  availableProcessors: 4
  freeMemory: 509455240
  totalMemory: 801112064
  uptime: 1381719
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-astro
  - binding-modbus
  - binding-mqtt
  - binding-network
  - binding-openweathermap
  - binding-shelly
  - binding-tibber
  - binding-tradfri
  - misc-openhabcloud
  - persistence-influxdb
  - persistence-rrd4j
  - transformation-jinja
  - transformation-jsonpath
  - transformation-regex
  - ui-basic
  - ui-habpanel
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: true
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
    pixelRatio: 1.5
    prefersColorScheme: dark
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 16
    language: de
    languages:
      - de
      - en-US
      - en
[...]

PR #2609, opened 2024-06-13
Sitemap editor: Fix issues with AND, buttongrid and numeric commands

mherwege:
Fixes https://github.com/openhab/openhab-webui/issues/2605
Also fixes issue with Buttongrid and non-numeric commands


florian-h05 (approved):

lolodomo:
@mherwege : does it cover new Button element?

mherwege:
@lolodomo No, it doesn't. And I won't be able to do that before the 4.2 release. It is a major refactoring, not just on the parser side, but mainly on the UI side. Ideally, I also don't want to be left with 2 totally different ways of defining a buttongrid in the UI, so the current approach may have to be replaced by a new approach that has the wider functionality. Of course, the old format should be parsed/converted automatically. In doing that, I actually may also need adapting the core part for this.
I started working on this, but the effort was quite big, and I have halted it for now. I will need to revisit.

This PR was just a fix to the already existing functionality.

PR #2610, opened 2024-06-14
Fix Main UI dark mode scrollbars in Chrome

wborn:
This fixes the issue that when using Chrome the scrollbars are still very bright and not dark.

### Before

Screenshot from 2024-06-15 00-33-01

### After

Screenshot from 2024-06-15 00-33-22


florian-h05 (approved):
LGTM, thanks!

PR #2611, opened 2024-06-16
Sitemap edit: Fix column width wrongly applied

### LINKED ISSUE #2574: Sitemap editor has a lot of unused space
## The problem

Today I pulled the latest snapshot Docker image. The Sitemap editor wastes a lot of space:

grafik

I tried to reload without cache, but that didn't help.

## Steps to reproduce

1. Open Sitemap editor

## Your environment

```yaml
runtimeInfo:
  version: 4.2.0
  buildString: "Build #4075"
```

Firefox 125

florian-h05:
Fixes #2574.

As I haven't been able to identify the root cause of the problem, let's use this workaround.


florian-h05:
@mueller-ma I finally fixed this issue, not sure what was the cause. Seems like Framework7 CSS did not properly work here, so I worked around that by "manually" setting the width from the components CSS.

PR #2626, opened 2024-06-25
Fix value mapped to NULL not shown in list label item

mherwege:

See description in https://github.com/openhab/openhab-core/pull/4203#issuecomment-2095792756

If a mapping is used to map a value to NULL, it is currently not shown in an oh_label_item.


florian-h05 (approved):
LGTM, thanks!

PR #2640, opened 2024-07-04
Developer sidebar: Fix script and scene "unpin all" links

JustinGeorgi:
The links to unpin scenes and scripts in the dev toolbar were set to 'rules' instead of the correct specifics, and so, didn't work.


florian-h05 (approved):
LGTM, thanks for the fix!

PR #2648, opened 2024-07-06
oh-sipclient: Add call status Item to report call status to server

florian-h05:
This adds the option to specify a String Item, to which the call status is sent.
This allows to track the usage of oh-sipclient from rules and scripts.

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-sipclient-card.md
+++ b/bundles/org.openhab.ui/doc/components/oh-sipclient-card.md
@@ -128,6 +128,11 @@ Usage is explained at the [`oh-sipclient` component docs](/docs/ui/components/oh
     Automatically dial the SIP address when loaded
   </PropDescription>
 </PropBlock>
+<PropBlock type="TEXT" name="stateItem" label="State Item" context="item">
+  <PropDescription>
+    String Item to publish the current SIP connection state to the openHAB server and make it accessible from rules etc.
+  </PropDescription>
+</PropBlock>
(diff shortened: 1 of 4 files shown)


digitaldan:
Seems like a cool feature!  One suggestion would be to have the item be able to be configured per browser (so using local storage) like we allow for other SIP configuration options, so add this to the localSettingsPopup part of the widget.    I have 5 (sometimes more) tablets that all pop up the SIP widget when a call comes in and i could see wanting to track those separately. 

On a bit of a tangent.  I have a Freeswitch PBX docker container i run that my SIP devices (like my grand stream gate phone)  connect to as well as some external phone numbers.  When a call comes in, its responsible for triggering an item update on openHAB which pops up my sip widgets.   Freeswitch keeps the incoming call ringing, even though my sip widgets are not registered yet, then  when they do pop up and register, it immediately rings them.   This is similar to how VOIP apps on mobile devices work, where the PBX sends a push notification that there is a call, which then wakes up the mobile sip app which immediately registers then gets sent the call.

florian-h05:
Agreed, sounds reasonable.

WDYT about the strings that are sent? Understandable and detailed enough?

digitaldan:
> WDYT about the strings that are sent? Understandable and detailed enough?

Yeah, LGTM!

florian-h05:
@digitaldan Can you please take a look at it? I have added support for local setting of the call state Item and extended its values to also include the caller ID.

digitaldan (approved):
LGTM!

PR #2652, opened 2024-07-07
Health menu: Remove sidebar and settings entry, instead use FAB

### LINKED ISSUE #2644: Health Check menu order
I just saw my MainUI and noticed the health check item. IMO, the item shouldn't be placed first under Settings, it should be last, since it's not something that people would need to access often.

I understand that this is just an opinion.

florian-h05:
Closes #2644.

This removes the health check entry from the sidebar and the settings menu and instead uses a floating action button (FAB) to allow access to the health checks page if there are any actual issues.

This is only a minor cosmetic change to a new feature, so it should be fine to merge that for RC2.


florian-h05:
Looks like this:

image

PR #2662, opened 2024-07-10
oh-context: Fix rendering failure when not in edit mode

florian-h05:
When not being in edit mode, oh-context did not render at all. Instead, an error was thrown:

```
[Vue warn]: Cannot set reactive property on undefined, null, or primitive value: undefined

found in

---> <OhContext> at src/components/widgets/system/oh-context.vue
```


florian-h05:
@JustinGeorgi FYI

PR #2511, opened 2024-03-29
Analyzer: Fix analyzer not initially visible on iOS 17.4

florian-h05:
When the analyzer was opened on iOS/iPadOS 17.4 devices, the screen stayed blank until the controls were opened.

ECharts logged the following warning to the console:

```
[ECharts] Can't get DOM width or height.Please check dom.clientWidth and dom.clientHeight.
They should not be 0.For example, you may need to call this in the callback of window.onload.
```

This is more or less a workaround, since I cannot really debug on iOS as I have no Mac and BrowserStack does not provide iOS 17.4 devices at the moment. I have checked that this workaround does not affect other devices.

Reported on the community: https://community.openhab.org/t/openhab-4-1-release-discussion/152252/188?u=florian-h05.

PR #2677, opened 2024-07-18
Charts: Fix issues with charts not displaying on iOS >= 17.4

florian-h05:
Follow-up for #2511.

I think I finally discovered the root cause for charts not displaying initially on iOS >= 17.4: The height of the chart was calculated to 0px. By using 100dvh (dynamic viewport height) and subtracting safe areas etc. instead of using 100%, I managed to fix this issue.


dbell37:
Thank you Florian for fixing this issue. :)

Boldfor:
Thanks to everyone involved. Works again as intended. 😃

Issue #2675, opened 2024-07-17
[BasicUI] Input widget converts doublequotes to `&quot;`


jimtng:
Test:

```java
String TestString1
```

```java
sitemap test {
  Input item=TestString1
}
```

Then open the sitemap on a browser, and enter `{"test":[]}`and press enter. BasicUI will send the correct string, but then converts that string to `{&quot;test&quot;:[]}` and displays it.

<img width="487" alt="image" src="https://github.com/user-attachments/assets/ed454b06-c858-4c60-b85b-2ea540196930">

It seems that I can't send another command by pressing enter, unless I modified the string. So I added a space and pressed enter, it would then send the command `{ &quot;test&quot;:[]}` (correctly as entered), but once again it urlencoded that string to `{ &amp;quot;test&amp;quot;:[] }` and displayed that.

Discovered/Reported here https://github.com/openhab/openhab-addons/issues/16479#issuecomment-2232653793

mherwege:
@jimtng I have been trying this in Chrome with a current snapshot (running OH from Eclipse IDE to debug this). I can't reproduce the issue.
What browser are you using?

jimtng:
Chrome Version 120.0.6099.199 (Official Build) (arm64) (macos 14.5)

mherwege:
Strange, I use Chrome Version 126.0.6478.183 (Official Build) (64-bit) on Windows. I tried with the exact same string and it works for me.

jimtng:
Strange indeed. I can reproduce it on Safari and Opera too. Currently running 4.3.0-SNAPSHOT - Build #4194

Just enter anything into the input field with either `&` or `"`

`&` turns into `&amp;` and `"` turns into `&quot;`

mherwege:
Even stranger now. I can reproduce it with &, but not with ".

mherwege:
I found the issue and created a PR.

PR #2685, opened 2024-07-25
[BasicUI] input widget, support html escape characters in input

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/2675

HTML escape characters in input widgets were not properly supported, leading to an escaped character sequence under certain conditions.

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -2318,7 +2318,7 @@
 		_t.prefixField = _t.parentNode.parentNode.querySelector(".mdl-form__input-prefix");
 		_t.postfixField = _t.parentNode.parentNode.querySelector(".mdl-form__input-postfix");
 		_t.prefix = _t.prefixField ? smarthome.UI.unEscapeHtml(_t.prefixField.innerHTML) : null;
-		_t.postfix = _t.prefixField ? smarthome.UI.unEscapeHtml(_t.postfixField.innerHTML) : null;
+		_t.postfix = _t.postfixField ? smarthome.UI.unEscapeHtml(_t.postfixField.innerHTML) : null;
 		_t.unitField =  _t.parentNode.parentNode.querySelector(".mdl-form__input-unit");
 		_t.verify = undefined;
 
@@ -2772,6 +2772,7 @@
 			return escapedText;
 		};
 
+		// See discussion at https://stackoverflow.com/questions/1912501/unescape-html-entities-in-javascript
 		_t.unEscapeHtml = function(html) {
 			var text = document.createElement("textarea");
 			text.innerHTML = html;
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -2773,7 +2773,7 @@
 		};
 
 		_t.unEscapeHtml = function(html) {
-			var text = document.createElement('textarea');
+			var text = document.createElement("textarea");
 			text.innerHTML = html;
 			return text.value;
 		};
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -2313,10 +2313,12 @@
...
 		_t.itemType = _t.parentNode.getAttribute(o.itemTypeAttribute);
 		_t.inputHint = _t.parentNode.getAttribute(o.inputHintAttribute);
-		_t.itemState = _t.parentNode.getAttribute(o.itemStateAttribute);
-		_t.unit = _t.parentNode.getAttribute(o.unitAttribute);
+		_t.itemState = smarthome.UI.unEscapeHtml(_t.parentNode.getAttribute(o.itemStateAttribute));
+		_t.unit = smarthome.UI.unEscapeHtml(_t.parentNode.getAttribute(o.unitAttribute));
 		_t.prefixField = _t.parentNode.parentNode.querySelector(".mdl-form__input-prefix");
 		_t.postfixField = _t.parentNode.parentNode.querySelector(".mdl-form__input-postfix");
+		_t.prefix = _t.prefixField ? smarthome.UI.unEscapeHtml(_t.prefixField.innerHTML) : null;
+		_t.postfix = _t.prefixField ? smarthome.UI.unEscapeHtml(_t.postfixField.innerHTML) : null;
```
lolodomo, line 2321: Shout be _t.postfixField as condition

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -2372,10 +2374,8 @@
 		}
 
 		function cleanValue(value) {
-			var prefix = _t.prefixField !== null ? _t.prefixField.innerHTML : "";
-			var postfix = _t.postfixField !== null ? _t.postfixField.innerHTML : "";
-			var newValue = value.startsWith(prefix) ? value.substr(prefix.length) : value;
-			newValue = value.endsWith(postfix) ? newValue.substr(0, newValue.lastIndexOf(postfix)) : newValue;
+			var newValue = (_t.prefix && value.startsWith(_t.prefix)) ? value.substr(_t.prefix.length) : value;
+			newValue = (_t.postfix && value.endsWith(_t.postfix)) ? newValue.substr(0, newValue.lastIndexOf(_t.postfix)) : newValue;
```
lolodomo, line 2378: Is condition `_t.prefix` equivalent to `_t.prefix !== null` ? I am just not 100% sure.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -2772,6 +2772,12 @@
 			return escapedText;
 		};
 
+		_t.unEscapeHtml = function(html) {
+			var text = document.createElement("textarea");
+			text.innerHTML = html;
+			return text.value;
```
lolodomo, line 2778: Is it a clean way to do ?
Doesn't this new element appear in the page after creation ?

lolodomo (changes requested):

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: No, it doesn’t. This is the most commended way I could find.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: null, undefined or empty will all produce false, which is OK. This avoids issues with the following calls on it.

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: Thanks, fixed.

bundles/org.openhab.ui.basic/web-src/smarthome.js
lolodomo: In the linked URL, someone proposed to remove the created element. So maybe something like that is even better ?
```
var text = document.createElement("textarea"),
    decodedHtml;;
text.innerHTML = html;
decodedHtml = text.value;
text.remove();
return decodedHtml;
```

mherwege: I also read that it is garbage collected anyway as no reference is held to it after returning from the function, and that remove is not available in all versions of js. The last point is probably not an issue anymore with versions currently used in browsers, but I don’t see why risking it.

PR #2687, opened 2024-07-26
Add a dirty indicator on the title bar

jimtng:
image


florian-h05 (approved):
LGTM, thanks for this nice addition!

FYI you forgot chart and tab edit, I have added the indicator there as well.

florian-h05:
I think this should be added to the docs, but I don't know where its best to put it.
@rkoshak Do you have an idea?

rkoshak:
Does this only apply to rules or things, items, and  transformations too?

If only rules I think the best place is under MainUI -> Settings -> Rules.

If all of these, a sentence probably needs to be added to each of these pages under MainUI -> Settings or a new section added to MainUI -> Settings to explain it.

florian-h05:
It applies to all of these. 
I guess let’s then put it into Main UI -> Settings.

florian-h05:
See https://github.com/openhab/openhab-docs/pull/2345 for documentation.

PR #2697, opened 2024-08-07
[BasicUI] Make button text not uppercased

jimtng:
The buttons on iOS and Android apps don't uppercase the text, so this PR makes BasicUI behave the same way.

Related discussion:
https://github.com/openhab/openhab-webui/pull/2388#issuecomment-2270883776
https://github.com/openhab/openhab-webui/pull/2388#issuecomment-2272304430

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -278,6 +278,7 @@
 		.mdl-button:focus {
 			box-shadow: none;
 			-webkit-box-shadow: none;
+			text-transform: unset;
 		}
 		.mdl-button-text {
 			html.ui-bigger-font & {


lolodomo:
I am fine with this change.

@kaikreuzer : no veto on your side ? 

@jimtng : please check that your change also applies to buttons inside a buttongrid and buttons in preferences page (theme).

kaikreuzer:
https://github.com/openhab/openhab-webui/pull/2388#issuecomment-2272304430 suggests to capitalize it when the Basic UI option "Capitalize Values" is on. I think this should at least be implemented here to keep at least a partial backward compatibility. Wdyt?

lolodomo:
I agree that would better.
Each time we change something, you will find someone complaining about it, so having an option for backward compatibility is even better.

jimtng:
> [#2388 (comment)](https://github.com/openhab/openhab-webui/pull/2388#issuecomment-2272304430) suggests to capitalize it when the Basic UI option "Capitalize Values" is on. I think this should at least be implemented here to keep at least a partial backward compatibility. Wdyt?

It happens to already be the case. When the UI option "Capitalize Values" is on, both the value and the buttons are capitalized, and when it's off, both are not.



> please check that your change also applies to buttons inside a buttongrid and buttons in preferences page (theme).

This doesn't affect Buttongrids. The buttons in a buttongrid was already not uppercased, and the "Capitalize Values" setting doesn't affect it, even prior to this PR.

The Theme in preferences page is a bit different. For some reason it doesn't want to be capitalized. Would you like me to spend more time (which may also involve additional code) to try to figure it out?

lolodomo:
> This doesn't affect Buttongrids. The buttons in a buttongrid was already not uppercased, and the "Capitalize Values" setting doesn't affect it, even prior to this PR.

Ok

> The Theme in preferences page is a bit different. For some reason it doesn't want to be capitalized. Would you like me to spend more time (which may also involve additional code) to try to figure it out?

No, forget it.

> It happens to already be the case. When the UI option "Capitalize Values" is on, both the value and the buttons are capitalized, and when it's off, both are not.

Ok, fine, I will test before merging.

lolodomo (approved):
LGTM, thank you.

Tested and text in buttons are now in uppercase only when the setting to show values in uppercase is enabled.

PR #2699, opened 2024-08-09
[BasicUI] Remove reading of frequency parameter of colorpicker element

lolodomo:
Note that %frequency% does not exist in colorpicker snippet.

Related to openhab/openhab-core#4338


kaikreuzer (approved):
Thanks!

PR #2703, opened 2024-08-11
[BasicUI] Add build date to resource requests

jimtng:
The goal is so users don't need to do a force refresh whenever openhab is updated.


jimtng:
Ideally the version number is replaced at build time, but I don't know how to get Maven to perform the replacement for the `$project.version` inside the snippet files.

florian-h05:
Possibly you can pass the version number to the build command (grunt or whatever is used here) and have the string replaced by the JS build.

jimtng:
`${project.version}` -> `4.3.0-SNAPSHOT` whereas I'm looking for this type of "version number": `4.3.0.202408100705`

jimtng:
I've added a commit that moved the injection into the gulp processing stage. However, since I still can't figure out how to get the same kind of version / build number, I just used `YYYYMMDDHHMM` timestamp.

bundles/org.openhab.ui.basic/snippets-src/main_static.html
```diff
@@ -0,0 +1,65 @@
+<!doctype html>
```
lolodomo, line 1: Why isn't it detected by Git as a file moving, like for main.html ?

lolodomo:
What annoys me a little with that solution is that you have then snippets source at two different places, main and main_static in snippets-src and all others in resources/snippets. Really not intuitive for developer.

bundles/org.openhab.ui.basic/snippets-src/main_static.html
jimtng: I wonder about that too

jimtng:
> What annoys me a little with that solution is that you have then snippets source at two different places, main and main_static in snippets-src and all others in resources/snippets. Really not intuitive for developer.

I am aware of this and I agree. Got a better idea? We can move everything into snippets-src and either run the replacement on all files, or just special-case the two files whilst copying the rest.

lolodomo:
> We can move everything into snippets-src and either run the replacement on all files, or just special-case the two files whilst copying the rest.

Yes, I believe it would be better with my preference to the second option.

bundles/org.openhab.ui.basic/snippets-src/main_static.html
lolodomo: Are you sure you did not change file encoding for example ?

jimtng:
I've moved all the files in snippets into snippets-src

bundles/org.openhab.ui.basic/.gitignore
```diff
@@ -1,6 +1,8 @@
 node_modules
 node
 npm_cache
+src/main/resources/snippets/*
```
lolodomo, line 4: Maybe *.html ?

bundles/org.openhab.ui.basic/gulpfile.js
```diff
@@ -3,17 +3,21 @@
...
 		uglify = require("gulp-uglify"),
 		eslint = require("gulp-eslint");
 
 	var
 		sources = {
 			js: [ "web-src/smarthome.js", "web-src/static.js", "web-src/settings.js" ],
-			sass: "web-src/smarthome.scss"
+			sass: "web-src/smarthome.scss",
+			html: {
+				all: "snippets-src/**",
```
lolodomo, line 16: ** ?
Maybe *.html ?

lolodomo (changes requested):

bundles/org.openhab.ui.basic/snippets-src/main_static.html
jimtng: I reset the commits and cleaned it up that way. It now shows up as moved.

bundles/org.openhab.ui.basic/.gitignore
jimtng: Not sure if that's a good idea. That means in the future people can start adding non .html files directly there, and we'll back to the situation where there are mixed files (generated + original) in there.

bundles/org.openhab.ui.basic/gulpfile.js
jimtng: ditto here, we don't want to confuse someone in the far future adding non .html and wondering why it didn't copy across. This way it's more predictable.

What are you trying to avoid by being more specific?

jimtng: I've renamed `html` to `snippets`. Perhaps that helps?

bundles/org.openhab.ui.basic/.gitignore
lolodomo: If you have a line with *, why do you add a specific line for readme.txt ?

bundles/org.openhab.ui.basic/gulpfile.js
lolodomo: What I do not understand is why you put a doubled *.

jimtng: It will cover subdirectories in case they're added in the future, to avoid surprises.

bundles/org.openhab.ui.basic/.gitignore
jimtng: It means ignore everything in that directory (because they're generated), but do not ignore (i.e. check in) readme.txt

lolodomo: Ok. I did not take attention to the "!" at the beginning of the next line !

bundles/org.openhab.ui.basic/gulpfile.js
lolodomo: Thank you for the explanation

lolodomo (approved):
LGTM, thank you.
@kaikreuzer : can you please have a second look and merge if that is fine for you too

lolodomo:
We have 3 PRs updating package-lock.json so we have to take care, merge one by one and update the file between each merge.

kaikreuzer (approved):
lgtm, thanks!

PR #2711, opened 2024-08-13
Widget actions: Add a HTTP action to perform HTTP requests

florian-h05:
This allows to perform HTTP requests from inside widget actions.
The use case might be limited, but for example this can be used to allow access to local only ressources/smart devices that should not be made available via remote access and hence cannot be integrated as Items into openHAB for security purpose.

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-aggregate-series.md
+++ b/bundles/org.openhab.ui/doc/components/oh-aggregate-series.md
@@ -157,18 +157,35 @@ prev: /docs/ui/components/
     <PropOption value="photos" label="Open photo browser" />
     <PropOption value="group" label="Group details" />
     <PropOption value="analyzer" label="Analyze item(s)" />
-    <PropOption value="url" label="External URL" />
+    <PropOption value="url" label="Navigate to external URL" />
+    <PropOption value="http" label="Send HTTP request" />
     <PropOption value="variable" label="Set Variable" />
   </PropOptions>
 </PropBlock>
 <PropBlock type="TEXT" name="actionUrl" label="Action URL" context="url">
   <PropDescription>
-    URL to navigate to
+    URL to navigate to or to send HTTP request to
   </PropDescription>
 </PropBlock>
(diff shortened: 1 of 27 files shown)


ghys:
All good but just to add some context on why I resisted adding such a feature for years:

I considered all HTTP requests from the client with a Page action as less controlled and secure than having the openHAB server do them. If you get access to the Pages API then you can easily make a Page silently make malicious requests if you're not actively monitoring that your page definition hasn't been compromised.

It's also true that you could have the Rules API do the same.

Note that this could be mitigated by adding [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (Content Security Policy) headers or meta tags, and only offering more secure (e.g. configuration in a file, similar to the Exec binding whitelist) ways to alter them (i.e. you could have a configuration that only allows local requests but also `*.example.com`).

ghys:
See here for when I added a CSP to HABPanel (and I got some flak for this) - `unsafe-inline` is to be avoided but I don't know if it's necessary for the Main UI (probably not):

https://github.com/openhab/openhab-webui/blob/a13ffe311e235ea37e0cc5528832be4d458ec4b2/bundles/org.openhab.ui.habpanel/web/index.html#L13

florian-h05:
@ghys Thanks for your feedback.
WRT to CSP: I see that Main UI already has a CSP, do you think we can adjust it in a way that a user cannot load scripts from external servers?

florian-h05:
Okay we can use CSP to limit to which domains a request can be sent by this action.
Should we limit that and advise users that want to use this action to overwrite our CSP by setting the header on a reverse proxy?

florian-h05:
Okay this seems to be very restrictive and not super user-friendly ...

ghys:
I totally forgot there was already a CSP... 

I wonder if we could have a configurable whitelist of allowed domains that would alter the CSP - maybe by generating the index.html dynamically.
Ideally it would not be configurable remotely, like the whitelist of the exec binding.

ghys:
> We can set the CSP as a header

Ah yes indeed... so maybe via a derived servlet to serve that index page (and the rest, possibly) - or equivalent.

PR #2714, opened 2024-08-14
Main UI: Improve Content-Security-Policy

florian-h05:
This new CSP is much more restrictive compared to the old one:
- Don't allow any origin as default
- Don't allow eval() usage
- Don't allow data:, gap:, content: and blob: schemes by default

### DIFF
--- a/bundles/org.openhab.ui/web/src/index.html
+++ b/bundles/org.openhab.ui/web/src/index.html
@@ -21,7 +21,9 @@
     * allow loading media from any source, and data:, blob: and media: URIs
     * allow connecting (through fetch(), XMLHttpRequest, WebSocket etc.) to the same origin, raw.githubusercontent.com (add-on logos etc.), Iconify icon sources, and any source
   -->
+  <% if (process.env.NODE_ENV !== 'development') { %>
   <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src * data:; media-src * data: blob: media:; connect-src 'self' raw.githubusercontent.com api.iconify.design api.unisvg.com api.simplesvg.com *;">
+  <% } %>
   <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no, minimal-ui, viewport-fit=cover">
 
   <meta name="theme-color" content="#e64a19">
--- a/bundles/org.openhab.ui/web/src/index.html
+++ b/bundles/org.openhab.ui/web/src/index.html
@@ -3,15 +3,25 @@
 <head>
   <meta charset="utf-8">
   <!--
-  Customize this policy to fit your own app's needs. For more guidance, see:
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/index.html
```diff
@@ -3,15 +3,25 @@
...
   -->
-  <meta http-equiv="Content-Security-Policy" content="default-src * 'self' 'unsafe-inline' 'unsafe-eval' data: gap: content: blob:; style-src 'self' 'unsafe-inline';">
+  <!-- Explanation:
+    * allow loading resources from the same origin, inline scripts and styles
+    * allow loading fonts from the same origin, and data: URIs
+    * allow loading images from any source, and data: URIs
+    * allow loading media from any source, and data:, blob: and media: URIs
+    * allow connecting (through fetch(), XMLHttpRequest, WebSocket etc.) to the same origin, raw.githubusercontent.com (add-on logos etc.), and any source
+  -->
+  <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src * data:; media-src * data: blob: media:; connect-src 'self' raw.githubusercontent.com *;">
```
florian-h05, line 24: By removing this wildcard here we could limit what the HTTP action is able to do:

```suggestion
  <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src * data:; media-src * data: blob: media:; connect-src 'self' raw.githubusercontent.com;">
```

However we have to keep in mind that this limits the SIP widget and probably the video widget as well.
It is possible to overwrite our CSP here by using a reverse proxy, however is this quite user-**un**friendly.

@ghys WDYT?

florian-h05:
@ghys I have tested this with my prod system by setting the header through my nginx, everything that is expected to work works fine. Can you please have a look?

bundles/org.openhab.ui/web/src/index.html
ghys: > that this limits the SIP widget and probably the video widget as well

That's kind of bad then, isn't it?

florian-h05: Yeah but as long as you have a reverse proxy in place this is easy to fix.

Either server everything from „self“, e.g. I have openhab.localnet for openHAB and my SIP WebSocket is at openhab.localnet/sip,
or overwrite the CSP by setting the header in the reverse proxy conf and add additional targets as needed.

ghys (approved):
> I have tested this with my prod system by setting the header through my nginx, everything that is expected to work works fine. Can you please have a look?

This would probably just confuse users unless properly documented ("you can only send HTTP requests to the server itself"), and probably be frustrating as well. I guess we'll just have to hope there won't be abuse. For marketplace widgets at least it's rather easy to delete them, at the minimum, if we discover they make malicious HTTP requests - this is explicitly banned in the [Marketplace Rules](https://community.openhab.org/t/about-the-add-on-marketplace-category/123408) anyways.
For changes targeting a specific instance/configuration however there's no good answer but then there are other ways (for example, with rules/rule templates).

Thanks!

PR #2716, opened 2024-08-15
Main UI: Allow use of eval() in CSP

florian-h05:
Regression from #2714.

eval() seems to be required by ECharts: When opening a chart, an CSP error is thrown that traces back to ECharts, but the chart still renders.

### DIFF
--- a/bundles/org.openhab.ui/web/src/index.html
+++ b/bundles/org.openhab.ui/web/src/index.html
@@ -15,14 +15,14 @@
         * Enable eval(): add 'unsafe-eval' to default-src
   -->
   <!-- Explanation:
-    * allow loading resources from the same origin, inline scripts and styles
+    * allow loading resources from the same origin, inline scripts (required for inline event listeners) and styles, and the use of eval() (required by ECharts)
     * allow loading fonts from the same origin, and data: URIs
     * allow loading images from any source, and data: URIs
     * allow loading media from any source, and data:, blob: and media: URIs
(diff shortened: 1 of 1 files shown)


florian-h05:
@ghys FYI

bundles/org.openhab.ui/web/src/index.html
```diff
@@ -15,14 +15,14 @@
...
-    * allow loading resources from the same origin, inline scripts and styles
+    * allow loading resources from the same origin, inline scripts (required for inline event listeners) and styles, and the use of eval() (required by ECharts)
     * allow loading fonts from the same origin, and data: URIs
     * allow loading images from any source, and data: URIs
     * allow loading media from any source, and data:, blob: and media: URIs
     * allow connecting (through fetch(), XMLHttpRequest, WebSocket etc.) to the same origin, raw.githubusercontent.com (add-on logos etc.), Iconify icon sources, and any source
   -->
-  <% if (process.env.NODE_ENV !== 'development') { %>
-  <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src * data:; media-src * data: blob: media:; connect-src 'self' raw.githubusercontent.com api.iconify.design api.unisvg.com api.simplesvg.com *;">
+  <% if (process.env.NODE_ENV === 'production') { %>
```
florian-h05, line 24: This is only cosmetic.

ghys:
something like this was to be expected - but fine ;)

thanks as always for the insight and stewardship!

PR #2719, opened 2024-08-19
[cometvisu] fix path check regressions

peuter:
1. Fix serving the index.html when "/" (or empty) path is requested
2. Fix allowing files beeing served from special cometvisu config folder


kaikreuzer (approved):
Thanks!

kaikreuzer:
@peuter Is this something critical that should be backported to 4.2.x or is it fine to only fix it on main?

peuter:
A Backport to 4.2 would be great. For one of the problems this PR fixes we do not have a workaround and it breaks a quite common and useful use case.

kaikreuzer:
Done!

PR #2729, opened 2024-08-24
Fixes logic to detect when a thing's config changes vs the whole thing structure

### LINKED ISSUE #2704: Thing handleConfigurationUpdate issue
Hi I'm working on a binding that i was hoping to take advantage of the `handleConfigurationUpdate` method to be able intercept configuration updates.   

However, in my testing, when i click "save" on a thing in the UI, the order of operations is `dispose() -> initialize() -> handleConfigurationUpdate()` .   I also tested this against the z-wave binding, which has similar functionality about intercepting updates, and its also behaving the same.

My understanding is that if this method is overridden, then any updates to the config should not trigger a `dispose() -> initialize()` on a thing and instead the `handleConfigurationUpdate()` override would be responsible for that if needed.  Looking at the core code, it seems `handleConfigurationUpdate()` handles the thing lifecycle after an update  if not overridden. 

Am i misunderstanding how this is suppose to work ? 

I'm on OH 4.3 nightlies right now, but i'm pretty sure this behavior is also in the 4.2  release.

digitaldan:
Fixes #2704

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -353,12 +353,12 @@ export default {
           let savedThingClone = cloneDeep(this.savedThing)
 
           // check if the configuration has changed between the thing and the original/saved version
-          delete thingClone.statusInfo
-          delete savedThingClone.statusInfo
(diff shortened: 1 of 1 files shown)


ghys (approved):
I'm sure you tested this ;)
So LGTM!
(until something else occurs in this area)

PR #2731, opened 2024-08-26
Developer sidebar: Focus on search when opened

### LINKED ISSUE #2727: Main UI developer sidebar -> automatic focus
## The problem

No problem - see below for suggestion

## Your suggestion

Suggest that the Alt-Shift-D to bring up the developer sidebar automatically changes focus to the search box in the sidebar.  Very frequently when I start the sidebar, I want to search.  This would make using the Main UI easier.  Others have expressed support for this - see: https://community.openhab.org/t/main-ui-developer-sidebar-automatic-focus/157951

## Your environment

Google Chrome: Version 127.0.6533.120
OH 4.2.1

```yaml

```

jimtng:
Resolve #2727

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
+++ b/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
@@ -1,10 +1,10 @@
 <template>
   <f7-block class="developer-sidebar">
     <f7-row :inner="false" v-if="!$theme.md">
-      <f7-searchbar style="width: 100%" custom-search placeholder="Search and Pin" :backdrop="false" @searchbar:search="search" @searchbar:clear="clearSearch" />
+      <f7-searchbar ref="searchbar" style="width: 100%" custom-search placeholder="Search and Pin" :backdrop="false" @searchbar:search="search" @searchbar:clear="clearSearch" />
     </f7-row>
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
```diff
@@ -450,6 +450,9 @@ export default {
   },
   mounted () {
     this.startEventSource()
+    setTimeout(() => {
```
florian-h05, line 453: This is usually only done for desktop, because on mobile this would bring up the on screen keyboard which would be very annoying.
And you could use nextTick instead of a timeout.

See https://github.com/openhab/openhab-webui/blob/77dbfcb198d20053b55373f32255f2a50eaa2ff7/bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-binding.vue#L85-L89

florian-h05 (changes requested):

bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
jimtng: Thanks for the tip!

florian-h05 (approved):
LGTM, thanks!

PR #2742, opened 2024-09-06
[blockly] Small improvements to blocks

mherwege:
When going through the available blocks, I saw potential for a few small improvements, which I collected in this PR:

- Persistence `previous...State` and `next...State` blocks: add boolean shadow block for skip same, default false
- Named `timer` blocks: add a `undefined` shadow block for `context`.
- `run rule or script` block: make the parameter dictionnary input a shadow block

The above does not change functionality, but I find it cleaner to always a default attached to the input, even if it is not required for the code. It avoids users thinking they need to put something in. Having a default indicates the default will work unless they need something specific.

In addition to that, for the `transform` block, I made the transformation dropdown dependent on the installed transformations.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
@@ -139,6 +139,10 @@ export default function defineOHBlocks_Persistence (f7, isGraalJs, persistenceSe
             .appendField('skip same ')
             .setAlign(Blockly.ALIGN_RIGHT)
             .setCheck(['Boolean'])
+          this.getInput('skipPrevious').setShadowDom(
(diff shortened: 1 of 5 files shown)


florian-h05 (approved):
LGTM, thanks!

Just one thing: The code of `transformationOptions()` can be simplified, I will push a commit for that soon.

stefan-hoehn:
@mherwege Has this been updated in the docs?

florian-h05:
I don‘t think these minor changes (basically add shadow blocks where they were missing) needs a doc update. Do you agree?

mherwege:
@stefan-hoehn @florian-h05 I agree I don't think this needs further documentation. The purpose of this was to make things clearer when building Blockly rules, so it does not need al the documentation.

stefan-hoehn:
Agreed. It would be just nice if the pics would be updated but I myself refrain from doing so often as it is not worth the effort.

PR #2752, opened 2024-09-12
State description metadata: Allow to select transformations from list

### LINKED ISSUE #1944: [MainUI] Select transformations from list
## The problem

Even when a user does everything through MainUI they have to remember and understand a text based syntax in order to select and apply a transformation to a Channel or a Profile. 

## Your suggestion

I know this is a huge job and may very well not be feasible without significant changes to core.

But now that we have a transformations API endpoint, it occurs to me we theoretically could use a selection dialog instead of needing to remember the syntax. 

Stuff to consider:

- Users should still be able to define a transformation inline
- Transformation chaining
- Transformation arguments (e.g. `JS(<scriptname>.js?arg=value)`)
- What happens when a transformation is removed?
- API needed to be implemented by developers of new transformations to handle any unique stuff about their transform and this selection dialog

## Your environment

```yaml
runtimeInfo:
  version: 4.0.0
  buildString: "Build #3512"
locale: en-US
systemInfo:
  configFolder: /openhab/conf
  javaVersion: 17.0.7
  javaVendor: Eclipse Adoptium
  javaVendorVersion: Temurin-17.0.7+7
  osName: Linux
  osVersion: 5.15.0-75-generic
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 517044976
  totalMemory: 1174405120
  startLevel: 100
bindings: null
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: true
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: macos
    pixelRatio: 1.600000023841858
    prefersColorScheme: light
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 12
    language: en-US
    languages:
      - en-US
      - en
    onLine: true
    platform: MacIntel
  screen:
    width: 1920
    height: 1080
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: light
    filled: true
    bars: filled
    homeNavbar: default
    homeBackground: default
  userAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36
    (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36
timestamp: 2023-06-28T14:39:21.041Z
```

florian-h05:
Resolves #1944.

When entering a state description pattern, an autocomplete list is provided containing the available transformations:
image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-itemdescription.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-itemdescription.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div v-if="ready">
     <config-sheet v-if="namespace === 'stateDescription'" :parameterGroups="[]" :parameters="stateDescriptionParameters" :configuration="metadata.config" />
     <f7-list>
       <f7-list-input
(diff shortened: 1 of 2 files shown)


florian-h05:
/cc @rkoshak WDYT?
Do you think this looks good, or should we adjust what's shown to the user? Note that we cannot hide the actual pattern that gets inserted into the field from the list, so the entries only get longer if we change something here.

rkoshak:
Is the "Label" of managed transfromations avaialble to show here? Given how distressingly common it is for people to just accept the randomly generated ID having access to the label could be very helpful. 

It's not clear if any of the examples you are showing are managed transformations.

florian-h05:
> Is the "Label" of managed transfromations avaialble to show here?

In the second example, the label is shown first and then the pattern that is to be inserted.
In my example, invert_pct.js is a managed transformation, ypu can recognize them by having the `config:`  part in the uid.

> Given how distressingly common it is for people to just accept the randomly generated ID having access to the label could be very helpful.

True, then I will make it like in my second screenshot and display the label.

florian-h05:
@rkoshak Done. It now shows the transformation label and then in brackets the pattern that gets inserted, the entries are sorted alphabetically by the transformation label.
I think I can merge this PR then.

rkoshak:
> ypu can recognize them by having the config: part in the uid.

Ah yes, now I see.

PR #2757, opened 2024-09-15
Proposed to add expanding group items in graph

jsjames:
The feature to plot all items in a group on a graph is not implemented and missing from the sitemap graph feature.  I have prototyped a solution in this PR - but before I spend more time to perfect/test, i wanted to get feedback on the implementation to ensure the overall plan is okay and get any suggestions:

- I have added a boolean config parameter call "groupItems" which can be set to true to expand the group.  If this field does not exist or is set to false, things operate as normal - with similar performance.
- When this field is set to true, it will wait for the first itemPromises to finish and then expand the neededItems to query the persistence (this wait only happens if the groupItems is set to true - so performance of original graphs should not be affected)
- Finally, I've modified the oh-time-series get to return multiple series items, one for each of the items in the group.  This works well in the line graphs I have tested.

Note, I have not tested other graph types to make sure there is no impact as I want to first get feedback/input on the proposed implementation.


jimtng:
got a screenshot?

jsjames:
> got a screenshot?

Here's a screenshot - this one still has the item gEnergy on the chart which is the group all of the other items belongs to.  The latest code removes the group item.

image

florian-h05:
Hi @jsjames,

nice feature proposal, you definitely have my support for this!
Wrt to the implementation outlined in the PR description, I think the overall plan is okay 👍 
Just one minor thing: I would rather name that new parameter `expandGroup`.

Let me know when you need help or this PR is ready for review!

florian-h05:
@jsjames What's the status here?

jsjames:
Hi Florian - Still on my list as this would be a good feature.  Just haven’t had a lot of time as I’m working on a couple bindings/etc and was helping get the python scripting up and running with Holger.

PR #2771, opened 2024-09-30
Thing add page: Add support for optional scan input

florian-h05:
Refs https://github.com/openhab/openhab-core/pull/4389.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-thing-type.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-thing-type.vue
@@ -22,7 +22,11 @@
             </f7-button>
           </div>
         </div>
-        <p class="margin-left margin-right" style="height: 30px" id="scan-progress" />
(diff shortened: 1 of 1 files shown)


lolodomo:
Tested and it works exactly as we expected. Top. @digitaldan you can now rely on this new feature and remove pairing code as thing configuration and as thing channel.

@florian-h05 I just saw a little bug in rendering with a text almost hidden:
image

florian-h05:
@lolodomo Thanks for noticing … this is cause by a „quick-and-dirty“ style fix I used, I should probably do it the clean way now.

lolodomo:
@florian-h05 : final step would be to also propose to scan a QR code. Is there something like that in "HTML"?

florian-h05:
I have already played around with QR code scanners in the browser in a little side project … so it is possible. These pairing code QR codes only contain the pairing code itself or additional data?

florian-h05:
I have already played around with QR code scanners in the browser in a little side project … so it is possible. These pairing code QR codes only contain the pairing code itself or additional data?

PR #2774, opened 2024-09-30
Z-Wave network map: Fix ECharts deprecation warnings

florian-h05:


florian-h05:
@stefan-hoehn IIRC you have Z-Wave devices. Can you please check if the network map still works?

stefan-hoehn (approved):
Yes, works like a charm with this PR:

image

PR #2796, opened 2024-10-03
Add version and commit to UI during production build

florian-h05:
This sets the version in the Framework7 parameters to the build version during production build, 
and adds the commit hash to the Vuex store so it can be displayed on the about page.


ghys:
This is what I wanted to do for a long time, I'm glad you finally did it!
:muscle:

florian-h05:
It was super annoying to always have to reload the PWA after an update, so I finally decided to tackle this 👍

PR #2800, opened 2024-10-06
Home page: Add options to set simple navbar/standard background the default on mobile/desktop

florian-h05:
This adds new boolean config parameters to the home page settings that allow setting the default navbar style and standard background for mobile and desktop devices.
The device-specific settings allow to override that default by using a segmented button with three options: The admin's default, option A and option B.

image

The segmented buttons are fully localized.


bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
```diff
@@ -7,7 +7,10 @@
   "about.darkMode.dark": "Dark",
   "about.navigationBarsStyle": "Navigation bars style",
   "about.miscellaneous": "Miscellaneous",
-  "about.miscellaneous.home.navbar": "Simple navigation bar on home page",
```
florian-h05, line 10: @ghys Do you know how I can remove the existing translations for this setting? 
I need to get rid of them because there is a new text. Does CrowdIn do that automatically or do I have to remove this string from CrowdIn and from the i18n assets?

florian-h05:
@ghys Please have a look at this PR.

florian-h05:
@ghys WDYT?

ghys (approved):
Sorry for stalling again :disappointed:  and thanks for all the work.

I think it's a great idea to use the home page component config to store default values for all devices (optionally) and still have local configuration for individual devices to override them or not.

In fact it could be applied to the other option pertaining to the home page - the choice between a white or light gray background color while in 'light' mode.
(when you have a simple navbar in light mode to match the non-home pages I think you also want your home page background color to match the other non-home pages, i.e. light gray by default.)

And I'm just spitballing here but it could be a principle applied to even more Main UI-related options not pertaining to the home page (then we'll have to choose where to store them because storing "page transitions or not" in the home page component isn't right):

I mean:
_"local settings and defaults set for you, unless there's global settings that are set, which will set the new default, but you can still override them locally"_

As for this PR, well I don't like the segmented buttons too much - in general, not in this particular case (in translations you might have long strings and it wouldn't quite work anymore). But I don't have better proposals so it's all good to me :+1:

bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
ghys: I have no idea but I can check.

florian-h05:
> In fact it could be applied to the other option pertaining to the home page - the choice between a white or light gray background color while in 'light' mode.

Good point, I have added that in my last commit.

> And I'm just spitballing here but it could be a principle applied to even more Main UI-related options not pertaining to the home page (then we'll have to choose where to store them because storing "page transitions or not" in the home page component isn't right):

In that case we could (mis)use the uicomponents endpoint to use a UI component to store that config ...

I agree with the segmented buttons, it is always a bit risky that translations get too long, but you said it, that's the best option here.

bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
florian-h05: Thanks! You can merge as soon as you want to - I will be away from keyboard next week.

PR #2832, opened 2024-10-25
Number parameter: Optionally allow any step size/number of decimals

florian-h05:


lolodomo:
@florian-h05 : a feedback from my tests.

I noticed that for parameters of type INTEGER, it is allowed to enter a decimal value.
No way to block the filling of the decimal character ?

PR #2833, opened 2024-10-25
Addons store: Multiple styling improvements

florian-h05:
Follow-up for #2216.

- Do not use addon swiper on iPad.
- Revert addon card size changes.
- Show more addon cards in swiper in landscape mode.


florian-h05:
Todo: Show more slider elements in landscape mode.

PR #2818, opened 2024-10-21
Thing page: Support invoking Thing actions & Viewing their output

florian-h05:
Refs https://github.com/openhab/openhab-core/pull/4392. 
Closes #2817.

This adds a new section "Actions" to the Thing tab of the Thing page, which provides a button for each UI-supported Thing action. 
Clicking on that button will open a popup, where action input can be configured and action output can be viewed.

All keys of the action output response object from REST are rendered as list Items, the labels are taken from the action output definitions and fallback to the key.
If the key is `qrCode` or its output type is defined as `qrCode`, its value is rendered as QR code. 

For actions without inputs or without outputs, messages are shown indicating that there is no such.

image

This PR marks the cold related to the old config actions as deprecated, my wish is to get rid of those config actions and have Thing actions used instead.

PR #2834, opened 2024-10-25
Thing actions: Sort by label, Show description, Hide unsupported actions

florian-h05:
Follow-up for #2818.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-action-popup.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-action-popup.vue
@@ -15,6 +15,9 @@
         </f7-nav-right>
       </f7-navbar>
       <f7-block class="no-margin no-padding">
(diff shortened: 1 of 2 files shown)


lolodomo:
@florian-h05 : I started to test the new snapshot and check what actions are shown on my different things. I discovered a very minor "issue": For my hue things, I see the title "Actions" but there is no action displayed. I will check the hue actions to confirm that they are not supported bu UI but I believe you should not show the "Actions" if there is no supported thing actions. WDYT ?

image

florian-h05:
True.
I already check the length of the actions array, but I need to filter it for UI support first before doing that. Will create a PR for that.

lolodomo:
Regarding the hue action, I can confirm it is not supported by UI because one of its input parameters is of type org.openhab.core.types.Command. What a strange choice...

florian-h05:
My plan is that once war are finished with all core and UI changes around Thing actions that I write to the docs how to design a action for „good“ UI support.

PR #2838, opened 2024-10-26
Config parameter: Implement `date` and `datetime` contexts

### LINKED ISSUE #2816: Add missing `datetime` and `date` config params
Main UI is missing config parameter support for the `datetime` and the `date` contexts, which are both defined by openHAB Core, see https://github.com/openhab/openhab-core/blob/a22349abf4b3106ec8a3eb9d36799e334cdfbf25/bundles/org.openhab.core.config.core/src/main/java/org/openhab/core/config/core/ConfigDescriptionParameter.java#L304.

There is need to implement those for https://github.com/openhab/openhab-core/pull/4392.

florian-h05:
Closes #2816.
Depends on https://github.com/openhab/openhab-core/pull/4428.


lolodomo:
@florian-h05 : a feedback from my tests.

For datetime and time, I see that we can enter only hours and minutes.

As a reminder, we decided to allow hours, minutes and seconds.

For type java.util/Date, it even leads to an error:
```
Input parameter 'dateParam': converting value '2024-10-31T08:07' into type java.util.Date failed! Input parameter is ignored
```

PR #2840, opened 2024-10-27
Thing action popup: Multiple improvements

florian-h05:
- Default to action input name if no label provided
- Use label for action output "result" if provided
- Display action output descriptions if provided


lolodomo:
Descr QuantityType<Temperature> parameter

In the popup, for a parameter that have `QuantityType<Temperature> parameter` as label and `Descr QuantityType<Temperature> parameter` as description, the label is properly displayed but the description is truncated to `Descr QuantityType parameter`.

image

PR #2851, opened 2024-11-01
[BasicUI] Implement new sitemap element Colortemperaturepicker

lolodomo:
Related to openhab/openhab-core#3891

### DIFF
--- a/bundles/org.openhab.ui.basic/snippets-src/colortemppicker.html
+++ b/bundles/org.openhab.ui.basic/snippets-src/colortemppicker.html
@@ -1,0 +1,37 @@
+<div class="mdl-form__row mdl-cell mdl-cell--%cells%-col mdl-cell--%cells_tablet%-col-tablet %visibility_class%">
+	<span class="mdl-form__icon">
+		%icon_snippet%
+	</span>
+	<span class="mdl-form__label">
+		%label%
+	</span>
+	<span class="mdl-form__value  mdl-form__value--colortemppicker">
+		%value%
+	</span>
+	<div
+		class="mdl-form__control mdl-form__colortemppicker"
+		data-control-type="colortemppicker"
+		data-item="%item%"
+		data-has-value="%has_value%"
+		data-state="%currentKelvin%"
+		data-color="%currentRGB%"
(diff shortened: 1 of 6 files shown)


lolodomo:
Here is the rendering of the new widget:

image

When clicking on the button, the following pop-up is displayed:

image

mueller-ma:
I'd prefer to have a preview of the state as actual color and as Kelvin, as I (and probably many other people) don't know how 4500 K look like. Suggestion:

382481287-49ab7a5d-ed81-4611-9901-c056b156c15e

lolodomo:
> I'd prefer to have a preview of the state as actual color and as Kelvin, as I (and probably many other people) don't know how 4500 K look like. Suggestion:

Very good idea, I will do that.
I will also keep the value.

lolodomo:
Conversion from Kelvin to RGB is not done in frontend. Backend is providing 21 RGB values to frontend covering the min/max range.
When the page is first opened, I can set the proper color as preview but when item state is updated, I can only use one of the 21 colours, the most approaching.
Of course, I can also increase this number if values.

The idea of this "preview" color could also be applied to the Colorpicker widget.

andrewfg:
> The idea of this "preview" color could also be applied to the Colorpicker widget.

Good idea!

andrewfg:
> when item state is updated, I can only use one of the 21 colours, the most approaching.

I guess there are two sub- cases..

1. when the state is updated from the picker itself
2. when the state is updated by notification from OH core

lolodomo:
When the color is updated from the picker itself, a command is sent to the item. The widget is updated when it receives a feedback state from the item.

lolodomo:
This is now ready for review.

Here is the rendering of the widget:

image

kaikreuzer (approved):

mueller-ma:
@lolodomo The "ok" button in the popup suggests that the value is updated when the button is pressed. But it seems that it's updated as soon as the slider is changed. Maybe rename the button to "close"?

lolodomo:
Yes, same behaviour as for the color picker. I agree that "Close" would be more appropriate in both pop-ups.

PR #2860, opened 2024-11-04
oh-colorpicker: Fix color picker sends commands on external state change

### LINKED ISSUE #1268: [mainui] [colorpicker] don't send commands on external state changes
OH 3.2 M4  MainUI

This is essentially a repeat of [https://github.com/openhab/openhab-webui/issues/789]#789 which was for knob widget, but here for color picker widget.

When the Item state updates and causes a change in widget display, unwanted commands are generated as though by user action.
This is a particular problem for slow-ramping devices issuing interim updates - the interim state update causes a new command to be issued to match the interim state, and the original command is trampled over.

Detail discussion in community here
https://community.openhab.org/t/colorpicker-with-zen31-controller-bounces-around/131390

Most essential data, events.log following single user click on picker
```
2022-01-08 18:59:50.352 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'HutchLEDStrip_ColorControl' received command 110,69,32
2022-01-08 18:59:51.352 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'HutchLEDStrip_ColorControl' changed from 248,69,32 to 244,55,27
2022-01-08 18:59:55.523 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'HutchLEDStrip_ColorControl' changed from 244,55,27 to 244,55,32
2022-01-08 18:59:55.538 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'HutchLEDStrip_ColorControl' received command 244,55,27
2022-01-08 18:59:57.835 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'HutchLEDStrip_ColorControl' changed from 244,55,32 to 244,55,26
2022-01-08 19:00:00.709 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'HutchLEDStrip_ColorControl' changed from 244,55,26 to 244,55,27
2022-01-08 19:00:01.755 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'HutchLEDStrip_ColorControl' changed from 244,55,27 to 244,55,26

```
Note especially second Item command, values matching earlier state change.
(I presume we don't get a command on _every_ change due to widget throttling.)

### LINKED ISSUE #2849: Color Temperature / Color widget wrongly sends second POST for Color item
@toomtech many thanks for the pcap. It shows something rather odd. See screenshot below. I am pretty sure that this is an error in Main UI. And to be specific the problem is NOT in the OH standard items slider screen, but IS in your Locations | Office widget. 

Your pcap shows the following sequence of events:

1. UI client commands CT item to 67%
2. OH server notifies CT item changed to 66.86%
3. OH server notifies Color item changed to HSB 38.101,56.52900,98.02 // note: S=**56.52900** !!
4. UI client commands Color item to HSB 39,39,98 // note: S=**39** !!
5. OH server notifies CT item changed to Undef
6. OH server notifies Color item changed to HSB 39.016,39.00500,98.02

Specific observation is that the S value massively changes between events 3. and 4. !! => It makes me suspect that your Client | Locations | Office widget is wrongly sending `HHB` instead of `HSB`. _=> We need to look into the widget code!_

image

_Originally posted by @andrewfg in https://github.com/openhab/openhab-addons/issues/17643#issuecomment-2448092063_

---

After the user issues a Color Temperature command, and OH sends respective Color Temperature and Color HSB state update responses, the UI Widget is issuing -- on its own initiative -- an extra Color HSB command. There are two problems here..

1. The Widget seems to be sending HHB rather than HSB
2. And -- far more important -- the Widget should not be issuing ANY commands on its own initiative.

florian-h05:
Fixes #1268.
Fixes #2849.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-colorpicker.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-colorpicker.vue
@@ -23,11 +23,10 @@ export default {
   data () {
     return {
       colorPicker: null,
+      ignoreInput: false,
       delayCommand: false,
-      delayUpdate: false,
       pendingCommand: null,
       lastCommand: null,
-      pendingUpdate: null,
       init: false
     }
   },
@@ -47,8 +46,8 @@ export default {
       if (state && state.split(',').length === 3) {
         let color = this.context.store[this.config.item].state.split(',')
         color[0] = parseInt(color[0])
-        color[1] = color[1] / 100
-        color[2] = color[2] / 100
+        color[1] = Math.round(color[1]) / 100
+        color[2] = Math.round(color[2]) / 100
         return color
       }
       return [0, 0, 0]
@@ -83,6 +82,8 @@ export default {
               return
             }
             if (!value.hsb) return
+            // Ignore input for a few millis after a new state has been received to prevent sending a command on external state change
+            if (vm.ignoreInput) return
             vm.sendCommand(value.hsb)
           }
         }
@@ -92,14 +93,13 @@ export default {
       console.debug('oh-colorpicker: Received command ' + hsb)
       const cmd = this.commandFromHSB(hsb)
       const state = this.commandFromHSB(this.color)
-      this.pendingCommand = [...hsb]
-      if (!this.areHSBEqual(this.roundedHSB(state), this.roundedHSB(this.context.store[this.config.item].state))) {
+      if (cmd !== state) {
+        this.pendingCommand = [...hsb]
         if (!this.delayCommand) {
           this.delayCommand = true
-          this.delayUpdate = true
-          console.debug(this.context.store[this.config.item].state + ' -> ' + state)
-          this.$store.dispatch('sendCommand', { itemName: this.config.item, cmd: state, updateState: true })
-          this.lastCommand = state
+          console.debug(state + ' -> ' + cmd)
+          this.$store.dispatch('sendCommand', { itemName: this.config.item, cmd, updateState: true })
+          this.lastCommand = cmd
           setTimeout(() => {
             const pendingCommand = [...this.pendingCommand]
             this.pendingCommand = null
@@ -109,10 +109,6 @@ export default {
             }
           }, 200)
         }
-        setTimeout(() => {
-          this.updateValue(this.pendingUpdate)
-          this.delayUpdate = false
-        }, 2000)
       }
     },
     commandFromHSB (hsb) {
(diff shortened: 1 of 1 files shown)


digitaldan:
this bug has caused me many so many headaches when testing Color items......thank you !

florian-h05:
@digitaldan You’re welcome — Can you please test the fix and also check if you notice any regressions?

andrewfg:
@florian-h05 many thanks for this. I see that your current code in this PR is focussed on supressing the second POST during a certain time window. This is certainly a good step forward. However in the case of #2849 there is a further refinement which I hope you can add to this PR. It concerns the case where there is both a color temperature picker and a color picker on the same UI widget. 

**Background**: Color temperatures are a very specific subset of the full CIE color space. To be specific they are colors lying on the Planckian locus in the CIE color XY space.

So when the UI widget sends a color temperature POST command, it causes the binding to respond with TWO event notifications for two attributes -- namely 1) the just changed color temperature (in Kelvin) plus 2) the corresponding color XY point on the Planckian locus.

The specific issue in #2849 is that the notification 2) is an accurate location that lies precisely on the Planckian locus so the notification comprises HSB values with double precision to several decimal places. However when the UI widget sends the second (unwanted) POST it rounds the double precision HSB values to the nearest integers which it POSTs back to the binding. This causes the color XY to move to a slightly different point in the CIE color space so it may 'fall off' the Planckian locus. And that in turn causes binding to notify that the color temperature has now become UNDEF.

So my request to you is for two things in this PR:

1. Try to suppress the second POST wherever possible (which is what you are already doing), plus
2. In the case that the notifications are delayed outside the time window so that the second POST is not suppressed, then please do not round the POST'ed values to the nearest integer, but simply repost the same values that were originally provided in the notification.

florian-h05:
The second POST should always be suppresed, the UI should never round an incoming state and POST it.

Have you been able to give it a try?

florian-h05:
If the Item receives an update, oh-colorpicker will ignore any change from the f7 color picker for 10 ms, which should be enough time to catch the f7 color picker changing because it was updated to the new state.
The rounding is only applied when actually using the color picker to control the Item.

andrewfg:
^
> Have you been able to give it a try?

Ok. Building it now..

florian-h05:
Succeeded with building?

andrewfg:
> Succeeded with building?

Yes. But.. Maybe I built it wrong with the original code. Or maybe your new code is still wrong..

https://github.com/user-attachments/assets/e76c4416-de2e-473a-a0f5-2b94a37d44f6

florian-h05:
Can you please check the Main UI Commit on the About page?

andrewfg:
> the Main UI Commit on the About page?

1612d507

florian-h05:
That’s not this PR, seems something went wrong when building or deploying this PR.

andrewfg:
I just tried another build. But it still doe not work. I think I have to wait untilthere is a proper snapshot and then I can test that.

florian-h05:
I built the UI bundle with my change for you:

[org.openhab.ui-4.3.0-SNAPSHOT.zip](https://github.com/user-attachments/files/17637307/org.openhab.ui-4.3.0-SNAPSHOT.zip)

I don't think it will work to just put it in the addons folder, instead put it somehwere else and install it with `bundle:update` through the Karaf console.

andrewfg:
> built the UI bundle with my change for you

Many thanks @florian-h05 .. I will test it tomorrow.

andrewfg:
@florian-h05 many apologies but I am really struggling with running the tests. Since I could not figure out the syntax of `bundle:update` on Windows, I ended up make a new linux install on an RPi. I installed it with OH latest snapshot dated '202411030259' and I updated the UI bundle with the jar you posted above dated '202411022308' (see below). However the UI Help About is still showing 'Main UI Commit 1612d507' and the UI is still sending unwanted POSTs for the color picker.

image

andrewfg (approved):
LGTM

florian-h05:
Great! As I haven‘t noticed any regressions, let‘s merge this now and have it tested a bit further in the snapshots before milestone 3 on sunday.

PR #2864, opened 2024-11-10
oh-canvas: Fix exception on empty initial svg config

stefan-hoehn:
This fixes a nasty bug on initial setup of an embedded SVG that has no configuration yet


stefan-hoehn:
@florian-h05  Wait a sec, there is more to come.

PR #2873, opened 2024-11-16
[BasicUI] Add preview color to Colorpicker widget

lolodomo:
Also clenup class ColorpickerRenderer

### DIFF
--- a/bundles/org.openhab.ui.basic/snippets-src/colorpicker.html
+++ b/bundles/org.openhab.ui.basic/snippets-src/colorpicker.html
@@ -20,17 +20,13 @@
 			<i class="material-icons">&#xE316;</i>
 		</button>
 		<button class="mdl-button mdl-button--raised mdl-js-button mdl-js-ripple-effect mdl-form__colorpicker--pick">
-			<!-- color_lens -->
-			<i class="material-icons">&#xE3B7;</i>
+			<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+				<rect width="100" height="100" x="0" y="0" rx="20" fill="Gray"/>
+			</svg>
 		</button>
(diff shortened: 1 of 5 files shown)


lolodomo:
The new rendering of the widget:

image

image

This is the same circle as in the Colortemperaturepicker widget.

kaikreuzer:
Why would we need the palette button here at all? It would feel more natural to me that I click on the colored circle if I want to change the color...

lolodomo:
I was not sure but it was also what I was thinking ;)
Note that we should keep something consistent with the new Colortemperaturepicker widget.

Do you think I should just keep the new circle or rather keep the rectangle button but with the current color inside?

kaikreuzer:
> Note that we should keep something consistent with the new Colortemperaturepicker widget.

Agreed. Having a look at it, I would actually say the same there: The picker button is redundant if the user could simply click on the colored circle.

> Do you think I should just keep the new circle or rather keep only the rectangle button but with the current color inside and no icon ?

I don't have a strong opinion. The circle already "invites" the user to click on it, but we could also stick to the rectagular look of the buttons - this might be the more consistent option.

lolodomo:
Kai, it is changed:

image

image

kaikreuzer (approved):
Looks nice, I like it!
Thanks. 😄

felixschndr:
This looks very nice, thank you very much @lolodomo! Maybe we can built something like this for the other clients, looking at the android app @mueller-ma ? That's be great

mueller-ma:
I'm pretty sure that there was a preview like this when I merged @maniac103 PR that implements this new widget.

maniac103:
Yes, the Android app does it like that already (displaying the color preview instead of the palette), with one tweak: if the light is off (brightness = 0) or the state is unknown, the palette is shown instead of the preview because a black preview looks somewhat ugly.

PR #2884, opened 2024-11-25
Add Log Viewer to MainUI

### LINKED ISSUE #2883: Add log viewer directly to the main UI
Currently frontail is available as an addon in openhabian. I just moved my main system to a new server that's not openhabian - Frontail requires a lot of dependencies so I wanted to avoid loading it - it also seems to be no longer developed, and it would be nice(r) IMHO to have a log viewer directly embedded within the Developer Tools menu in the main UI. I guess given there is a websocket `/logs` endpoint in the core, this was also considered as the way forward...

I just wanted to raise this issue for visibility / comment. **Is anyone working on such a viewer**? I've not found anything from a quick search other than these closed issues.

#1185
#1109

I very quickly hacked something together last week as an initial concept to display the logs and manipulate logging levels - it needs a LOT more work to make it what I'd personally like to see. If someone is already working on a viewer or has plans, or has a minimum product etc, then I won't spend too much time on it, but if not I might look to put together a minimum solution. Conversely, if someone does have something running, maybe they'd like to share...

image

cdjackson:
Closes #2883.
This adds a basic log viewer into the Main UI, and displays the logs from openHAB log file using a websocket connection.

This is probably a bit rough still, but it's functional so I welcome feedback.  There are a few issues that need to be worked through -:

* The websocket address is currently hard coded - I couldn't work out how to get the dev proxy to proxy the websocket data (it always wants to default to http).
* This looks a bit messy in dark mode - probably someone with a better understanding of the UI CSS could easily change some of this to make it work properly with the themes
* Performance may be improved? The system can become swamped with very heave debug logging. Possibly active highlights could be moved into a separate array (might help if there are a lot of inactive highlights if a user creates a large list).

## Overview 

This adds the log viewer to the Developer Tools -:

image

The main logger window is reasonably simple, and a little busy. Log entries are styled with a background color (again, dark mode may not work so well) - red=error, yellow=warn, greed=info, grey text = trace, no change = debug. The window scrolls horizontally - the time column remains fixed, but the log level, class and message scroll left/right. This allows the user to effectively remove these columns if they desire by simply setting the horizontal scroll point accordingly.

image

Across the top in the navbar are a bunch of status and configuration data/buttons. These are (from left to right) -:

* Time of the oldest and latest log entries held in the local buffer.
* Number of log entries currently displayed. This is less than the number of logs available due to the filtering.
* Number of log entries currently in the buffer.
* Total number of entries before old logs are removed. This is currently hard coded at 5000 - this seemed to work ok in my testing, but possibly a larger number might also be fine.
* Filter input box - this filters the display to display only text in either the logger name or message containing the filter text (case insensitive).
* Play button: Continues receiving/processing logs if paused / stopped.
* Pause: Continues to receive logs from openHAB, but doesn't add them to the display, so the display is static. When Play is pressed, the filter is updated, so any recent entries will be added.
* Stop: Stops receiving data from the server. The websocket connection is disconnected - this ensures there is no change to the local buffer.
* Down: Scrolls to the latest entry and keeps the bottom / latest entry visible - scrolling the screen up. Scrolling down to the bottom should also do this.
* Download: Downloads a CSV of the filtered log file for further processing.
* Copy: Copies the log as HTML to the clipboard. This allows pasting of the filtered log complete with highlighting.
* Delete: Clears the log buffer.
* Exclamation: Shows `ERROR` logs even if they would otherwise be filtered out.
* Pen: Configures highlighting (more on this below).
* Cog: Configured logging (more on this below).
 
## Highlighting 

The UI allows the user to add any number of text elements that will be highlighted. The user can specify a colour, and enable and disable if the highlighting is active or not. The list of highlights is saved locally.

image

image
 
## Log Settings 

The user can directly manipulate log levels. This allows the user to simply change the log level for existing packages, or delete the log entry (restoring it to default) or adding a new logger for a custom package. If adding a custom package, this gets added as `WARN` level (it uses the log level provided in the `ROOT` logger), which can then be changed in the dropdown.

image


cdjackson:
Thanks for the feedback @mherwege 

> For larger screens (desktop), it might be useful to not have it full screen, but at the bottom of the screen

While I understand the idea, I might leave this one for another day (or someone who knows the UI and how to do this a bit more than me :) ).

> In my view, the logging class takes up too much space, hiding the log message. It could potentially be shortened with ... at the start.

I should have said above - the window scrolls horizontally. The time column remains fixed, but you can scroll to the right and cut off the log level and class if you want. Arguably neither of these are "important" - the level is already denoted by background colour, and I agree that the class is secondary as well. That said, it might be useful to reduce the size of the class, but I'm also mindful of not doing too much processing on incoming logs as performance will become an issue for heavy logging applications...

> An extra option to wrap the log messages can also be useful. That can avoid horizontal scrolling (I don't know if it is even possible at this time).

I'll have a look at this. For me, having a single line (unwrapped) makes it easier to view when "stuff" is scrolling past quite quickly since everything is much more uniform. As above, horizontal scrolling is already implemented.

digitaldan:
Very excited as well!   This should really help with troubleshooting, thanks @cdjackson  !

florian-h05:
This looks like a great addition! :rocket: 
Many thanks @cdjackson, please let me know once you think this is ready for review.

ghys:
Even without testing I can feel you've put some thought into this and it looks great!

The "Highlight" feature is brilliant.

I also wish this could be a tool that you could invoke at any time when you need it - like the Developer Sidebar with <kbd>Shift-Alt-D</kbd>.

<kbd>Shift-Alt-L</kbd> could work, maybe? (it's actually a pain to find a convenient keyboard shortcut that doesn't conflict with the browser's or whatever else). But it's just wishful thinking, the fullscreen page will definitely do for now.

I'll be sure to test it soon.

First remark: I suspect this busy navigation bar at the top is designed for desktops only and won't fare well on mobile... I'm normally (very) picky about this but that's (relatively) OK because the Developer Tools aren't on the same _"it-has-to-work-on-mobile"_ standard as the rest.

image


Side note - can we drop the checks for the webpack-stats one day, these errors reported twice are annoying now.

image

Why hasn't the RelativeCI bot commented...? :thinking:

cdjackson:
> Even without testing I can feel you've put some thought into this and it looks great!

Yep - it's come together from a server development mindset - ie I plan to use this myself :) On the other side, you might find some of the coding is a little less optimal ;) - it's been a while since I've done too much in js/vue...

> I also wish this could be a tool that you could invoke at any time when you need it 

I agree - I might need to leave that to you :). I'd suggest to get this merged and then do that?

> First remark: I suspect this busy navigation bar at the top is designed for desktops only and won't fare well on mobile...

I'm sure that's right - and I've never tested it. I would say that in reality, viewing logs in a meaningful way on a mobile device is likely to be problematic at best, but if we wanted to support that a bit better, we should be able to hide some of the buttons and status info on different screen sizes (things like the number of logs, and buttons like download are certainly contenders for removal on small devices). I did try that, but couldn't find classes that worked!

Performance may be an issue - certainly with heavy logging this currently can't keep up. I think in the longer term we might want to think more about server side filtering/processing. I had a look at a few open source loggers (eg frontail) and as far as I can tell (from a quick look/read of their repos), they all seem to do a chunk of this on the server side, where-as at the moment the OH server just sends all logs straight out the socket as they are logged. I think some buffering and filtering might be useful, but again, I think that's a discussion for another day unless you have any thoughts on this...

cdjackson:
I've improved dark mode so that it's at least usable. I accept the comments above by @rkoshak about colours in general, but at least this is now usable in dark mode -:

image

I've added a toggle filter with the exclamation mark in the navbar. If enabled, this will show `ERROR` level logs even if they would otherwise have been filtered out. I found this super useful where I've got a filter for a script - this filter would otherwise not display any errors from the script due to the way I'm filtering it, so this ensures errors are always displayed.

I reduced the number of log entries to 2000 - maybe we can make this configurable but I read on another OS logger that it's what they found as a good compromise. It's clearly a trade here with performance versus visibility and as mentioned above, in the slightly longer term we might want to add some more functionality to the core to better support logging. I'm not against doing this myself, but I suggest we do this in a second phase as this logger is perfectly usable, with some performance limits, as it is...

I'll do a rebase shortly and a bit of final testing then @ghys you can tear it apart :)

cdjackson:
@ghys, @florian-h05 I think this is now ok for review. Note that I've not worked out how to proxy the websocket or get the correct URL automatically, so this is still hard coded. Hopefully this is simple and you can suggest a fix.

florian-h05:
Thanks for the info :+1: 

I don't think I will have the time to review this before our feature freeze, and anyway I would propose to not merge such a major addition as short before the 4.3 release, better have more time for testing. 
So I have added this to our openHAB 5 feature planning :rocket:

cdjackson:
> I don't think I will have the time to review this before our feature freeze

No problem - I appreciate the feedback.

> anyway I would propose to not merge such a major addition as short before the 4.3 release, better have more time for testing.

I would just counter this by saying that, yes, this is in one view a "major addition", but in another it's very self contained. It doesn't impact at all on other systems, so the risk of merging it is relatively low. It has no links to the wider UI - it just receives the websocket data and displays it. I think it would be a shame to wait for OH5 (another 6 to 12 months?) just because of the perceived risk of a "major change" that may cause wider problems in the UI.

Anyway, your time is a different issue, so no problem if it doesn't happen due to time :) ...

maniac103:
> I've improved dark mode so that it's at least usable. I accept the comments above by @rkoshak about colours in general, but at least this is now usable in dark mode -:

Looking at that screenshot the colors look pretty bold/strong. I wonder whether using foreground/font colors instead of the background colors would make for a nicer UX by not making single lines jump out as much, but still be distinguishable?

cdjackson:
@maniac103 I fully admit I didn't spend a lot of time styling this - it was more about making it usable and the dark mode styles are basically consistent with the light mode.  

There are quite a few combinations of colours here due to the fact that the colour is used to denote the log level - so there's 5 or 6 options. I think that we do want "bold/strong" colours for things to stand out (certainly for errors / warnings) - in my world, the majority of logging is `DEBUG`, so I set that as the background colour, with info/error/warning with a strong background colour change and trace with a grey font.

However, as @rkoshak mentioned above, that may not be the best styling as probably `INFO` is more common for most users.  If people want to suggest various combinations of colours for the different levels I'm certainly happy to make changes to go with the consensus, but I think at the moment I'd personally prefer to get this merged so there's something to build on.

bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
```diff
@@ -0,0 +1,738 @@
...
+    removeHighlight (index) {
+      this.highlightFilters.splice(index, 1)
+      this.prefilterHighlights()
+    },
+    openColorPopover (index, event) {
+      this.currentHighlightColorItemIndex = index
+      this.currentHighlightColor = this.highlightFilters[index].color
+      this.$f7.popover.open('#color-picker-popover', event.target)
+    },
+    selectHighterColor (color) {
```
maniac103, line 637: `selectHighlightColor`?

lolodomo:
Of course it depends on availability of reviewers but I would be also in favour of putting that enhancement in OH 4.3 if possible.

ghys:
If I'm not mistaken we still have a 1-week window to merge this, if @florian-h05 isn't available I'll do the review next week.

florian-h05:
@ghys You are correct, we have a one week window left to merge this.
I am very likely not available next week, so please review it!

ghys:
I have finally been able to review this.

Took the liberty of committing changes that fixed some styles and cosmetics along with refactorings, which made the log viewer largely mobile-friendly:

image

image

I'm not finished, and I hope there are no regressions, I may continue tomorrow, wish I could do more like having the logs in a drawer as discussed above or restyle the table rows but I'm not sure I will have time. But at least since it's mobile-friendly you can perhaps do changes on your computer and check the logs on your tablet or phone at the same time.

Here's what I encountered when testing:
1. the webpack config was pretty wrong (it targeted the WebSocket used for the dev server's own purpose to the openHAB main WebSocket which are both at `/ws`), took me some time to figure out the weird stuff that was happening :) Also when we decide to use that openHAB `/ws` WebSocket in the UI we might have a problem since it shares the URL with the dev server's
2. I tested with a fresh instance (only got through the setup wizard, no things, no items) and as it happens when the WebSocket doesn't have any traffic for 10 seconds, it times out:
```
21:03:21.642 [INFO ] [ab.core.io.websocket.log.LogWebSocket] - WebSocket error: java.util.concurrent.TimeoutException: Idle timeout expired: 10001/10000 ms
```
this would probably not happen to users who have configured their system (I simulated a message every 2s in the screenshots above with a rule), but it's still not a good experience for new users to have these disconnections and error messages in the log. Also there's no reconnection mechanism so the log viewer would just stop after 10 seconds and have to be restarted manually
3. I tried to do some keep-alive mechanism by sending a 'ping' message ever 9s but it doesn't seem to work.

cdjackson:
Thanks @ghys - I will try and run this up today to check for any regression, but I appreciate your time in a/ reviewing, and b/ improving.

cdjackson:
The idle timeout is set in core at 10 seconds. The "ping" message throws an exception, although this is caught in the handler and there's no return and I don't see why it doesn't work (see below - it does work for me)...

Anyway, I changed the "ping" to "[]" and the exception goes away, and I don't get timeouts. 

I then put the "ping" back, and I also don't get the timeouts!  The ping messages cause a `WARN` message (by default) and that should (??) keep the connection alive in itself as there's always logging coming through.  I set `org.openhab` to `ERROR` - this stops pretty much all logging and I had no logging for a minute, and changed `org.openhab` back to `INFO` and got logging again straight away (see below). I don't think there's any reconnection here.

So, I'm slightly confused why your ping doesn't work - it seems to be fine here, and if I comment it back out, and increase the `org.openhab` logging to `ERROR` again, then I see the timeouts and the socket disconnects.

My suggestion for now is to leave the keepalive enabled, but to use `[]` to avoid the exception. I plan to produce a PR on core that changes this to instead of using a list of package names, can instead take a DTO that allows us to pass other information. I've already started this, but don't plan on providing the PR in the short term (ie before the release).

image

cdjackson:
I noticed that this now becomes VERY slow once we hit the maximum number of logs - I guess it is the element removal that's taking the time (??) - or maybe it decides to re-render the whole table? I don't recall this being slow like this previously, but maybe it was just as bad... 

Otherwise this appears to work the same as before your changes @ghys

ghys:
> So, I'm slightly confused why your ping doesn't work - it seems to be fine here

So I guess it did work, I thought it didn't because I tested when there was both a connection to openHAB's `/ws` and `/ws/logs` websockets due to the webpack misconfiguration.
I thought the latter logged a timeout and it was the former so I figured sending a 'ping' didn't work, although I saw these "Failed to parse 'ping'" messages.

Good catch on the `.send('[]')` to avoid the exception.

ghys:
> I noticed that this now becomes VERY slow once we hit the maximum number of logs - I guess it is the element removal that's taking the time (??) - or maybe it decides to re-render the whole table? I don't recall this being slow like this previously, but maybe it was just as bad...

It's a common problem when you use Vue to filter, it has appalling performance after a few hundred elements.

The Framework7 searchbar component has the ability to bypass all that and just filter elements in the DOM given their parent and item CSS classes, which is way faster - https://v5.framework7.io/vue/searchbar#searchbar-properties. That's how they're used in other pages. However getting them to work can be quite the challenge.

florian-h05:
IIRC does the lot websocket itself support filtering, see https://github.com/openhab/openhab-core/pull/3859.
I guess server-side filtering outperforms the web and also saves bandwidth.

cdjackson:
Yes, the websocket supports filtering, but IMHO this is not super useful. Since there is no local buffering on the server, if you filter using websocket, then you can't change the filter to see something different (which is often the case when debugging). You might as well just change the log level (although of course then it's also not logged to file).

As above, I plan to look at changing the server side - but that's a job for another day ;) I think a combination of server side buffering and filtering is probably a strong combination

I've not looked at the dom filtering mentioned above yet.

ghys:
I'm just remembering them now but this would be a textbook case for a so-called "Virtual List":
https://v5.framework7.io/vue/virtual-list
https://v5.framework7.io/docs/virtual-list
Especially since the items' heights are fixed and can be determined in advance.

They're made for this exact purpose of when you have huge amounts of elements (the examples have 10000).

I'm not sure if we can still do the scroll-to-bottom/auto-scroll features easily with these (there's a `virtualList.scrollToItem(index);` method though) and the filtering would probably need a partial reimplementation.

cdjackson:
Thanks @ghys - I'll take a look at this tonight (your Saturday morning).

@florian-h05 - one other point about the server side filtering that I probably wasn't clear on above - it can only filter on the logger name (ie package). Hence my point that you might as well just use the log levels... A lot of filtering is more useful (again, IMHO) if it's done on the message rather than package... I have started to mod the server side, but it's not for the next release ;)

ghys:
@cdjackson this:
https://github.com/cdjackson/openhab-webui/compare/log-viewer...ghys:openhab-webui:log-viewer-virtual-list
(commit https://github.com/ghys/openhab-webui/commit/b361d29d8ddca4ac2c39ff31a1512f73638e6e64)
is a very early implementation with a virtual list.

I couldn't reproduce the tabular layout easily unfortunately, and the filtering hasn't been changed so it's broken. Maybe you could give it a spin to check if it's better performance-wise, otherwise we can abandon the idea.

cdjackson:
Thanks @ghys 

I did a quick test of your change, and if anything I'd say it has worse performance...  I ran with some moderate log levels and by the time it got to 2000 entries, I'd say (a little subjectively) that it was about 5 seconds behind in the logs it was displaying. You could see that the display was scrolling consistently/all the time - ie not in response to the incoming logs which could can have maybe 1/2 a second between incoming entries.  However, performance didn't noticeably change when we hit the 2000 limit and it started removing old logs...

With the normal table and the same logging, it seemed to keep up with the incoming logs. As opposed to the above test, the logs were noticeably "jittery" as new entries came in at random intervals, so this seemed more "real-time".  However (!!) the big problem here is the issue identified earlier in that as soon as we hit the 2000 limit, and we're removing old entries, performance drops to "zero" - it takes about a second maybe to process a new entry. My guess is that the removal of an entry in the table causes the framework to re-evaluate the whole table.

I have to go out for a while today so won't get the opportunity to look at this further until this evening. I think if we can find a way to avoid the performance hit when the buffer is full, then the original code seems best so far (unless you think there's further optimisation)...

ghys:
@cdjackson thanks for trying, too bad - I agree optimizing the original implementation should be best.

I tried reaching the max log size and you're right, removing the top row of the table seems to be the reason.

It's pretty clear in this performance recording when the 2001st element is added. Most of the time seems to be spent or rendering/recalculating styles.

image

One obvious solution I think would be to display the logs in reverse order i.e. adding recent entries at the top, also there would be nothing to do to "stay at the top", but it's less intuitive.

ghys:
I think I had a marginal performance improvement by not using Vue to render the table, but the bottleneck seems to be elsewhere (in the Dom7 module).

image

I think if there isn't any further objections we can merge this before the feature freeze.

cdjackson:
@ghys I somehow missed your message above, but saw that you'd pushed the direct dom manipulation.  I found a reference to batched processing and manipulating the data within the `requestAnimationFrame` method. This seems to help further - it still can't keep up with really heavy logging, but it does seem to improve on your earlier improvement.

cdjackson:
note that I left your debug that adds 1980 logs on startup enabled...

cdjackson:
I just noticed that the clear log doesn't work now. I can look at it tomorrow, but since the badge with the number of logs is cleared, I suspect that it's just that the dom needs clearing.

ghys (approved):
I made the last fixes and we'll have to leave it at that.
Any further fixes can still be made before the release.
Also the help text for the developer tools will have to be amended:

image

Thanks!

florian-h05:
> Also the help text for the developer tools will have to be amended:

That text can be found in the docs repo:
https://github.com/openhab/openhab-docs/blob/main/mainui/developer/index.md

florian-h05:
@cdjackson Thousand thanks for work, and @ghys many thanks for the review and the fixes on this!

cdjackson:
Awesome - thanks @ghys for your help with getting this over the line.

I think there's still some work to do as I'd like to improve the performance for heavy logging (probably with some core changes) but I think as it stands this is a great start and will hopefully cover most user needs.

mherwege:
> I think there's still some work to do as I'd like to improve the performance for heavy logging (probably with some core changes) but I think as it stands this is a great start and will hopefully cover most user needs.

Great we finally have an integrated log viewer.

One thing that immediately struck me when I started testing this, is that the server sends over all logs, also what has been explicitely excluded in the log4j configuration file. I have the systeminfo binding running and it creats some heavy event logging on things like CPU usage. I normally don't want to see this in the log, so explicitely excluded things like this from the logger configuration.

When you start looking at doing something server side, it would probably be good to have filters you can configure very upstream in the logging pipeline, so it never hits the websocket connection. It would definitely also make the usage in the logger view easier.

PR #2893, opened 2024-12-02
[blockly] Riemann sum persistence extension and js tern definitions

mherwege:
Depends on https://github.com/openhab/openhab-js/pull/401
Depends on https://github.com/openhab/openhab-core/pull/4461

This will also require a new js scripting library release.

This PR extends the Blockly block getting a statistical value from persistence to be aligned with core and js scripting extensions.
The added statistical function is Riemann sum. This is an approximation of the integral value and can e.g. be used to calculate an approximation for energy consumption (in kWh) when the instantanous power (in W) is persisted.
The existing sum method calculates a naive sum, ignoring the time dimension and is not applicable for this. It could be used only if the persistence interval is constant by using the sum and multiplying with the interval duration.

There are multiple types of Riemann sum calculations depending on which value is used as an approximation in each bucket. The ones implemented in core (and js scripting) are: left, right, trapezoidal and midpoint. If the Riemann sum statistical method block is selected in the block, a parameter for this will also be shown, defaulting to left.

As average, variance and deviation statistical methods are based on Riemann sum calculations (the current average calculation assumes Riemann sums of type left), these methods now also have this extra type input parameter (defaulting to left).

Note that in most cases, the trapezoidal (or midpoint) methods would result in better accuracy. However, for backward compatibility reasons, the default has been kept on left.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
@@ -311,7 +311,7 @@ export default function defineOHBlocks_Persistence (f7, persistenceServices) {
     skipPrevious = (skipPrevious === 'undefined') ? false : skipPrevious
 
     let riemannType = block.getFieldValue('riemannType')
-    riemannType = (riemannType === 'undefined') ? '' : (', ' + riemannType)
+    riemannType = (riemannType === 'undefined') ? '' : `, ${itemCode}.persistence.${riemannType}`
 
     const persistenceExtension = (persistenceName === '\'default\'') ? '' : `, ${persistenceName}`
 
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
@@ -25,6 +25,7 @@ export default function defineOHBlocks_Persistence (f7, persistenceServices) {
           ['historic evolution rate', 'evolutionRateSince'], ['future evolution rate', 'evolutionRateUntil'], ['state evolution rate between', 'evolutionRateBetween'],
           ['historic state minimum', 'minimumSince'], ['future state minimum', 'minimumUntil'], ['state minimum between', 'minimumBetween'],
           ['historic state maximum', 'maximumSince'], ['future state maximum', 'maximumUntil'], ['state maximum between', 'maximumBetween'],
+          ['historic state Riemann sum', 'riemannSumSince'], ['future state Riemann sum', 'riemannSumUntil'], ['state Riemann sum between', 'riemannSumBetween'],
           ['historic state sum', 'sumSince'], ['future state sum', 'sumUntil'], ['state sum between', 'sumBetween'],
           ['historic state updates count', 'countSince'], ['future state updates count', 'countUntil'], ['state updates count between', 'countBetween'],
           ['historic state changes count', 'countStateChangesSince'], ['future state changes count', 'countStateChangesUntil'], ['state changes count between', 'countStateChangesBetween'],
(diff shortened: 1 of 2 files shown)


mherwege:
@stefan-hoehn FYI, not to be merged yet, as waiting for the core and javascript enhancements.

stefan-hoehn:
Can you add what the extension is about. We should avoid overloading blocks with too many things (not saying you do so 😉)

mherwege:
> Can you add what the extension is about. We should avoid overloading blocks with too many things (not saying you do so 😉)

I have added some description in the first post. This is about keeping the persistence blocks in line with the (still proposed) core and js scripting extensions.
Agreed it should not be overloaded. But I believe this is an important part of rules, and being able to use all variations of persistence extensions in Blocky feels important to me.

stefan-hoehn:
And let's not forget to update the blockly docs after that ;-)

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
```diff
@@ -282,9 +306,12 @@ export default function defineOHBlocks_Persistence (f7, persistenceServices) {
...
     const dayInfoSince = javascriptGenerator.valueToCode(block, 'dayInfoSince', javascriptGenerator.ORDER_NONE)
     const dayInfoUntil = javascriptGenerator.valueToCode(block, 'dayInfoUntil', javascriptGenerator.ORDER_NONE)
-    const dayInfo = dayInfoSince + ((dayInfoSince && dayInfoUntil) ? ' ,' : '') + dayInfoUntil
+    const dayInfo = dayInfoSince + ((dayInfoSince && dayInfoUntil) ? ', ' : '') + dayInfoUntil
     let skipPrevious = javascriptGenerator.valueToCode(block, 'skipPrevious', javascriptGenerator.ORDER_NONE)
-    skipPrevious = ((skipPrevious === 'undefined') ? false : skipPrevious)
+    skipPrevious = (skipPrevious === 'undefined') ? false : skipPrevious
+
+    let riemannType = block.getFieldValue('riemannType')
+    riemannType = (riemannType === 'undefined') ? '' : `, ${itemCode}.persistence.${riemannType}`
```
florian-h05, line 314: Let's use static access to the RiemannType:

```suggestion
    riemannType = (riemannType === 'undefined') ? '' : `, items.${riemannType}`
```

bundles/org.openhab.ui/web/src/assets/openhab-js-tern-defs.json
florian-h05, line 1: It would be great if you could define the RiemannType in the tern defs and that it is static accesible on the `items` namespace, but TBH I am not sure if that's really possible as JS does not know enums.
What might be a trick to get it working though is to define RiemannType as object with the enum values as string properties.

florian-h05 (changes requested):
Overall LGTM, thanks! Please take a look at my comments.

@stefan-hoehn Can you please take a look at the Blockly changes?

bundles/org.openhab.ui/web/src/assets/openhab-js-tern-defs.json
mherwege: I think I understand what you mean, but I honestly don't know much about tern definitions in the first place. My change is just by analogy. So I am unsure on how to atack this.

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-persistence.js
mherwege: done

bundles/org.openhab.ui/web/src/assets/openhab-js-tern-defs.json
florian-h05: I have just pushed a commit addressing this :+1:

florian-h05 (approved):
LGTM, thanks!

Waiting for @stefan-hoehn to take a look at the Blockly code.

bundles/org.openhab.ui/web/src/assets/openhab-js-tern-defs.json
mherwege: Thanks

stefan-hoehn:
The following block

image

generates

> console.info(('rieman sum garden temp' + String(items.getItem('garten_temperature').persistence.riemannSumSince(time.ZonedDateTime.now().minusHours(1), items.RiemannType.LEFT)?.numericState)));

Unfortunately, I can't test the functionality as I currently have no 5.x backend yet.

stefan-hoehn (approved):

stefan-hoehn:
Sorry, I only noticed now that you already did it 🙏🏻
https://github.com/openhab/openhab-docs/pull/2477

mherwege:
> but I think the blockly docs update is still missing?

@stefan-hoehn I did update it in the same PR you already merged. But must admit I did not redo the images. And it largely links to the broader documentation for more info. Let me know if this is enough.

stefan-hoehn:
Sure, let's leave the images like that. As long as you have mentioned the entries of the drop down and the new field that comes up when choosing the Riemann properties then that's fine.

PR #2894, opened 2024-12-03
[BasicUI] Fix unit determination for min/max from item state description

lolodomo:
Concerns the Colortemperaturepiucker element.

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ColortemppickerRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ColortemppickerRenderer.java
@@ -207,8 +207,24 @@ public class ColortemppickerRenderer extends AbstractWidgetRenderer {
     }
 
     private boolean isUnitInKelvin(StateDescription stateDescription) {
-        // If no pattern or pattern with no unit, assume unit for min and max is Kelvin
-        Unit<?> unit = UnitUtils.parseUnit(stateDescription.getPattern());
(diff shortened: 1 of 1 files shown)


lolodomo:
> we need to add a proper range unit to the state descriptions

Of course but it is probably too late for OH 4.3. So I propose something working well for OH 4.3, our priority.
When something better is available in OH core, of course I will adjust Basic UI.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ColortemppickerRenderer.java
andrewfg, line 1: Instead of hard coding this constant 1000 perhaps you can use the min value from the state description of org.openhab.core.thing.DefaultSystemChannelTypeProvider.SYSTEM_COLOR_TEMPERATURE_ABS

lolodomo: I also have this min value in the class, I should have use it.

lolodomo: By the way, that is a good idea to get default min/max from system channel type, I will change that.

lolodomo: Done

kaikreuzer (approved):

PR #2902, opened 2024-12-07
Allow custom css UI file

JustinGeorgi:
This PR checks for a `theme.css` file served from, OH_CONF/html. If that file exits it links it to the MainUI page after f7 css calculations have been made so that user directives are not over-riden by any f7 settings.

The ability to set MainUI site-wide css has come up several times in forum questions and requests. I think I remember at some point @ghys had some hesitancy with regards to enabling a site-wide css file like this, but I don't remember what the objection was and I can't find where I thought this comes from anymore so I may be misremembering. Either way, this small change just enables an advanced user to create a single css file.

However, if this starts to get a lot of use, I can see refactoring many of the current styles to use a series of `--oh-` specific css variables which would allow for easy creation and sharing of different MainUI themes (something that has just come up again in the most recent wishlist thread on the forums).

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -632,6 +632,15 @@ export default {
       this.updateThemeOptions()
       this.$f7.data.themeOptions = this.themeOptions
 
+      fetch('/static/theme.css').then(resp => {
+        if (resp.ok) {
+          const themeLink = document.createElement('link')
+          themeLink.rel = 'stylesheet'
+          themeLink.href = '/static/theme.css'
+          document.head.appendChild(themeLink)
+        }
+      })
+
       if (!this.user) {
         this.tryExchangeAuthorizationCode().then((user) => {
           this.loggedIn = true


ghys:
@JustinGeorgi  I don't remember where I made my reservations either but in general I'm very conservative when it comes to letting a simple remote HTTP request (either through the UI or directly) make catastrophic changes (direct JS execution or CSS overrides) that would render the UI inoperable or worse, doing malicious stuff.

I'm also not a fan of just making a request and checking whether there's a 404 response or not. I think it's better to check whether the file exists server-side and add it to a index.html template (which would have to be rendered dynamically).

JustinGeorgi:
> I'm also not a fan of just making a request and checking whether there's a 404 response or not. I think it's better to check whether the file exists server-side and add it to a index.html template (which would have to be rendered dynamically).

I'm in complete agreement here, unfortunately what I understand to be the necessary changes to core to implement this method correctly are beyond my current capabilities and also beyond my current time to learn.

I don't even completely disagree with your conservative viewpoint here, but I do think css is a very limited (and 100% easily reversible) risk. So, in my opinion, the pros outweigh the cons here (especially as there are more potentially dangerous alternatives to achieve this if a user is determined but inexperienced).

All that said, feel free to close this if you really think it's a non-starter. This was just a quick proof of concept while I was working on something else, so there's little lost.

florian-h05:
My proposal would be to modify Main UI's `index.html` when serving it through the `UIServlet` and add a directive there to include the css file from the openHAB server.

PR #2904, opened 2024-12-08
Sitemap editor: Add drag & drop

mherwege:
This PR implements drag and drop in the sitemap tree, to make it easier to graphically rearange sitemaps. To achieve this, I have included [Vue.Draggable](https://github.com/SortableJS/Vue.Draggable), by itself based on SortableJS.

Implementing this for the sitemap editor is only a first step. I think it should be possible to also use this library to implement drag and drop in the model tree, making it a lot easier to reconfigure the model tree. The level of complexity there is higher, so I started with trying it on the sitemap tree.

This PR obviously adds another dependency. Let me know if you believe it is worth the added size before I continue on this.

Here is how it looks like:
SitemapDragAndDrop

This will also require further testing on different browsers and mobile.

### DIFF
--- a/bundles/org.openhab.ui/web/.vscode/settings.json
+++ b/bundles/org.openhab.ui/web/.vscode/settings.json
@@ -3,7 +3,7 @@
   "vetur.format.defaultFormatter.html": "js-beautify-html",
   "javascript.format.insertSpaceBeforeFunctionParenthesis": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": false
+    "source.fixAll.eslint": "never"
   },
   "eslint.validate": ["vue", "html", "javascript"]
 }
--- a/bundles/org.openhab.ui/web/package-lock.json
+++ b/bundles/org.openhab.ui/web/package-lock.json
@@ -62,6 +62,7 @@
         "vue-qrcode": "^0.3.5",
         "vue-round-slider": "^1.0.1",
         "vue2-leaflet": "^2.7.1",
+        "vuedraggable": "^2.24.3",
         "vuetrend": "^0.3.4",
         "vuex": "^3.6.2",
         "yaml": "^2.5.0"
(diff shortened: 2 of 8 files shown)


ghys:
Looks promising!

> This PR obviously adds another dependency. Let me know if you believe it is worth the added size before I continue on this.

https://app.relative-ci.com/projects/ZNG5hy4VeSJQVQcq1Kvu/jobs/2537-ytd15fXORFKYrg9tYWdQ/packages?bp=%7B%22filters%22%3A%22changed-1%22%7D&dt=table&dtg=

image

(the webpack-stats above is in the "report" artifact in the Checks tab under "Main UI").

So, there's far worse. However it shouldn't be part of the initial JS, not sure why it isn't part of the bundle where treeview-item.vue is...

mherwege:
> So, there's far worse. However it shouldn't be part of the initial JS, not sure why it isn't part of the bundle where treeview-item.vue is...

If I understand it well, treeview-item.vue is loaded from the beginning:
https://github.com/openhab/openhab-webui/blame/320c3f57b207ea00620f7bcdf4d1b209c2474f10/bundles/org.openhab.ui/web/src/js/app.js#L10

I am not sure why that is the case.

ghys:
Ah you're right, this was the only solution I found for those recursive components who reference themselves in their own template.
You could probably try the async syntax for referencing the draggable component like:
https://github.com/openhab/openhab-webui/blob/320c3f57b207ea00620f7bcdf4d1b209c2474f10/bundles/org.openhab.ui/web/src/components/widgets/system/oh-knob.vue#L15
that should move vuedraggable to another bundle (then make sure it isn't loaded until a treeview item is created).

mherwege:
> that should move vuedraggable to another bundle (then make sure it isn't loaded until a treeview item is created).

Is working now. I similar cleanup will be required for the model treeview.
I also noticed there are some sitemap components still loaded initially. While small, they are not used and I don't think anyone is intending to reimplement BasicUI in Vue. Should I remove all of that as well?

mherwege:
And there even is a better solution.

mherwege:
I have now tested on a number of browsers, Android and iOS. It is working as expected for me. So this is ready for review as far as I am concerned.
I will work on the model treeview in another PR.

ghys:
> And there even is a better solution. Just name the component if it is directly calling itself

I knew there was something obvious I was missing 5 years ago :smile: 
We should do the same at least for the model treeview item.
The initial JS and CSS weights have actually decreased now:
https://app.relative-ci.com/projects/ZNG5hy4VeSJQVQcq1Kvu/jobs/2540-Az3UtyhTQjJloP5QiSVL

I still have to test it on actual phones etc but I think it works well.
Not sure if we should wait for the same to be possible in the model treeview as well.
The sitemap treeview is a good "lower impact" test.
But this is arguably a new feature and we can't merge those currently until after the release.

mherwege:
> I knew there was something obvious I was missing 5 years ago 😄
> We should do the same at least for the model treeview item.

Already working on that, but it is more difficult. I am confident we can get this working for 5.0.

> I still have to test it on actual phones etc but I think it works well.

I did some tests with Browserstack, but not on a real mobile.

I indeed think sitemaps is low impact and I did see it as a first test before attacking the model tree. I stands on its own, so there is no real link and it can be merged when you see fit.

ghys:
> >We should do the same at least for the model treeview item.

> Already working on that, but it is more difficult. I am confident we can get this working for 5.0.

For now I only meant giving the component a name and moving it out of the entry point :)

ghys:
> I also noticed there are some sitemap components still loaded initially. While small, they are not used and I don't think anyone is intending to reimplement BasicUI in Vue. Should I remove all of that as well?

Yep this was partially sitemap viewer in the UI that still exists in code "just in case" even if it shouldn't be compiled in. I agree that the sitemap-viewing UI is Basic UI and there's no need to duplicate the functionality in Main UI. So at least this sitemap item component declaration should be removed, if not everything. I forgot it was still declared here.

mherwege:
@ghys I created a separate PR https://github.com/openhab/openhab-webui/pull/2915 that moves the imports and kills the sitemap element, without any further functional change, separating it out from this PR.

florian-h05:
@mherwege Can you please rebase and resolve conflicts?

mherwege:
@florian-h05 Rebase done

florian-h05 (approved):
LGTM, thanks!
This is really a game changer for editing sitemaps :rocket: 
Now we need this for the semantic model as well and layout pages ;-)

The bundle report looks good IMO, no increase in initial size so everything fine.
I have even tested this on iPhone, works great there as well!

PR #2905, opened 2024-12-09
Log viewer: only render lines around visible area

ghys:
This should fix the performance problems in the log viewer when the number of items reaches the max buffer size.

It works by calculating the range of line indexes currently visible in the scroll area and redraw only those along with those around them (50 on each side). The rest are replaced by 2 padded empty spaces of the appropriate heights, so that the table appears to be of the expected total height.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
@@ -232,6 +232,9 @@
     background #f1f1f1
     z-index 1
 
+  tr.table-rows
+    height 31px
+
   tr.error
     background-color rgb(255, 96, 96)
     color black
@@ -547,13 +550,17 @@ export default {
               this.logStart = removedElement.time
               if (removedElement.visible) {
                 this.filterCount--
-                this.$refs.dataTable.firstChild.removeChild(this.$refs.dataTable.firstChild.firstChild)
+                let firstRow = this.$refs.dataTable.firstChild.firstChild
+                if (firstRow.className === 'padder') firstRow = firstRow.nextSibling
+                this.$refs.dataTable.firstChild.removeChild(firstRow)
               }
             }
           })
           this.batchLogs.length = 0
           if (this.autoScroll) {
             this.$nextTick(() => this.scrollToBottom())
+          } else {
+            this.$nextTick(() => this.handleScroll())
           }
 
           this.batchUpdatePending = false
@@ -593,6 +600,7 @@ export default {
         // Delay manual scroll detection to avoid autoscrolling being defeated when new logs arrive
         this.scrollTime = Date.now() + 250
       }
+      this.redrawPartOfTable()
     },
     handleScroll () {
       const tableContainer = this.$refs.tableContainer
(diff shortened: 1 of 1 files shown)


florian-h05 (commented):
Code looks good, I am curious to try this out later.

cdjackson:
This works generally fine for me. It still can't keep up with the heavy logging if I enable low level logs from Zigbee, but there's no noticeable performance change now when when we hit 2000 entries, and I think it is generally faster with the heavy logging.

One problem I noticed is that if you scroll away from the bottom, the log should be fixed, but it still scrolls.  Maybe it always did this when we've hit the 2000 limit (I'm not sure), but I guess ideally this should be fixed so that users can view the log without having to scroll to keep a section in view.

cdjackson:
Just for info/reference. Since I keep talking about "heavy logging with the zigbee binding" I thought I'd comment out all UI functionality and just see if the websocket side performs - it doesn't...  What I did was to just increment `filterCount` and return in the `addLogEntry` - so no array additions, filtering etc. Moving this return down to after the filtering makes not apparent difference to performance (ie the code now does everything up to, but excluding, the adding of the entry to the array).

So, that tells me that the performance issues are a wider system issue - not the UI (or not just the UI). IIRC `logtail` was able to keep up with this logging (I don't have it installed now), and as I commented on the previous issue, core changes will be needed to make this perform more widely. I recall that logtail was noticeably jittery with this low level zigbee logging, and I suspect that the server side is batching up logs to improve websocket performance.  I've already started to look at the server side buffering in the core, but have no plans on a PR before the release.

Bottom line - as far as I can tell, with this PR the performance of the UI is pretty good and the UI is not the limiting factor in the system.

ghys:
Thanks @cdjackson for the feedback.

It's my own implementation of the "virtual list" I tried last week, and works generally similarly.
The goal is to limit the number of DOM elements affected by a forced style recalculation, so despite the fact that we rebuild the table (partially) when you scroll means a hit on performance (it probably could be improved as well), the goal was to deal with around 200 rows rendered at most, which is more manageable than 2000, and it is no matter how many entries you have in memory. I think you could even increase that 2000-entries limit and it would not make a difference.

But if there are a lot of log entries coming in a short time frame this might become a problem although the batch processing should help.

> One problem I noticed is that if you scroll away from the bottom, the log should be fixed, but it still scrolls.

I tested that and didn't encounter this but obviously it's prone to bugs - it's not a "real" table after all, just one that gets partially rendered with paddings when you scroll. Probably could be fixed.

> Bottom line - as far as I can tell, with this PR the performance of the UI is pretty good and the UI is not the limiting factor in the system.

Lots of moving parts, but the goal of this PR was to fix the appalling performance once you have to remove log entries at the top of the table, which I think was achieved.

cdjackson:
> Lots of moving parts, but the goal of this PR was to fix the appalling performance once you have to remove log entries at the top of the table, which I think was achieved.

Understood and agreed - it certainly solves the problem when the table is full so for me this is definitely a big improvement. I was also thinking that this might allow the 2000 to be increased - I'll probably try that at some point, but I think for the upcoming release we now have a very usable viewer that will cover most needs.

Just one other comment for info - I've now tested this without the dev proxy and it still doesn't keep up with the low-level zigbee debug. I'll look at this further later (as above).

Regarding the scrolling issue I mentioned that you don't see. I note there's a `TODO` in the code and IIRC I think I put that there during my initial development as a result of the same issue, so I suspect it's not a new issue with this PR. It's interesting though that you don't see the problem - did you try it with 2000+ entries - maybe it only occurs then? I need to do some more testing - I saw this at some point and added the `TODO`, but didn't notice it again until today, so maybe it only occurs when the table is maxed out and we're removing elements off the top...

florian-h05:
Looking at the browser console while testing this, I have noticed the extensive console logging.
This is something we should probably remove at some point as well, as console logging does impact performance.

With the change from this PR, scroll behaviour is (obviously) not as smooth as it was before (as already said above) but still very smooth so no problem, just to note that.

I am just waiting for my system to reach the 2000 logs limit, but from what I read here I would say once I have tried out that, we can merge this if you agree.

florian-h05:
I can also observe the behaviour that the logs starts scrolling once the 2000 logs are reached, but that is no big issue for now.

florian-h05 (approved):
@ghys Feel free to merge yourself.
If you don't need the logging, can you remove it or comment it out?

cdjackson:
Thanks @florian-h05 so I think probably that's why I'd not observed this after adding the `TODO` - it's only an issue when we're above 2000 logs and entries are being removed off the top....  I'll have a look to see if I can improve this, but I agree it's not major at this stage.

ghys:
> If you don't need the logging, can you remove it or comment it out?

Certainly, but I thought the debug level was not a concern since it's not even shown by default.
I'll just have a few more testing and then I'll remove them.

florian-h05:
That’s a good point about the debug level, not sure how it behaves in that case.
Performance impact is much smaller when the console is closed, but not nothing for console.log.

ghys:
> Thanks @florian-h05 so I think probably that's why I'd not observed this after adding the TODO - it's only an issue when we're above 2000 logs and entries are being removed off the top.... I'll have a look to see if I can improve this, but I agree it's not major at this stage.

I think this is in fact to be expected. Once you reach the limit you remove the top row for every row you add at the bottom, so if the scroll position stays the same it's quite normal to have the logs appear to "scroll up".

florian-h05 (approved):
LGTM, thanks!

PR #2907, opened 2024-12-09
Refactor websocket code to `$oh` namespace & Pass access token through header

florian-h05:
Depens on https://github.com/openhab/openhab-core/pull/4515.

This refactors the WebSocket connection code from #2884 to the `$oh` namespace, same as it is for the SSE logic.
It also passes the access token as WebSocket subprotocol so it is sent with the `Sec-WebSocket-Protocol` header.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/index.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/index.js
@@ -1,6 +1,7 @@
 import api from './api'
 import auth from './auth'
 import sse from './sse'
+import ws from './ws'
 import media from './media'
 import speech from './speech'
 import utils from './utils'
@@ -9,6 +10,7 @@ export default {
   api,
   auth,
   sse,
+  ws,
   media,
   speech,
   utils
--- a/bundles/org.openhab.ui/web/src/js/openhab/ws.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/ws.js
@@ -11,8 +11,7 @@ const openWSConnections = []
 
 function newWSConnection (path, messageCallback, readyCallback, errorCallback, heartbeatCallback, heartbeatInterval) {
   // Create a new WebSocket connection
-  const socket = new WebSocket(path + `?access_token=${getAccessToken()}`)
-  socket.path = path
+  const socket = new WebSocket(path, [`org.openhab.ws.accessToken.base64.${btoa(getAccessToken())}`, 'org.openhab.ws.protocol.default'])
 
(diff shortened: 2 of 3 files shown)


florian-h05:
@ghys Can you please have a look at this PR?

ghys (approved):
It's good but I'd still prefer the access token to be passed in headers instead of the URL query: https://github.com/openhab/openhab-core/issues/4490
(at least when possible)

This reduces the odds that it could appear in a log.
Even if ephemeral, tokens should never be logged.

If it's urgent please merge but otherwise I don't see why the rush.

florian-h05:
@ghys I have created a PR to address https://github.com/openhab/openhab-core/issues/4490 and adjusted this PR here.

cdjackson:
@florian-h05 is it possible to provide backward compatibility with the old url parameter based authentication so we can continue to run this on OH4 for a while? I've got an OH5 test system which is fine for testing UI, but it would really be nice for now if we can also proxy this across to a live OH4.x system, and currently that doesn't work after this change.

I tried adding the parameter back (so putting the token in both the header and parameter) and that didn't work - I had to remove the header to make it work on OH4. 

I think it would be good to keep compatibility for testing for a little while at least if possible?

florian-h05:
> is it possible to provide backward compatibility with the old url parameter based authentication so we can continue to run this on OH4 for a while?

That is difficult to implement, as we would need to first check if the header is supported or not ... or it would require at least some amount of boilerplate code.

cdjackson:
I see in other issues you mention having a switch to switch between SSE and WS - could we do something similar? Sorry - I know it's a pain, but at least for me my OH5 system is pretty much bare bones so there's no significant logging coming out, and testing the log function is difficult. I can keep trying to apply a patch, but that's also painful...

florian-h05:
Yeah, it's a pain ... what if the UI would use the access token as query param in development mode, so dev against openHAB 4 would work?

cdjackson:
For me, and for now, that would be fine. I don't know about others, but I normally try to run some level of snapshot on my production system, but just at the moment with the 4->5 transition I'm not doing that due to a much higher potential for breaking changes.

You talk about dev mode - are you then talking about making the change somehow to the proxy in the npm script? Or is there a UI side development mode switch that I'm not aware of?

florian-h05:
You can check whether you are in dev mode inside the UI, see https://github.com/openhab/openhab-webui/pull/2980.
You basically only need to revive my PR https://github.com/openhab/openhab-webui/pull/2980.

Issue #2395, opened 2024-02-23
[MainUI] Excessive fetching of inbox rest resource from settings/things


seime:
## The problem

After opening `Settings` -> `Things` page, the UI starts reloading the `inbox` resource excessively. This leads to a frozen UI and overheated CPU.

After initial load:
<img width="772" alt="CleanShot 2024-02-23 at 08 37 12@2x" src="https://github.com/openhab/openhab-webui/assets/1565156/23271870-56b1-4f13-915c-36248315d9e2">

With a mouse click (anywhere?) it starts issuing a bunch of calls to the `inbox` endpoint in an endless loop.

<img width="1640" alt="CleanShot 2024-02-23 at 08 39 04@2x" src="https://github.com/openhab/openhab-webui/assets/1565156/7466b994-6663-4f0e-bc57-a68df1144725">

The loading of the `inbox` seems to happen in batches of 100+. 

## Expected behavior

A single call to `inbox`, not hundreds (if browser window left open, thousands and a CPU fan going mad)

## Steps to reproduce

1. Open `Settings` -> `Things`
2. After first load, just click on a random place on the page (that isn't a link to another view)

-> Endless loop starts.

## Your environment

```yaml
runtimeInfo:
  version: 4.1.1
  buildString: Release Build
locale: en-NO
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 17.0.10
  javaVendor: Debian
  osName: Linux
  osVersion: 6.1.21-v8+
  osArchitecture: aarch64
  availableProcessors: 4
  freeMemory: 1490051368
  totalMemory: 2147483648
  uptime: 85271
  startLevel: 70

clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: true
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: macos
    pixelRatio: 2
    prefersColorScheme: light
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 8
    language: en-GB
    languages:
      - en-GB
      - nb-NO
      - nb
      - en-US
      - en
      - no
      - hr
[...]

PR #2918, opened 2024-12-14
[MainUI] Do not reload inbox for inboxupdatedevents

seime:
See https://github.com/openhab/openhab-webui/issues/2395

Skip inboxupdatedevents as they do not change the inbox count (which is the only information actually used after loading this potentially very large structure of data)


florian-h05 (commented):
Thanks for putting this up!

Instead of ignoring the InboxUpdatedEvent client side, we can say we don't want to listen to it and the server will not sent it anymore.

I have done that in #2942, which therefore replaces this PR.

florian-h05:
Superseded by #2942.

PR #2919, opened 2024-12-14
Update docs with missing npm lint scripts

seime:


florian-h05 (approved):
LGTM. As this is documentation only, we can safely merge it during the code freeze.

PR #2920, opened 2024-12-14
[MainUI] Add dialog to inbox that shows thing file template

seime:
Makes it easier for everyone using file configuration to adopt discovered inbox things.


florian-h05 (changes requested):
Let's wait for the result of https://github.com/openhab/openhab-core/issues/4509 before merging anything here.

PR #2921, opened 2024-12-14
[MainUI] Add thing action menu to add new thing page

seime:
When adding from inbox, there are a few thing options to choose from, ie setting a custom id (and also showing thing file details after #2920 .

Add some of the same actions to route via binding

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-thing-type.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-thing-type.vue
@@ -34,7 +34,7 @@
           <f7-list-item v-for="entry in scanResults"
                         :key="entry.thingUID"
                         :link="true"
-                        @click="approve(entry)"
+                        @click="openEntryActions(entry)"
                         media-item
                         :title="entry.label"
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-thing-type.vue
```diff
@@ -224,29 +224,99 @@ export default {
...
+                      },
+                      null,
+                      entry.label)
+                  },
+                  null,
+                  entry.thingUID.substring(entry.thingUID.lastIndexOf(':') + 1))
+              }
+            },
+            {
+              text: 'Show equivalent Thing file configuration (basic properties)',
```
florian-h05, line 276: Can you please remove this from this PR? See https://github.com/openhab/openhab-webui/pull/2920#pullrequestreview-2519664477.

I would also suggest to refactor the `openEntryActions` and `approveEntry` actions to a mixin, which then both `inbox-list.vue` and `choose-thing-type.vue` can import. This reduces code duplication and avoids further issues where those two provide different options.

florian-h05 (changes requested):

florian-h05:
Why have you closed?
I have created #2973 which picks up your work here, but uses a mixin.

PR #2932, opened 2024-12-19
`oh-clock-card`: Remove (broken) background property in favour of style config

florian-h05:
This avoids confusion because `background` is not working anymore.
Use `style: background: ` instead.


sebastiencastermans:
> Use `style: background: ` instead.

Hi ! It would be great to update the documentation [here](https://www.openhab.org/docs/ui/components/oh-clock-card.html) :)...

florian-h05:
You just have to click on the link below the page …

image

PR #2690, opened 2024-07-26
Link add: Support linking Number channel to Switch Item with profile

florian-h05:
Closes #1478.

Issue #1478, opened 2022-08-24
Number channel cannot be linked to Switch Item (with profile)


ssalonen:
## The problem

When linking a Number channel, it is impossible to select Switch item to link to

I use profile to transform the numbers to ON/OFF type.

As a workaround, it is possible to change item type temporarily to Number, do the channel link, and revert item type back to Switch.

## Expected behavior

UI would not filter out items of different type when linking channels. Thanks to profiles, otherwise "incompatible" types should be linkable.

## Steps to reproduce

1. 

## Your environment

```yaml
runtimeInfo:
  version: 3.3.0
  buildString: Release Build
locale: fi-FI
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 11.0.15
  javaVendor: Raspbian
  osName: Linux
  osVersion: 5.15.32-v7l+
  osArchitecture: arm
  availableProcessors: 4
  freeMemory: 81427080
  totalMemory: 372244480
  startLevel: 100
bindings:
  - fmiweather
  - modbus
  - network
  - shelly
  - systeminfo
  - zwave
clientInfo:
  device:
    ios: true
    android: false
    androidChrome: false
    desktop: false
    iphone: true
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    os: ios
    osVersion: 15.6.1
    webView: true
    webview: true
    standalone: true
    pixelRatio: 3
    prefersColorScheme: dark
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 4
    language: en-GB
    languages:
      - en-GB
    onLine: true
    platform: iPhone
  screen:
    width: 390
    height: 844
    colorDepth: 32
  support:
    touch: true
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: true
    intersectionObserver: true
  themeOptions:
    dark: dark
    filled: true
    pageTransitionAnimation: default
    bars: light
    homeNavbar: default
    homeBackground: default
    expandableCardAnimation: default
  userAgent: Mozilla/5.0 (iPhone; CPU iPhone OS 15_6_1 like Mac OS X)
[...]

PR #2936, opened 2024-12-22
Fix profile selection when linking a channel to an item

jimtng:
Regression from #2690

Reported here:
https://community.openhab.org/t/enocean-impossible-to-link-a-rockerswitch-channel-with-an-item-in-main-ui-there-is-no-profile-available-for-the-selected-item/160987

When creating a Thing channel link to an item, the profile selection are disabled. This presents two problems:

- It made linking a trigger channel to an item not possible, because a profile must be selected, but they're disabled.
- Linking a non-trigger channel to a new item is possible, but selecting a profile is not possible at link creation. The user has to create the link without a profile first, then go back to edit the link in order to assign a profile.

This PR fixes both these problems.

Note this needs to be backported to 4.3.x

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
@@ -182,9 +182,12 @@ export default {
     },
     compatibleProfileTypes () {
       let currentItemType = this.currentItem && this.currentItem.type ? this.currentItem.type : ''
-      return this.profileTypes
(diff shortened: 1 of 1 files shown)


jimtng:
@ssalonen would you mind testing this to see whether the changes in this PR stilll works for #1478 ?

The test JAR can be found here if you don't want to build it yourself:
https://github.com/jimtng/openhab-webui/releases/tag/ui-link-fix-4.3.1-1

florian-h05:
@jimtng See #2941, this fixes the same issue and a bit more and refactors the code a bit.
I will therefore close your PR in favour of mine.

ssalonen:
I am unable to test this due to lack of time, sorry

PR #2938, opened 2024-12-22
Fix enabled profiles in link editor appear disabled

jimtng:
Reported here:
https://community.openhab.org/t/enocean-impossible-to-link-a-rockerswitch-channel-with-an-item-in-main-ui-there-is-no-profile-available-for-the-selected-item/160987/40

This problem exists in 4.3.0 too, so not sure if we want to backport this?


florian-h05 (approved):
LGTM, thanks for this styling fix.
I will backport this one!

PR #2941, opened 2024-12-22
Link add/edit: Fix profile selection

florian-h05:
Regression from #2690.

Reported here:
https://community.openhab.org/t/enocean-impossible-to-link-a-rockerswitch-channel-with-an-item-in-main-ui-there-is-no-profile-available-for-the-selected-item/160987

When creating a Thing channel link to an item, the profile selection are disabled. This presents two problems:

- It made linking a trigger channel to an item not possible, because a profile must be selected, but they're disabled.
- Linking a non-trigger channel to a new item is possible, but selecting a profile is not possible at link creation. The user has to create the link without a profile first, then go back to edit the link in order to assign a profile.

When editing a channel link, profile selection was possible for unsupported profiles, e.g. when editing a link between a Switch Item and a Number channel, one could edit the link to use the default or follow profiles.

This PR fixes these problems.

Note this needs to be backported to 4.3.x

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-mixin.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-mixin.js
@@ -1,18 +1,18 @@
 export default {
   methods: {
     /**
-     * Check whether the type of the given Item does not match the type of the given channel.
-     * @param {object} channel
+     * Check whether the type of the given Item does match the type of the given channel.
      * @param {object} item
+     * @param {object} channel
      * @return {boolean}
      */
-    itemTypeIsNotChannelType (channel, item) {
-      if (!channel || !channel.itemType) return false
-      if (!item || !item.type) return false
+    itemTypeIsChannelType (item, channel) {
+      if (!channel || !channel.itemType) return true
+      if (!item || !item.type) return true
       if (channel.itemType.startsWith('Number')) {
-        return !item.type.startsWith('Number')
+        return item.type.startsWith('Number')
       }
-      return channel.itemType !== item.type
+      return channel.itemType === item.type
     },
     /**
      * Check whether the given profileType is compatible with the given Item for the given channel.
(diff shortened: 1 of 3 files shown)


florian-h05:
Supersedes #2936.

florian-h05:
Wrt to https://github.com/openhab/openhab-webui/pull/2936#issuecomment-2558457840:

I have tested this and it still works :+1:

florian-h05:
@jimtng Can you please test this PR?

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-mixin.js
```diff
@@ -0,0 +1,31 @@
+export default {
+  methods: {
+    /**
+     * Check whether the type of the given Item does not match the type of the given channel.
+     * @param {object} channel
+     * @param {object} item
+     * @return {boolean}
+     */
+    itemTypeIsNotChannelType (channel, item) {
```
jimtng, line 9: This makes code like `if (!itemTypeIsNotChannelType())` much harder to read.

I'd prefer it if this method is called `itemTypeIsChannelType`, with the corresponding inverse of the logic of course.

jimtng (commented):
Thanks for the refactoring! Just one stylistic request below.

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-mixin.js
jimtng: To be even more nitpicky:
`channelTypeIsItemType` so the method name matches the order of the parameters, or reverse the parameter order to match the method name.

jimtng:
> @jimtng Can you please test this PR?

It seems to work fine, but I'm not the one who's affected by this, so I can't say with 100% certainty. Could you build a jar and ask the people here to test: https://community.openhab.org/t/enocean-impossible-to-link-a-rockerswitch-channel-with-an-item-in-main-ui-there-is-no-profile-available-for-the-selected-item/160987

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-mixin.js
florian-h05: Makes sense, I will change that.

florian-h05:
Merged as it has been confirmed in the linked community thread that the fix works.

lsiepel:
I would classify this as critical as this makes the profile selections not useable at all.

PR #2946, opened 2024-12-24
Enable additional chart types for `oh-data-series`

Nadahar:
This enables the use of all imported ECharts in (custom) widgets using  `oh-data-series`.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/chart/index.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/chart/index.js
@@ -286,7 +286,7 @@ export default {
     props: {
       parameterGroups: [actionGroup()],
       parameters: [
-        seriesTypeParameter('gauge', 'pie'),
+        seriesTypeParameter('line', 'bar', 'heatmap', 'scatter', 'gauge', 'pie'),
         ...actionParams()
       ]
     }
--- a/bundles/org.openhab.ui/doc/components/oh-data-series.md
+++ b/bundles/org.openhab.ui/doc/components/oh-data-series.md
@@ -33,6 +33,10 @@ prev: /docs/ui/components/
     The type of the series.<br/><em>Note: <code>heatmap</code> needs a configured visual map or uses the default and is not supported for time series!</em>
   </PropDescription>
   <PropOptions>
+    <PropOption value="line" label="Line" />
+    <PropOption value="bar" label="Bar" />
+    <PropOption value="heatmap" label="Heatmap" />
+    <PropOption value="scatter" label="Scatter" />
     <PropOption value="gauge" label="Gauge" />
     <PropOption value="pie" label="Pie" />
   </PropOptions>


bundles/org.openhab.ui/web/src/assets/definitions/widgets/chart/index.js
```diff
@@ -286,7 +286,7 @@ export default {
     props: {
       parameterGroups: [actionGroup()],
       parameters: [
-        seriesTypeParameter('gauge', 'pie'),
+        seriesTypeParameter('bar', 'gauge', 'heatmap', 'line', 'pie', 'scatter'),
```
florian-h05, line 289: As the order is shown in the docs and the UI: I would rather order it similar to below, i.e.:

line, bar, heatmap, scatter, gauge, pie

Nadahar: I changed the order to alphabetical exactly because I realized that the order would be visible. I can change it to whatever order you'd like, but I'm not sure I "get" what order you're suggesting.

Nadahar: If the order is that of the order in which they are documented at ECharts, the order should be:

line, bar, pie, scatter, heatmap, gauge

I have no idea how they came up with that order though - maybe it's simply the order in which the components were created/added?

florian-h05: I was referring to the order in line 302 and 317 of this file.

> maybe it's simply the order in which the components were created/added?

I guess yes, but I would like to keep everything consistent with the status quo.

florian-h05 (approved):
LGTM, thanks!

We can apply the exception from DCO here as this is a one liner code-wise and the docs are autogenerated.

BrettLHolmes:
@florian-h05  - apologies if commenting here is incorrect.  I am trying to do some testing of bar charts for the openhab series and the theory is it should be simple and should be able to get it working as i tested in echarts.  - I tested here by changing bar to pie and the echart changes.  When i try the reverse from a openHAB echart pie (to bar) i get a blank screen result.

OpenHAB 5.0 Stable Release

Am i missing  something

On Apache I change the bar to pie (& vice verse) and this works
https://echarts.apache.org/examples/en/editor.html?c=bar-simple
on openhab i change a working simple pie to bar and get black screen
https://community.openhab.org/t/openhab-5-0-milestone-discussion/162686/74

@Nadahar can you provide any advice please.

florian-h05:
Ask @Nadahar for help, I have only reviewed this PR.

Nadahar:
It's been a while since I dealt with this, so I don't quite remember the details, but I did make a widget using a bar chart, so I got it working at least.

Your data probably can't be the same for a bar and a pie chart..? Have you defined an axis, for example?

Nadahar:
I found this widget among my files from back then, when I paste this into OH and set up a temperature Item, it does draw something:
```yaml
uid: widget_test_bar
tags:
  - bar
  - temperature
props:
  parameters:
    - context: item
      description: The current value
      label: Item
      name: item
      required: true
      type: TEXT
    - context: number
      description: The min range value
      label: MinRange
      name: minRange
      required: false
      type: TEXT
    - context: number
      description: The max range value
      label: MaxRange
      name: maxRange
      required: false
      type: TEXT
    - context: number
      description: The min value
      label: Min
      name: min
      required: false
      type: TEXT
    - context: number
      description: The max value
      label: Max
      name: max
      required: false
      type: TEXT
    - context: string
      description: The name
      label: Name
      name: name
      required: false
      type: TEXT
    - context: number
      description: The height of widget
      label: Height
      name: height
      required: false
      type: TEXT
  parameterGroups: []
timestamp: Dec 20, 2024, 3:10:02 AM
component: oh-chart
config:
  chartType: day
  height: =props.height
  periodVisible: false
slots:
  grid:
    - component: oh-chart-grid
      config:
        borderWidth: 0
        includeLabels: false
        width: 10
  series:
    - component: oh-aggregate-series
      config:
        aggregationFunction: last
        barCategoryGap: 0
        barGap: 0
        barWidth: 10
        dimension1: date
        gridIndex: 0
        item: =props.item
        itemStyle:
          color: red
        offset: 5
        showBackground: false
        type: bar
        xAxisIndex: 0
        yAxisIndex: 0
  tooltip:
    - component: oh-chart-tooltip
      config:
        orient: vertical
        show: true
        trigger: axis
  xAxis:
    - component: oh-value-axis
      config:
        axisLabel:
          color: red
          show: true
        axisLine:
          show: true
        axisPointer:
          show: true
        axisTick:
          interval: 1
          lineStyle:
            opacity: 1
          show: true
        gridIndex: 0
        show: false
  yAxis:
    - component: oh-value-axis
      config:
        axisLabel:
          align: right
          margin: 23
          show: true
          verticalAlign: bottom
        axisLine:
          onZero: true
          show: true
        axisTick:
          length: 40
        gridIndex: 0
        inverse: false
        max: 30
        min: -30
        minorTick:
          length: 20
          show: true
        offset: 0
        scale: false
        show: true
        splitNumber: 10
        startValue: -20
```

Nadahar:
It should be said that I've never tried to use echarts for series in OH - my dabbling with echarts was all about widgets.

BrettLHolmes:
@Nadahar thanks a million for the sample it has given me even more that I was looking for
Indeed you are correct as well if the Axix is not defined it does not display - I will do some testing and post in the community for reference - Thanks

PR #2948, opened 2024-12-27
Interactive SVG add String Type support

stefan-hoehn:
Addition to interactive SVG layout:

Adds String type support by allowing to compare a string to a specific value that is perceived as state ON:

Here an example: If the state of the String is "BD/DVD", highlight the button of the receiver:

image

image


florian-h05 (approved):
LGTM, thanks!

PR #2953, opened 2024-12-28
Page: Add ability to edit current tab of tabbed page

### LINKED ISSUE #2899: Add "Edit page" link to tabbed pages
## The problem
I often use tabbed pages to display different charts on one page. But when I want to make some changes to one particular tab page, I can’t navigate to edit this page directly (e.g. change chart configuration). If I use the edit button in top right corner I get to the tabbed layout, where I can manage the tabs, but it is not possible to go directly to page of one particular tab. I have to check the page UID and find that page manually on Pages list or using developer toolbar.

It would be really nice to have some shortcut for this use case.

## Your suggestion

I have two suggestions.

The easiest one would be to add second Edit link on tabbed pages, where one would lead to edit of tabbed layout page and second would lead to edit page of currently opened tab.

Something like this:
image

Second option is to add one entry into submenu of tabbed layout page to list of tabs. That link would lead to edit page of that particular tab.

Something like this:
image

## Your environment

```yaml

```

florian-h05:
Closes #2899.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/page/page-view.vue
+++ b/bundles/org.openhab.ui/web/src/pages/page/page-view.vue
@@ -6,7 +6,7 @@
       </f7-nav-left>
       <f7-nav-title>{{ (page) ? page.config.label : '' }}</f7-nav-title>
       <f7-nav-right>
-        <f7-link v-if="isAdmin" icon-md="material:edit" :href="'/settings/pages/' + pageType + '/' + uid">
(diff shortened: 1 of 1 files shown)


ghys:
At last! :laughing: 
This had bugged me for years as this was an obvious improvement. To tell the truth I was holding on to see someone doing a PR about it (there is/was a "good first PR" GitHub tag to let first-comers know what to work on to get acquainted with the code).
So I'm both relieved it's finally done and disappointed a maintainer had to do it.

PR #2958, opened 2024-12-31
Allow item names to start with an underscore

jimtng:
Reported in https://community.openhab.org/t/openhab-4-3-item-name-starting-with-underscore/161244

Regression from #2658


florian-h05 (approved):
LGTM, thanks for the fix.

florian-h05:
Backported to 4.3.x.

PR #2965, opened 2025-01-01
Widget expressions: Fix handling of Item name being `undefined` & Align behaviour for @ and @@ shorthands in that case

### LINKED ISSUE #2956: [WARN ] [e.internal.SseItemStatesEventBuilder] - Attempting to send a state update of an item which doesn’t exist: undefined
This is a fairly frequently occuring warning, and often the origin/cause is not apparent

Trouble then is, it's hard to find out *what* is the item and which widget is it used in.

Please make that information part of the error message to help with debugging.

florian-h05:
Fixes #2956.
This fix creates the limitation that an Item must not be named `undefined` - for an Item with this name, state tracking won't work.


ghys (approved):
> This fix creates the limitation that an Item must not be named undefined

Well I think we can live with that :)
Thanks, these useless log messages happened to me too and getting rid of them is a good thing.

PR #2967, opened 2025-01-02
Wrap lines in log messages with line feeds

### LINKED ISSUE #2961: [Log Viewer] The logs from exceptions errors are converted to a single line
## The problem
Logs from exception errors (such as stack traces) are being converted from multiple lines to a single line.

Reading such messages in a single line is inconvenient and may force users to unnecessarily copy the text and paste it on a different tool to analyze the problem.

## Your suggestion

Logs with internal formatting to split the text into multiple lines should not be modified to fit in a single line.

## Your environment

```yaml
runtimeInfo:
  version: 4.3.0
  buildString: Release Build
locale: en-DE
systemInfo:
  configFolder: /etc/openhab
  logFolder: /var/log/openhab
  javaVersion: 17.0.13
  javaVendor: Red Hat, Inc.
  javaVendorVersion: (Red_Hat-17.0.13.0.11-1)
  osName: Linux
  osVersion: 5.4.60-v8.1.el8
  osArchitecture: aarch64
  availableProcessors: 4
  freeMemory: 309470264
  totalMemory: 730857472
  uptime: 489198
  startLevel: 100
addons:
  - binding-astro
  - binding-hue
  - binding-irobot
  - binding-netatmo
  - binding-openweathermap
  - binding-roku
  - binding-sonos
  - binding-systeminfo
  - binding-telegram
  - binding-tplinksmarthome
  - binding-unifi
  - binding-upnpcontrol
  - misc-openhabcloud
  - persistence-jdbc-mysql
  - transformation-map
  - ui-habot
  - voice-marytts
clientInfo:
  device:
    desktop: true
    edge: false
    ie: false
    firefox: false
    macos: true
    webView: false
    webview: false
    standalone: false
    os: macos
    pixelRatio: 2
    prefersColorScheme: dark
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 8
    language: en-US
    languages:
      - en-US
    onLine: true
    platform: MacIntel
  screen:
    width: 1470
    height: 956
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: dark
    filled: true
    bars: light
    homeNavbar: default
    homeBackground: default
    blocklyRenderer: null
  userAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15
    (KHTML, like Gecko) Version/18.1.1 Safari/605.1.15
timestamp: 2025-01-01T09:16:29.042Z

```

cdjackson:
This wraps log messages where the message has a line feed sequence. The main case for this is exception / stack traces.

image

Closes #2961

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
@@ -125,10 +125,8 @@
                  :class="{ 'disabled-link': !stateConnected || !stateProcessing, 'no-margin-left': $device.ios }"
                  @click="loggingPause" />
         <f7-link icon-ios="f7:stop_fill" icon-aurora="f7:stop_fill" icon-md="material:stop_fill"
-                 :icon-color="!stateConnected ? 'gray' : ''"
-                 :tooltip="!$device.ios ? 'Stop receiving logs' : ''"
-                 :class="{ 'disabled-link': !stateConnected, 'no-margin-left': $device.ios }"
-                 @click="loggingStop" />
+                 :icon-color="!stateConnected ? 'gray' : ''" :tooltip="!$device.ios ? 'Stop receiving logs' : ''"
+                 :class="{ 'disabled-link': !stateConnected, 'no-margin-left': $device.ios }" @click="loggingStop" />
       </f7-nav-right>
 
       <f7-subnavbar :inner="false" style="padding-right: var(--f7-safe-area-right)">
@@ -158,8 +156,8 @@
                :class="{ 'disabled-link': filterCount == 0 }" @click="downloadCSV" />
       <f7-link icon-f7="rectangle_on_rectangle" tooltip="Copy filtered log to clipboard"
                :class="{ 'disabled-link': filterCount == 0 }" @click="copyTableToClipboard" />
-      <f7-link icon-f7="trash" tooltip="Clear the log buffer"
-               :class="{ 'disabled-link': tableData.length == 0 }" @click="clearLog" />
+      <f7-link icon-f7="trash" tooltip="Clear the log buffer" :class="{ 'disabled-link': tableData.length == 0 }"
+               @click="clearLog" />
       <f7-link @click="toggleErrorDisplay" tooltip="Always show error level logs">
         <f7-icon v-if="showErrors" f7="exclamationmark_triangle_fill" />
         <f7-icon v-else f7="exclamationmark_triangle" />
@@ -220,7 +218,7 @@
     table-layout auto
 
   td.nowrap
-    padding 5px
+    padding 0px
     text-align left
     white-space nowrap
 
@@ -233,6 +231,7 @@
 
   tr.table-rows
     height 31px
+    vertical-align top
 
   tr.error
     background-color rgb(255, 96, 96)
@@ -486,7 +485,7 @@ export default {
         milliseconds: ms,
         level: logEntry.level.toUpperCase(),
         loggerName: logEntry.loggerName,
-        message: logEntry.message
+        message: logEntry.message.replace(/\n/g, '<br>')
       }
 
       this.batchLogs.push(entry)


bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
```diff
@@ -486,7 +485,7 @@ export default {
         milliseconds: ms,
         level: logEntry.level.toUpperCase(),
         loggerName: logEntry.loggerName,
-        message: logEntry.message
+        message: logEntry.message.replace(/\n/g, '<br>')
```
florian-h05, line 488: I would add some indentation here as well, for example 2 spaces.

wborn:
Hello @openhab/webui-maintainers. Can you make the GHA CI build required instead of the Jenkins CI build? I disabled the PR builds in Jenkins for repos that already have a GHA CI build as a first step to migrate everything to GHA.

florian-h05:
@wborn Done. We have to check if the requirement still works when upgrading the action and its name changes.

bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
cdjackson: I've not looked at a "real" stack trace - doesn't it have tabs (or whitespace of some description) after the break. In theory it should render with some indentation.

While I understand your thinking, I'm a little hesitant to add "random" indentation. In principal, this isn't stack trace specific - it should (IMHO) faithfully represent whatever was received in the logger message - so if the message has indentation, then we should also have some, but if not, we shouldn't add any extra (IMHO).

WDYT?

florian-h05: > I've not looked at a "real" stack trace - doesn't it have tabs (or whitespace of some description) after the break. In theory it should render with some indentation.

I was only looking at your screenshot above and wondered where the indentation was.
I think normally the exception message itself has intendation.
Will check this.

ghys:
I think this will break the "virtual table" from #2905 where every line is expected to be exactly 31 pixels high.

Maybe another way to deal with these multi-line messages is to only display the first line and let the user know there's more. They can then click on the row (or an icon) to reveal the entire message in a popup.

cdjackson:
Interestingly I don't see any issues with this - what do you expect to happen @ghys ? Is it during the removal that lines might get out of sync or something?

cdjackson:
Looking at the code, it seems that the index of the rows might be wrong - but does that really matter? I mean, the user will scroll to the lines they want to see, so this is kind of self compensating - right?

cdjackson:
So what I can notice is when a stack trace comes in, the table only scrolls up 1 line rather than 4 (or whatever). This seems to trigger the detection that we're not at the bottom of the list, and the user needs to scroll down. After this, it works fine.

Personally, I think this is ok. I don't think there should be loads of multi-line messages - although if there were it might get a bit messy.

I'm also ok with @ghys suggestion to only show the first line and then have a popup if that's preferred.

ghys:
> Interestingly I don't see any issues with this - what do you expect to happen @ghys ? Is it during the removal that lines might get out of sync or something?

What I feared is that the scrolling would act up - I didn't test it - but glad to hear it's not so bad.

It still would perhaps have inconsistent scrollbar behaviour because with a fixed row height, a table with  _x_ rows is a predictable 31 * _x_ pixels high and doesn't depend on the height of individual entries, so you can easily calculate the height of the padding empty spaces on either side to maintain that height and the illusion of scrolling through a table with all entries rendered.

florian-h05:
Using the following UI rule and watching the logs, rendering sometimes flickers to a black or white screen:

```yaml
configuration: {}
triggers:
  - id: "1"
    configuration:
      cronExpression: "* * * * * ? *"
    type: timer.GenericCronTrigger
conditions: []
actions:
  - inputs: {}
    id: "2"
    configuration:
      type: application/javascript
      script: console.warn('Hello\nworld\nor\nearth!')
    type: script.ScriptAction
```

florian-h05:
@cdjackson This has been superseded by #3160, correct,

cdjackson:
Yes, that's correct. While this worked ok for me, the concern from @ghys is that it could break, so I think it's probably best to drop this.

florian-h05:
For me it also didn’t work, the table flickered massively.

PR #2968, opened 2025-01-02
Add level icons to log viewer

cdjackson:
Following a discussion on the forum where it was pointed out that the colour coding of the log levels was not suitable for colour blind people, I propose to add an icon representing the log level.

image

The icon is right at the beginning of the line so that it remains in view if the log is scrolled to the right.

I'm open to other icons or thoughts on this...

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
@@ -125,10 +125,8 @@
                  :class="{ 'disabled-link': !stateConnected || !stateProcessing, 'no-margin-left': $device.ios }"
                  @click="loggingPause" />
         <f7-link icon-ios="f7:stop_fill" icon-aurora="f7:stop_fill" icon-md="material:stop_fill"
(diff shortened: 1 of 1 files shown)


ghys:
I think I would do 
image
for warnings (instead of the flag)
and still keep the filled
image
for errors.

Perhaps also
image
for debug (not completely sure, I don't really like it) and
image
for trace.

cdjackson:
I tried the ant icon for debug - that was my first thought, but it looked kinda messy when there's a page full of them.

florian-h05:
I chose to try it with the ant and `line_horizontal_3` for trace and I think it looks good:

image

Even a page full of debug looks okay IMO.

I have pushed these two icon changes, if you are fine with them please let me know and we can merge.

rkoshak:
My only concern, and it's tiny and can be ignored, is that the three horizontal lines usually indicates a "hamburger" menu. Using it in this case shouldn't be confusing, but it's a possibility.

florian-h05:
@cdjackson WDYT about the icons?

cdjackson:
It's good for me - thanks.

ghys (approved):
Then so be it!

PR #2969, opened 2025-01-02
Fix Schedule page shows UTC instead of local timezone

jimtng:
Reported in:
https://community.openhab.org/t/shown-time-in-schedule-view-is-wrong/161288
https://community.openhab.org/t/schedule-page-in-mainui-off-by-one-hour-since-4-3/161307

Needs to be backported to 4.3


florian-h05 (approved):
LGTM, thanks!

florian-h05:
Backported :+1:

Issue #2728, opened 2024-08-23
Semantic model tree closes completely after each change to a variable after saving


vodair:
## The problem
The semantic model tree closes completely after every change after saving. For example, if you want to change the default or widget index, you have to click through the entire tree again after each save.

For example: 

![opera_Neh2k8ftRb](https://github.com/user-attachments/assets/cb371496-5f86-4522-bbea-60ee01f0d049)

## Expected behavior

If you make a change to a variable or other changes (metadata, etc.), the entire tree should remain open

## Steps to reproduce

See attached gif.

## Your environment

OpenHab:
Proxmox LXC (Proxmox 8.2.2)

Linux Distribution:
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian

Tested with:
Windows 10
Opera 112.0.5197.53

Reproduced with
Firefox: 129.0.2 (64-Bit)

Boldfor:
Noticed this as well (though not specific to 4.2.1). Happy tontest as well.

CappuccinoCake:
+1 for me

calle2010:
As a workaround, I expand the whole tree using the button on the upper right and then filter the items I want to see, which usually is possible by part of the name (e. g. "kitchen"). The tree stays open in the expanded state and the filter is still applied.

FROeHlyEisvogel:
Same issue here with openHAB 4.2.2 but also noticed it some versions before.

j-simian:
+1 in 4.3.2. I may give this a go if I find some time as it seems like a small feature (I've not worked with openhab/vue before, but this sounds like a case of just throwing the state of folded/unfolded nodes into localstorage and loading that on refresh)

mherwege:
This should be solved with https://github.com/openhab/openhab-webui/pull/2970, so no need to create a separate PR.

Issue #969, opened 2021-03-22
OpenHab 3 -> Semantic Model: Items shoul appear in every group they are members of


NCC1701M:
## The problem
I have the following model:
```
Shutters
├── FirstFloor
│   ├── Bedroom
│   ├── Livingroom
│   └── Kitchen
└── ControlGroups
    └── DownOnSunset
        ├── Bedroom
        └── Livingroom
```
`FirstFloor`, `ControlGroups` and `DownOnSunset` are groups. As you can see, the shutters `Bedroom` and `Livingroom` are in the groups `FirstFloor` and `DownOnSunset` but in the model the only are display in one of those groups. In which group can change when you modify one of those groups (e.g. by adding a new member).

## Expected behavior
Every item should be displayed in every group they are a member of. Even if this means that the item is more than once in this list.

## Steps to reproduce
1. Create a group
2. Add a item to the group
3. Create a different group
4. Add the same item to the new group

You will see, that the item is only display in one group. But how can you tell, that it is also in the second group? Only by selecting the group in the list, click on the Item on the right side of the screen and take a look at `Direct Group Members`

## Your environment

```yaml
runtimeInfo:
  version: 3.0.1
  buildString: Release Build
locale: de_DE
systemInfo:
  configFolder: /openhab/conf
  userdataFolder: /openhab/userdata
  logFolder: /openhab/userdata/logs
  javaVersion: 11.0.10
  javaVendor: Azul Systems, Inc.
  javaVendorVersion: Zulu11.45+27-CA
  osName: Linux
  osVersion: 5.10.17-v7l+
  osArchitecture: arm
  availableProcessors: 4
  freeMemory: 10072952
  totalMemory: 139206656
bindings:
  - astro
  - homematic
  - hue
  - network
  - openweathermap
  - telegram
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
    pixelRatio: 1
    prefersColorScheme: dark
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: 8
    hardwareConcurrency: 12
    language: de-DE
    languages:
      - de-DE
      - de
      - en-US
[...]

PR #2970, opened 2025-01-03
Model editor: Add drag & drop

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/2728
Closes https://github.com/openhab/openhab-webui/issues/969

Allow drag and drop in the model view:

1. Drag and drop inside the semantic model
2. Move items into the semantic model
3. Move between non-semantic groups (and duplicate)
4. Ask for creation of location, equipment or point if item does not have a semantic class yet when moving into the semantic model

This PR requires further testing, but I do want to already provide it for others to start evaluating, as there is a fair amount of change. I probably have missed some boundary cases of changes in the model that are or are not possible.

Edit: I have tested this on different devices, and from my perspective, it works well. But again, I may have missed something. Therefore I would appreciate others to test.
I have put extensive debug logging into the code to be able to debug potential issues.

Here are a few gifs of what it looks like:

ModelMove1

ModelMove2

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model-mixin.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model-mixin.js
@@ -1,4 +1,5 @@
 import { compareItems } from '@/components/widgets/widget-order'
+import cloneDeep from 'lodash/cloneDeep'
 
 function compareModelItems (o1, o2) {
   return compareItems(o1.item || o1, o2.item || o2)
@@ -22,11 +23,13 @@ export default {
       links: [],
       locations: [],
       rootLocations: [],
-      equipment: {},
+      equipment: [],
       rootEquipment: [],
       rootPoints: [],
       rootGroups: [],
       rootItems: [],
+
+      expandedTreeviewItems: [],
 
       previousSelection: null,
       selectedItem: null
@@ -39,8 +42,10 @@ export default {
      * @returns {Promise<void>}
      */
     loadModel () {
-      if (this.loading) return
+      if (this.loading) return Promise.resolve()
       this.loading = true
+
+      this.saveExpanded()
 
       const items = this.$oh.api.get('/rest/items?staticDataOnly=true&metadata=.+')
       const links = this.$oh.api.get('/rest/links')
@@ -152,6 +157,20 @@ export default {
             item.classList.add('treeview-item-opened')
           } else {
             item.classList.remove('treeview-item-opened')
+          }
+        }
+      })
+    },
+    saveExpanded () {
+      this.expandedTreeviewItems = [...document.querySelectorAll('.treeview-item-opened')]
+    },
+    restoreExpanded () {
+      const treeviewItems = document.querySelectorAll('.treeview-item')
+
+      treeviewItems.forEach(item => {
+        if (item.classList.contains('treeview-item')) {
+          if (this.expanded || this.expandedTreeviewItems.includes(item)) {
+            item.classList.add('treeview-item-opened')
           }
         }
       })
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model-mixin.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model-mixin.js
@@ -130,7 +130,7 @@ export default {
 
       if (this.includeNonSemantic) {
         parent.children.groups = this.items
-          .filter((i) => i.type === 'Group' && (!i.metadata || (i.metadata && !i.metadata.semantics)) && i.groupNames.indexOf(parent.item.name) >= 0)
+          .filter((i) => i.type === 'Group' && !(parent.item.metadata && parent.item.metadata.semantics) && i.groupNames.indexOf(parent.item.name) >= 0)
           .map(this.modelItem).sort(compareModelItems)
         parent.children.groups.forEach(this.getChildren)
         if (parent.item.metadata && parent.item.metadata.semantics) {
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model-mixin.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model-mixin.js
@@ -1,12 +1,11 @@
 import { compareItems } from '@/components/widgets/widget-order'
-import cloneDeep from 'lodash/cloneDeep'
 
 function compareModelItems (o1, o2) {
   return compareItems(o1.item || o1, o2.item || o2)
 }
 
 /**
- * Mixin for model page and model picker popup.
+ * Mixin for the model page and model picker popup.
  *
  * The component using this mixin has to provide the following methods:
  * - `selectItem(item)`: Called when an item is selected.
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
@@ -73,7 +73,7 @@
           <f7-block v-show="!empty" strong class="semantic-tree" no-gap @click.native="clearSelection">
             <model-treeview :rootNodes="[rootLocations, rootEquipment, rootPoints, rootGroups, rootItems].flat()"
                             :includeItemName="includeItemName" :includeItemTags="includeItemTags" :canDragDrop="true"
-                            @selected="selectItem" :selected="selectedItem" @reload="load"/>
+                            @selected="selectItem" :selected="selectedItem" @reload="load" />
           </f7-block>
         </f7-col>
         <f7-col width="100" medium="50" class="details-pane">
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
@@ -73,7 +73,7 @@
           <f7-block v-show="!empty" strong class="semantic-tree" no-gap @click.native="clearSelection">
             <model-treeview :rootNodes="[rootLocations, rootEquipment, rootPoints, rootGroups, rootItems].flat()"
                             :includeItemName="includeItemName" :includeItemTags="includeItemTags" :canDragDrop="true"
-                            @selected="selectItem" :selected="selectedItem" />
+                            @selected="selectItem" :selected="selectedItem" @reload="load"/>
           </f7-block>
         </f7-col>
         <f7-col width="100" medium="50" class="details-pane">
@@ -378,6 +378,7 @@ export default {
     clearSelection (ev) {
       if (ev.target && ev.currentTarget && ev.target === ev.currentTarget) {
         this.selectedItem = null
+        this.detailsOpened = false
       }
     },
     toggleNonSemantic () {
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
@@ -71,13 +71,9 @@
           </f7-block>
 
           <f7-block v-show="!empty" strong class="semantic-tree" no-gap @click.native="clearSelection">
-            <!-- <empty-state-placeholder v-if="empty" icon="list_bullet_indent" title="model.title" text="model.text" /> -->
-            <f7-treeview>
-              <model-treeview-item v-for="node in [rootLocations, rootEquipment, rootPoints, rootGroups, rootItems].flat()"
-                                   :key="node.item.name" :model="node"
-                                   :includeItemName="includeItemName" :includeItemTags="includeItemTags"
-                                   @selected="selectItem" :selected="selectedItem" />
-            </f7-treeview>
+            <model-treeview :rootNodes="[rootLocations, rootEquipment, rootPoints, rootGroups, rootItems].flat()"
+                            :includeItemName="includeItemName" :includeItemTags="includeItemTags" :canDragDrop="true"
+                            @selected="selectItem" :selected="selectedItem" />
           </f7-block>
         </f7-col>
         <f7-col width="100" medium="50" class="details-pane">
@@ -177,15 +173,6 @@
 .semantic-tree
   padding 0
   border-right 1px solid var(--f7-block-strong-border-color)
-  .treeview
-    --f7-treeview-item-height 40px
-    .treeview-item-label
-      font-size 10pt
-      white-space nowrap
-      overflow-x hidden
-    .semantic-class
-      font-size 8pt
-      color var(--f7-list-item-footer-text-color)
 .model-details-sheet
   .toolbar
     --f7-theme-color var(--f7-color-blue)
@@ -234,6 +221,7 @@
 
 <script>
 import ModelDetailsPane from '@/components/model/details-pane.vue'
+import ModelTreeview from '@/components/model/model-treeview.vue'
 import AddFromThing from './add-from-thing.vue'
 import AddFromTemplate from './add-from-template.vue'
 
@@ -241,7 +229,6 @@ import ItemStatePreview from '@/components/item/item-state-preview.vue'
 import ItemDetails from '@/components/model/item-details.vue'
 import MetadataMenu from '@/components/item/metadata/item-metadata-menu.vue'
 import LinkDetails from '@/components/model/link-details.vue'
-import ModelTreeviewItem from '@/components/model/treeview-item.vue'
 
 import ModelMixin from '@/pages/settings/model/model-mixin'
 
@@ -250,11 +237,11 @@ export default {
   components: {
     'empty-state-placeholder': () => import('@/components/empty-state-placeholder.vue'),
     ModelDetailsPane,
+    ModelTreeview,
     ItemStatePreview,
     ItemDetails,
     MetadataMenu,
-    LinkDetails,
-    ModelTreeviewItem
+    LinkDetails
   },
   data () {
     if (!this.$f7.data.model) this.$f7.data.model = {}
@@ -336,7 +323,7 @@ export default {
             this.$refs.searchbar.f7Searchbar.$inputEl[0].focus()
           }
           this.$refs.searchbar?.f7Searchbar.search(this.$f7.data.lastModelSearchQuery || '')
-          this.applyExpandedOption()
+          this.restoreExpanded()
         })
         if (!this.eventSource) this.startEventSource()
       })
--- a/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
@@ -35,7 +35,7 @@
       </div>
       <f7-link class="right details-link padding-right" ref="detailsLink" @click="detailsOpened = true" icon-f7="chevron_up" />
     </f7-toolbar>
-    <f7-toolbar v-else bottom class="toolbar-details" style="height: calc(50px + var(--f7-safe-area-bottom))">
+    <f7-toolbar v-else bottom class="toolbar-details">
       <f7-link :disabled="selectedItem != null" class="left" @click="selectedItem = null">
         Clear
       </f7-link>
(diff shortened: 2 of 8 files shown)


digitaldan:
That's super cool.

ghys:
Indeed and I can't imagine openHAB 5 without it :)

ghys:
First comment from watching the GIFs alone would be to make sure there's a confirm message for every drag and drop as it might be too easy to screw up and make an involuntary change. This could be an additional dialog or a Cancel button to the existing ones - like the one asking whether to ask an Equipment or a Point.

mherwege:
> First comment from watching the GIFs alone would be to make sure there's a confirm message for every drag and drop as it might be too easy to screw up and make an involuntary change. This could be an additional dialog or a Cancel button to the existing ones - like the one asking whether to add an Equipment or a Point.

I have added some confirmation dialogs.

florian-h05:
@ghys Do you want to review?

ghys:
@florian-h05 yeah if you don't mind, this is personal for me (like the log viewer) since I always wanted to do it but never got around to it. So I'm glad it got implemented by @mherwege and I'll take care of this one and get it done.
I think you have plenty on your plate already :)

ghys:
@mherwege I'm testing it and unfortunately I do have a major remark which IMO is a must fix...

The tree view now adds arrows/chevrons besides items even if they don't have children, it's super confusing:

Before:
image

After:
image

This leads to unnecessary clicks on them because you'd think there's something to expand when there's not.

Actually it's the case for the sitemap tree view now too (I missed it).

I think it should be fixed because it leads to confusion for example:

image

Since the Text widget here is an "expanded empty group" we could easily think the Switch is one of its children, when it's actually on the same level as a child of the Frame.

image

It is possible to at least show those after you've started the drag operation only (so you can insert into an empty group)?

florian-h05:
> Actually it's the case for the sitemap tree view now too (I missed it).

Fixed in #3007.

mherwege:
> @mherwege I'm testing it and unfortunately I do have a major remark which IMO is a must fix...
> 
> The tree view now adds arrows/chevrons besides items even if they don't have children, it's super confusing:

Yes, fair enough. And it looks like @florian-h05 already found a fix for it in the sitemap editor. I will look into doing something similar here, but need to find some time for doing it. It doesn't look to be a major change, so should be easy enough to include without changing everything. I don't think it should impact other review comments you may have.

mherwege:
I created a fix in this PR.

florian-h05:
@ghys Can you please have another look at this?

jimtng:
@mherwege I tried checking out this branch and testing it out but I can't seem to make it do drag/drop :(

mherwege:
@jimtng I am not sure what the issue is. I can only imagine something went wrong in the rebase. I can’t try it out now, but will soon.

How did you test? I tend to have a running OH version and then run mainUI from the development environment on port 8081. That should automatically redirect and use the modified code.

jimtng:
I know I'm looking at the correct version. I checked out your branch, and ran my usual `npm start` for the development version of the webui which I accessed using the special port.

I added console.log in the vue/js code to make sure it is the version I'm looking at. I logged the `canDragDrop` variable/prop, and even removed the check for `editable` to make sure that my file-based stuff isn't causing the problem.

image

I'll keep looking. There must be something that I've missed.

mherwege:
@jimtng I just checked out my branch again and tried it on a snapshot from last week. It functions for me.
One thing I noticed, if you move with the mouse, don't go to fast. If you click and move immediately, it may not pick up the event. You need to have a slight pause before dragging. The reason is that touch scrolling on a longer list would get in trouble otherwise. When you try to scroll, it would move things.

jimtng:
> One thing I noticed, if you move with the mouse, don't go to fast. If you click and move immediately, it may not pick up the event. You need to have a slight pause before dragging.

BINGO!!! This was the problem indeed! Once I've paused before dragging, it worked. Funny I tried and tried before and never could get it to work.

jimtng:
I've just tried to move a point into an equipment, and got an error

"Cannot move semantic item "xxxxx" from non-semantic group "model root" into semantic group "yyyyy"

After this, I can no longer drag/drop. I had to refresh for the drag drop to start again.

mherwege:
> "Cannot move semantic item "xxxxx" from non-semantic group "model root" into semantic group "yyyyy"

I can reproduce that, will look to fix it.

> After this, I can no longer drag/drop. I had to refresh for the drag drop to start again.

But this didn't happen for me. I could just drag/drop again after closing the warning message.

mherwege:
> I might sound like an incompetent noob who doesn't know how to use the mouse or how to do drag and drop, and honestly I feel that way when using this model drag/drop. I have been using GUI / drag drop just as much as the next person. I think some tweaking is needed here to make it feel more natural.

No, not at all. Try again if it feels better now. I tweaked the parameters.

jimtng:
> But it only goes one level up in its check

I think only one level is enough. In my version, it doesn't care if it belongs to the same location or different location.

jimtng:
> Try again if it feels better now. I tweaked the parameters.

Thanks! Drag Start works more naturally now. However I still see problems once it's released. It takes a while - one second perhaps before it displays a confirmation dialog, and after clicking OK, more delays are happening and things seem to become sluggish after that. Furthermore, starting another drag becomes difficult again. A refresh fixes the problem.

mherwege:
> Thanks! Drag Start works more naturally now. However I still see problems once it's released. It takes a while - one second perhaps before it displays a confirmation dialog, and after clicking OK, more delays are happening and things seem to become sluggish after that. Furthermore, starting another drag becomes difficult again. A refresh fixes the problem.

What browser are you using? I didn't notice that so far. I am using Firefox mainly for testing, but also do some testing on Chrome.

mherwege:
I tried again with Chrome and I don't see any slugishness.

Could you try the sitemap editor drag/drop as well? Do you see the same slugishness there?

jimtng:
> Could you try the sitemap editor drag/drop as well? Do you see the same slugishness there?

I just did. No sluggishness there, but I noticed that when I dropped something randomly onto another random thing, that thing would disappear on me. Seems like a bug, but that's for another topic.

I had another play with firefox vs chrome. I somehow ended up with this (same equipment appearing twice)

image

And after this, I noticed my point disappeared, and then I can no longer refresh my model page. I suspect there's a circular reference causing an infinite recursion:

```
Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at Be (app.836faffdf20d01332dd4.js:2:1412361)
    at new e (app.836faffdf20d01332dd4.js:2:1412032)
    at Me (app.836faffdf20d01332dd4.js:2:1412336)
    at Be (app.836faffdf20d01332dd4.js:2:1412568)
    at new e (app.836faffdf20d01332dd4.js:2:1412032)
    at Me (app.836faffdf20d01332dd4.js:2:1412336)
    at e.observeArray (app.836faffdf20d01332dd4.js:2:1412136)
    at new e (app.836faffdf20d01332dd4.js:2:1411960)
    at Me (app.836faffdf20d01332dd4.js:2:1412336)
    at Object.set [as equipment] (app.836faffdf20d01332dd4.js:2:1412907)Understand this error
```

Sorry this is a different issue separate from the sluggishness.

mherwege:
> And after this, I noticed my point disappeared, and then I can no longer refresh my model page. I suspect there's a circular reference causing an infinite recursion:

Yes, looks like a bug. I will build in a check.

mherwege:
> testeq1 was set to become its own parent, and a point somehow became the parent too

Whatever I tried, I never succeeded dropping something onto itself. However, I built in a check to abort the operation if it does happen.

mherwege:
When you test, could you run the application in debug mode. The error messages would be much more meaningful.

Could the circular reference be caused by testing with 2 browsers at the same time? If one completes a drag/drop, the items will be refreshed and events will be pushed to the second browser that may still be in the middle of a drag/drop operation.

mherwege:
> I just did. No sluggishness there, but I noticed that when I dropped something randomly onto another random thing, that thing would disappear on me. Seems like a bug, but that's for another topic.

Another topic indeed, but did you not see a 'Move invalid' banner at the bottom? If the move is invalid, it would keep it in the original position. There is no dialog to click away though, but the question is what is most appropriate. The model rules are so complex that a dialog is in place. But it also slowes down the flow.

jimtng:
Both Firefox and Chrome. When I dropped / released the mouse, there's a considerable delay >1s before the confirmation dialog opens. Then whichever button I clicked (Cancel / OK), there's some further delay before the dialog goes away.

Also the confirmation dialog won't go away if I clicked cancel on this:
image

mherwege:
> Both Firefox and Chrome. When I dropped / released the mouse, there's a considerable delay >3s before the confirmation dialog opens. Then whichever button I clicked (Cancel / OK), there's some further delay before the dialog goes away.

How large is the model you are testing this on? The checks do a considerable amount of work filtering the model. And as a change is saved, the model will be requeried. Also the debug logs clone the objects to easily see them in the debugger, but that should not have an impact when not having the development tools open.

mherwege:
This may be one reason for the delay when checking:

```
      if (node.item.type === 'Group' && node.class === '') {
        const semanticNode = this.nestedNodes(node).find((n) => n.class !== '')
        if (semanticNode) {
          const message = 'Cannot insert non-semantic group "' + this.itemLabel(node.item) +
            '" with semantic child "' + this.itemLabel(semanticNode.item) +
            '" into semantic group "' + this.itemLabel(parentNode.item) + '"'
          console.debug('Add rejected: ' + message)
          this.$f7.dialog.alert(message).open()
          this.restoreModelUpdate()
          return
        }
      }
```

And `nestedNodes` is defined as:

```
    nestedNodes (node, nodes) {
      nodes = nodes || []
      const children = [...node.children.locations, ...node.children.equipment, ...node.children.points, ...node.children.groups, ...node.children.items]
      nodes.push(...children)
      children.forEach((child) => this.nestedNodes(child, nodes))
      return nodes
    }
```

It goes through all the children in all child levels to see if there is a semantic child, because it can't insert a non-semantic group in a semantic group then. It's this kind of logic that could potentially be time consuming. And there are probably other examples like this in the code. This is just one example.
This one could potentially be optimized by not building the full list, but getting out when it finds the first one. But I first would want to be sure if this is the real issue before investing time in that.

jimtng:
> it can't insert a non-semantic group in a semantic group then

In my case my non-semantic tags weren't shown, i.e. "Show non-semantic" checkbox isn't checked.

mherwege:
I tried some performance improvements:

- It will not build the full nested children list anymore.
- The list of children is now a computed property, reducing how often it gets rebuild.
- I keep a list of nested children that are updated with the parent to limit the number of item save REST calls (e.g. needed when an item becomes a location, so its children, and its children's children..., also need to become semantic).

mherwege:
> In my case my non-semantic tags weren't shown, i.e. "Show non-semantic" checkbox isn't checked.

So, you should indeed not be hitting the code as shown above.

It would be helpful to do these tests with debug info, so it gets easier to analyze.

jimtng:
I've noticed your new commit and I've tested it. There's a very slight improvement, but overall still slow. I am suspecting (my apologies I still haven't looked at the code) that it's slow because you might be resetting / reloading the whole thing:
- when cancel is clicked
- after drop (when OK is clicked)

jimtng:
I'm wondering why _you_ are not seeing the slowness. Are you testing it against an adequate number of items in your semantic model, i.e. how an "average/advanced user" might have? I've just counted mine, I have 443 items/groups with semantic tag of some sort and a total of 1112 items including the non-semantic ones.

mherwege:
> it's slow because you might be resetting / reloading the whole thing

This is indeed the case. I have relied on the model editor as it was before this addition of drag/drop. You were doing edits through the properties window then, and every change save would also trigger a reload of the model. I have not fundamentally touched this logic. So the slowness would already have been there. It is obviously more apparent with drag and drop as many items may be impacted at the same time, and every time something is saved, or the original has to be restored, it rebuilds.

As for cancel, while it would rebuild, it should use the REST API cache, so there would not be traffic with new data. The performance impact with cancel should be much less.

> Are you testing it against an adequate number of items in your semantic model, i.e. how an "average/advanced user" might have?

My test have been against smaller test data sets. And that may explain the performance issue. The items are also all managed items.

mherwege:
I tested against around 350 items now. And I indeed do see delays now:
- delay before opening a dialog: there is no REST call or model rebuild involved here. So all improvements would have to be in the drag/drop code.
- delay on cancel or save: there are REST calls involved, but the delay is not fundamentally longer than the delay before the dialog opens. So the question remains if it is because the model is being rebuilt.

I imagine to improve performance on this, it will be a major refactoring of the model tree as a whole (not just drag/drop). I am not sure if I am capable of doing that.

jimtng:
One suggestion, unrelated to the delay issue: While dragging, please allow user to press ESC to cancel the drag.

Also this doesn't make sense to me:

image

If you moved the item to model root, how can you also keep it in the original group? An item that belongs in the "root", by definition, has no parent group that it belongs to, because there is no "root" group to belong to, so it can only be achieved by having no parent groups.

mherwege:
> One suggestion, unrelated to the delay issue: While dragging, please allow user to press ESC to cancel the drag.

ESC should now work on all of the confirmation dialogs. And there should always be at least one dialog. Or do you mean something else?


> If you moved the item to model root, how can you also keep it in the original group?

Thanks for the catch. I will correct that.

jimtng:
Not the dialogue. I meant while you're still dragging around, before any dialog pops up, while mouse is still down. Press ESC to cancel drag mode so when mouse is released, no dialogs will pop up.

mherwege:
@jimtng I didn't find a way to make the ESC key work while dragging so far.

mherwege:
> * For it to be a copy process, the user needs to press Cmd while dragging. This should change the cursor to a +

That would make it impossible to do on touch devices.
I tried to make sure to map the enter key to the most logical option. And as much as possible, there should only be one dialog.

There are multiple issues about cancel on the Vue sortable github. The main readme has broken links to examples. I will keep on looking, but I am not sure I will succeed.

jimtng:
> There are multiple issues about cancel on the Vue sortable github. The main readme has broken links to examples. I will keep on looking, but I am not sure I will succeed.

It isn't super important. Just a nice-to-have.

The performance / delay has improved a lot, although still not ideal.

jimtng:
It seems that just clicking around (selecting an item in the model tree) without dragging is significantly slower (almost feels like non responsive) compared to normal (outside the PR).

mherwege:
@jimtng I think ESC now works.
I am still struggling with the performance aspect. The code is now full of debug logging. At logs timings from the first detection of a drag start. It looks like the time spent inside the code I developed is limited, and most time is spent before the first detection of a drag start and after finishing my part of the code, hence inside the Sortable.js code is suppose. I don't know what I can do about that. It does look like the performance is a lot worse if the structure is deeply nested.

jimtng:
> @jimtng Can you check if the latest is more performant for you? I believe I improved it somewhat (without being perfect).

There seem to be a _significant_ improvement in certain parts of the operation.
However, it seems that there are other parts of the operation that are still "strange" and slow. So what you said about it being more performant but not perfect is accurate.

jimtng:
I could somehow end up in this weird state, the only way to recover is refresh

image

The TestPoint1 that I was dragging, somehow ended up "froze" on the screen like that even after I've released the mouse button. Unfortunately I don't know how to reliably cause this situation.

jimtng:
This looks a bit strange

image

I'm dragging an equipment so this message shouldn't have showed up.

jimtng:
> I just need to make sure I then don't loose anything if it was manually set and you start moving things around.

Maybe just do a warning if a point was originally in both equipment + location and it was being dragged to root.
- If it's dragged onto another equipment, we could just move the equipment, retaining the location?
- If it's dragged to another location, just move the location, retaining the equipment

But if a point only belongs to _either_ equipment or location but not both, then move them as usual without retaining the original parent nor prompting about it.

mherwege:
> Maybe just do a warning if a point was originally in both equipment + location
> 
> But if a point only belongs to _either_ equipment or location but not both, then move them as usual without retaining the original parent nor prompting about it.

I implemented this suggestion.

mherwege:
> It looks like at some point, a drag start was (mistakenly) triggered. I wasn't intentionally trying to drag at all in here. But that dragstart might've been the trigger that caused the inability to select any other items in the model.

Yes, I noticed this as well. I had put in some logic not to open the details when dragging, as this inadvertently happened without this, but missed the case when there is no move in the end. I fixed that.

jimtng:
With the latest update, I no longer experience "frozen" selection. But sometimes I still noticed "glitchy" item as I was clicking around (selecting different items). It looked like the item was about to be dragged, so it moved a bit, even though I wasn't trying to drag. 

Maybe it's a bit too sensitive? Is there some sort of "mouse down and move a minimum of X pixels before it's considered a dragstart"? I'm not sure if this is the cause, so I'm just making a guess here.

mherwege:
> Maybe it's a bit too sensitive? Is there some sort of "mouse down and move a minimum of X pixels before it's considered a dragstart"? I'm not sure if this is the cause, so I'm just making a guess here.

That's indeed the reason. I just changed it to have some move threshold.

jimtng:
> > Maybe it's a bit too sensitive? Is there some sort of "mouse down and move a minimum of X pixels before it's considered a dragstart"? I'm not sure if this is the cause, so I'm just making a guess here.
> 
> That's indeed the reason. I just changed it to have some move threshold.

This seems fine now.

I found another issue:

image

This should have never been allowed. It should just be ignored, and the mouse cursor changed to a circle with diagonal line  `not-allowed`

image

I think it doesn't currently allow it to be dropped into a read-only (file-based) semantic groups, so perhaps just use that same mechanism (instead of changing the cursor)

mherwege:
> I think it doesn't currently allow it to be dropped into a read-only (file-based) semantic groups, so perhaps just use that same mechanism (instead of changing the cursor)

An editable item should be allowed to be dropped in a non-editable semantic group, as the changes will only need to be persisted for the item, not the group you drop into. But I need to think this through in more detail, as I have not spent much time on this corner case.

> This should have never been allowed. It should just be ignored, and the mouse cursor changed to a circle with diagonal line `not-allowed`

I tried to make this more dynamic at one point, but failed, so settled for a warning. I can try again. A warning is still better than silently reverting in my view.

florian-h05:
@ghys Do you still want to review this PR or should I take over?

mherwege:
> Question: I am running 4.x on the backend and pulled the PR. Is is supposed to work with 4.x as well as the backend or is the logic only in the frontend? The reason I am asking is that it doesn't work for me at the moment.

All of the logic for DnD is in the frontend webui code, but every change (drop) will immediately update the model through the REST api. That is also what happens without this PR. I never tested on a OH 4 backend. I didn’t think there are REST API changes that would impact this, but I can also not be sure.

What browser are you using? If you are running this with the browser development tools, do you see something in the browser console?

stefan-hoehn:
> What browser (and browser version) are you using? If you are running this with the browser development tools, do you see something in the browser console?

Sorry, Mark, I have been occupied a lot in the last weeks and will be until mid of June, so I overlooked this.
I am using chrome on MacOs.

I would really love this feature go forward. Would it make sense to have a remote session to investigate that on my machine?

bundles/org.openhab.ui/web/src/pages/settings/model/model.vue
```diff
@@ -35,7 +35,7 @@
       </div>
       <f7-link class="right details-link padding-right" ref="detailsLink" @click="detailsOpened = true" icon-f7="chevron_up" />
     </f7-toolbar>
-    <f7-toolbar v-else bottom class="toolbar-details" style="height: calc(50px + var(--f7-safe-area-bottom))">
```
florian-h05, line 38: To check: Does this revert the fix from #2861?

bundles/org.openhab.ui/web/src/pages/settings/model/model-mixin.js
```diff
@@ -1,4 +1,5 @@
 import { compareItems } from '@/components/widgets/widget-order'
+import cloneDeep from 'lodash/cloneDeep'
```
florian-h05: I don't think this import is needed here.

florian-h05: Removed it myself :+1:

florian-h05 (approved):
LGTM, thanks! I have tested this on desktop and on iPhone, fantastic work!

PR #2975, opened 2025-01-04
[cometvisu] Upgrade openapi generator to 7.0.1

holgerfriedrich:
* Upgrade openapi-generator-maven-plugin from 4.3.0 to 7.0.1

This is a bit complicated, since the code generator changed.
Our yaml file did no longer work, as genrated classes in `src-gen` did no longer extend from the proper base classes.

With the tweaks it seems to generate as expected, the output looks very similar (though not identical).
Hopefully, this does not break anything.

btw: Could not go beyond 7.0.1 due to OpenAPITools/openapi-generator#18596. Did not find out how to circumvent this problem.

### DIFF
--- a/bundles/org.openhab.ui.cometvisu/src/main/resources/openapi.yaml
+++ b/bundles/org.openhab.ui.cometvisu/src/main/resources/openapi.yaml
@@ -238,7 +238,7 @@ paths:
           content:
             application/json:
               schema:
-                oneOf:
+                allOf:
                   - $ref: '#/components/schemas/HiddenConfig'
                   - $ref: '#/components/schemas/ConfigSection'
                   - $ref: '#/components/schemas/ConfigOptionValue'
@@ -477,7 +477,7 @@ components:
       additionalProperties:
         type: string
     ReadResponse:
-      oneOf:
+      allOf:
         - $ref: '#/components/schemas/DirContent'
     DirContent:
       type: array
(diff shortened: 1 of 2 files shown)


florian-h05:
@peuter Please have a look.

peuter:
Hm, I'm not sure that this will work. Basically this yaml file is just a copy of https://github.com/CometVisu/CometVisu/blob/develop/source/rest/openapi.yaml

So both would need to be changed to keep them in sync. At least the first change from oneOf to allOf looks problematic, because it should only be oneOf the three options in the response. I need to check if the change breaks something. Give me some time to check this.

bundles/org.openhab.ui.cometvisu/src/main/resources/openapi.yaml
```diff
@@ -238,7 +238,7 @@ paths:
           content:
             application/json:
               schema:
-                oneOf:
+                allOf:
```
peuter, line 241: Is there a special reason for this change or did you just replace all oneOfs? I don't see that this change is needed, my local build works without it. If you revert this change, I would approve this PR.

peuter (changes requested):

bundles/org.openhab.ui.cometvisu/src/main/resources/openapi.yaml
holgerfriedrich: Great, thanks. I think I just replaced all of them. Actually this gave me some headache when I updated the generator.

You are right, it generates fine even without this. Reverted.

peuter (approved):

holgerfriedrich:
@florian-h05 would you take care of the merge?

PR #2976, opened 2025-01-04
WS client: Remove trailing `=` from base64-encoded access token

florian-h05:
Follow-up for #2907.

Fixes WS client broken due to `=` being part of the base74-encoded token, which is no allowed WS subprotocol value.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/ws.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/ws.js
@@ -10,8 +10,9 @@ const HEARTBEAT_MESSAGE = `{
 const openWSConnections = []
 
 function newWSConnection (path, messageCallback, readyCallback, errorCallback, heartbeatCallback, heartbeatInterval) {
+  const encodedToken = btoa(getAccessToken()).replace(/=*$/, '')
   // Create a new WebSocket connection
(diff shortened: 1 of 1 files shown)


ghys:
Thanks!

It seems to me though that OH Cloud doesn't yet have support for WebSockets;

https://community.openhab.org/t/openhab-4-3-release-discussion/160888/14?u=ysc

Until it has, I would consider it a showstopper for investing in them.

cc @cdjackson @digitaldan

florian-h05:
There is a PR to add that: https://github.com/openhab/openhab-cloud/pull/447

florian-h05:
@ghys I would like to replace SSE step-by-step with WS to get rid of the max parallel HTTP connections limitation. As my plan was to provide a mechanism to let the user switch back to SSE if having trouble with WS, I would suggest to have the UI automatically use SSE for openHAB Cloud, in else WS where implemented.

I would start with the UI‘s global SSE connection used if Web Audio or UI Command Item are enabled.

WDYT?

ghys:
@florian-h05 this is great, and I think the plan is perfectly reasonable and ultimately necessary.

I hope WS connections would eventually work over OH Cloud as well (or that we could have full-featured http/2) but it's a good idea to start with this SSE connection dedicated to web audio and UI commands, because it's relatively low-impact (it doesn't even happen if you don't have either of these rather obscure options enabled) and it's actually unlikely to be used non-locally anyway.
This allows to test the whole new thing without much impact and then use that experience to gradually replace the others.

PR #2980, opened 2025-01-04
WS client: Fix connection fails when using dev server

florian-h05:
Webpack Dev Server/http-proxy-middleware cannot handle our own pseudo websocket protocols used to send the access token.


florian-h05:
@ghys FYI

florian-h05:
Okay seems this was my fault and something was wrong with my dev installation of openHAB.

PR #2982, opened 2025-01-05
Tag Input: Fix automatic add on blur

jimtng:
User inadvertently losing the entered new tag when they haven't realised that they needed to press Enter. It seems that onblur was originally intended to auto-add the typed tag, e.g. when user immediately clicked "Save" (top right), but it was attached to the wrong element, thus the event didn't fire.

I've also changed keypress to keyup event because the keypress is missing the last letter otherwise. It wasn't a problem in the past because it only worked with Enter key (being the last keypress). Missing the last letter becomes a problem when adding it in the onblur event. So keyup solved that.

Reported here: https://community.openhab.org/t/how-to-reuse-code-between-ui-rules/161028/35

I also removed `@keyPressed.native="keyPressed"` but TBH I'm not sure why it was needed.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/tags/tag-input.vue
+++ b/bundles/org.openhab.ui/web/src/components/tags/tag-input.vue
@@ -17,11 +17,9 @@
         placeholder="Add tag"
         :value="pendingTag"
         @input="pendingTag = $event.target.value"
-        @blur="addTag()"
-        @keyPressed.native="keyPressed"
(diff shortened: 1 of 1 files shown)


florian-h05:
Can you please resolve the conflicts?

jimtng:
Rebased

florian-h05 (approved):
LGTM, thanks!

PR #2983, opened 2025-01-05
Tag Input: Prevent user from adding a semantic tag

jimtng:
When the tag input is supposed to filter out semantic tags, also prevent user from entering them.

It is up to the caller of tag-input form to determine whether it can accept semantic tags or not, by specifying it in the `showSemanticTags` property.

image


florian-h05 (approved):
LGTM, thanks - this is really helpful!

PR #2989, opened 2025-01-07
Fix JavaDoc output directory

holgerfriedrich:
Regression, fix issue introduced by #2972.
JavaDoc output dir changed from target/site to target/reports.
This will get it back to the known location.


florian-h05 (approved):
LGTM, thanks.

PR #2998, opened 2025-01-09
Rule edit: Add descriptions to triggers, actions & conditions sections

florian-h05:
Closes https://github.com/openhab/openhab-core/issues/4503.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
@@ -99,9 +99,14 @@
               </f7-button>
             </div>
             <div v-for="section in ['triggers', 'actions', 'conditions']" :key="section">
-              <f7-block-title medium style="margin-bottom: var(--f7-list-margin-vertical)" v-if="isEditable || rule[section].length > 0">
-                {{ SECTION_LABELS[section][0] }}
-              </f7-block-title>
+              <template v-if="isEditable || rule[section].length > 0">
+                <f7-block-title medium>
+                  {{ SECTION_LABELS[section][0] }}
+                </f7-block-title>
+                <f7-block-footer class="no-margin-top margin-horizontal" style="margin-bottom: var(--f7-list-margin-vertical)">
+                  {{ SECTION_LABELS[section][1] }}
+                </f7-block-footer>
+              </template>
               <f7-list sortable swipeout media-list @sortable:sort="(ev) => reorderModule(ev, section)">
                 <f7-list-item media
                               :title="mod.label || suggestedModuleTitle(mod, null, section)"
@@ -118,7 +123,7 @@
                 </f7-list-item>
               </f7-list>
               <f7-list v-if="isEditable">
-                <f7-list-item link no-chevron media-item :color="($theme.dark) ? 'black' : 'white'" :subtitle="SECTION_LABELS[section][1]" @click="addModule(section)">
+                <f7-list-item link no-chevron media-item :color="($theme.dark) ? 'black' : 'white'" :subtitle="SECTION_LABELS[section][2]" @click="addModule(section)">
                   <f7-icon slot="media" color="green" aurora="f7:plus_circle_fill" ios="f7:plus_circle_fill" md="material:control_point" />
                 </f7-list-item>
                 <!-- <f7-list-button :color="(showModuleControls) ? 'gray' : 'blue'" :title="sectionLabels[section][1]"></f7-list-button> -->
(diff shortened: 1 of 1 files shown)


ghys:
Why not at first, but after once you've read them they're probably going to be uselessly taking up valuable real estate...?

And I would argue rules are pretty used and straightforward to understand, compared to for instance:

image

florian-h05:
See https://github.com/openhab/openhab-core/issues/4503, rules seem to be not as well understood and as straight forward … what about adding a short header to the rules editor with a link to the docs?

IMG_5572

ghys:
> See https://github.com/openhab/openhab-core/issues/4503, rules seem to be not as well understood and as straight forward

Right, I missed this also (@rkoshak's comment https://github.com/openhab/openhab-core/issues/4503#issuecomment-2565718432).

It's fair enough but I stand by my comment that once you get the gist of rules, you want the most concise page that you can get (maybe it's just me) and everything getting in the way of that is just clutter.

I know that I insisted to have a description of persistence strategies on their page but I think those are seldom accessed compared to rules, and much more obscure, so they deserve some text to describe them.

florian-h05:
Hmm agreed.
I have just checked what docs are shown when the help icon in the upper right corner is clicked, and there is a link to the rules concepts page that explains the parts of a rule - so let's not clutter the rules page here, instead expect new users to read the docs.

rkoshak:
I can't say that I agree. 

First of all, the space for a description is there whether it's used or not. So by not having a description we are not saving any space, we just don't have words between the header and the field. The rules page takes up just as much space on the screen, it's just less effectively utilized.

Secondly, one of the biggest complaints from users is that OH "hides" too much, making reading the docs the only way to learn something which could be made more obvious. This comment makes it appear that's by design. Why do we have descriptions anywhere at all? The UI can become very concise and terse if the only place to learn what anything does is in the docs.

Thirdly, no one reads the docs. This is more true of those who have used OH for awhile. Because conditions are something that do not exist for for based rules and because they are put last in the list, many just skip over it.

Just in the past year I've had to explain "you cannot use comparisons in the trigger, that's what considering are for" many times, even to *very* experienced users who should have known better. Sometimes even more experienced users are all "OH yeah, I forgot about conditions."

Adding the descriptions is a low effort low impact way to improve the situation. 

Another improvement would be to put conditions second, not third as that is how many of not most users think about their rules. 

Regardless, I've had my say and ain't say any more. But when it comes to again on the forum my response will have to be "the maintainers have determined you should read the docs.".

florian-h05:
I have now optimised the styling a bit, and now the conditions only take up one list entry together, I think it is reasonable to allow them to do so.
Let's add those descriptions and hope they help.

florian-h05:
New PR: #3008

ghys:
> Regardless, I've had my say and ain't say any more. But when it comes to again on the forum my response will have to be "the maintainers have determined you should read the docs.".

@rkoshak I can feel that you reacted quite strongly - I know because I've reacted quite strongly to other things I cared about before.

So just know that I completely get it. I've read your arguments carefully and I think they are sound and perhaps we can do something? There's no need to beef :hugs:

rkoshak:
No beef and I didn't mean it to come across quite so strongly. But I am disappointed and have concerns with the reasoning. If "we don't need to show it because it's in the docs"  is taken to its extreme we end up with a very compact, very pretty, and almost unusable UI. 

But I also recognize as a non-maintainer I can do little more than complain and this is not a topic I intend to keep coming back to. To continue to complain when the maintainers have given their opinion is annoying and counterproductive.

I'm not sure how the new PR will look on the screen (I don't know how to read the code and convert that to how it's rendered that well) but it sounds like it's good.

florian-h05:
> If "we don't need to show it because it's in the docs" is taken to its extreme we end up with a very compact, very pretty, and almost unusable UI.

Well, parts of the UI have descriptions added by the maintainers (e.g. persistence page) because we found it necessary as persistence config is no „daily business“ and needs some description.
For rules, I would have thought it is clear for what triggers, actions and conditions are, but I have to admit I am not very active in the forum wrt to generic questions, mainly focused on specific topics, e.g. JS Scripting. Your reports suggest this assumption is wrong.

> I'm not sure how the new PR will look on the screen (I don't know how to read the code and convert that to how it's rendered that well) but it sounds like it's good.

Basically it is the same as this PR, just with less margin between the header and the description to save as much space without looking weird — I wanted to reopen this PR but GitHub didn‘t allow me to do so because I force-pushed, so I had to open a new PR.

I would say let‘s merge my new PR, the descriptions take up one list entry of space which is like nothing as long as you are not on a mobile phone.

rkoshak:
> Your reports suggest this assumption is wrong.

I can't figure it out myself why it's so much of a surprise to people. So many users are completely unaware that the "but only if..." section of the rule even exists. At least once a quarter, usually more often, I field exactly this questions that caused this issue to be opened.

It's discussed in the concepts section of the docs, Getting Started, and elsewhere as well. Docs are clearly not enough. 

> I would say let‘s merge my new PR, the descriptions take up one list entry of space which is like nothing as long as you are not on a mobile phone.

:+1:

PR #3001, opened 2025-01-10
Settings menu: Always show things count even when inbox is not empty

jimtng:
Change from:
image

To:
image

When there is nothing in inbox:
image


florian-h05 (approved):
LGTM, thanks!

PR #3003, opened 2025-01-12
Thing details: Add "Install Binding" button for HANDLER_MISSING_ERROR

jimtng:
image

Which leads to:

image


ghys (approved):
Makes sense!
LGTM, thanks!

PR #3004, opened 2025-01-12
Item Details: Add a link to search for the item in the developer bar

jimtng:
This makes it super convenient. Just click on the link and it's pre-searched for you.

image

Which leads to:

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -148,7 +148,7 @@
     </f7-panel>
 
     <f7-panel v-if="showDeveloperDock" right :visible-breakpoint="1280" resizable>
-      <developer-dock :dock="activeDock" :helpTab="activeHelpTab" :toolTab="activeToolTab" />
+      <developer-dock :dock="activeDock" :helpTab="activeHelpTab" :toolTab="activeToolTab" :searchFor="developerSearch" />
     </f7-panel>
 
     <f7-block v-if="!ready && communicationFailureMsg" class="block-narrow">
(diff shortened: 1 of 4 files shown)


ghys (approved):
Seems indeed quite useful, but would probably not work as intended if you look for usages of an item whose name is also part of other item's names.
For example `Kitchen_Lights` is included `Kitchen_Lights_Table`.
Since you clicked on the link for a particular item you'd think the results would only pertain to it and it only.

(This is not strictly speaking an issue of this PR since you can also search for "usage" manually in the searchbar and have the same problem).

jimtng:
> probably not work as intended if you look for usages of an item whose name is also part of other item's names.
> For example `Kitchen_Lights` is included `Kitchen_Lights_Table`.
> Since you clicked on the link for a particular item you'd think the results would pertain to it and it only.

Good point! We could improve the searching (perhaps globally?) so when something is enclosed in double quotes, it would do a "full word search" so `Kitchen_Lights_Table` is not found when you're searching for `"Kitchen_Lights"`

ghys:
That would be best, but I'm also willing to just consider it a simple link to put the item name in the dev sidebar's searchbox and call it a day :) Your choice how to proceed.

jimtng:
> a simple link to put the item name in the dev sidebar's searchbox and call it a day

You mean just as is now? I think this will do for now and if the double quoted search thing gets implemented, it's a matter of a minor change for the link later on.

ghys:
@florian-h05 WDYT?

florian-h05 (approved):
Looks good :+1: 

Wrt to the quotes: Instead of having such "magic search query strings", I would rather have a small checkbox with an icon like VS Code 
image
to enable such a full-word match.

ghys:
I remember when Google almost removed the literal search ability, so ironic we'd talk about instating it :)
https://news.ycombinator.com/item?id=37554736

PR #3007, opened 2025-01-12
Sitemap treeview: Fix arrows/chevrons shown if not expandable

florian-h05:
Regression from #2904.
Reported in https://github.com/openhab/openhab-webui/pull/2970#issuecomment-2585663931.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/treeview-item.vue
+++ b/bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/treeview-item.vue
@@ -66,6 +66,7 @@ export default {
     },
     onStart (event) {
       console.debug('Drag start event:', event)
+      this.$set(this.localMoveState, 'moving', true)
       this.$set(this.localMoveState, 'widget', this.widget.slots.widgets[event.oldIndex])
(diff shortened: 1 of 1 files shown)


mherwege:
Thanks. I didn't test yet, but LGTM

ghys:
Too late for 4.3.2?

florian-h05:
#2904 is no part of 4.3.x at all, no need to backport this at all.

ghys:
Indeed, sorry.

mherwege:
@florian-h05 Did you test this? I tried replicating this logic for the model tree-view, but the expansion icons still show.

florian-h05:
Yes, I tested this and it worked fine.

mherwege:
OK, I now understand why this works here, but not in the model treeview. The difference is that I made sure there is an empty slot in the children in this case.
I tested this, and an alternative approach with a toggle that also works, which I have to use in the model treeview. My review proposal uses in here uses this approach.

mherwege (changes requested):
Remove `v-if="canHaveChildren"` from line 10.
Add `:toggle="canHaveChildren"` around line 6.

florian-h05:
Let's align the implementation - I have just implemented your review comments.

PR #3010, opened 2025-01-13
Item Add: Navigate to the new Item once added

jimtng:
This PR changes the navigation / where the user is sent to after creating/adding a new item.

Currently, if you go to Settings -> Items -> Add -> Save, you'll be brought back to the Items list (back navigation). 

- If the user wants to add another item - it's easy, they can click `+` again to add another item. No problems.
- But if the user wants to edit that new item, e.g. to edit metadata, add a link, etc. they would need to scroll around, or do a search for the item, click on it to get to that new item.

This PR changes this behavior so that after the new item is saved/added, the user is redirected to the Item Details page **of that new item**, so they can immediately work on it.
- If they want to add more item, they can click `< Back` which will bring them to the Add Item page, so they can enter new details and create a new item.
- If they want to go to the items list, they can just click cancel or Back again.

I think this workflow is more streamlined whether the user wants to add one item or multiple items.

This PR doesn't change the navigation behavior when **editing** an item. As previously done, the user would be sent back to the previous page, which would be the Item details. In other words: Settings -> Items (list) -> Item Details -> Edit -> Save -> Back to Item Details.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
@@ -217,7 +217,11 @@ export default {
         }
 
         this.dirty = false
-        this.$f7router.back()
+        if (this.createMode) {
+          this.$f7router.navigate('/settings/items/' + this.item.name)
+        } else {
+          this.$f7router.back()
+        }
       }).catch((err) => {
         this.$f7.toast.create({
           text: 'Item not saved: ' + err,


florian-h05 (approved):
Seems reasonable.
WDYT @ghys ?

rkoshak:
Would it be worth requesting something similar for Things too, at least when creating a Thing manually (e.g. Generic MQTT Thing)? I find that even more annoying than the Item behavior, perhaps because I rarely create an Item like that. Maybe a separate issue should be opened for that.

florian-h05:
Good point, Rich 👍
Instead of opening an issue, let’s ask Jim if he wants to change that right now.

jimtng:
Maybe open a separate issue..it will be in a different pr. Thanks

rkoshak:
https://github.com/openhab/openhab-webui/issues/3013

PR #3011, opened 2025-01-13
Log viewer: Fixed column width for timestamp, level & logger name

florian-h05:
When using the log viewer, I have regularly noticed that the time column is wider than required, as well as the logger name column taking up much space and making the message column move if a new log with a longer logger name comes in.
This PR sets the width of the time, level and logger name columns to a fixed value and limits the logger name length to the last 40 chars, which is enough to identify the logger.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
@@ -225,11 +225,26 @@
   td.sticky
     position sticky
     left 0
+    width 105px
     color black
     background #f1f1f1
     z-index 1
     white-space nowrap
     overflow hidden
+
+  td.level
+    width 50px
+
+  td.logger
+    width 280px
+
+  span.logger
+    width 280px
+    display block
+    direction rtl
+    overflow hidden
+    text-overflow ellipsis
+    white-space nowrap
 
   tr.table-rows
     height 31px
@@ -335,6 +350,7 @@ export default {
       batchLogs: [],
       nextId: 0,
       maxEntries: 2000,
+      maxLoggerNameLength: 40,
       logStart: '--:--:--',
       logEnd: '--:--:--',
       currentHighlightColorItemIndex: null,
@@ -468,8 +484,8 @@ export default {
           break
       }
       tr.innerHTML = '<td class="sticky"><i class="icon f7-icons" style="font-size: 18px;">' + icon + `</i> ${entity.time}<span class="milliseconds">${entity.milliseconds}</span></td>` +
-        `<td>${entity.level}</td>` +
-        `<td>${entity.loggerName}</td>` +
+        `<td class="level">${entity.level}</td>` +
+        `<td class="logger"><span class="logger">${entity.loggerName}</span></td>` +
         `<td class="nowrap">${this.highlightText(entity.message)}</td>`
       return tr
     },


cdjackson:
I'm reasonably ok with this. I do find the truncation of the logger slightly annoying, but on balance it's fine.

The time column seems to change width - depending on the message. ie if the message is long, then the time column is shrunk...

image

image

florian-h05:
> The time column seems to change width - depending on the message. ie if the message is long, then the time column is shrunk...

That’s because this PR was developed on top of your logger icon PR, the icon will avoid that behaviour.

florian-h05:
Setting width, and text-align right together with overflow hidden doesn't work.

florian-h05:
I will merge this soon, we can always come back and improve later if we find a better solution, but for now this already is an improvement.

mherwege:
This style does align right and cut, put I can't get ellipsis to work:
```
  td.logger
    width 280px
    overflow hidden
    text-align right
    display flex
    justify-content flex-end
    align-items center
    height 29px
```

mherwege:
I think I solved it:

instead of a `td.logger` style, replace it with:

```
  span.logger
    width 280px
    display block
    align-items center
    direction rtl
    overflow hidden
    text-overflow ellipsis
```

and use this for the HTML:

```
      tr.innerHTML = '<td class="sticky"><i class="icon f7-icons" style="font-size: 18px;">' + icon + `</i> ${entity.time}<span class="milliseconds">${entity.milliseconds}</span></td>` +
        `<td>${entity.level}</td>` +
        `<td><span class="logger">${entity.loggerName}</span></td>` +
        `<td class="nowrap">${this.highlightText(entity.message)}</td>`
```

florian-h05:
Thanks for the help @mherwege, got it working now with some extra styling.

florian-h05:
[Bildschirmaufnahme_20250115_184520.webm](https://github.com/user-attachments/assets/22c18a2d-d41b-43e3-be4c-6665abba190d)

bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
```diff
@@ -335,6 +350,7 @@ export default {
       batchLogs: [],
       nextId: 0,
       maxEntries: 2000,
```
mherwege, line 352: This is not needed anymore.

bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
```diff
@@ -335,6 +350,7 @@ export default {
...
+      maxLoggerNameLength: 40,
```
mherwege, line 353: This is not needed anymore.

bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
```diff
@@ -225,12 +225,27 @@
...
     background #f1f1f1
     z-index 1
     white-space nowrap
     overflow hidden
 
+  td.level
+    width 50px
+
+  td.logger
+    width 280px
```
mherwege, line 239: I didn’t need this when testing in Firefox. It will not harm either.

florian-h05: I needed this when developing in Chromium.

mherwege: For some strange reason, when I was trying this, things worked in Firefox, but I couldn’t even get the websocket connection to start in Chrome, so no way to test (there was an error in the console). That is obviously another issue and I will keep an eye out for it.

cdjackson:
In my opinion, the logger name now looks strange. The right alignment makes it difficult to read - it's ok if they are all truncated, but it's strange if some fit and some don't.

image

For me, it should be left aligned, with the left side truncated if it's too long. I think this is what the original PR here did...

florian-h05:
> For me, it should be left aligned, with the left side truncated if it's too long. 

I agree, will change that: #3026

cdjackson:
Thanks @florian-h05 I'll take a look.

PR #3022, opened 2025-01-15
Update oh-slider handing of steps when rounding state values

### LINKED ISSUE #3019: [MainUI] oh-slider inoperable and fails in toStepFixed with RangeError: Precision XXX out of range
**Proposed fix and background on why this happened is included at the end.** Long story short, oh-slider has a long-standing bug in its handing of stateDescription.step. A recent change in mqtt.generic seemed to make this more common due to newly _very_ precise stateDescription.step values being passed to the slider from mqtt Dimmer channels -> Items -> UI.

Workaround at end.

## The problem

Background: I have been debugging an issue with rendering of ColorTemperature and Brightness items (that are linked to MQTT.generic things after openhab/openhab-addons#17813 (i.e. 4.3.x). See below for details on the relationship with that PR). When I attempt to adjust one of these items in MainUI I see a Dimmer that looks like this no matter the state of the item is (in this case mt item is at 19%): 

image

The item details from the API Explorer:

```json
{
  "link": "https://<<redacted>>/rest/items/BedsideTableLamp_Brightness",
  "state": "19.00",
  "stateDescription": {
    "minimum": 0,
    "maximum": 100,
    "step": 0.3937007874015748,
    "pattern": "%.0f %%",
    "readOnly": false,
    "options": []
  },
  "metadata": {
    "ga": {
      "value": "lightBrightness"
    },
    "semantics": {
      "value": "Point_Control",
      "config": {
        "relatesTo": "Property_Level",
        "isPointOf": "BedsideTableLamp"
      }
    }
  },
  "editable": true,
  "type": "Dimmer",
  "name": "BedsideTableLamp_Brightness",
  "label": "Brightness",
  "category": "slider",
  "tags": [
    "Control",
    "Level"
  ],
  "groupNames": [
    "BedsideTableLamp"
  ]
}
```

(See Additional Details below for details on why that step showing here are below where I've identified the root cause.)

Clicking/interacting with the slider at any location causes a single update of the item to the existing value (in this case 19) with no visual update to the slider. Any additional clicks do nothing.

## Expected behavior

I expect the slider to look like this and be completely interactive. 

image

or, if that isn't possible, I expect the slider to not load at all, and ideally be replaced with an error message. Ideally something somewhat detailed like "Error rendering Slider with config {dumped config}". (in this case something as simple as "Error when loading slider label" would have helped me significantly when debugging this.)

## Steps to reproduce

1. Create an item
2. Note that the Slider renders as expected.
3. Apply State Description min=0, max=100, step=1.234 (note, `234` may be replaced with any string of numbers above 100).
4. Clear cache and refresh page
5. Note that the slider does not have a number line and is no longer interactive (it will not slide).

Minimal broken item:

```json
{
  "link": "https://<<redacted>>/rest/items/Test_Broken_Dimmer",
  "state": "NULL",
  "stateDescription": {
    "minimum": 0,
    "maximum": 100,
    "step": 1.234,
    "readOnly": false,
    "options": []
  },
  "metadata": {
    "stateDescription": {
      "value": " ",
      "config": {
        "step": "1.234",
        "min": "0",
        "max": "100"
      }
    }
  },
  "editable": true,
  "type": "Dimmer",
  "name": "Test_Broken_Dimmer",
  "label": "Broken Dimmer with Step",
  "category": "",
  "tags": [],
  "groupNames": []
}
```

## Your environment

```yaml
runtimeInfo:
  version: 4.3.2
  buildString: Release Build
locale: en-US
systemInfo:
  configFolder: /openhab/conf
  javaVersion: 17.0.13
  javaVendor: Debian
  osName: Linux
  osVersion: 6.1.0-27-amd64
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 295457088
  totalMemory: 538968064
  uptime: 98828
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-airquality
  - binding-androiddebugbridge
  - binding-astro
  - binding-chromecast
  - binding-hue
  - binding-icalendar
  - binding-ipcamera
  - binding-jellyfin
  - binding-mqtt
  - binding-nanoleaf
  - binding-openweathermap
  - binding-pushover
  - binding-roku
  - binding-tplinksmarthome
  - binding-unifi
  - binding-vizio
  - binding-zwave
  - misc-metrics
  - misc-openhabcloud
  - persistence-influxdb
  - persistence-rrd4j
  - transformation-jsonpath
  - transformation-map
  - ui-basic
  - ui-habpanel
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: true
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
    pixelRatio: 1.6666666666666667
    prefersColorScheme: dark
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 4
    language: en-US
    languages:
      - en-US
      - en
    onLine: true
    platform: Win32
  screen:
    width: 1800
    height: 1200
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: dark
    filled: true
    bars: light
    homeNavbar: default
    homeBackground: default
    blocklyRenderer: null
  userAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:133.0) Gecko/20100101
    Firefox/133.0
timestamp: 2025-01-14T10:38:52.994Z
```

## Browser console

Specifically these are the errors from my minimal repro.

```txt
RangeError: precision 234 out of range
    toStepFixed https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    formatScaleLabel https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    value https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    Tn https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    value https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    t https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    r https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    <<<callstack trimmed for conciseness, I hope you don't mind>>>
app.4b7e0bc536a621ffbe03.js:2:1437118

Uncaught TypeError: e.$refs.rangeslider.f7Range is undefined
    mounted https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    setTimeout handler*mounted https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    Kn https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    kn https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    insert https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    S https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    Bi https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    _update https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    a https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    get https://<<redacted>>/js/app.4b7e0bc536a621ffbe03.js:2
    <<<trimmed again>>>
app.4b7e0bc536a621ffbe03.js:2:1396663

```

## Browser network traffic

All traffic is successful (2XX), and relevant Item details are listed above. If anything else from the Network tab can help I'd be happy to provide.

## Additional information

Taking a look at toStepFixed, the RangeError seems to be coming from toFixed. This code is trying to use the number of decimals in the step to round the incoming number, but instead it is passing the decimal portion of the step value directly. (i.e. a step of 1.245 is setting nbDecimals to 245, not 3). The fix should be adding `.length` after `[1]`. I'm happy to open a PR myself if a contributor on this repo is able to verify it for me (this would be my first contribution and I'd need to ramp up to the dev environment first).

In oh-slider.vue, the step is used to round the 
```vue
    toStepFixed (value) {
      // uses the number of decimals in the step config to round the provided number
      const nbDecimals = this.config.step ? Number(this.config.step).toString().replace(',', '.').split('.')[1] : 0
      return parseFloat(Number(value).toFixed(nbDecimals))
    },
```

https://github.com/openhab/openhab-webui/blame/7ef5b6fe1c29aeca733d4012b5720538a07a245f/bundles/org.openhab.ui/web/src/components/widgets/system/oh-slider.vue#L58

### Root cause

The change to oh-slider above was made 4 years ago. I can't speak for other addons, but in the case of mqtt I believe it didn't raise issues because step values were not automatically this precise before a recent change in 4.3.x. In this case this happened because my Item is linked to this channel from mqtt.generic: 

```txt
 Type dimmer : brightness      "brightness" [ stateTopic = "zigbee2mqtt/HueWhiteAndColor0AE7/brightness", commandTopic = "zigbee2mqtt/HueWhiteAndColor0AE7/set/brightness", min = 0, max = 254, step = 1 ]
```

Prior to openhab/openhab-addons#17813, this step would have been 1 (matching the channel) or default to 1 if no step was provided. After this change the step is calculated based on the range and channel step (see [here](https://github.com/openhab/openhab-addons/blob/184ef673a2aad92d2c83bb4af4e2088c67a4187b/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/PercentageValue.java#L72), which is set on the Item at the bottom of that file. In my case 0.3937007874015748 is (1 * 100 / 254). 

I suspect that most of the time other bindings also don't set such oddly precise step values and haven't run into this, and I also suspect that humans don't write stateDescriptions with such odd step values either.

### Workarounds for others

If you come across this before the change is available, the workaround that's worked for me is manually creating a StateDescription with a more reasonable Step value (i.e. one that has a decimal part that is less than 100, like 1.99, 25.45, etc.). Its a bit painful to make one for each item, but it seems to address the issue after a forced reload of the UI.

vidia:
Fixes #3019, see that bug for precise details on what led to this change and the details of the broken behavior.

Updating oh-slider to round the format label, scale, etc based on the step provided (which comes from the StateDescription). Before this change, a step value such as 1.234 would try to fix the value like:

Number(...).toFixed(234)

which causes a RangeError (toFixed requires inputs ranging 0..100). It needs to be checking the length of the decimal portion of the string, not the value itself.

This change attempts to keep as close to the current behavior, and the intention mentioned by the comments. In the above example it fix the state to 3 digits (instead of 234).

# Testing

Tested using local MainUI server via NPM checking my various items that had been broken by 4.3.X.

Confirmed that this addresses the console errors listed in the bug as well as makes the sliders interactive again when using very precise step values.

Tested with integer steps (fixing to 0 decimal digits), undefined steps (fixing to 0 decimal digits), and very precise steps generated from my mqtt.generic items (0.3937007874015748, fixing to 16). All work as expected and the sliders are functional again.

 Signed-off-by: David Tschida <dmtschida1@gmail.com> (github: vidia)

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-slider.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-slider.vue
@@ -55,8 +55,8 @@ export default {
     },
     toStepFixed (value) {
       // uses the number of decimals in the step config to round the provided number
-      const nbDecimals = this.config.step ? Number(this.config.step).toString().replace(',', '.').split('.')[1] : 0
-      return parseFloat(Number(value).toFixed(nbDecimals))
+      const nbDecimals = this.config.step ? Number(this.config.step).toString().replace(',', '.').split('.')[1]?.length : 0
+      return parseFloat(Number(value).toFixed(nbDecimals ?? 0))
     },
     onChange (newValue) {
       if (isNaN(this.value)) return
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
@@ -50,9 +50,8 @@ export default {
     toStepFixed (value) {
       // uses the number of decimals in the step config to round the provided number
       if (!this.config.step) return value
-      const nbDecimals = Number(this.config.step).toString().replace(',', '.').split('.')[1]
-      // do NOT convert to number, instead return string, otherwise formatting wouldn't work
-      return parseFloat(value).toFixed(nbDecimals ? nbDecimals.length : 0)
+      const nbDecimals = Number(this.config.step).toString().replace(',', '.').split('.')[1]?.length
+      return parseFloat(Number(value).toFixed(nbDecimals ?? 0))
     },
     onPlusMinusClick () {
       setTimeout(() => {
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
@@ -50,8 +50,9 @@ export default {
     toStepFixed (value) {
       // uses the number of decimals in the step config to round the provided number
       if (!this.config.step) return value
-      const nbDecimals = Number(this.config.step).toString().replace(',', '.').split('.')[1]?.length
-      return parseFloat(Number(value).toFixed(nbDecimals ?? 0))
+      const nbDecimals = Number(this.config.step).toString().replace(',', '.').split('.')[1]
+      // do NOT convert to number, instead return string, otherwise formatting wouldn't work
+      return parseFloat(value).toFixed(nbDecimals ? nbDecimals.length : 0)
     },
     onPlusMinusClick () {
       setTimeout(() => {


florian-h05 (approved):
LGTM, thanks and congrats to your first contribution!

florian-h05:
FYI I have cherry-picked this to 4.3.x so it will be part of openHAB 4.3.3 if there is another patch release.

vidia:
Thanks for taking a look, and thanks for the updated naming.

PR #3027, opened 2025-01-19
Channel Settings: Make long Channel UID fully visible

### LINKED ISSUE #3023: MainUI: Channel UID has fixed width and does not show full UID details of long UIDs.
## Problem
Some items have very long Channel UIDs and due to the fixed display width some information is then truncated.

Image
Here "valve-tappet-position" is not completely shown and the copy button is missing.

## Expected behavior

Channel UID with should be dynamically enlarged or rest of UID should be shown in a second line

## Steps to reproduce

1.  Create an Item with very long Channel UID.

## Your environment

```yaml

```

## Browser console

```txt

```

jimtng:
Fix #3023

image

image


florian-h05 (approved):
LGTM, thanks!

PR #3028, opened 2025-01-19
Addons Store: Extend search to the add-on descriptions

### LINKED ISSUE #2870: Extend search of Add-on store to descriptions
## The problem

https://github.com/openhab/openhab-addons/issues/17682#issuecomment-2469187287

> We have two add-ons, ism8 and wolfsmartset (you will need one or the other depending on the smart home interface installed in the heating system). When a user types "wolf" into the search field of the add-on store, I would like to see both add-ons shown.
> 
> Current behaviour is as follows: Whey I search for "wolf", I get only the latter.

The search mechanism seems to look at the name of the binding, but not at the description (which contains Wolf for both add-ons).

## Your suggestion

Extending the search to the description field.
This may lead to a lot of false positives. Therefore, we might restrict this feature, e.g., use the description only if the search term has at least 3 letters.

## Your environment

```yaml

```

## Additional information

We discussed adding the term "Wolf" to the binding name in openhab/openhab-addons#17747. This would basically mean that we rename the binding - not my preferred solution.

I do not know about the current implementation of the search and how easy it would be to extend it to the description.

jimtng:
Resolve #2870

Related: https://github.com/openhab/openhab-core/pull/4564

image

TODO: Go through addon descriptions and refine the wordings, e.g. a lot of them start with `This is the binding for...`

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
+++ b/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
@@ -370,7 +370,7 @@ export default {
         results = results.filter((a) => a.type === this.currentTab)
       }
       query = query.toLowerCase()
-      results = results.filter((a) => a.id.includes(query) || a.label.toLowerCase().includes(query))
+      results = results.filter((a) => a.id.includes(query) || a.label.toLowerCase().includes(query) || a.description?.toLowerCase()?.includes(query))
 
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!

holgerfriedrich:
Thanks!

PR #3030, opened 2025-01-20
[logviewer] Support history

cdjackson:
This is the UI side of the enhancements to a) receive historical logs from the OH server, and b) improve performance by receiving arrays of log entries rather than just single entries.

It does work with the existing log system since it will work with both single log entries (as currently used) and arrays of logs.  Due to the change in format, the filter requests do log a parse error since it now uses a DTO with a number of bits of data rather than just an array of loggers, but this doesn't impact anything else. Therefore it is safe to merge this before the core is merged (although it's probably best to wait a little to ensure there's no major comments to the core side!).

This also fixes a minor bug where the number of filtered logs may be incorrectly displayed.

With this and the core changes, I can display full zigbee debug - it's about 2000 lines in 15 seconds (or put another way, the log buffer shows about 15 seconds of data). The log remains responsive (I can change the logging levels back). Possibly the log buffer can/should be increased more than 2000 entries, but that's for another day...

Core PR coming soon...

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
@@ -355,6 +355,7 @@ export default {
       logEnd: '--:--:--',
       currentHighlightColorItemIndex: null,
       currentHighlightColor: '#FF5252',
+      lastSequence: 0,
       colors: [
         '#FF0000', // Red
         '#00FF00', // Green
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
core PR has been merged.

florian-h05 (commented):
LGTM, thanks!

@kaikreuzer You were a bit too fast with merging, I would have preferred to merge this myself so I can take a look before merging 😉

kaikreuzer:
@florian-h05 Sorry, will leave it in future to you. 😄

PR #3034, opened 2025-01-22
Link to New Item: Prefix with Thing's label

### LINKED ISSUE #3032: [MainUI] use better default label for new items
## The problem

When linking a new Item to a thing channel, a default label is automatically entered. However - depending on the thing type - the default label must be edited or it will cause confusion. Example: my Tradfri power outlets have one channel called "Operation", which is also the auto-generated label of the new Item. Having 5 Items called "Operation" isn't useful, so I manually copy+paste the thing label to the Item label field.

## Your suggestion

Prefix all auto-generated Item labels with the Thing label, e.g. "Floor lamp Operation".

## Additional information

Especially for "simple" things, like these outlets, it might be even better to use only the thing label as item label ("Floor lamp").

jimtng:
Resolve #3032


florian-h05 (approved):
LGTM, thanks.

PR #3039, opened 2025-01-27
Widget Config and Widget Code: Live update, dirty checking, movable popup

jimtng:
- Make Widget Config and Widget Code Editor popups draggable by the navbar, so they can be moved out of the way to view the widget being edited
- Preview changes immediately on the widget
- The changes can be reverted if desired by clicking the "Reset" link in the navbar.
- The changes are reverted if the popup is closed without saving.
- Implement dirty checking confirmation
- Since the Widget Config Editor is used in many other places, this also applies to them

Here's an animated GIF to show how it works
popup_move

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/movable-popup-mixin.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/movable-popup-mixin.js
@@ -17,24 +17,29 @@ export default {
   },
   methods: {
     /**
-     * Make the given popupRef movable by dragging the draggableRef element
-     * If draggableRef is not provided, the popupRef will be used as the draggable element
-     * 
-     * This method needs to be called in the popup:open event handler, and
-     * cleanupMovablePopup() must be called in the popup:close event handler.
-     * 
-     * The popupRef must be a reference to the popup element, e.g.
-     *   <f7-popup ref="popupRef"
-     * And the draggableRef must be a reference to the element that will be used to 
+     * Make the given `popupRef` movable by dragging the `draggableRef` element.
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.ui/web/src/pages/settings/movable-popup-mixin.js
```diff
@@ -0,0 +1,105 @@
...
+        popupRef: null,
+        draggableRef: null,
+        resizeObserver: null
+      }
+    }
+  },
+  methods: {
+    // Make the given popupRef movable by dragging the draggableRef element
+    // If draggableRef is not provided, the popupRef will be used as the draggable element
+    initializeMovablePopup (popupRef, draggableRef = null) {
```
florian-h05, line 21: Can you please add JSDoc here to explain the usage of this mixin?

florian-h05 (commented):
LGTM, thanks! This is again a great UX enhancement after the browser title PR!

Can you please have a look at my comment about JSDoc?

bundles/org.openhab.ui/web/src/pages/settings/movable-popup-mixin.js
jimtng: I'm not familiar with jsdoc... but I gave it a go.

jimtng: Thanks! I think we need to insert proper <code> fences and {@link} but I'm not familiar with it and have never worked with jsdoc before. Is there a mechanism to render and preview it?

florian-h05 (approved):

PR #3042, opened 2025-01-28
Script edit: Display conditions

jsjames:
Added support so conditions will be displayed when script rules are viewed.

// Triggers:
// - At 05:55:55

// Conditions:
// - Only If the time is between 06:00 and 13:00
// - Only If the time is between 03:30 and 14:00
// - Only If the time is between 09:00 and 14:00


florian-h05 (approved):
Nice enhancement, thanks!

PR #3043, opened 2025-01-28
Fix oh-icon icon name type in widget def & docs

jimtng:


ghys (changes requested):
Thanks, good catch, but you modified a generated doc:

```
<!-- DO NOT REMOVE the following comments -->
<!-- GENERATED props -->
### General
<div class="props">
<PropGroup label="General">
<PropBlock type="BOOLEAN" name="icon" label="Icon">
<PropBlock type="TEXT" name="icon" label="Icon">
...
```


You want to modify https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/icon.js instead first (pb to pt), then regenerate the docs with the script described in
https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/CONTRIBUTING.md#documentation--resources

jimtng:
Thanks! Learnt something new :)

In the process of running generate.js there were other changes, which I submitted as a separate commit.

jimtng:
> In the process of running generate.js there were other changes, which I submitted as a separate commit.

Ughh, it seems that those changes shouldn't be there. I've removed them, so that this PR only contains changes to icon

florian-h05 (approved):
LGTM, thanks!

PR #3045, opened 2025-01-29
Item Detail: Fix search in developer sidebar

jimtng:
Fix the following scenario:
- "search for usages of this item" for Item A is clicked
- Sidebar opens and sets `searchFor=itemA` preloads the search as requested
- User then types new query in the sidebar, meanwhile the searchFor prop remains unchanged
- Clicking on "search for usages of this item" for item A again, caused no changes in `searchFor`, as a result, the watch for this prop didn't fire

This PR fixes this situation by resetting the searchFor prop, therefore causing the watch to perform the search again.


florian-h05 (commented):
LGTM, thanks!

PR #3047, opened 2025-01-30
Location Picker: Set default to system location, otherwise set to worldwide zoom level

### LINKED ISSUE #2999: [MainUI] Map picker should center on location set in regional settings
## The problem

When the map picker first comes up to choose a lat/lon, it is fully zoomed out centered over Europe. 
## Your suggestion

It would be more user-friendly if the map picker used the location set in the regional settings (if it's set, current behavior is fine if not) and use a lot more zoomed in level. If a user is choosing a location for their home automation, they are likely wanting something reasonably close to home so something showing 100 mi/200 km (I know that's not the same) or so would be about right.

As a bonus, if it can remember where it was the last time it was used that would be nice. I just created a rule template that uses a location context. It's great that it lets users use the map picker but they need to pick two points and when opening the new map one needs to zoom and scroll all over again.

## Your environment

```yaml
runtimeInfo:
  version: 4.3.1
  buildString: Release Build
locale: en-US
systemInfo:
  configFolder: /openhab/conf
  javaVersion: 17.0.13
  javaVendor: Debian
  osName: Linux
  osVersion: 6.8.0-51-generic
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 76209400
  totalMemory: 489684992
  uptime: 2746
  startLevel: 70
addons:
  - automation-jsscripting
  - automation-pidcontroller
  - binding-amazonechocontrol
  - binding-astro
  - binding-chromecast
  - binding-http
  - binding-ipcamera
  - binding-mail
  - binding-mqtt
  - binding-network
  - binding-networkupstools
  - binding-openweathermap
  - binding-shelly
  - binding-vesync
  - binding-zigbee
  - binding-zwave
  - misc-openhabcloud
  - persistence-mapdb
  - persistence-rrd4j
  - transformation-basicprofiles
  - transformation-jinja
  - transformation-jsonpath
  - transformation-map
  - transformation-regex
  - ui-basic
  - ui-habot
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    pixelRatio: 0.8999999761581421
    prefersColorScheme: light
  isSecureContext: true
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: 8
    hardwareConcurrency: 16
    language: en-US
    languages:
      - en-US
      - en
    onLine: true
    platform: Linux x86_64
  screen:
    width: 2560
    height: 1600
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: light
    filled: true
    bars: light
    homeNavbar: default
    homeBackground: default
    blocklyRenderer: null
  userAgent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like
    Gecko) Chrome/131.0.0.0 Safari/537.36
timestamp: 2025-01-09T23:36:44.489Z
```

jimtng:
Resolve #2999

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
@@ -54,16 +54,20 @@ export default {
       this.center = (this.value) ? latLng(this.value.split(',')) : latLng(48, 6)
       this.showMap = true
       if (!this.value) {
-        this.$oh.api.get('/rest/services/org.openhab.i18n/config').then((data) => {
-          if (data.location) {
-            this.center = latLng(data.location.split(','))
-            this.zoom = 15
-            this.showMap = false
-            this.$nextTick(() => {
-              this.showMap = true
-            })
-          }
-        })
+        this.$oh.api.get('/rest/services/org.openhab.i18n/config')
+          .then((data) => {
+            if (data.location) {
+              this.center = latLng(data.location.split(','))
+              this.zoom = 15
+              this.showMap = false
+              this.$nextTick(() => {
+                this.showMap = true
+              })
+            }
+          })
+          .catch(() => {
+            // silently ignore if the request is not permitted for the user
+          })
       }
     })
   },
--- a/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
@@ -65,8 +65,11 @@ export default {
               })
             }
           })
-          .catch(() => {
+          .catch((err) => {
             // silently ignore if the request is not permitted for the user
+            if (!(err === 'Forbidden' || err === 403)) {
+              return Promise.reject(err)
+            }
           })
       }
     })
--- a/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
@@ -36,7 +36,7 @@ export default {
   data () {
     return {
       showMap: false,
-      zoom: 4,
+      zoom: 1,
       center: latLng(48, 6),
       // url: 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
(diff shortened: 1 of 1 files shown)


jimtng:
Waiting for feedback

jimtng:
this is now ready

ghys:
Just to make you aware, because the location picker is potentially user-facing and `/rest/services` is only for admins, you can have 403 errors when you do call that from this widget.

jimtng:
> because the location picker is potentially user-facing and `/rest/services` is only for admins, you can have 403 errors when you do call that from this widget.

Thanks, I wasn't aware of that! I'm struggling to find where/how to create a user-facing page that would invoke the location picker. As far as I can see, the location picker (popup) is only called by parameter-location, which in turn is used by config-parameter, all of these are (currently) admin-only stuff.

In the event that it ever gets used in user-facing, wouldn't the 403 error be hidden in the javascript console, and not visible to user? In which case, they'd simply get a worldwide view, instead of the current mostly-Europe-only view.

A possible fallback from this is to query the location from the browser, but it would trigger a permission prompt. Perhaps we can add that later when the need arises.

ghys:
> In the event that it ever gets used in user-facing, wouldn't the 403 error be hidden in the javascript console, and not visible to user? In which case, they'd simply get a worldwide view, instead of the current mostly-Europe-only view.

Yes I suppose, but in general it shouldn't be "let's try something and fall back if it fails?"

jimtng:
I can add an `isAdmin` check before calling the REST, but this creates a "hard coded logic", assuming that this rest call is always going to be admin-only. Whilst the risk is small, if one day this particular endpoint is opened up to non-admin (e.g. for read-only access), then we'd have multiple places to update.

Your call - I can add an `isAdmin` here check if you prefer.

jimtng:
Alternatively I can just catch the exception and fail silently?

florian-h05:
I would tend to catch the 403 and fail silently.

PR #3050, opened 2025-02-02
oh-input: Fix editing a number item with unit & Several enhancements

jimtng:
When oh-input is set to a Number item with unit, the f7-input in 'numeric' mode cannot set its numeric value to the item state because it contains the unit.

This PR:

- Display the number's unit after the input field for numeric type
- Make numeric input field right-aligned
- Improve input-elements alignment
- Accept Enter key to send the value, with or without a sendButton
- Honor useDisplayState config
- Add `min` and `max` config
- If the stateDescription on the item specifies `minimum`, `maximum`, and `step`, they will be used as the default, but it can be overridden by the widget's config
- Tested against Dimensionless Percent
- Tested against patterns with multiple format specifiers, e.g. `Test format %1$.4f and %1$.2f %unit%`. The last format specifier will be used.

image

### DIFF
--- a/bundles/org.openhab.ui/doc/components/oh-input-card.md
+++ b/bundles/org.openhab.ui/doc/components/oh-input-card.md
@@ -120,6 +120,16 @@ Display an input in a card
     Use the formatted state as the value for the input control
   </PropDescription>
 </PropBlock>
+<PropBlock type="DECIMAL" name="min" label="Minimum">
+  <PropDescription>
+    Minimum allowed value when type set to number
+  </PropDescription>
+</PropBlock>
+<PropBlock type="DECIMAL" name="max" label="Maximum">
+  <PropDescription>
+    Maximum allowed value when type set to number
+  </PropDescription>
+</PropBlock>
 <PropBlock type="DECIMAL" name="step" label="Step">
   <PropDescription>
     Step value when type set to number, any if left empty
--- a/bundles/org.openhab.ui/doc/components/oh-input-item.md
+++ b/bundles/org.openhab.ui/doc/components/oh-input-item.md
@@ -125,6 +125,16 @@ Display an input field in a list
     Use the formatted state as the value for the input control
   </PropDescription>
 </PropBlock>
+<PropBlock type="DECIMAL" name="min" label="Minimum">
+  <PropDescription>
+    Minimum allowed value when type set to number
+  </PropDescription>
+</PropBlock>
+<PropBlock type="DECIMAL" name="max" label="Maximum">
+  <PropDescription>
+    Maximum allowed value when type set to number
+  </PropDescription>
+</PropBlock>
 <PropBlock type="DECIMAL" name="step" label="Step">
   <PropDescription>
     Step value when type set to number, any if left empty
--- a/bundles/org.openhab.ui/doc/components/oh-input.md
+++ b/bundles/org.openhab.ui/doc/components/oh-input.md
@@ -88,6 +88,16 @@ Displays an input field, used to set a variable
     Use the formatted state as the value for the input control
   </PropDescription>
 </PropBlock>
+<PropBlock type="DECIMAL" name="min" label="Minimum">
+  <PropDescription>
+    Minimum allowed value when type set to number
+  </PropDescription>
+</PropBlock>
+<PropBlock type="DECIMAL" name="max" label="Maximum">
+  <PropDescription>
+    Maximum allowed value when type set to number
+  </PropDescription>
+</PropBlock>
 <PropBlock type="DECIMAL" name="step" label="Step">
   <PropDescription>
     Step value when type set to number, any if left empty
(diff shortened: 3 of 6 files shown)


mherwege:
In the sitemap UI, we went the extra mile to split of the unit in that case and show it separately behind the input field. Maybe it is worth looking at that.

jimtng:
indeed, more work is needed here, so we can also see the unit. Some layout improvements would be nice too.

mherwege:
Will this respect the number of decimal positions from the state description?

jimtng:
> Will this respect the number of decimal positions from the state description?

Thanks! No it currently would not, but I'll work on that too.

Sample here with state desc `%.2f %%`

image

jimtng:
@mherwege I've updated the first post above with the full detail of the changes + latest screenshot

mherwege:
Where does the unit come from? If it is the item unit, that is meant to be for internal use, not representation. That will mix things again, while we separated them by introducing the item unit. The unit used should be from the state description. E.g. your unit is default (W) but you want to show and change in kW.
With sitemap widgets, there is also the possibility to override the state description in the label for that one wiget. So it may even be necessary to introduce an extra widget parameter for the oh-input-widget.

jimtng:
> Where does the unit come from? If it is the item unit, that is meant to be for internal use, not representation. That will mix things again

Thanks! I'll make the adjustment.



> With sitemap widgets, there is also the possibility to override the state description in the label for that one wiget. So it may even be necessary to introduce an extra widget parameter for the oh-input-widget.

Yes, but this needs to be done in core so it can perform unit conversion. The ui can't do that (or do we need to create a unit conversion REST endpoint?)

Say the item's unit _and_ stateDescription is `W` and the widget specified unit is `kW`. How does the widget display a value of say `1000 W` in its specified unit of `kW`?

jimtng:
OK, the next iteration will honor the unit specified in displayState, and if it contains %unit%, then fall back into the main item's unit.

image

mherwege:
Great. Sorry to go through iterations on this. Can you also test with other locales that use , instead of . as the decimal separator? I believe that was the next challenge I ran into with basicUI.

jimtng:
@mherwege This is a bit more complicated than it may seem.

What do we do when the state description pattern is something like this: 
```
I am part of stateDescription pattern %1$.2f W %1$.3f blah kW
```
image



Note it contains two patterns, one with two decimal digits, the second one with two decimal digits. I believe we should use the last one for editing. I know this is a crazy pattern, but it is possible



> Great. Sorry to go through iterations on this. My memory seems to come back in fragments. Can you also test with other locales that use , instead of . as the decimal separator? I believe that was the next challenge I ran into with basicUI.
> And then also check with Dimensionless unit ONE and unit % for the different configuration options.

Thank you for sharing your experiences! Saves me from finding all these bugs later on down the line.

I am still working through the above, and I'll incorporate your latest cases too.

mherwege:
> Note it contains two patterns, one with two decimal digits, the second one with two decimal digits. I believe we should use the last one for editing. I know this is a crazy pattern, but it is possible

Yes, I am aware. I and also made an assumption there of only having on pattern. I also think there are places in core where it would fail when there is more than one.

jimtng:
> I also think there are places in core where it would fail when there is more than one.

indeed, I believe there are many places that would get tripped up with this, in webui, basicui, too.

Thankfully, I think I can rely on the fact that the _unit_ is always the last token in the pattern, and that this isn't possible:
`%1$.1f W %1$.4f kW` (i.e. two different units in the same pattern)

jimtng:
@mherwege

Finnish locale, amongst others, use a space as group (thousand) separator, whilst some others use comma / period for group separator / decimal and sometimes in reverse.

Additionally... `f7-input` doesn't want to deal with group separators, so they'd have to be stripped somehow.

jimtng:
And then.... scientific notation `%e`

jimtng:
As it is now, this PR will work fine provided that no funky patterns are used. i.e. no: numbers preceding the pattern not separated by space, no thousand separators, etc.

mherwege:
@jimtng

> Finnish locale, amongst others, use a space as group (thousand) separator, whilst some others use comma / period for group separator / decimal and sometimes in reverse.

Yes, I know this is tricky. In BasicUI, a few regex expressions are used for this: https://github.com/openhab/openhab-webui/blob/0041fb699fc40b99c9b2f057e6812155d9c53e5f/bundles/org.openhab.ui.basic/web-src/smarthome.js#L2654
I believe these do a fair job on , vs . and scientific notation. But they are not perfect. Spaces for thousand separators will fail, although it might be possible to add that.

> Additionally... `f7-input` doesn't want to deal with group separators, so they'd have to be stripped somehow.

I can't say what the problem is with that. Is that true for all browsers? When developing, I noticed there are some browser variations. Firefox seemed to be particularly challenging.

florian-h05:
@jimtng Can you please provide the Items you used to test?

jimtng:
I can't seem to find them, nor the pages / widgets that I used to test this :(

florian-h05 (approved):
LGTM, thanks!
I tested it with my prod system where I have several different inputs and additionally with the following Items:

```
Group test_number_input

Number:Dimensionless Number_Dimensionless (test_number_input)  {listWidget="oh-input-item"[type="number",outline=true]}
Number:Dimensionless Number_Percent (test_number_input)  {listWidget="oh-input-item"[type="number",outline=true], unit="%"}
Number:Power Number_Power "Power" (test_number_input) {listWidget="oh-input-item"[type="number",outline=true]}
Number:Power Number_Power_DisplayUnit "Power with display unit [%.1f kW]" (test_number_input) {listWidget="oh-input-item"[type="number",outline=true,useDisplayState=true,sendButton=true]}
```

jimtng:
Thanks for the fixes!

PR #3051, opened 2025-02-02
Script edit: Add support for new python add-on & Refactor automation-languages definition

jsjames:
- Added support in script-edit.vue for new/imminent graal python binding.
- removed hard-coding of commentChar and pulled from restructured annotation-language.js definition
- cleaned up the annotation-language definition implementation

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/automation-languages.js
+++ b/bundles/org.openhab.ui/web/src/assets/automation-languages.js
@@ -1,12 +1,28 @@
-export const AddonNames = {
-  'application/javascript': 'JavaScript Scripting',
-  'application/javascript;version=ECMAScript-5.1': 'JavaScript Scripting (Nashorn)',
-  'application/x-groovy': 'Groovy Scripting',
-  'application/x-python': 'Jython Scripting',
-  'application/x-ruby': 'JRuby Scripting'
-}
-
-export const DocumentationLinks = {
-  'application/javascript': '/link/jsscripting-script',
-  'application/x-ruby': '/link/jrubyscripting-script'
+export const AUTOMATION_LANGUAGES = {
+  'application/javascript': {
+    'name': 'JavaScript Scripting',
+    'commentChar': '//',
+    'documentationLink': '/link/jsscripting-script'
+  },
+  'application/javascript;version=ECMAScript-5.1': {
+    'name': 'JavaScript Scripting (Nashorn)',
+    'commentChar': '//'
+  },
+  'application/x-groovy': {
+    'name': 'Groovy Scripting',
+    'commentChar': '//'
+  },
+  'application/x-python': {
+    'name': 'Jython Scripting',
+    'commentChar': '#'
+  },
+  'application/x-ruby': {
+    'name': 'JRuby Scripting',
+    'commentChar': '#',
+    'documentationLink': '/link/jrubyscripting-script'
+  },
+  'application/x-python3': {
+    'name': 'Python Scripting',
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.ui/web/src/assets/automation-languages.js
```diff
@@ -1,12 +1,28 @@
...
-  'application/x-python': 'Jython Scripting',
-  'application/x-ruby': 'JRuby Scripting'
-}
-
-export const DocumentationLinks = {
-  'application/javascript': '/link/jsscripting-script',
-  'application/x-ruby': '/link/jrubyscripting-script'
+export const AUTOMATION_LANGUAGES = {
+  'application/javascript': {
+    'name': 'JavaScript Scripting',
```
jimtng, line 3: @florian-h05 what's the generally preferred syntax for defining objects? i.e. I would prefer

```javascript
export const AUTOMATION_LANGUAGES = {
  'application/javascript': {
  name: '......',
  commentChar: '....'
}}
```

instead of `'name': '....'` etc.

Also `ObjectName.key1.subkey` as opposed to `ObjectName['key1']['subkey']`

bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
```diff
@@ -396,7 +396,13 @@ export default {
...
     },
     mimeTypeDescription (mode) {
-      return AddonNames[mode] || mode
+      return AUTOMATION_LANGUAGES[mode]['name'] || mode
+    },
+    documentationLink (mode) {
+      if ('documentationLink' in AUTOMATION_LANGUAGES[mode]) {
+        return AUTOMATION_LANGUAGES[mode]['documentationLink']
+      }
+      return null
```
jimtng, line 405:
```suggestion
      return AUTOMATION_LANGUAGES[mode].documentationLink
```

jimtng (commented):
This is more of a question from myself as an inexperienced javascript coder rather than a request for change

bundles/org.openhab.ui/web/src/assets/automation-languages.js
florian-h05: I prefer and I think its common to not use quotes as long as they aren‘t required, like in your example.
When accessing object properties, you usually use the dot notation, the `[string]` notation is only used if dot notation won‘t work due to special chars or you need to use vars.

I would have requested to change that when doing a full review, just had a swift look at the code.

bundles/org.openhab.ui/web/src/pages/settings/rules/script/script-edit.vue
florian-h05: This will return `undefined` instead of `null`, which however shouldn‘t be a problem when checking for the return value without a comparison operator, e.g. `if (someValue) do something` or `if (!someValue) do nothing`.

jsjames: Without the check if documentationLink exists, I get the following error in Chrome: TypeError: Cannot read properties of undefined (reading 'documentationLink')

jsjames: this works if I use this form: return AUTOMATION_LANGUAGES[mode]['documentationLink']

florian-h05 (approved):
LGTM, thanks!
I have reverted the package-lock.json changes.

PR #3053, opened 2025-02-04
Addons Store: Show tab bar when left panel is hidden

jimtng:
When the left panel is hidden, there's no way to see uninstalled add-ons, so we should show the tab bar.

This should be backported.


florian-h05 (approved):
LGTM, thanks!

PR #3055, opened 2025-02-06
Sitemap editor: fix negative number parameters

mherwege:
See: https://community.openhab.org/t/basic-ui-setpoint-negative-values/161011

Parameters with negative number where not parsed correctly. This is because hyphen is defined as a token and therefore already recognized before recognizing the number. Changing the order does not help as it breaks chart period parsing.

This fix avoids recognizing numbers as tokens and handles numbers in the lexer.

Tests where added and all existing tests still pass.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
+++ b/bundles/org.openhab.ui/web/src/assets/sitemap-lexer.nearley
@@ -46,7 +46,9 @@
     comma:            ',',
     colon:            ':',
     hyphen:           '-',
-    number:           /[+-]?[0-9]+(?:\.[0-9]*)?/,
+    plus:             '+',
+    dot:              '.',
+    unsignedint:      /[0-9]+/,
     string:           { match: /"(?:\\["\\]|[^\n"\\])*"/, value: x => x.slice(1, -1) }
   })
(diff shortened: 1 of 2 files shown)


jimtng:
Would this work on browsers with a locale that uses a comma as the decimal point?

mherwege:
@jimtng

> Would this work on browsers with a locale that uses a comma as the decimal point?

This is about the sitemap editor and how it parses sitemap xtext syntax, not about displayed values. So, this is not impacted by locale.
The parser first tokenizes and splits on tokens for efficiency. The nearly code was too aggressive at that. I took care to exactly match what is done in the core sitemap parser.

I assume you had our other discussion about input fields and uom in mind.

jimtng:
Oh good point, it just needs to be the same number syntax that core will use to parse the sitemap. I assume core will require a dot, regardless of _its_ locale?

mherwege:
> Oh good point, it just needs to be the same number syntax that core will use to parse the sitemap. I assume core will require a dot, regardless of _its_ locale?

Yes, it does. the sitemap syntax does not depend on locale.

florian-h05 (approved):
LGTM, thanks!

PR #3056, opened 2025-02-06
Layout edit: Move page settings to bottom of screen in edit mode

jimtng:
This frees up screen real estate to see more widgets

In Edit mode, remove the page settings from the top of editor page
image

image

In createMode, it is still shown at the top as before, but a note added
image


florian-h05 (approved):
Overall LGTM, just one styling issue:

image

The page settings use the full available width, which looks awkward.
I will fix that to use the default behaviour, e.g. as used by the script settings.

PR #3058, opened 2025-02-07
Use a more specific browser title for each page

### LINKED ISSUE #3054: configurable HTML title for pages
I’d like to be able to define the HTML page title that the openHAB dashboard appears with in browser tabs.
(for now, it's always a static string "openhab").

The default should be changed to use the page’s label, but please also offer an option to override it on all pages, especially on system-generated pages like and including the main dashboard page.

Related: #3037

jimtng:
Resolve #3054

- Refactor how the current url is checked inside app.vue. This reduces the amount of `String.indexOf()` calls, and it makes implementing this PR easier.
- Fix a bug where `Settings` menu item in the sidebar isn't highlighted when viewing the Transformation and Persistence sections.
- Add a new `browserTitle` config for pages other than the `overview` page
- Hide "Sidebar & Visibility" and "Tags" input for the `overview page`. This frees up some screen real estate for editing widgets
- Every page has a custom browser page title:
  - Pages will use its `browserTitle` || `label` config
  - Overview page will just use `label` since it's not used elsewhere
  - For other pages, use the section menu's text, so if/when this gets internationalized, the title will also change
  - Settings -> System Settings will use `Settings - <navbar title> - openHAB` and so it's also automatically localized as the navbar title gets localized.
  - Settings -> Addon settings will use `<navbartitle> - openHAB` because the navbar title always starts with "Configure....". Adding "Settings - " in front of it would seem redundant.
  - Things/Items/Widgets/Add-on Store/Pages/Rules/Scripts/Scenes will show `[X] - [uid] - openHAB` e.g. `Things - mqtt:topic:xxx - openHAB`

browser_title3

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/app.vue
+++ b/bundles/org.openhab.ui/web/src/components/app.vue
@@ -637,9 +637,12 @@ export default {
       // Some special cases where the title should be different
       if (this.currentPath.page) {
         title.unshift(this.$store.getters.page(this.currentPath.page?.$key)?.config?.label)
-      } else if (this.currentPath.overview || this.currentPath.locations || this.currentPath.equipment || this.currentPath.properties) {
+      } else if (this.currentPath.overview) {
         const config = this.$store.getters.page('overview')?.config
-        title.unshift(config?.browserTitle || config?.label)
+        const localizedTitle = this.$t(`home.${this.currentPath.$key}.title`)
+        title.unshift(config?.browserTitle || (config?.label === 'Overview' ? localizedTitle : config?.label))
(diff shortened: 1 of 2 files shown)


florian-h05:
@jimtng Can you please resolve conflicts?

jimtng:
rebased

florian-h05 (approved):
This is a big UX enhancement, I wonder no one had the idea before - thanks!

I have fixed a few minor things, e.g. for the persistence and transaformations settings "Settings" was used as title, and for the model tabs I decided to go with their names.

PR #3060, opened 2025-02-07
UID Input: Add `Fix UID` button to normalize UID input

### LINKED ISSUE #3052: autocorrect custom UIDs - convert humancase to snakecase
## The problem
it is easy to first make descriptive label then paste it into UID input
but when i paste name of my thing/item - i need to manually valid my own uids - remove spaces etc

## Your suggestion
like correcting branches name in VSCode - auto convert pasted text to snake_case (or camel case?)

jimtng:
Resolve #3052 

image

After clicking the hammer icon:

image

Also add automatic item name generation as the label is typed (only when name hasn't been entered)
uid_fix_sync

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/utils.js
+++ b/bundles/org.openhab.ui/web/src/js/openhab/utils.js
@@ -3,10 +3,10 @@ import diacritic from 'diacritic'
 
 export default {
   normalizeLabel: (label) => {
-    return diacritic.clean(label.normalize('NFKD')).replace(/\s+/g, '_').replace(/[^0-9a-z_]/gi, '').replace(/^([0-9])/, '_$1')
+    return diacritic.clean(label.normalize('NFKD')).trim().replace(/\s+/g, '_').replace(/[^0-9a-z_]/gi, '').replace(/^([0-9])/, '_$1')
   },
   normalizeLabelForThingId: (label) => {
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.ui/web/src/js/openhab/utils.js
```diff
@@ -3,7 +3,12 @@ import diacritic from 'diacritic'
 
 export default {
   normalizeLabel: (label) => {
-    return diacritic.clean(label.normalize('NFKD')).replace(/\s+/g, '_').replace(/[^0-9a-z_]/gi, '')
+    return diacritic.clean(label.normalize('NFKD')).replace(/\s+/g, '_').replace(/[^0-9a-z_]/gi, '').replace(/^([0-9])/, '_$1')
```
mherwege, line 6:
```suggestion
    return diacritic.clean(label.normalize('NFKD')).replace(/\s+/g, '_').replace(/[^0-9a-z_]/gi, '').replace(/^([0-9-])/, '_$1')
```

jimtng: The previous regex removes `-` so this is not necessary here.

However, I might need to have two versions of this,  one specific for thinguid that allows dashes.

jimtng: I've added a separate check for thing/channel id

florian-h05 (approved):
Nice addition - thanks!

PR #3044, opened 2025-01-29
Tabbed Page: Support Iconify and openHAB icons for tabs

jimtng:
Resolve #2937

image

With badge:
image

image

```yaml
config:
  label: tabbed test
  sidebar: true
tabs:
  - component: oh-tab
    config:
      badgeColor: green
      icon: if:f7:archivebox-fill
      page: page:photos
      title: Iconify XXXXXXXXXX
    slots:
      default: []
  - component: oh-tab
    config:
      icon: f7:archivebox_fill
      page: page:photos
      title: F7 XXXXXXXXXX
    slots:
      default: []
  - component: oh-tab
    config:
      badgeColor: blue
      icon: oh:siren
      page: page:home
      title: OH Icon
    slots:
      default: []
```

PR #3064, opened 2025-02-15
Tabbed Page: Fix label too large

florian-h05:
Regression from #3044.

Noticed on my iPhone that the font size of the label was way too large.


florian-h05:
@jimtng FYI

jimtng:
Thanks for the fix!

PR #3046, opened 2025-01-29
oh-image-card: Fix inconsistent image size when action is defined

jimtng:
Fix #3002

Before:
image

After:
image

PR #3065, opened 2025-02-16
Fix oh-image-card styling affects all cards

florian-h05:
Regression from #3046.


florian-h05:
Thanks for the hint, done :+1:

PR #3070, opened 2025-02-19
Persistence edit: Support configuring aliases

mherwege:
Configuration in the UI for persistence alias changes introduced in https://github.com/openhab/openhab-core/pull/4363


mherwege:
https://github.com/openhab/openhab-core/pull/4363 was merged a while ago, so this is ready for review and merge as well.

florian-h05:
I would prefer to the aliases integrated right into the persistence edit page without having the extra dialog to modify them. This way it would be enough to use the Item picker to add a new Item, then set its alias right besides.
This will also align the style with the other configuration.

mherwege:
> I would prefer to the aliases integrated right into the persistence edit page without having the extra dialog to modify them. This way it would be enough to use the Item picker to add a new Item, then set its alias right besides. This will also align the style with the other configuration.

That doesn’t work when you have no strategy defined for an item. An it is possible to just use extension methods to store and retrieve. That’s part of the reason why I moved it in a separate section in the configuration file in the first place.
Also you can set configure persistence with an item group, but still want aliases for the individual items.
Strategies also has a separate section not directly linked to items. I see this as similar.

florian-h05:
I am not sure we are talking about the same thing here ...
I meant that I would prefer to have the aliases editable directly in their section, and not only displayed in their section with a button to an edit popup.
I did not mean to integrate the aliases into the popup where you select items and assign a policy and a filter to them.

mherwege:
@florian-h05 This was a bit more challenging than anticipated. I wanted it to look integrated and still be able to select multiple items from the picker in one picker open. I hope the last commits satisfy that requirement.

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
```diff
@@ -539,16 +674,32 @@ export default {
...
+        console.debug(`Removing ${module}:`)
+        console.debug(self.persistence[module][index])
+        self.persistence[module].splice(index, 1)
+        this.checkDirty()
+      })
+    },
+    deleteModuleKey (ev, module, key) {
       if (!this.editable) return
+      let swipeoutElement = ev.target
+      const self = this
```
florian-h05, line 694: Normally there is no need for `self = this` "hacks" when using array functions in callbacks, which is the case here.
Why did you use this "hack" here?

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
```diff
@@ -539,16 +674,32 @@ export default {
       this.checkDirty()
     },
     deleteModule (ev, module, index) {
+      if (!this.editable) return
       let swipeoutElement = ev.target
+      const self = this
```
florian-h05, line 679: Normally there is no need for `self = this` "hacks" when using array functions in callbacks, which is the case here.
Why did you use this "hack" here?

BTW: `this.checkDirty` is used, so you anyway expect `this` to work fine.

florian-h05 (changes requested):
Looks good now, thanks!
Three minor comments though.

bundles/org.openhab.ui/web/src/pages/settings/persistence/persistence-edit.vue
mherwege: I changed it back. But it is strange. While it works without assigning to `self`, I cannot see the value of `this.persistence` on line 686 in the (Firefox) debugger. It is shown as undefined. I needed to see it when debugging.

florian-h05 (approved):

PR #3078, opened 2025-02-25
blockly: set value to key of dictionary

### LINKED ISSUE #3075: Blockly missing block to update a dictionary value
OH 4.3.2

I hope I am reporting this in the right place.

Unless I missed it, I don’t see a way to update the value of a dictionary key, the equivalent in javascript of:

mydict["mykey"] = "newvalue";

I had to use the “inline script” block to do this. It seems that these are the only two blocks for dictionaries:

Image

This limits the use of dictionaries.

stefan-hoehn:
fixes #3075 

Provides a block that allows to update a dictionary that was assigned to a (named) variable:

image


florian-h05 (approved):
LGTM, thanks!

PR #3081, opened 2025-03-01
Add Alt-Shift-M shortcut key to toggle menu when unpinned

jimtng:
It has always been a pain point for me when the menu is not pinned, getting back to it is very difficult, involving several "back" clicks, and sometimes it also gets lost and it became impossible without a full browser refresh.

This PR adds a shortcut Alt-Shift-M (similar to Alt-Shift-D) to show the menu whereever you are.


florian-h05 (approved):
Nice, thanks!

PR #3082, opened 2025-03-01
[blockly ]oh_taggeditems block: Fix multiple tags in vars & Add array support

### LINKED ISSUE #3077: Blockly “get items with tag” does not work with multiple tags when using variable
OH 4.3.3

I am not sure I am putting this in the right place and it might not be considered a bug depending on the point of view, but this is a trap unless the documentation mentions the limitation. Once you understand how it works behind the scenes, you understand why it works that way in blockly.

The blocky documentation says for “get items with tag”:

_“multiple tags can be provided which then need to be separated with a comma”_

It works but with a caveat, it only works when hardcoding tags directly in the block between the quotes like this

Image

which gives

`items = items.getItemsByTag('tag1','tag2');`

But this

Image

gives this as expected

tags = 'tag1,tag2';
items = items.getItemsByTag(tags);

but does not work as one might expect, it returns nothing.

The explanation is clear, getItemsByTag(string tag1, string tag2, ...) expects one string argument per tag, so when using a variable only one string is passed and not parsed. The parsing is done by blockly which cannot work when using a variable. A list (array) is not accepted as a parameter and gives a type conversion error.

The issue per se is not blockly but the way getItemsByTag() works.

Possible solutions:

1. getItemsByTag() could accept arrays, this way there would be nothing to change in blockly.
2. getItemsByTag() could do the parsing internally, but this would be in contradiction with the current way it works, you can't have it both ways.
3. Jumping on the previous proposal, have two functions: getItemsByTag() and getItemsByTags() where the latter could accept a string with comma separated tags.
4. Just update the documentation saying that multiple tags only works for hardcoded tags ;(

Solution 1 is probably the easiest and cleanest solution. In all cases the documentation should be amended for users to properly use the block.

stefan-hoehn:
fixes #3077

- allows multiple tags variables
- adds array support
- more resilient by trimming away spaces in tags

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-items.js
@@ -88,7 +88,7 @@ export default function (f7) {
     init: function () {
       this.appendValueInput('tagName')
         .appendField('get items with tag')
-        .setCheck('String')
+        .setCheck(['String', 'Array'])
       this.setInputsInline(false)
       this.setOutput(true, 'Array')
       this.setColour(0)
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!

PR #3092, opened 2025-03-08
Parameter-text redesign for multiple text values

Nadahar:
I was considering adding a configuration parameter with multiple strings and predefined `options`, where you could also add free-text values that weren't among the `options`. The current implementation of `parameter-text.vue` doesn't handle that too well, so I did some redesign of the component.

The configuration parameter turned out to be a dead end, and is no longer considered. But, the modification of `parameter-text.vue` might still be valuable if other configuration parameters use this combination.

Here is a demonstration of the changes done by this PR, with the abandoned configuration parameter:

https://youtu.be/w1j7M4muyLs


florian-h05 (approved):
LGTM, thanks!
I think this improvement will come very handy in many situations!

I will soon push a minor code change.

PR #3111, opened 2025-03-16
GH Action: Detection of new files

holgerfriedrich:
This restores the behavior before tj-actions was removed. New untracked files created during build will break the build.

Follow-up to #3106. Consistent to what we merged on core/distro.


florian-h05 (approved):
LGTM, thanks!

PR #3112, opened 2025-03-18
Addon Card: Give title has more room to avoid overflow

jimtng:
Before:

image

After:

image

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
@@ -65,8 +65,9 @@
     .addon-card-label
       text-overflow ellipsis
       overflow clip
-      white-space nowrap
       color var(--f7-text-color)
+      max-height 3.4rem
+      line-height 1.1
     .addon-card-title-after
       .preloader-inner .preloader-inner-left, .preloader-inner .preloader-inner-right, .preloader-inner .preloader-inner-line
         margin-left inherit !important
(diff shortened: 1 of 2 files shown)


florian-h05:
That kinda looks awkward.
I would rather move the button down besides the openHAB verifiy text.

jimtng:
That's what I tried first. But in marketplace it's occupied by 3 things: views, likes, and something else. They're too wide and adding a button makes them wrap to the next line

florian-h05:
@ghys WDYT?

ghys:
The only thing that looked good IMO was to have the button consistently aligned with the title in both widgets (card and list).

image

Problem is that the card also needs a fixed height so the clipped titles were a compromise.

Maybe "line clamping" with `line-clamp` (`-webkit-line-clamp`) (https://developer.mozilla.org/en-US/docs/Web/CSS/line-clamp) to allow 2 lines (max) for both types is worth a try. It could also make sense for UI widget cards, their titles tend to be longer.

jimtng:
I've updated the original post with a new screenshot. This time the button stays as it was.

florian-h05 (approved):
LGTM now, thanks!

An additional improvement idea would be to remove the "Binding" from the add-on label. I think that is redundant information, as bindings will only be displayed in sections where Binding is already part of the section header. WDYT?

jimtng:
> An additional improvement idea would be to remove the "Binding" from the add-on label. I think that is redundant information, as bindings will only be displayed in sections where Binding is already part of the section header. WDYT?

Good idea, but in the main Addons store page, one can search like this

image

florian-h05:
What about only removing the „Binding“ from the big cards?
Those aren’t used for search results IIRC.

jimtng:
@florian-h05 the title of this PR seems a bit odd to me. Maybe you meant "Give title more room....."?

PR #3114, opened 2025-03-19
Config sheet: Fix loading of multi-line default values

jimtng:
A config parameter with multiple="true" (a multiline / list) can have default values specified in the config.xml by separating them with commas.

For example,

```
  "parameters": [
    {
      "defaultValues": [
        "a",
        "b",
        "c"
      ],
      "default": "a,b,c",
    ...
```

`default` gets split up into `defaultValues` in core by https://github.com/openhab/openhab-core/blob/baf79c5f27a7f69b70c21241c8aae68c2e8ca05a/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/config/EnrichedConfigDescriptionParameterDTO.java#L49

There is an attempt to "join" the values here: https://github.com/openhab/openhab-webui/blob/d714a1cd1599ebba6c9a171a06cab446e437cf13/bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue#L36

However the value it is trying to join isn't the array (defaultValue), but the single unsplit String value `default`.

This PR corrects that and sets `value` to the the correct list/array default value when `multiple="true"` so that `join('\n')` would work correctly.

Also related: https://github.com/openhab/openhab-core/pull/4657

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
@@ -90,7 +90,13 @@ export default {
       const conf = Object.assign({}, this.configuration)
       this.parameters.forEach((p) => {
         if (conf[p.name] === undefined && p.default !== undefined) {
-          conf[p.name] = typeof p.default === 'function' ? p.default(this.configuration) : p.default
+          if (typeof p.default === 'function') {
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM. thanks!

PR #3116, opened 2025-03-22
Main UI mvn build: Various enhancements to improve DX in Eclipse IDE

florian-h05:
- Do not always clean install, do only so in mvn `clean` phase.
- Built web app directly to mvn target dir.
- Perform all Main UI related executions in the `prepare-package` phase to allow faster Java code compilation.

This improves the behaviour when using Eclipse IDE, as the M2E plugin causes a recompilation "loop" due to the Main UI build writing to a mvn source folder.

### DIFF
--- a/bundles/org.openhab.ui/pom.xml
+++ b/bundles/org.openhab.ui/pom.xml
@@ -76,6 +76,7 @@
             <goals>
               <goal>npm</goal>
             </goals>
+            <phase>generate-resources</phase>
             <configuration>
               <arguments>install --save false</arguments>
             </configuration>
@@ -87,7 +88,7 @@
             <goals>
               <goal>npm</goal>
             </goals>
-            <phase>validate</phase>
+            <phase>test</phase>
             <configuration>
               <arguments>run lint</arguments>
             </configuration>
@@ -110,28 +111,9 @@
             <goals>
               <goal>npm</goal>
             </goals>
+            <phase>compile</phase>
             <configuration>
-              <arguments>run build-prod ${project.version}</arguments>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>build-helper-maven-plugin</artifactId>
-        <executions>
-          <execution>
-            <goals>
-              <goal>add-resource</goal>
-            </goals>
-            <phase>generate-resources</phase>
-            <configuration>
-              <resources>
-                <resource>
-                  <directory>web/www</directory>
-                  <targetPath>app</targetPath>
-                </resource>
-              </resources>
+              <arguments>run build:mvn ${project.version}</arguments>
             </configuration>
           </execution>
         </executions>
--- a/bundles/org.openhab.ui/pom.xml
+++ b/bundles/org.openhab.ui/pom.xml
@@ -55,13 +55,15 @@
           <workingDirectory>web</workingDirectory>
         </configuration>
 
+        <!-- run all executions in the prepare-package phase, to build the UI at the latest possible time and avoid building it when only compiling Java code -->
+        <!-- this helps Eclipse IDE to compile Java code faster -->
         <executions>
           <execution>
             <id>Install node and npm</id>
             <goals>
               <goal>install-node-and-npm</goal>
             </goals>
-            <phase>initialize</phase>
+            <phase>prepare-package</phase>
           </execution>
 
           <execution>
@@ -69,7 +71,7 @@
             <goals>
               <goal>npm</goal>
             </goals>
-            <phase>generate-sources</phase>
+            <phase>prepare-package</phase>
             <configuration>
               <arguments>install --save false</arguments>
             </configuration>
@@ -80,7 +82,7 @@
             <goals>
               <goal>npm</goal>
             </goals>
-            <phase>generate-resources</phase>
+            <phase>prepare-package</phase>
             <configuration>
               <arguments>run build:mvn ${project.version}</arguments>
             </configuration>
--- a/bundles/org.openhab.ui/pom.xml
+++ b/bundles/org.openhab.ui/pom.xml
@@ -61,7 +61,7 @@
             <goals>
               <goal>install-node-and-npm</goal>
             </goals>
-            <phase>generate-resources</phase>
+            <phase>initialize</phase>
           </execution>
 
           <execution>
@@ -69,7 +69,7 @@
             <goals>
               <goal>npm</goal>
             </goals>
-            <phase>generate-resources</phase>
+            <phase>generate-sources</phase>
             <configuration>
               <arguments>install --save false</arguments>
             </configuration>
@@ -80,7 +80,7 @@
             <goals>
               <goal>npm</goal>
             </goals>
-            <phase>compile</phase>
+            <phase>generate-resources</phase>
             <configuration>
               <arguments>run build:mvn ${project.version}</arguments>
             </configuration>
--- a/bundles/org.openhab.ui/pom.xml
+++ b/bundles/org.openhab.ui/pom.xml
@@ -32,6 +32,26 @@
 
   <build>
     <plugins>
+      <!-- Delete web/node_modules in Maven clean goal -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-clean-plugin</artifactId>
+        <configuration>
+          <filesets>
+            <fileset>
+              <directory>web/node_modules</directory>
+            </fileset>
+          </filesets>
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>clean</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- Build Main UI -->
       <plugin>
         <groupId>com.github.eirslett</groupId>
         <artifactId>frontend-maven-plugin</artifactId>
@@ -52,12 +72,36 @@
           </execution>
 
           <execution>
-            <id>npm ci</id>
+            <id>npm install</id>
             <goals>
               <goal>npm</goal>
             </goals>
             <configuration>
-              <arguments>ci</arguments>
+              <arguments>install --save false</arguments>
+            </configuration>
+          </execution>
+
+          <!-- Run ESLint during the Maven initialize phase -->
+          <execution>
+            <id>npm lint</id>
+            <goals>
+              <goal>npm</goal>
+            </goals>
+            <phase>validate</phase>
+            <configuration>
+              <arguments>run lint</arguments>
+            </configuration>
+          </execution>
+
+          <!-- Run unit tests during the Maven test phase -->
+          <execution>
+            <id>npm test</id>
+            <goals>
+              <goal>npm</goal>
+            </goals>
+            <phase>test</phase>
+            <configuration>
+              <arguments>run test:unit</arguments>
             </configuration>
           </execution>
 
--- a/bundles/org.openhab.ui/pom.xml
+++ b/bundles/org.openhab.ui/pom.xml
@@ -43,13 +43,6 @@
             </fileset>
           </filesets>
         </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>clean</goal>
-            </goals>
-          </execution>
-        </executions>
       </plugin>
       <!-- Build Main UI -->
       <plugin>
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.ui/pom.xml
```diff
@@ -52,42 +72,48 @@
...
+              <arguments>run test:unit</arguments>
+            </configuration>
+          </execution>
+
+          <execution>
+            <id>npm run build-prod</id>
+            <goals>
+              <goal>npm</goal>
+            </goals>
+            <phase>compile</phase>
```
Nadahar, line 114: I don't remember why at the moment, but I had assigned this to the `prepare-package` phase. `compile` is for compiling the Java code (generating the `.class` files) and doesn't need the webpack output at that stage. You can run `mvn compile` to just do the compilation without the packaging. Packaging is where the JAR is created, and these files must be generated before that so that they are packaged as a part of the JAR. I assume that's why I put it there.

Did you experience problems using the `prepare-package` phase, or did you just think that `compile` is a "better fit"?

bundles/org.openhab.ui/web/build/build.js
```diff
@@ -10,12 +10,14 @@ const config = require('./webpack.config.js');
 
 const env = process.env.NODE_ENV || 'development';
 const target = process.env.TARGET || 'web';
+const maven = process.env.MAVEN || false;
+const outPath = maven ? '../target/classes/app' : 'www';
```
Nadahar, line 14: Did you verify that this works without the `./`? I just assumed that `rm` required it since it was already there, but it doesn't actually make sense because as far as I know, the "`./` convention" only applies when you execute files. If you're on Linux/macOS, you can easily check that the folder is actually deleted.

Nadahar (commented):
Is there a reason why you left `version` at `4.2.0` in `package.json`, or did you just don't think it belonged here?

This version number should probably either be generated from some other source, or added to the procedure (that must exist) for changing version numbers..?

All in all, this looks good to me, except that as I wrote on the forum, I found that Eclipse keeps launching the stuff in the `test` phase as well. Therefore, adding the tests and the linting to the Maven build makes it even slower in Eclipse.

I'm not sure if there's a way to prevent that, M2E isn't exactly "well documented". While it's nice to have those things in the Maven build, if a separate GitHub workflow is going to be used anyway and we can't find a way to not make Eclipse run them every time the source changes, maybe it's better all in all to not include them?

I'll do some more experimentation to see if I can get Eclipse to not run them.

Nadahar:
I've tried to look at why the test/lint is run during Eclipse build.

It's actually quite complicated, but to try to simplify it a bit, the "surefire" Maven plugin that is used to run Java tests is set to "ignore" for all Eclipse builds by default. That's why we don't have the "problem" that tests are being run when Eclipse builds with Java projects.

You can make custom [m2e lifecycle mapping](https://eclipse.dev/m2e/documentation/m2e-execution-not-covered.html#ignore-plugin-goal) in the project POM that specifies what should and shouldn't be run by m2e. However, the "filtering mechanism" only allows specifying a combination of a plugin and a goal. This doesn't work for us, because `frontend-maven-plugin` is used to invoke all `npm` operations with one goal: `npm`.

The `frontend-maven-plugin` has [its own embedded m2e configuration](https://github.com/eirslett/frontend-maven-plugin/blob/b7b060fa13ca47e69172ee6d06194ffd41c35084/frontend-maven-plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml#L2) that defines that all goals except the "install" goals (install node and npm etc.) are run on incremental builds. We can override this default, but [we can't differentiate between different phases](https://bugs.eclipse.org/bugs/show_bug.cgi?id=450798), which is what would have allowed this to work. I must say that it's quite arrogant of the developers to call this a "corner case", but regardless.

I can't see that there's any way to make m2e run the "webpack" operation but not the "lint" and "test", as long as these are all executed by the same plugin using the same goal (although in different phases). The "suggestion" from the m2e issue that one could use (Maven) profiles for this, isn't a practical solution as far as I can tell. While defining profiles with "different rules" in the POM is fine, executing them requires extra command line options. We can't have that the WebUI will require "extra parameters" to build versus all the other bundles, so unless there is some magic that have eluded me, this seems like a no-go.

We *could* make a separate profile that contained the lint and test calls, and use that profile when running the GitHub builder, but that means that the tests and linting wouldn't be performed when doing "normal" builds of the whole project (unless the profile was explicitly included).

I don't know what is the best solution here, I don't know "the importance" of including the tests and the linting in the Maven build, if that would simplyfy the GitHub workflow for example. But, it's certainly not ideal to have Eclipse launch them, adding another minute or so, to each build cycle.

bundles/org.openhab.ui/pom.xml
```diff
@@ -32,6 +32,26 @@
...
+          <filesets>
+            <fileset>
+              <directory>web/node_modules</directory>
+            </fileset>
+          </filesets>
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>clean</goal>
```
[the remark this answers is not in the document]
florian-h05: True, I will change that.

bundles/org.openhab.ui/web/build/build.js
florian-h05: I have tested it on Linux, works fine without the `./` as expected.

florian-h05:
> Is there a reason why you left version at 4.2.0 in package.json, or did you just don't think it belonged here?

I just though this does not belong into this PR.

> This version number should probably either be generated from some other source, or added to the procedure (that must exist) for changing version numbers..?

Third option would be to set it to 0.0.0, as we don't publish Main UI to npm (would make absolutely no sense) the version number in the package.json is not used anywhere, and this avoids confusion. I know projects doing this.

> I don't know what is the best solution here, I don't know "the importance" of including the tests and the linting in the Maven build, if that would simplyfy the GitHub workflow for example. But, it's certainly not ideal to have Eclipse launch them, adding another minute or so, to each build cycle.

I would say let's remove them again from the Maven build. It is not super important to have them here, as we have the Main UI workflows that run them as well and do some additional stuff as well.

bundles/org.openhab.ui/pom.xml
florian-h05: I am neither happy with prepare-package nor with compile, I think my latest selection is best in line with how the phases should be used (see https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html):

- initialize: Install Node and npm
- prepare-sources: Install dependencies
- prepare-resources: Build Main UI

florian-h05:
@Nadahar From my POV this is finished. Let me know what you think and I can merge it soon.

Nadahar:
I'll do some final checks/tests and come back.

bundles/org.openhab.ui/pom.xml
Nadahar: I take a more "pragmatic approach" than merely looking at "correctness" when it comes to the Maven phases. They are run in a specific sequence (according to the life cycle), and my primary concern is where something is done.

Maven is "Java centric" in that it designed around building a Java project, so these phases must be seen in that context. I'm not sure where you find `prepare-sources` and `prepare-resources`, but in case you mean `process-` , those are phases, but IMO they aren't right ones. They are both before the `compile` phase, which means that they would both run if somebody execute `mvn compile`. 

Both "sources" and "resources" are Java-specific in this context, a Java resource is a resource that is "magically available" for the Java code, and thus they must be prepared before compilation. The same goes for the sources of course.

In this context, the "web files" are just something that is "bundled in the JAR" that is otherwise "unrelated" to the Java build process. Therefore I find it best to do them as late as possible before they are needed, so that they are included in a few goals as possible (where they don't do anything "useful" anyway).

florian-h05: > I'm not sure where you find prepare-sources and prepare-resources

Before the `compile` phase.

> In this context, the "web files" are just something that is "bundled in the JAR" that is otherwise "unrelated" to the Java build process. 

Why would you compile the Java part of Main UI without having the actual Main UI resources available?

Nadahar: If you change anything in the Java code (there are some Java code there as well, running the servlet), the Java code must be recompiled (generate new .class files). That can be for syntax checking, code analysis etc., and doesn't need the JAR file "to be useful".

Nadahar: Just for information, all goals have default phases, specified by the Maven plugin. If the phase isn't stated, the default phase is used, which explains what phases are currently used (`generate-resources` for both goals in use):

https://github.com/eirslett/frontend-maven-plugin/blob/b7b060fa13ca47e69172ee6d06194ffd41c35084/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java#L18

https://github.com/eirslett/frontend-maven-plugin/blob/b7b060fa13ca47e69172ee6d06194ffd41c35084/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/NpmMojo.java#L17

Nadahar: `Install Node and npm` is skipped effectively if not needed, so it doesn't matter that much when that is done. The others take time so they do matter, and I think it's "safer" to have them in different phases to make sure that they're always run in the right order, even though having them in the same phase usually seems to work in that they are invoked in the order they are declared in the POM. I'm just thinking that it might not always be the case that one is finished before the other starts in multithreaded builds, so having them in different phases is probably useful to make sure that dependencies are done installing before the "webpack" run is started.

I left `npm install` at `generate-resources` for that reason, even though it's before `compile` and as such, less than ideal. It looks like it should be possible to run this at `process-classes` instead, to still have it run before the "webpack run" yet still after `compile`. All the "test" phases are off-limits, because the build is often done with tests skipped, which skips all those phases.

bundles/org.openhab.ui/pom.xml
```diff
@@ -48,16 +61,17 @@
             <goals>
               <goal>install-node-and-npm</goal>
             </goals>
-            <phase>generate-resources</phase>
+            <phase>initialize</phase>
```
Nadahar, line 64:
```suggestion
```

bundles/org.openhab.ui/pom.xml
```diff
@@ -66,28 +80,9 @@
...
-      </plugin>
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>build-helper-maven-plugin</artifactId>
-        <executions>
-          <execution>
-            <goals>
-              <goal>add-resource</goal>
-            </goals>
             <phase>generate-resources</phase>
```
Nadahar, line 83:
```suggestion
            <phase>prepare-package</phase>
```

bundles/org.openhab.ui/pom.xml
```diff
@@ -48,16 +61,17 @@
...
+            <phase>initialize</phase>
           </execution>
 
           <execution>
-            <id>npm ci</id>
+            <id>npm install</id>
             <goals>
               <goal>npm</goal>
             </goals>
+            <phase>generate-sources</phase>
```
Nadahar, line 72:
```suggestion
            <phase>process-classes</phase>
```

Nadahar (approved):
I've been testing, and this is working as far as I can tell.

I do suggest changing the phases though, because I can't see any benefit of including the generation of the "web files" earlier in the life-cycle than when they are needed.

I've tested when running OH from Eclipse, in case the files were accessed directly from the `target/classes/app` folder, but they are not. Deleting these files have no effect, while deleting the JAR file breaks the web UI. So, as far as I can tell, the only reason the files in `target/classes/app` exists is so that they become a part of the JAR file. Therefore, it makes no sense to generate them if the JAR file isn't generated.

This is the result of `mvn compile` with the phases as they are now:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:40 min
[INFO] Finished at: 2025-03-26T04:30:22+01:00
[INFO] ------------------------------------------------------------------------
```

This is the result of `mvn compile` on the same machine with under the same circumstances with my suggested phases:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  32.727 s
[INFO] Finished at: 2025-03-26T04:42:43+01:00
[INFO] ------------------------------------------------------------------------
```

The same difference will apply if running all other goals that terminates earlier than `package`, like `mvn test` for example.

florian-h05:
> I do suggest changing the phases though, because I can't see any benefit of including the generation of the "web files" earlier in the life-cycle than when they are needed.

I decided to follow your paragmatic approach and take it one step further: I run all Main UI related tasks in the `prepare-package` phase, which works fine when testing locally.

PR #3123, opened 2025-03-31
Persistence no default

mherwege:
Depends on: https://github.com/openhab/openhab-core/pull/4682

This PR removes the default strategy from the persistence configuration and implements a health check on the persistence configuration.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/health/health-overview.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/health/health-overview.vue
@@ -16,7 +16,7 @@
       </f7-col>
     </f7-block>
 
-    <f7-block class="block-narrow">
+    <f7-block v-if="orphanLinksCount" class="block-narrow">
       <f7-col>
         <f7-list media-list>
           <f7-list-item
@@ -32,6 +32,22 @@
         </f7-list>
       </f7-col>
     </f7-block>
+    <f7-block v-if="persistenceProblemsCount" class="block-narrow">
+      <f7-col>
+        <f7-list media-list>
+          <f7-list-item
+            media-item
+            link="persistence/"
+            title="Persistence Problems"
+            :badge="persistenceProblemsCount > 0 ? persistenceProblemsCount : undefined"
+            :after="persistenceProblemsCount > 0 ? undefined : persistenceProblemsCount"
+            :badge-color="persistenceProblemsCount ? 'red' : 'blue'"
+            :footer="objectsSubtitles.persistenceProblems">
+            <f7-icon slot="media" f7="link" color="gray" />
+          </f7-list-item>
+        </f7-list>
+      </f7-col>
+    </f7-block>
   </f7-page>
 </template>
 
@@ -40,10 +56,11 @@ export default {
   data () {
     return {
       objectsSubtitles: {
-        orphanLinks:
-          'Items pointing to non-existent thing channels or vica versa'
+        orphanLinks: 'Items pointing to non-existent thing channels or vica versa',
+        persistenceProblems: 'Persistence configurations missing items or strategies'
       },
       orphanLinksCount: '',
+      persistenceProblemsCount: '',
 
       expandedTypes: {
         systemSettings: this.$f7.width >= 1450
@@ -62,10 +79,16 @@ export default {
   },
   methods: {
     loadCounters () {
+      let self = this
       if (!this.apiEndpoints) return
       if (this.$store.getters.apiEndpoint('links')) {
         this.$oh.api.get('/rest/links/orphans').then((data) => {
-          this.orphanLinksCount = data.length.toString()
+          self.orphanLinksCount = data.length.toString()
+        })
+      }
+      if (this.$store.getters.apiEndpoint('persistence')) {
+        this.$oh.api.get('/rest/persistence/persistencehealth').then((data) => {
+          self.persistenceProblemsCount = data.length.toString()
         })
       }
     },
--- a/bundles/org.openhab.ui/web/src/pages/settings/persistence/strategy-picker.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/persistence/strategy-picker.vue
@@ -2,9 +2,9 @@
   <f7-list class="strategy-picker-container" v-if="strategies">
     <f7-list-item :title="title" :smart-select="disabled !== true" :smart-select-params="smartSelectParams"
                   ref="smartSelect" class="defaults-picker">
-      <select v-if="disabled !== true" :name="name" multiple @change="select">
+      <select v-if="disabled !== true" :name="name" multiple>
         <option v-for="s in strategies" :key="s" :value="s"
-                :selected="value.includes(s)">
+                :selected="value.length ? value.includes(s) : suggested.includes(s)">
           {{ s }}
         </option>
       </select>
@@ -28,8 +28,10 @@
 </style>
 
 <script>
+import cloneDeep from 'lodash/cloneDeep'
+
 export default {
-  props: ['title', 'name', 'strategies', 'value', 'disabled'],
+  props: ['title', 'name', 'strategies', 'value', 'suggested', 'disabled'],
   emits: ['strategiesSelected'],
   data () {
     return {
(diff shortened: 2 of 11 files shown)


kaikreuzer:
@florian-h05 Might I ask you for a 10 minute speed review? If there isn't anything that you think is a problem, I would merge it now together with https://github.com/openhab/openhab-core/pull/4682.

kaikreuzer (approved):
@florian-h05 As you seem to be afk right now, allow me to merge - if there's any issue with it, let's fix it in a follow-up PR.

bundles/org.openhab.ui/web/src/pages/settings/health/health-overview.vue
```diff
@@ -94,6 +115,11 @@ export default {
           this.semanticsProblemCount = data.length || 0
         })
       }
+      if (this.$store.getters.apiEndpoint('persistence')) {
+        this.$oh.api.get('/rest/persistence/persistencehealth').then((data) => {
+          self.persistenceProblemsCount = data.length || 0
```
florian-h05: Self shouldn’t be required in an arrow function, they don’t bind their own this.

bundles/org.openhab.ui/web/src/pages/settings/persistence/strategy-picker.vue
```diff
@@ -37,29 +37,62 @@
...
     disabled: Boolean
   },
   emits: ['strategies-selected'],
   data () {
     return {
       smartSelectParams: {
         view: f7.view.main,
-        openIn: 'popup'
+        openIn: 'popup',
+        virtualList: true,
```
florian-h05: Why use a virtual list? We don't need to render a huge amount of entries, so we shouldn't need a virtual list here.

helmar74:
Could someone hav alook in my comments here:
https://community.openhab.org/t/openhab-5-1-milestone-discussion/166385/217

I think there'S a critical issue with persistence settings in RC1
Looks like filebased configuration is not used anymore

lolodomo:
> Looks like filebased configuration is not used anymore

Mine seems to be still used.

helmar74:
> > Looks like filebased configuration is not used anymore
> 
> Mine seems to be still used.

Can you share a screenshot from WebUI Persistence settings?

kaikreuzer:
> Looks like filebased configuration is not used anymore

From the logs that you provided, it seems that the upgrader now adds a managed configuration despite the fact that you already have a textual configuration in place and during startup it picks the "new" managed config and thus it does not regard your "real" textual one anymore:

```
2025-12-20 09:51:24.172 [WARN ] [ore.common.registry.AbstractRegistry] - Cannot add "PersistenceServiceConfiguration" with key "rrd4j". It exists already from provider "ManagedPersistenceServiceConfigurationProvider"! Failed to add a second with the same UID from provider "PersistenceModelManager"!
```
If that's the case, that's indeed a very critical issue!

helmar74:
Yes it is. Should I delete the added configuration in WebUI in order to fix?

kaikreuzer:
I would guess it will resolve your issue, yes. So sure, please have a try. But we will have to investigate this...

lolodomo:
> Yes it is. Should I delete the added configuration in WebUI in order to fix?

Stop OH and then look at in userdata/jsondb a find the json file that contains the persistence configuration. Delete this file and then restart openHAB. Sorry I have not in mind the name of the generated file.

Edit: file is org.openhab.core.persistence.PersistenceServiceConfiguration.json

I have not such file myself.

kaikreuzer:
I can confirm the issue. I have a textual inmemory.persist file. When calling the upgradetool with `--force` option, it logs:
```
[main] INFO org.openhab.core.tools.internal.PersistenceUpgrader - inmemory: added strategy configurations for persistence service without configuration
```
which should not happen.

During startup, I then get the warning:
```
2025-12-20 11:35:05.263 [WARN ] [ore.common.registry.AbstractRegistry] - Cannot add "PersistenceServiceConfiguration" with key "inmemory". It exists already from provider "PersistenceModelManager"! Failed to add a second with the same UID from provider "ManagedPersistenceServiceConfigurationProvider"!
```

Looking at the PR that I merged yesterday, it indeed does not respect textual configs:

https://github.com/openhab/openhab-core/pull/5212/changes#diff-8e4cccaa92760e57af3bcc119fd6d1e731b88ba3c76582021d080152789242baR83-R88

helmar74:
For you information.
I deleted rrd4j and mapdb persistence configuration in WebUI, now filebased configuration is used.
Just as "historical" information:
1. I updated from 5.0.3 (production) to 5.1.0M4 on monday
2. Today I updated from 5.1.0M4 to 5.1.0RC1

What I also saw, that the permissions for /var/lib/openhab/jsondb/org.openhab.core.persistence.PersistenceServiceConfiguration.json changed to root:root:

`-rw-r--r-- 1 root    root       1237 20. Dez 09:05 org.openhab.core.persistence.PersistenceServiceConfiguration.json`

Maybe this happend thru update to RC1 and caused the issue?
... running on Raspberry PI with openhabian installation. Updates where done with apt update && apt upgrade

Here's my startup log after deleting persistence settings in WebUI:
```
2025-12-20 11:42:06.957 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading DSL model 'telegram.things'
2025-12-20 11:42:07.164 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading DSL model 'mapdb.persist'
2025-12-20 11:42:07.268 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading DSL model 'rrd4j.persist'
2025-12-20 11:42:07.499 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading DSL model 'Testrule.rules'```

Just ping me, if I should try something. Would be happy to contribute here

kaikreuzer:
> Looking at the PR that I merged yesterday, it indeed does not respect textual configs:
>
> https://github.com/openhab/openhab-core/pull/5212/changes#diff-8e4cccaa92760e57af3bcc119fd6d1e731b88ba3c76582021d080152789242baR83-R88

This was not the right reference, but this is the place where the file is created with root user rights; this might need to be fixed as well...

Not sure yet, why unmanaged configs are not correctly identified and managed configs are created. Need to investigate.

mherwege:
The upgrader tool should only add a managed persistence configuration for the specific service if there is no .persist file for the service. If there was a persist file and it still does, there still is a bug in the changes to the upgrade tool done yesterday. The upgrade tool lists the existing .persist files and should not add a managed persistence file for these.

mherwege:
> > Looking at the PR that I merged yesterday, it indeed does not respect textual configs:
> > https://github.com/openhab/openhab-core/pull/5212/changes#diff-8e4cccaa92760e57af3bcc119fd6d1e731b88ba3c76582021d080152789242baR83-R88
> 
> This was not the right reference, but this is the place where the file is created with root user rights; this might need to be fixed as well...
> Not sure yet, why unmanaged configs are not correctly identified and managed configs are created. Need to investigate.

The exclusion of the unmanaged ones should be done on line 70 in that file.

florian-h05:
I've found the bug and seems I have a fix for it.

florian-h05:
See https://github.com/openhab/openhab-core/pull/5213.

kaikreuzer:
> What I also saw, that the permissions for /var/lib/openhab/jsondb/org.openhab.core.persistence.PersistenceServiceConfiguration.json changed to root:root

@helmar74 When you say "changed to root:root", does this mean that this file already existed before? Does it contain any information?

helmar74:
As far as I can see, the file wasn't there with version 5.0.3 Prod:
image


it wasn't there with version 5.1.0M4:
image


So I think it was created with Update to 5.1.0 RC

kaikreuzer:
> So I think it was created with Update to 5.1.0 RC

Ok, that's what I expected. So permissions were not changed, but the newly created file was simply created by the root user.

helmar74:
Yes it looks like. Update was done with 
`sudo apt upgrade`
as root right's are required on openhabian based installations for apt upgrade

kaikreuzer:
I would imagine that this can be an issue, because the runtime might not have write permission to that file, so doing changed through the Main UI might fail.
I am not aware of any other upgrader that already creates new files and how we could deal with setting correct permissions.
The situation here is hopefully a not so common edge case, so it might not be critical to be fixed in 5.1 today.
But if you have any insights or opinions, @mherwege, @florian-h05, @mstormi or others, please chime in!

mherwege:
It looks like @florian-h05 beat me with a solution on this. The change in distro is probably better as it effectively changes ownership. My solution just changes file permissions.

florian-h05:
I just wanted to comment that I've created https://github.com/openhab/openhab-distro/pull/1847 but you were faster to comment.
I decided to modify the update script in distro as permissions and ownership are OS-specific and I don't want to mess with that in the Java code. In distro, we already have working scripts for handling ownership, let's modify them instead.

PR #3124, opened 2025-04-02
Create equipment from model: Use semanticEquipmentTag of Thing

mherwege:
Depends on https://github.com/openhab/openhab-core/pull/4617

If the thing provides a default `semanticEquipmentTag`, this will be used when creating an equipment from thing instead of the default `Equipment`.

@andrewfg FYI


florian-h05 (approved):
LGTM, thanks!

PR #2129, opened 2023-10-10
Thing and triggerchannel pickers

d51x:
discussion #2128

PR #3126, opened 2025-04-05
Thing & Channel pickers: Add Thing location/Channel description

### LINKED ISSUE #2128: Thing- and TriggerChannel- pickers with footer text
## The problem

When you have many things and some channels with only names then it is very difficult to looking for a thing / channel when creating/editing rules/scripts

## My suggestion

* modify Thing picker to show locations for things in it. It's allow simplify navigate trough list of things
* modify TriggerChannel picker to show channels description for every channel if has. It gives me a better understanding of what kind of channel it is

But these modifications are only available with redefining of renderItem method of f7SmartSelect, because ` <select><option>` doesn't allow any html-tags and makes a plain text from html

My PR is #2129 

### Thing Picker with location in footer 

image

### TriggerChannel picker with channel description in footer

image
image

florian-h05:
Closes #2128.
Replaces abandonded PR #2129.

Displays the Thing location (if any) in the Thing picker and the channel description in the channel picker,
which improves UX greatly as it allows to distinguish Things easier and know the channel functionality without having to lookup documentation.


ghys:
How does it look in the end? Please post some screenshots ;)

florian-h05:
There you go:

image

image

PR #3130, opened 2025-04-06
File Definition: Add Copy Thing YAML file definition

jimtng:
To support https://github.com/openhab/openhab-core/pull/4691

- Refactor file definition code into a mixin to create a consistent UI in inbox, things-list, items-list, item-details, thing, details, single, and multi-selection.
- Add `Copy` button in Things List which shows up when multi-selection is active. This makes it in line with Items list and Inbox
- Standardize the list selection ui for inbox, things, items list. Action buttons/links are closer together in the center now.
- Remove the selection counter off the buttons to make room. The selection counter is shown in the list title.
- Use one "Copy" button which opens a dialog to select the file definition format to export/copy to clipboard

image

image

image

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/file-definition-mixin.js
+++ b/bundles/org.openhab.ui/web/src/pages/settings/file-definition-mixin.js
@@ -3,6 +3,43 @@ import Clipboard from 'v-clipboard'
 
 Vue.use(Clipboard)
 
+function executeFileDefinitionCopy (vue, type, typeLabel, objectIds, copiedObjectsLabel, fileFormatLabel, mediaType) {
+  const progressDialog = vue.$f7.dialog.progress(`Loading ${typeLabel} ${fileFormatLabel} definition...`)
+
+  const path = `/rest/file-format/${type}s`
+  let apiCalls = []
+  if (objectIds !== null) {
+    apiCalls = objectIds.map((id) => vue.$oh.api.getPlain({
+      url: path + '/' + id,
+      headers: { accept: mediaType }
+    }))
+  } else {
+    apiCalls = [vue.$oh.api.getPlain({
+      url: path,
+      headers: { accept: mediaType }
+    })]
+  }
+
+  Promise.all(apiCalls)
+    .then(definitions => {
+      const definition = definitions.join('\n')
+      progressDialog.close()
+      if (vue.$clipboard(definition)) {
+        vue.$f7.toast.create({
+          text: `${typeLabel} ${fileFormatLabel} definition copied to clipboard:\n${copiedObjectsLabel}`,
+          destroyOnClose: true,
+          closeTimeout: 2000
+        }).open()
+      } else {
+        vue.$f7.dialog.alert(`Error copying ${typeLabel} ${fileFormatLabel} definition to the clipboard`, 'Error')
+      }
+    })
+    .catch(error => {
+      progressDialog.close()
+      vue.$f7.dialog.alert(`Error copying ${typeLabel} ${fileFormatLabel} definition: ${error}`, 'Error')
+    })
+}
+
 export default {
   created () {
     // Define the ObjectType enum to be used when calling the copyFileDefinitionToClipboard method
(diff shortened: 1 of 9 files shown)


lolodomo:
Core PR is now merged.

jimtng:
@florian-h05 could we merge this before M2 release?

Currently it will offer YAML format everywhere a DSL format is offered. This means when selecting YAML for Items file, it will issue an error (from the server) "Not Acceptable", but once the YAML Item REST is implemented in core, it will simply return it without any further modifications on the mainui (hopefully).

jimtng:
@lolodomo I've posted some screenshots in the original post above.

lolodomo:
I imagine we also have the popup to select the file format when we are on the setup page of a thing ?

jimtng:
> I imagine we also have the popup to select the file format when we are on the setup page of a thing ?

What are you referring to as "setup page"? Please elaborate. Did you mean Thing-details? If so, then yes. 

In general everything has "Copy File Definition" (or just "Copy" where applicable). When that is clicked, the popup dialog with `cancel, dsl, yaml` would pop up. This was refactored and collected into a mixin that everyone calls into, so that they all behave the same way.

jimtng:
@lolodomo An idea just came to mind - this is for the future, not this current implementation.

In MainUI, the user can create a new Thing, or Item as it may, by entering the details into the Thing-Creation/Item Creation form, then clicking "Create Thing", e.g.
image

Perhaps we can also offer a button that says "Copy File Definition" next to "Create Thing". Basically we're using the UI to construct a file definition for us, without using Managed Thing / Item.

Of course the implementation would need to be figured out. Right now the /rest/file-format/ can only do so based on an actual existing thing/item, so this wouldn't work. Or it could do a quick `Create -> Copy -> Delete` to get around that without adding anything more in core, this would be very easy for me to implement in the UI - for another PR if we like this idea.

WDYT?

lolodomo:
@jimtng this will be possible and easy once this core PR is finished:
https://github.com/openhab/openhab-core/pull/4630
Don't look at its current state as I plan to change it to only one new API able to transform from an input format to an output format. Supported formats will include DSL file format, YAML file format and a REST JSON format.

bundles/org.openhab.ui/web/src/pages/settings/file-definition-mixin.js
```diff
@@ -0,0 +1,97 @@
+import Vue from 'vue'
+import Clipboard from 'v-clipboard'
+
+Vue.use(Clipboard)
+
+export default {
+  data () {
+    return {
+      ObjectType: Object.freeze({
```
florian-h05, line 9: As this is a constant it is better to set it in the created method, data is mainly meant for reactive data:

```js
this.OBJECT_TYPE = ....
```

bundles/org.openhab.ui/web/src/pages/settings/file-definition-mixin.js
```diff
@@ -0,0 +1,97 @@
...
+              text: 'YAML',
+              color: 'blue',
+              onClick: () => this.executeFileDefinitionCopy(type, typeLabel, objectIds, copiedObjectsLabel, 'YAML', 'application/yaml')
+            }
+          ]
+        })
+        .open()
+    },
+    // This is a "private" method that is called by the copyFileDefinitionToClipboard method
+    executeFileDefinitionCopy (type, typeLabel, objectIds, copiedObjectsLabel, fileFormatLabel, mediaType) {
```
florian-h05, line 60: You can try to define this as `function` before defining export default, this way it is really private.

florian-h05 (changes requested):
Overall LGTM, I haven't had the chance to test it though.
Please check my comments.

bundles/org.openhab.ui/web/src/pages/settings/file-definition-mixin.js
jimtng: thanks for the tip. done.

bundles/org.openhab.ui/web/src/pages/settings/file-definition-mixin.js
jimtng: how do I have access to `this`? I suppose the caller can pass it?

jimtng: I've moved it out

jimtng:
@lolodomo I've just managed to actually tried this against the core PR. I noticed that when I selected multiple Things, I got repeated `version: 2` and `things:` preamble - one for each thing. Example:

```
version: 2
things:
  mqtt:homeassistant:broker2:5002915a4df9:
    bridge: mqtt:broker:broker2
    config:
      topics:
        - sensor/bedroom3-switch/esphome_version
        - sensor/bedroom3-switch/ip-address
        - sensor/bedroom3-switch/rssi
        - sensor/bedroom3-switch/signal
        - sensor/bedroom3-switch/uptime
        - switch/bedroom3-switch/reboot
        - switch/bedroom3-switch/switch1

version: 2
things:
  mqtt:homeassistant:broker2:983dae500fa4:
    bridge: mqtt:broker:broker2
    config:
      topics:
        - sensor/doorbell/esphome_version
        - sensor/doorbell/ip-address
        - sensor/doorbell/rssi
        - sensor/doorbell/signal
        - sensor/doorbell/uptime
        - switch/doorbell/reboot
```

Should we exclude the `version` and `thing` keys from the generated yaml in core, so they don't get sent out at all?

Or do you think they should get filtered out on the UI side and only allow the first to be included?

lolodomo:
Because you called the API once for each item.
We don't have currently an API to generate the DSL or YAML file format for a list of things, we just have for all things in registry or for only one thing.

lolodomo:
Until this is enhancement, the best could be in UI to not provide the copy feature for a list of items/things.

lolodomo:
> Or do you think they should get filtered out on the UI side and only allow the first to be included?

I think we should just enhance the API.

lolodomo:
> Until this is enhancement on API side, the best could be in UI to not provide the copy feature for a list of items/things when more than one is selected.

Or just change nothing, this is not the final version, we can change that later.

jimtng:
It seems that both core and ui side need to be fine tuned further for multiple things request.

I agree, I'd rather have this merged now, especially since it includes a few other improvements, and then refine it later.
If we wait, resolving merge conflicts later will be harder.

florian-h05 (approved):

florian-h05:
Then let's merge now, and you can open follow-up PRs to fix those remaining issues. I think it still is a nice enhancement, even without being perfect yet, many thanks @lolodomo and @jimtng!!

PR #3142, opened 2025-04-08
Remove ESLintPlugin from webpack build

Nadahar:
Removed `ESLintPlugin` execution from the webpack build as discussed [here](https://community.openhab.org/t/using-eclipse-with-openhab-webui/162923/30).

@florian-h05 I'm not sure if this is how you envisioned it, we could remove `VueLoaderPlugin()` completely, but I figured that it might be better to leave it there as a "skeleton" if needed in the future?


florian-h05 (changes requested):
Please remove it from the package.json as well.

WRT to the VueLoaderPlugin: I would keep it as skeleton.

Nadahar:
> Please remove it from the package.json as well.

@florian-h05 I've never modified `package.json` before, so I'm not sure that I did it correctly, as `package-lock.json` must be changed correspondingly. I let `npm install` do the changes to `package-lock.json`, and I am hoping that it turned out correct. I've tried running the lint manually, building for production and starting the dev server, and those all worked fine, so hopefully it is OK.

PR #3143, opened 2025-04-09
[BasicUI] Shrink labels before values

### LINKED ISSUE #3094: [BasicUI] On narrow screens show the value uncut, strip only the label
For a sitemap with BasicIU if the space is not enough to show both label and value, in OH 4.2.2 the value is cut.

Here with 9 A there is enough space:

Image

For 10 A the horizontal space is not enough:

Image

I think it is more important to show the value completely and use the remaining space to show the label, instead of cutting both.

jimtng:
Resolve #3094

Here's an animated gif:
label-value-width

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -224,14 +224,18 @@
 		}
 		@include flex-shrink(0);
 		@include flex-grow(2);
-		// Set flex-basis to 0 to prioritize values over labels (see __text below)
-		@include flex-2011(2 0 0);
+		@include flex-2011(2 2 auto);
 		white-space: nowrap;
 		overflow: hidden;
 		text-overflow: ellipsis;
 		html.ui-bigger-font & {
 			font-size: 18px;
 		}
+	}
+	&__label:has(+ .mdl-form__text) {
+		// this applies to labels that are followed by a text input
+		// Set flex-basis to 0 to prioritize __text over __label (see __text below)
+		@include flex-2011(2 0 0);
 	}
 	&__label-multiline {
 		min-width: 3em;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -224,7 +224,8 @@
 		}
 		@include flex-shrink(0);
 		@include flex-grow(2);
-		@include flex-2011(2 2 auto);
+		// Set flex-basis to 0 to prioritize values over labels (see __text below)
+		@include flex-2011(2 0 0);
 		white-space: nowrap;
 		overflow: hidden;
 		text-overflow: ellipsis;
@@ -501,7 +502,10 @@
 			font-size: 18px;
 		}
 		overflow: hidden;
+		// Make the label shrink first before the text (see __label above)
+		// See https://stackoverflow.com/a/67858285/15409057
 		flex-shrink: 2;
+		flex-grow: 0;
 		text-overflow: ellipsis;
 	}
 	&__webview,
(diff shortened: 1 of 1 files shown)


lolodomo:
Thank you for that PR.
I will review and test.
The change is limited to text elements?

jimtng:
> The change is only limited to text elements?

I think so, but it's best to test and make sure everything is still fine.

I'd really love to see a "standard" sitemap containing as complete combination of things as possible, included in `CONTRIBUTING.md` so we can all use it as a basis for testing.

lolodomo:
You applied a change also to the label rendering so with a potential impact on any sitemap element.
So it requires tests to be sure that other widgets than the text widget are not impacted.

jimtng:
I tried it with a bunch of sitemap elements:

Before:
image


After:
image


It event made the multiline buttons look better

jimtng:
So the label is smaller when coupled with: text input, slider, multiline buttons

jimtng:
Also noted (unrelated to this issue) the Value is a bit far from the slider. It should be pushed closer to it I think.

jimtng:
For the record, I used this JRuby script to generate the test sitemap:

```ruby
items.build do
  group_item TestGroup do
    string_item TestLongString, state: "abc" * 10
    string_item TestString, state: "ABC"
    number_item TestNumber, state: 30
    switch_item TestSwitch
    number_item ColorTemp, state: 3000 | "K"
  end
end

label = "Test 3 abcd efgh ijkl mno pqr stuv wxyz [%s]"

sitemaps.build do
  sitemap "test2" do
    default item: TestLongString, label: label
    switch item: TestSwitch, label: label
    switch item: TestSwitch, label: label, mappings: %w[ON OFF ABC DEF AAA BBB CCCCC]
    switch item: TestString, label: label, mappings: %w[ON OFF ABC DEF AAA BBB CCCCC]
    group item: TestGroup, label: label
    selection item: TestString, label: label, mappings: %w[ON OFF ABC DEF AAA BBB CCCCC]
    setpoint item: TestString, label: label
    slider item: TestNumber, label: label
    colortemperaturepicker item: ColorTemp, label: label
    input item: TestString, label: label
  end
end

```

jimtng:
@lolodomo would you like me to work on this further to keep the impact only to text widget, or do you think the impact on other widgets are acceptable?

lolodomo:
I have not yet a clear opinion!
What is your opinion ?

jimtng:
as far as I can see, 3 other widgets are affected:
- sliders
- input field
- multiline buttons
 
I think we should leave this as is, and make another PR to shrink the slider and input field a little. 

Although I'm unsure about slider, because I think it's actually nicer to have a longer slider (finer control). 

As for the input field, I think the ideal solution would be the ability to specify its width (in terms of approx how many characters maybe - translates to `em`), but that would be a much bigger change.

Another alternative is to allow for label above the controls, then that would solve all these problems, including the extremely complex adjustment we did for multiline buttons in the past. Perhaps at least offer it as an option? I think android does it this way.

The impact on multiline buttons seem to actually be positive.

lolodomo:
Is the slider size variable with that change, depending on label?
We should keep a fixed size.

jimtng:
The next commit will make it so this only changes labels that are followed by text, so it won't affect any other combinations at all.

Before:
image


After:
image

bundles/org.openhab.ui.basic/web-src/_layout.scss
```diff
@@ -232,6 +232,11 @@
 			font-size: 18px;
 		}
 	}
+	&__label:has(+ .mdl-form__text) {
+		// this applies to labels that are followed by a text input
```
lolodomo, line 236: Not "text input".
Maybe rather something like "this applies only to labels for sitemap text element" ?

lolodomo:
> The next commit will make it so this only changes labels that are followed by text, so it won't affect any other combinations at all.

Yes, let's go in that direction and at least we will be able to merge this one quickly.

I have just a comment about one of your added comment.

bundles/org.openhab.ui.basic/web-src/_layout.scss
jimtng: comment fixed. thanks for catching it.

lolodomo (approved):
LGTM, thank you

PR #3144, opened 2025-04-09
Items List: Add select all / deselect all

jimtng:


florian-h05 (approved):
LGTM thanks!
I have refactored the code to use computed data instead of methods, this is the cleaner way, and I fixed the problem that computed initially did not work.

PR #3147, opened 2025-04-11
[BasicUI] Add a setting to render buttons with rounded corners

jimtng:
image

image

Default:
image

Rounded:
image

Pill:
image


lolodomo:
I will review and test.

jimtng:
> Does it apply to all buttons including buttons in a buttongrid?

yes, buttongrid and switch buttons



> Does this also change the rendering of buttons in the settings page itself?

No but that's a good idea. I've added this.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/PageRenderer.java
```diff
@@ -303,6 +303,17 @@ public CharSequence renderSettings() throws RenderException {
                 localizeText("@text/ui.config.basic.adjustedColors.description"), "adjustedColors",
                 "openhab.ui.basic:adjustedColors", "enabled", "disabled", true, sb);
 
+        buttons = new StringBuilder();
+        buildButton("Default", "", buttons);
+        buildButton("Rounded", "rounded", buttons);
+        buildButton("Pill", "pill", buttons);
```
lolodomo, line 309: Please localize the buttons label.

lolodomo: I have no idea how I will translate "Pill" in French, I do not understand what it means in our context.

lolodomo:
Great, it works perfectly.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/PageRenderer.java
jimtng: I've localized them.

There's a picture at the _bottom_ of this page that shows a "Pill"
https://www.havardbrynjulfsen.design/writing/tips/pill-buttons

Posting some other photos here in case that link goes away
image

jimtng: Maybe I should call it "Pill-Shaped"? But that might make the translation (e.g. German) to be even longer. I tried google translate but I think a native speaker could pick a better translation.

lolodomo: Ok, I see but I am not sure it will be clear for everybody.
Why not rather "Light rounded" for rounded and "More rounded" for pill ?

lolodomo: And default could be replaced by "square" ?

jimtng: How about this:

image

lolodomo: Do we keep "rounded" and "pill" as values for the local storage ?
Or do we align them a little with the new labels ?

jimtng: I've changed the value names to say "fully-rounded". Also found a bug when "Square" is selected, `ui-button-null` class was incorrectly set.

lolodomo (approved):
LGTM, thank you

PR #3148, opened 2025-04-12
[BasicUI] Remove top/bottom margins in buttons-multiline

jimtng:
Before:
image
condensed mode:
image

After:

image

condensed mode:
image

```java
sitemap test {
    Text item=TestString1 label="Test"
    Switch item=TestSwitch1 label="Test" mappings=[ON="ON Label", OFF="OFF label"]
    Text item=TestString1  label="Test"
    Switch item=TestSwitch1 label="Test" mappings=[ON="ON Label", OFF="OFF label"]
}
```

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -515,10 +515,8 @@
 		padding-bottom: 2px;
 	}
 	&__buttons-multiline {
-		margin: 6px 0;
 		gap: 4px;
 		html.ui-layout-condensed & {
-			margin: 0;
 			gap: 3px;
 		}
 		padding-top: 2px;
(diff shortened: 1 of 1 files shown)


lolodomo:
I never saw that...
Did you check when condensed is applied and when it is not applied ?

jimtng:
> Did you check when condensed is applied and when it is not applied ?

Ohh.... the current fix doesn't fix "condensed" mode!

lolodomo:
> Ohh.... the current fix doesn't fix "condensed" mode!

Yes but as the height is different in condensed mode, it seems to me important to check the result in both modes.

jimtng:
condensed mode fixed now.

lolodomo (approved):
LGTM, thank you

lolodomo:
I am not sure if this is the consequence of your change but here is a strange rendering I just discovered.
We loose the alignment on one line.

image

My feeling is that it was not like that before but I am not 100% sure.

jimtng:
I compiled the version before my PR:
image

And this is with the PR
image

PR #3149, opened 2025-04-14
Item edit: Relabel category to icon

### LINKED ISSUE #3146: Rename "category" to "icon" for item
Currently the field ``Category`` in the item context is used to configure an icon.

Image

Since this is not a category but an icon it should be renamed accordingly to make things more clear.
See also [docs](https://www.openhab.org/docs/configuration/items.html#icons)

spacemanspiff2007:
Closes #3146.

Changed label from "Category" to "Icon" to make usage more clear.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -52,7 +52,7 @@
         <group-form ref="groupForm" v-if="itemType === 'Group'" :item="item" :createMode="createMode" />
       </f7-list-group>
       <f7-list-group v-if="!hideCategory">
-        <f7-list-input ref="category" label="Category" autocomplete="off" type="text" placeholder="temperature, firstfloor..." :value="itemCategory"
+        <f7-list-input ref="category" label="Icon" autocomplete="off" type="text" placeholder="temperature, firstfloor..." :value="itemCategory"
                        @input="itemCategory = $event.target.value" :disabled="!editable" :clear-button="editable">
           <div slot="root-end" style="margin-left: calc(35% + 14px)">
             <oh-icon :icon="itemCategory" :state="(createMode || itemType === 'Image') ? null : item.state" height="32" width="32" />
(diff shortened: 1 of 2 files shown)


florian-h05:
Can you please link the related core PR?

spacemanspiff2007:
This just changes a description which is UI only. What would you expect a core PR to do?

florian-h05:
Ah sorry, thought the ItemDTO has changed.

spacemanspiff2007:
I'd love to change that, too - but that's way above my skill level :wink: 
So I did what I could and that's the description

florian-h05:
Can you please also update `toYaml` and `fromYaml` in `item-edit.vue` accordingly?

florian-h05:
Looks good 👍 

I don’t think it makes sense to implement backward compatibility, as the UI YAML mainly is a representation format and nothing that actually stores the data. I would consider the UI YAML more like an additional „UI“ than a file format.

florian-h05 (approved):
LGTM, thanks and congratulations to your first Main UI contribution :tada:

florian-h05:
@spacemanspiff2007 Can you please provide a proper sign-off, i.e. containing your real name and a mail address where you can be reached (i.e. no GitHub no-reply)?

lolodomo:
This should be coordinate with the change in core framework. If not, this will lead to UI displaying icon while the REST API is providing category. Not very consistent.

florian-h05:
Is there a core change in the making for this?

spacemanspiff2007:
I've created an [issue](https://github.com/openhab/openhab-core/issues/4716) to request a rename of the field in the DTO but so far no comment.

The description/label is for understanding what the field does and thus does not have to be the same like the field name in the DTO. It's more important that the UI aligns with the current docs than with the DTO and the docs call this field [icon](https://www.openhab.org/docs/configuration/items.html#icons).
There are also many cases where the UI does not align with the DTO field name (even in the same dialogue it's "Semantic Class", "Non-Semantic Tags" and "Parent Groups")

spacemanspiff2007:
@florian-h05 I tired signing off by just adding the appropriate comment. Is it okay like that?

florian-h05:
Yeah it’s fine like that, thanks.
It would have also been okay to just comment the sign off, I can add it to the commit message when merging.

florian-h05:
> It's more important that the UI aligns with the current docs than with the DTO and the docs call this field [icon](https://www.openhab.org/docs/configuration/items.html#icons).

I absolutely agree 👍

Issue #1427, opened 2022-06-23
Firmware update management is not implemented in OH3


cdjackson:
The firmware API, which reads the list of available firmwares for a thing, and allows the user to select a firmware, and then provides the status information when a firmware is loading is not implemented in OH3. This is a core feature of OH so probably should have support included in the UI as it was in OH2.

PR #3158, opened 2025-04-20
Show thing firmware status

cdjackson:
This adds a block to the thing information to provide the status of firmware. It just shows if there are firmwares available, the current firmware version and the status.

I plan to add a button on each line of firmware to allow the user to start the upload but wanted to open this PR for visibility / comment.

Refs #1427.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -13,8 +13,7 @@
       <f7-link @click="switchTab('thing')" :tab-link-active="currentTab === 'thing'" class="tab-link">
         Thing
       </f7-link>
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!
I am not too sure about the look of the available firmware versions, but as I cannot test this let's squeeze this into M2 and come back later, as you already proposed.

FYI I have added a v-if because if no firmwareStatus is available this change would have broken the Thing details page.

florian-h05:
BTW, checking the docs how I could implement this for my bindings, I have noticed we are missing docs: https://next.openhab.org/docs/developer/bindings/#firmware-information-firmware-update
If you know about that, it would be great if you could write the missing part.

cdjackson:
Yes, I also noticed the missing docs. It's been a (very!) long time since I looked at this and had to remind myself how it works. I'm also knocking the zigbee firmware update back into shape, so hopefully I'll try and write something about this if I can work it all out :)

PR #3159, opened 2025-04-20
Sidebar Search: Fix rules/scenes/scripts search result count

### LINKED ISSUE #3157: [MainUI] Developer sidebar Rule count is incorrect when initially searching
## The problem

When searching via the developer sidebar the count is incorrect if the number of hits is above 5.  Let's say there are 10 rules that match, what you see for the count is 5.  But if you expand the Rules list, you see all 10 and the count changes to 10.

Contrast this to Items.  If there are 10 hits for Items, you see a count of 10 initially with 5 rules listed.  If you expand the Items list, you still see a count of 10, and you see all Items.  This is what I expect for Rules too.

## Expected behavior

Rules should show the actual count of Rules that match, not 5 initially.  It should match the behavior of Items in the developer sidebar.

## Steps to reproduce

Search for something that has more than 5 hits in the developer sidebar for both Rules and Items.  Note that you'll see the correct count for Items, but not for Rules.

1. 

## Your environment

runtimeInfo:
  version: 5.0.0.M1
  buildString: Milestone Build

I believe this occurs in 4.3 builds as well.

```yaml

```

## Browser console

```txt

```

## Additional information

See also: https://community.openhab.org/t/developer-side-bar-search-hit-count-for-rules/163710/2

jimtng:
Fix #3157
Backport if possible

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/developer/search-results.vue
+++ b/bundles/org.openhab.ui/web/src/components/developer/search-results.vue
@@ -48,9 +48,9 @@
       </f7-list>
     </f7-block>
     <!-- Rules -->
-    <f7-block class="no-margin no-padding" v-if="filteredSearchResults.rules.length">
+    <f7-block class="no-margin no-padding" v-if="searchResults.rules.length">
       <f7-block-title class="padding-left">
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!
I am normally only backporting critical stuff, but this one is a no brainer as there is no risk of breaking things.

PR #3162, opened 2025-04-20
[MainUI] Adds dialog view for thing properties with overflow key/values

digitaldan:
This adds an icon to properties who have values longer then can be displayed on the thing details page.  Clicking this icon brings up the key and value of the property in a dialog box.  This icon only shows if the value overflows.

Has been tested on both desktop and mobile screen sizes.
 
image

Clicking on the "i" icon brings this up.....

image


florian-h05 (approved):
LGTM, thanks!

PR #3163, opened 2025-04-21
Copy File Definition: Process multiple objects with single api request

jimtng:
Depends on core PR: https://github.com/openhab/openhab-core/pull/4734

This solves the problem with YAML format returning the `version` and `things` (or `items`) keys for each item in multi-select requests. By combining them into a single request, only one key will be returned at the top.

See discussion in https://github.com/openhab/openhab-webui/pull/3130#issuecomment-2817126752


lolodomo:
Now that the core PR is merged, the feature is broken until this UI PR is also merged.

florian-h05 (approved):
Code LGTM, thanks!
I will trust you that this works.

jimtng:
I've just tried the latest snapshot. It seems to work.

PR #3164, opened 2025-04-21
Developer Sidebar: make pinned objects persistent

jimtng:
Make pinned objects persistent within the browser session, so they don't disappear when closing / reopening the developer sidebar, or when switching to Help tab and back.


florian-h05 (approved):
LGTM, this will greatly improve UX!

PR #3169, opened 2025-04-29
Add mime type & code snippet for PY transformation

HolgerHees:
@florian-h05 can you please review


HolgerHees:
@florian-h05 you are right, it is PY... is changed now.

PR #3171, opened 2025-04-30
Refactor Semantic Tags

jimtng:
Store the full details of semantic tags so that:
We have the labels, synonyms, hierarchy (parent), description, etc.
This will allow richer tags handling, e.g. to display tag description and hierarchy.


jimtng:
Oops I didn't see #3165 :(

PR #3174, opened 2025-04-30
Rule Module Popup: Hide Save link on inline script language selection

jimtng:
Clicking Save here would create an invalid module with a missing language, so just hide it.

image


florian-h05 (approved):
LGTM, thanks!

PR #3183, opened 2025-05-04
[BasicUI] Reload the chart with higher dimensions when zooming

### LINKED ISSUE #3139: [BasicUI] 'Zooming' charts does not actually show more content

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

Charts in BasicUI have a 'zoom' feature where they are shown larger than they normally are. The normal size is the chart backend's default size (640x320), while the 'zoomed' size seems to be double of that (1280x640). In that case the same image is used, though, which leads to a blurry look and questionable usefulness.
Consider the following screenshot which shows a normal and a zoomed chart:

Image

## Expected behavior

Zoomed charts should not use the same image, but load an image tailored for the larger size. Consider the following screenshot where I manually appended `&w=1280&h=640` to the URL of the zoomed chart using developer tools:

Image

## Steps to reproduce

1. Add a chart to a sitemap
2. Open sitemap in BasicUI
3. Use 'zoom' button in chart toolbar (shown when clicking the chart)

## Your environment
- OH 5.0 M1
- Firefox browser on Linux (shouldn't matter though)

## Additional comments
Maybe the zoom button could display an even larger image, outside/on top of the sitemap content?

lolodomo:
When zooming, instead of just upscaling the previously loaded chart, a new chart is built using available width as new chart width. Upscale is kept and will be used in case the user defined a maximum width for charts that is lower than the available width.

Closes #3139

### DIFF
--- a/bundles/org.openhab.ui.basic/snippets-src/chart.html
+++ b/bundles/org.openhab.ui.basic/snippets-src/chart.html
@@ -21,7 +21,7 @@
 				%period_rows%
 			</div>
 		</script>
-		<button id="button3-%widget_id%" class="mdl-button mdl-button--raised mdl-js-button mdl-js-ripple-effect image-upscale-button">
+		<button id="button3-%widget_id%" class="mdl-button mdl-button--raised mdl-js-button mdl-js-ripple-effect chart-upscale-button">
 			<!-- fullscreen -->
 			<i class="material-icons">&#xE5D0;</i>
 		</button>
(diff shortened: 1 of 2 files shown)


lolodomo:
As an example, here is the unchanged rendering when the chart is not yet zoomed with default size of chart being 480x240:

image

New rendering when zooming:

image

Here is how it was before (just an image upscaled):

image

kaikreuzer (approved):
Thanks!

PR #3184, opened 2025-05-05
Sitemap editor: Small fixes & Add duplicate element functionality

mherwege:
This fixes:
- Scrollbar for sitemap tree not visible on narrow screens
- Details sheet not closed resulting in browser error
- Sitemap marked dirty when no icon defined on one of the sitemap elements

This adds a sitemap element duplicate function. This will duplicate a sitemap element with all of its sitemap children elements and can drastically speeds up sitemap creation or editing. Before, the easiest way was doing this in the code editor. With this change it can easily be done in the treeview.


florian-h05 (approved):
LGTM thanks!
The issue with everything getting dirty when no icon was defined was really driving me crazy.

florian-h05:
A few issues remain or are regressions:

- I cannot scroll through the three on iOS, it immediately moves elements
- When selecting an item, I get a dirty warning
- When collapsing the tree, the editor gets dirty

mherwege:
> A few issues remain or are regressions:
> 
> * I cannot scroll through the three on iOS, it immediately moves elements
> * When selecting an item, I get a dirty warning
> * When collapsing the tree, the editor gets dirty

I will have a look.

PR #3193, opened 2025-05-16
Fix Python syntax highlighting

Nadahar:
The MIME types for Jython/Python were changed, breaking syntax highlighting in the script editor: https://github.com/openhab/openhab-addons/pull/18616

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
@@ -181,7 +181,13 @@ export default {
       if (mode.startsWith('application/javascript') || mode === 'js') return 'text/javascript'
       if (mode === 'application/vnd.openhab.dsl.rule') return 'text/x-java'
       if (mode === 'application/x-groovy' || mode === 'groovy') return 'text/x-groovy'
-      if (mode === 'application/x-python' || mode === 'py') return 'text/x-python'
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
```diff
@@ -181,7 +181,13 @@ export default {
       if (mode.startsWith('application/javascript') || mode === 'js') return 'text/javascript'
       if (mode === 'application/vnd.openhab.dsl.rule') return 'text/x-java'
       if (mode === 'application/x-groovy' || mode === 'groovy') return 'text/x-groovy'
-      if (mode === 'application/x-python' || mode === 'py') return 'text/x-python'
+      if (mode === 'application/x-python2' || mode === 'py2') {
+        return {
+          name: 'text/x-python',
+          version: 2
```
florian-h05, line 187: Is the code able to handle this object instead of the standard string?

Nadahar: Yes, I have tested this before I submit it 😉 

There are several options that can be specified, just sending a string for `mode` seems to be "the exception" where the string value is then interpreted as `name`. Python 3 is the default, which is why you don't need to send the `version` when you want that.

florian-h05 (approved):
LGTM, thanks!

PR #3194, opened 2025-05-19
Sitemap editor: Various fixes

mherwege:
Relates to: https://github.com/openhab/openhab-webui/pull/3184#issuecomment-2886630684

- I cannot scroll through the three on iOS, it immediately moves elements: this should be fixed, but I could only test in BrowserStack. Performance is not great in BrowserStack, so it would be better to test on a real device.
- When selecting an item, I get a dirty warning: this doesn't happen for me. I can't reproduced, but may actually be impacted by the fix for the next item.
- When collapsing the tree, the editor gets dirty: this is fixed.

Drag and drop also struggled when there were mutiple equal trees. When dragging one of these, it could not find the proper new position. This has been fixed as well.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/sitemap-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/sitemap-mixin.js
@@ -78,7 +78,7 @@ export default {
     }
   },
   methods: {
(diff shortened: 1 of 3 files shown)


florian-h05 (approved):
LGTM, thanks!
I have tested this on my iPhone 14, works fine :heavy_check_mark:

PR #3196, opened 2025-05-21
Analyzer: Add comment to silent series option

florian-h05:


ghys:
I can't find this option anymore either in the ECharts documentation but I think it was to make the chart "non-clickable".

https://stackoverflow.com/a/75170769

It might have been renamed.

ghys:
Got it, it's related to the items with "discrete" values like Switch which draw areas on the chart:

image

As you see when you hover over the area it takes precedence and it's hard to have the tooltip display the line chart value.

When you check "silent" on the area series you make sure that the area doesn't "count" when displaying the tooltip:

image

PR #3197, opened 2025-05-23
Regenerate rule templates

Nadahar:
This PR depends on https://github.com/openhab/openhab-core/pull/4718.

It does several things, but the major points are:
* It enables regeneration of rules based on rule templates from the UI. Regeneration can be desirable if the rule template has been updated and you want the rule(s) to reflect the change, or if you wish to change one of the parameters, for example to make the rule work with a different Item.
* It restores the display of read-only/unmanaged rules in the UI. It is currently broken because all read-only rules are forwarded to the script editor, effectively hiding the rule editor, unless you modify the URL manually. This is very relevant for https://github.com/openhab/openhab-core/pull/4633, whose rules will otherwise be quite meaningless in the UI (they will still work, but you won't be able to see what they do from the UI).
* It provides a third tab "Source" to the rule editor that shows the "source script" used to create the rule, if the rule is supplied via a JSR223 based scripting add-on, *if* the "source script" is embedded as metadata with the rule. This is the current practice with most JSR223 based scripting add-on provided rules, and is the only situation where showing read-only rules in the script editor makes any sense. The "Source" tab is thus a replacement for removing the forwarding to the script editor. Rules that have embedded "source scripts" will by default open in the "Source" tab, so that the difference for these rules will be minimal, while it also allows other read-only rules to work.

In addition to the above points, there are quite a few bug fixes that I have come over while debugging/testing this.

There's a lot that could be said about the details here, but I know from experience that if I attempt to describe everything, people won't read it. So, I tried to make this description brief, and can instead elaborate on any subjects on demand.


Nadahar:
Testing this is a bit involved, especially since https://github.com/openhab/openhab-core/pull/4718 hasn't been merged yet. But even when it is, you need rules, scripting add-ons etc. installed and configured to be able to test what it does.

So, I've created a little video that shows the gist of it: https://youtu.be/AhH_3iVEIJY

Instead of making long pauses in the video, I've assumed that the viewer will just pause playback as needed to study the details. The last rule in the video, the JSR223 based Ruby rule without embedded source, has very little information to show. That's merely a reflection of the facts, OH has very little information about the JSR223 based rules if they don't embed the source.

It should also be noted that the way it works today, without this PR, is that all read-only rules without embedded source/action, only shows up as a blank script editor, which isn't very helpful. Currently, the only read-only rules that "work" in MainUI are JSR223 based ones with embedded source/actions, where the source is opened directly in `script-edit`. The information that isn't shown in the script editor simply isn't available in the UI.

Nadahar:
@florian-h05 FYI: Now that https://github.com/openhab/openhab-core/pull/4718 has been merged, rule templates won't work without this PR. There is a change in how they behave that was necessary to enable regeneration, in the Core side has changed, but not MainUI.

Nadahar:
I've addressed both of the issues found by Copilot. My testing didn't show any difference in behavior, but it's probably cleaner/more correct.

edit: The "low confidence" issue found by Copilot is wrong, this change is very much intentional.

ghys:
Hi @Nadahar, first off congratulations - this PR and the related features look great! Also the explanation video was very much appreciated.

I tend to test first and look at the code afterwards so these are my first remarks design-wise:

- The new button on the bottom toolbar doesn't really look good on narrow screens:

image

I can imagine it could be useful to regenerate multiple rules from their template in one go, but I don't like how it messes up the toolbar so I would hide it until we can discuss how to display it:

image

- For the JSR223 rules which add a Source tab with this PR:

image

I would argue that in this case the "Code'" tab isn't really needed, and the toolbar before the code in the new "Source" tab is redundant because it's already display in the first tab.

What I would suggest to do is just show the contents of the Source tab in the Code tab and keep only 2 tabs, as in the "Code" of the rule would just be the JSR223 source - it is not-editable anyway so having the actual Ruby source (in this example) as the Code makes sense IMO.

ghys (approved):
I have only disabled the new toolbar button in the list until we have a good way to display it, but other than that, although I would prefer the new Source tab to be merged with the Code tab eventually, I can't see why this couldn't be merged.

Nadahar:
> * The new button on the bottom toolbar doesn't really look good on narrow screens:

I absolutely agree - F7/layout is what I find "most challenging" to figure out how to best handle, and I don't know if there's a way to handle this when the screen becomes too narrow (by reducing to just icons for example), or if we just have to limit this to 3 items on the toolbar. That said, Maybe "Enable" and "Disable" could be combined by some clever logic, one of them often isn't necessary. The exception is if you select multiple rules where some are enabled and others are disabled. But, in that case, the "regenerate" button will be hidden because it's only displayed when *one* rule is selected.

I agree that it could potentially be useful to be able to regenerate multiple rules in once go - the reason I "blocked" that option is that for *managed* rules, you are taken to the "rule edit" page when you click "regenerate". This is to give the user the option to change any of the parameters, or maybe the updated template requires some previously unspecified parameter. It would be hard to make a sensible solution where the user would "understand the process" if he/she was queued through a series of such pages. I think that would just be confusing.

For *unmanaged* rules however, you can't edit the parameters in the UI (you will have to do it in the file instead), so for those you could regenerate many at once. But, I decided that making a logic that figured out all this was too complicated/potentially time-consuming to justify, considering that I don't expect regeneration to be a frequent action anyway.

It doesn't have to exist at that toolbar at all of course, but I think the best solution might be to simply hide "Enable" or "Disable" if not rule where that action is possible is selected. Because "regenerate" only shows with just one rule selected, it would never be more than 3 elements on the toolbar.

Nadahar:
> I have only disabled the new toolbar button in the list until we have a good way to display it, but other than that, although I would prefer the new Source tab to be merged with the Code tab eventually, I can't see why this couldn't be merged.

Oh... ok. I was expecting to change these things, but I can make a separate PR then 😉

ghys:
> Oh... ok. I was expecting to change these things, but I can make a separate PR then 😉

Yes we had a time constraint so it had to make an upcoming release... :)
Feel free to follow up with a new PR!

Nadahar:
Regarding the "Code" and "Source" tab, I agree that not much useful will be found in the "Code" tab. I was just thinking consistency, to not "confuse". But, the "Code" tab could probably just be hidden whenever the "Source" tab is visible. I think it's important to keep the distinction between "Code" and "Source" though, because what is shown is different. The "Code" tab shows the rule "as OH sees it", the "Source" tab only displays whatever text is embedded in the rule.

Regarding the toolbar with MIME type/rule status and enable/disable/run, I just did it because there was such a toolbar in the "script editor". But, as you point out, it's already there in the other tab, and since these rules are read-only anyway, the value of having the toolbar there is probably non-existing. I can remove the toolbar if that's what's considered the best way (to save screen real-estate on small screens), or we optionally could just hide it for small screens.

rkoshak:
> It would be hard to make a sensible solution where the user would "understand the process" if he/she was queued through a series of such pages. I think that would just be confusing.

I've seen this handled by simply not doing that. If you regenerate more than one, you just keep the current settings and don't give the user the option to override. Only go to that edit page when refreshing only one rule. 

There will be lots of cases where all the user wants to do is upgrade their rules from the template and keep the same properties. That will handle that use case nicely. I have half a dozen rules that are based on one template. I actually want to be able to update them all in one go without changing any properties. 

That should make the logic a lot simpler too since it doesn't matter if they are managed or unmanaged. All that matters is if it's one or more than one.

Something to consider for the new PR.

Nadahar:
> I've seen this handled by simply not doing that. If you regenerate more than one, you just keep the current settings and don't give the user the option to override. Only go to that edit page when refreshing only one rule.

That's one way to solve it, I don't think it should be difficult technically, but the reason I didn't do that initially is that I considered it "unintuitive". If that's the way it works, I feel that it should be explained somewhere, but I don't know where/how this information should be made available to the user. Any ideas on that?

Another problem is that I've already started on the follow-up PR (not created the PR itself yet, but I've done some of the coding), and I went for what I thought was a good idea by hiding "enable" or "disable" when they aren't relevant:

rules-list-toolbar-1

rules-list-toolbar-2

rules-list-toolbar-3

rules-list-toolbar-4

rules-list-toolbar-5

I also shortened the text to just "Regenerate", and together these things makes sure that the toolbar is never "overcrowded". That only holds true as long as "Regenerate" is only visible when one is selected, because otherwise, if you select some enabled and some disabled rules, both will show, and we'll still run out of space.

I see that it could be handy to be able to regenerate many in one go, so I'm in no way opposed to it, I just don't know how to solve the whole "toolbar overcrowding" issue. Ideas as welcome!

rkoshak:
> I feel that it should be explained somewhere, but I don't know where/how this information should be made available to the user. Any ideas on that?

We can start with the docs. We can also add it to the context sensitive help and the developer sidebar.

> I see that it could be handy to be able to regenerate many in one go, so I'm in no way opposed to it, I just don't know how to solve the whole "toolbar overcrowding" issue. Ideas as welcome!

Leave what you've done so far which is good. Just make it look like the Remove and Disable buttons where you get a count of how many are selected.

Nadahar:
> Leave what you've done so far which is good. Just make it look like the Remove and Disable buttons where you get a count of how many are selected.

But, that still doesn't allow regenerating multiple rules at once...

If we found a way to let the UI handle regeneration of multiple rules, I was thinking of something "crazy" like automatically selecting all rules from the same template if you click the template tag. It's so obscure that nobody will ever discover it if it isn't explained, but I imagine that the functionality itself could be handy.

rkoshak:
> It's so obscure that nobody will ever discover it if it isn't explained

That's what the docs are for. With a phone sized screen there isn't room to explain everything in place. So either you have to depend on the docs or give up and don't provide the feature. I would much rather have the feature.

Nadahar:
> That's what the docs are for. With a phone sized screen there isn't room to explain everything in place. So either you have to depend on the docs or give up and don't provide the feature. I would much rather have the feature.

Fair enough, I just have a strong "urge" to make sure things are as "self documenting"/intuitive as possible, as I suspect that things that require reading of documentation to be figured out will rarely be used.

But, before I can make any of this, I *must* find a way around the "overcrowding". I'm struggling to find a way. The things I've done already, reduces the likelihood over "overcrowding" quite a bit, because you'd have to select some disabled and some enabled rules, where some of them are managed (so that "Remove" is an option) *and* some of them are template-based to be able to experience "overcrowding" in the first place. And, you'd need to be on a very small screen, less than 400 px or so wide. Would it be "acceptable" if these circumstances made the labels disappear, so that you only had the icons on the toolbar? It should happen very rarely.

Nadahar:
I have found another way to handle the "overcrowding" that might be acceptable: Change the layout of the toolbar if the screen width is below 480 px:

rules-list-toolbar-6

That allows all four to display fine on small screens. Any views?

Nadahar:
I've now implemented what we discussed, except the selection of all rules with a given template by clicking the template badge: Multiple rules can be regenerated, the "edit rule" page is only opened up if only one rule is selected, and that rule is managed. Otherwise, the rule(s) is/are regenerated using the existing parameters. *If* the template has changed so that a change to the parameters is required, the regenerated rules will fail/error.

rules-list-toolbar-7

If this is seen as a good solution, I can start working on trying to select all based on a specific template if the template badge is clicked.

rkoshak:
"Refresh" is a few characters shorter than "Regenerate". That can save some space. Beyond that I've no complaints. 

I suggest creating the PR

Nadahar:
> "Refresh" is a few characters shorter than "Regenerate".

Yes, but it isn't very descriptive of what actually happens. It doesn't just "update" the rule, it discards the existing rule and makes a brand new one, using the previous parameters and the template. There is enough screen space for "Regenerate" as it is now, so the only "benefit" of using "Refresh" is that the length of the words are "more balanced". But, let's not forget that this is in English only. When translated to other languages, it could look completely different.

Nadahar:
I've now made the "template badge click" functionality. If you ctrl+click a template badge, it will enter "selection mode" and select all rules based on this template. If you're already in "selection mode", clicking a template badge will toggle selection of all rules with this template. In case these are partially selected, it will check them all if the majority is unchecked and vice versa. On a tie (equal number checked and unchecked), they will all be unchecked.

This will need to be documented to be of much use, but at least it's there. Does it sound OK?

rkoshak:
For tracking purposes I think al lot of this discussion should be on an issue or PR associated with the work that is ongoing rather than on this closed PR. 

But the behavior sounds OK to me. I like having the option to select all the rules created from one template.

Nadahar:
@rkoshak The reason I haven't created the PR yet is because I haven't cleaned up the commits, they are just "roughly thrown together" and not really "presentable". I know that I can create a draft PR, but GitHub still lists all the changes/rebasing/reorganization I do to the commits. I'm very "established" in my routines, and they work well for me, but doesn't make much sense to an outsider. Therefore, it makes more sense to with until it's a bit closer to finalized before I push it.

Making a separate issue just for a "fix PR" for this PR is a bit overkill IMO. But, I can create a draft PR and then reference the discussion here, so that it's possible to make sense of it later.

PR #3199, opened 2025-05-24
Model editor: drag drop fixes

mherwege:
This is a follow-up from https://github.com/openhab/openhab-webui/pull/2970

This applies 2 fixes:

1. Extra non-semantic tag was added with value equal to semantic class when semantic item was moved.
2. On larger models, SortableJS and Vue DOM updates sometimes got out of sync. This is solved by forcing SortableJS to use the fallback that does not directly update the DOM.

@stefan-hoehn Maybe you can check this out to see if fix 2 solves the issues you where seeing.


stefan-hoehn:
Still doesn't fix it for me @florian-h05  @mherwege

Issue #433, opened 2020-10-23
[MainUI] Allow selecting existing Items to add as points/properties from the Model config page


rkoshak:
When someone is starting from scratch or has already built a semantic model in OH 2 this isn't a big deal. But if someone imports a bunch of Things and Items, or creates a bunch of Things and Items without building the model it becomes awkward adding those already existing Items to the model. One has to leave the model page, find the parent Group Item to add the Items to the right Location or Equipment, then go to the individual Items and edit them to add the semantic tags which takes a ton of clicks.

mherwege:
Would be solved with model drag & drop: https://github.com/openhab/openhab-webui/pull/2970

rkoshak:
I don't think so. I'm referring to when  you choose "Add equipment to thing" or "Add points to model" from the Thing's Channels page.

![Image](https://github.com/user-attachments/assets/e5b87997-afc0-4e62-87d7-4777c1d06233)

Which leads to the form where you select the Location, select/create the Location, then choose the Channels to link to *new* Items.

![Image](https://github.com/user-attachments/assets/10f34c7a-b293-4732-9e77-268295184265)

I want to link the Channel to an existing Item from here instead of only having the option to create a new Item. 

#2970 helps with different use cases but not with this initial linking step.

mherwege:
OK, sorry, I misunderstood.

mherwege:
> One has to leave the model page, find the parent Group Item to add the Items to the right Location or Equipment, then go to the individual Items and edit them to add the semantic tags which takes a ton of clicks.

Model drag/drop would allow you to do it from the model tree. You just drag the item to the right location or equipment. No need to add semantic tags manually.

rkoshak:
> Model drag/drop would allow you to do it from the model tree. You just drag the item to the right location or equipment. No need to add semantic tags manually.

True, but that's not the main focus of this issue. Though the way I worded the original issue I agree it kind of sounds like it. The whole issue of situating the Items into the model are indeed solved by the drag and drop PR. But it still would be great if we could select existing Items and not just create new Items from this form above and beyond the whole model discussion.

mherwege:
@rkoshak Would it be OK to be allowed to pick (one of) the existing item(s) already linked to the channel selected? Or would you want to also allow linking existing items, not linked to a channel or to another channel potententially on another thing? I can see the first to make a lot of sense. I am more doubtful about the second.

mherwege:
[...]

PR #3200, opened 2025-05-25
Add to model from thing, allow picking existing items and groups

mherwege:
Resolves https://github.com/openhab/openhab-webui/issues/433

Adding to the model from thing channels has been enhanced in a few ways:

- Add Points to Model: now allows to pick an existing item already linked to the channel. The item will get default semantic tags from the channel defaults if it does not have semantic tags already.
- Add Equipment to Model: ability to pick an existing group for the equipment. This group will get the default equipment semantic tag from the thing if it does not have a semantic tag already. Groups tagged as locations cannot be selected to guard model consistency. Adding points in this case is enhanced just as above.

The purpose of this enhancement is to ease the creation of a semantic model for users that already have items linked to channels and group structure that could be used as a basis for equipment definition.

This PR does not do anything to ease the creation of items linked to channels outside of the model, another requested, but secondary, feature in the above issue.

When doing this, I ran into some reactivity issues with oh-icon, where switching icon and state at the same time caused update issues. This was caused by conflicts in the 2 watchers updating the same properties. I fixed that by grouping these 2 elements in 1 object with a single watcher.

I will do some further tests before this is ready for review.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
@@ -1,6 +1,6 @@
 <template>
   <ul class="item-picker-container">
-    <f7-list-item :title="title" :disabled="disabled" smart-select :smart-select-params="smartSelectParams" v-if="ready" ref="smartSelect" class="item-picker">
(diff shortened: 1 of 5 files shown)


ghys:
Sorry for the delay, it will be reviewed soon!

florian-h05 (approved):
LGTM, thanks!

One minor issue to fix later though:

When clicking on a channel in the add-to-model view, I get the following warning:

image

PR #3209, opened 2025-06-01
Add blockly "rule enable" block

### LINKED ISSUE #3096: [Blockly] New Block for enabling and disabling rules
## The problem

Currently it's not possible to change the enabled-status of a rule within blockly.

## Your suggestion

Introduce a new block "set enabled/disabled [rule]"
It may also be useful to have "get enabled-status [rule]" (i.e., so you can write code that's semantically equivalent to "if rule X is enabled, then ABC, else DEF")

## Your environment

```yaml
runtimeInfo:
  version: 5.0.0.M1
  buildString: Milestone Build
locale: en-CA
systemInfo:
  configFolder: /etc/openhab
  logFolder: /var/log/openhab
  javaVersion: 21.0.6
  javaVendor: Debian
  osName: Linux
  osVersion: 6.1.0-31-amd64
  osArchitecture: amd64
  availableProcessors: 6
  freeMemory: 119667808
  totalMemory: 230686720
  uptime: 510119
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-astro
  - binding-harmonyhub
  - binding-http
  - binding-mqtt
  - persistence-inmemory
  - persistence-mapdb
  - persistence-rrd4j
  - transformation-jsonpath
  - transformation-map
  - ui-basic
```

## Additional information

This thread has a helpful reference to an alternative method for accomplishing this, but it's a big chunk of inline scripting: https://community.openhab.org/t/how-to-disable-a-rule-from-a-blockly-script/144548/4

stefan-hoehn:
fixes #3096

Adds a new block to allow enabling/disabling rules.
Variables of type Boolean and String are supported. A String with "true" or "enabled" will be converted to true, all other values will result into false.

image

I also added Boolean to typed variables as this was missing:

image

I used "as" because I think this is the most intuitive way of describing what the block does, in particular when you use "enabled". If anyone has a better idea to make the block design better, then please provide feedback. Using a dropdown with Note that "Enable/Disable" ruleUID would be nicer but would not allow to use variables.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
@@ -348,7 +348,7 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
     const ruleUID = javascriptGenerator.valueToCode(block, 'ruleUID', javascriptGenerator.ORDER_ATOMIC)
     const enableValue = javascriptGenerator.valueToCode(block, 'enable', javascriptGenerator.ORDER_ATOMIC)
     const enableType = blockGetCheckedInputType(block, 'enable')
-    let enable = (enableType === 'Boolean') ? enableValue : (enableValue === '\'true\'')
+    let enable = (enableType === 'Boolean') ? enableValue : (enableValue === '\'true\'' || enableValue === '\'enabled\'')
     return `rules.setEnabled(${ruleUID}, ${enable});\n`
   }
 }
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
```diff
@@ -324,4 +325,30 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
...
     return code
   }
+
+  Blockly.Blocks['oh_rule_enable'] = {
+    init: function () {
+      this.appendValueInput('ruleUID')
+        .setCheck('String')
+        .appendField('Set Rule ')
+      this.appendValueInput('enable')
+        .appendField('as ')
```
ghys, line 335: Not sure why the extra space is needed at the end - and in other places - since it seems to be perfectly fine without (with equal padding on both sides), but as it is apparently widespread in the code and the de-facto convention then okay (maybe I'm missing something!)

ghys (approved):
Thanks!

PR #3223, opened 2025-06-10
Model treeview: Fixes for drag and drop

mherwege:
This fixes 2 issues:

- When picking items from model, the already checked or selected items would not appear if these are non-semantic. It would also return a bad selection (selecting the already selected semantic item twice). This PR fixes it by enabling showNonSemantic if any of the selected or checked items are non-semantic and expanding the tree to show all selected or checked items.
- Drag/drop sometimes fails when moving into a subtree of a sibling of a node being moved, as the added and removed events where not generated, but a moved event is created instead. This fix also interprets the moved event. To make things more easily understood for the users when dragging, the ghost element is now hidden, as it sometimes showed in another place than where it would be dropped.
- Drag/drop performance on large models was sometimes very slugish. Some performance improvements where made that should improve this.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/model/model-picker-popup.vue
+++ b/bundles/org.openhab.ui/web/src/components/model/model-picker-popup.vue
@@ -179,6 +179,7 @@ export default {
         this.$nextTick(() => {
           this.initSearchbar = true
           this.restoreExpanded()
+          this.expandSelected()
         })
       })
     },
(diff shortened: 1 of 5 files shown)


stefan-hoehn:
With the latest change, Mark achieved a huge performance improvement. Thanks Mark 🥳

We should try to merge this ASAP.

stefan-hoehn (approved):
L Great TM

florian-h05 (approved):
Code LGTM and works for my very small development system. 
I will trust Stefan's testing :+1: 

Thanks Mark & Stefan!

mherwege:
Thanks @florian-h05

PR #3226, opened 2025-06-11
Make Main UI log levels configurable

mherwege:
While developing drag and drop, I inserted a lot of console.debug logs in the javascript code. Even when the browser console is closed, these still consume a considerable amount of time (especially because it deep clones objects to be able to trace back the state of the objects during drag and drop).
I did not want to completely remove the logs, as the functionality is complex and I expect these logs may proof useful when receiving user feedback and debugging.
Therefore, I rather implemented a UI wide log level to set the log level and only log to the console when the level is above a set value. The levels mimic the OH server side log levels.

The log level for the browser can be set in an extra developer tools screen entry, and are persisted in the browser store when set. I set the default logging level to INFO (will include ERROR and WARN), thereby disabling `console.log`, `console.debug` and `console.trace`.

The impact is UI wide. To test, you can look at the Log Viewer with the console open, with log level Off or at Debug level. You will notice extra logs when at Debug level.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/routes.js
+++ b/bundles/org.openhab.ui/web/src/js/routes.js
@@ -93,40 +93,6 @@ const loadAsync = (page, props) => {
   }
 }
 
-// Avoid excessive console logging, make it configurable from the UI
-const LOG_LEVELS = ['OFF', 'ERROR', 'WARN', 'INFO', 'DEBUG', 'TRACE']
-if (!window._originalConsole) {
-  window._originalConsole = {
-    info: console.info,
-    warn: console.warn,
-    error: console.error,
-    debug: console.debug,
-    trace: console.trace,
-    time: console.time,
-    timeEnd: console.timeEnd,
-    timeLog: console.timeLog
-  }
-}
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.ui/web/src/js/routes.js
florian-h05, line 1: IMO this should rather go into app.js

mherwege: I changed that.

stefan-hoehn:
@mherwege Do you mind adding a screenshot that explain "The log level for the browser can be set in an extra developer tools screen entry" ? Or even add this somehow to the documentation? Otherwise I would guess that people wouldn't find where and how to set it (like me...)

mherwege:
> @mherwege Do you mind adding a screenshot that explain "The log level for the browser can be set in an extra developer tools screen entry" ? Or even add this somehow to the documentation? Otherwise I would guess that people wouldn't find where and how to set it (like me...)

Sure, I would, but don't have access to a computer for the next few weeks.

Issue #3217, opened 2025-06-06
Show the full semantic hierarchy on Item's page


rkoshak:
## The problem

The representation of the semantic model and an Item's position within it is different in the Model view and the individual Item's view. In the model view there is a nice tree structure with icons presents a lot of information about the semantic model in an intuitive way that doesn't take up much space. On the Item's page there are three or four fields showing the semantic information as text and the semantic significance of the parent Groups is not shown.

## Your suggestion

Present a little snippet of the tree structure as it's presented Model view. All the pertinent semantic information is shown in the tree with the icons and small print showing the tags. For example, one of my lights would look something like this:

```
Wuithering Heights (House)
  |_Ground Floor (Groundfloor)
    |_Family Room (FamilyRoom)
      |_Family Room Lights (Lightbulb)
        |_Family Room Counch Lamps (Switch/Light)
```
The "Family Room Lights" Item may or may not show up in the "Direct Parent Groups" list. It probably should appear there even if the information is redundant.

Not only will more information be shown about the Item's position in the semantic model in about the same amount of space, the consistency with how the semantic model is presented throughout MainUI reduces complexity and the amount of mental translation required on end users to understand their own config.

## Your environment

```yaml
runtimeInfo:
  version: 5.0.0.M2
  buildString: Milestone Build
locale: en-US
systemInfo:
  configFolder: /openhab/conf
  userdataFolder: /openhab/userdata
  logFolder: /openhab/userdata/logs
  javaVersion: 21.0.7
  javaVendor: Eclipse Adoptium
  javaVendorVersion: Temurin-21.0.7+6
  osName: Linux
  osVersion: 6.8.0-60-generic
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 172026720
  totalMemory: 1336934400
  uptime: 414561
  startLevel: 70
addons:
  - automation-jsscripting
  - automation-pidcontroller
  - binding-amazonechocontrol
  - binding-astro
  - binding-chromecast
  - binding-exec
  - binding-http
  - binding-ipcamera
  - binding-mail
  - binding-mqtt
  - binding-nest
  - binding-network
  - binding-networkupstools
  - binding-openweathermap
  - binding-shelly
  - binding-vesync
  - binding-zigbee
  - binding-zwave
  - misc-openhabcloud
  - persistence-mapdb
  - persistence-rrd4j
  - transformation-basicprofiles
  - transformation-jinja
  - transformation-jsonpath
  - transformation-map
[...]

Issue #3211, opened 2025-06-03
Terminology alignment semantic information in Item status and edit screen.


mherwege:
This continuous the discussion started in an unrelated PR: https://github.com/openhab/openhab-webui/pull/3165#issuecomment-2930357686.

It was pointed out by @andrewfg that the terminology used in the Item status and the Item edit screens are not aligned.

In the status screen, we have:
- `class`
- `isPointOf`
- `hasPoint`
- `relatesTo`
- `isPartOf`
- `hasLocation`

These match the relationships in the semantic model: https://www.openhab.org/docs/tutorial/model.html#introduction-to-the-ontology-and-relationships. The values directly come from the REST API semantic config data. `class` comes from `semantics.value`. All other keys are as is in the API and are not interpreted or changed for visualisation by the UI. `class` directly comes from the API, including underscores.
Note the `hasPoint` field. An Equipment can have multiple Points. As far as I can tell, only one of the Points is identified on the Equipment group Item with the `hasPoint` label. This may be a REST API issue.

Setting up the ontology and relationships involves:
- tagging the Item with an appropriate semantic tag from a class (Location, Equipment, Point).
- if the semantic tag is a Point tag, tag the Item with a Property tag.
- create the appropriate group memberships.

For the full rules, see the documentation.
Therefore, configuring this in the edit screen requires setting one or 2 semantic tags and defining groups. The information in the status screen will be derived from that. Currently, the tagging is in one or two fields called:
- Semantic Class: configure a Location, Equipment or Point tag for the Item.
- Semantic Property (only appears when the selected semantic tag is a Point tag).

My initial thought is the terminology in the status screen and edit screen can simply be aligned. On further tought, it is not as straightforward, as these do carry a different meaning. Not all of the semantics are configured with tags (group membership has an important role in this as well). The result is shown in the status screen, but the edit screen just has fields required for configuration.
Also the labels in the status screen closely match the documentation and how the semantic model is explained there. I think it is good to keep that consistency.

I propose the following to better align the terminology. For the status screen:
- hide `hasPoint`: this is limited value as there may be multiple points, and only one is populated now anyway.
- remove `class` and replace with one of 4 conditional fields (`Location`, `Equipment`, `Point`). In the value for this field, replace the underscores by `->` to show the hierarchical structure of the semantic value and make the corresponding tags stand out more.
- rename `relatesTo` to `Property`. Also for this, replace the underscores by `->` to show the hierarchical structure of the semantic value.
[...]

PR #3227, opened 2025-06-12
Visualize semantic model tree in item details screen

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/3217

This PR extends the item details screen to show part of the model the item is in, from the top, down to all its children. The tree is opened up to the level of the item. Clicking on a level in the tree, will take you to the related item.

With this PR, the terminology discussion from https://github.com/openhab/openhab-webui/issues/3211 becomes less critical as the terminology is replaced by a graphical representation of the model with all information in it.

Below is how it looks like.

Note the Parent Group and Child Group sections will contain all semantic and non-semantic direct parent and child groups.

image


ghys:
> A concern may be this is introducing a load of the full model for each item.

Absolutely a concern, as you've identified :)

> The REST API call only retrieves static data, therefore I believe it is cached and should not continuously pass the full payload over the network. (...) In my tests with over 1000 items, I didn't see a strong negative performance impact

I'm not sure we can rely only on the fact that the data retrieved repeatedly would be cached to say "it's all okay".
Sometimes there could be proxy servers between openHAB and the client, and the cache mechanisms might be screwed up so we could have a problem.

mherwege:
> I'm not sure we can rely only on the fact that the data retrieved repeatedly would be cached to say "it's all okay".
> Sometimes there could be proxy servers between openHAB and the client, and the cache mechanisms might be screwed up so we could have a problem.

So what are the options then?

1. I abandon this.
2. I extend the REST API to return this local tree for use in this view. It does mean it will effectively return data for each item opened and caching will not have any effect.

Any other suggestions?

ghys:
> So what are the options then?

>    I abandon this.

Please don't ragequit, it's a good feature, I just struggle how to implement it correctly :)

I have understood recently that you can't rely on client caches only.

> I extend the REST API to return this local tree for use in this view. It does mean it will effectively return data for each item opened and caching will not have any effect.

It could work but it's a very specific API, will it have an use outside of that screen?

mherwege:
> Please don't ragequit, it's a good feature, I just struggle how to implement it correctly :)

That’s not at all what I meant. I am much less experienced then you though, and I don’t see an easy solution.


> It could work but it's a very specific API, will it have an use outside of that screen?

There actually already is a very specific one I believe is mostly there for this screen: get all children (recursive), including the dynamic data. It could be extended to get the parents as well. I was already looking into that, also because there still is a mismatch. On the item screen, the direct children have a state (from that API), but not the parents. I get these trough the model static data request, so without state.

mherwege:
@ghys With the last commit, I now use an adjusted REST API to only retrieve the item model information. Therefore this requires https://github.com/openhab/openhab-core/pull/4861

mherwege:
I rebased and fixed an issue. The core PR has been merged, so this is now ready for review.

florian-h05:
@mherwege I am sorry, but you have to rebase again due to the upgrade PR.

mherwege:
@florian-h05 This was more involved then I expected. I am still learning Vue3 and I struggled a bit with the various changes that had been done impacting the code in this PR. I believe it is stable and working again. I hope I didn't destroy anything else. So ready for review.

florian-h05 (approved):
LGTM, thanks!

PR #3232, opened 2025-06-18
Regenerate rule templates: Follow up for #3197

Nadahar:
#3197 had to be merged with some minor unresolved issues because of a deadline. This is a follow-up PR to address these issues.

The content of this follow-up PR has already been discussed in #3197 from this post and out: https://github.com/openhab/openhab-webui/pull/3197#issuecomment-2974261143

Further discussion should continue here.

As of now, I believe I have addressed the first "unresolved issue", the overcrowding of the `rules-list` toolbar under certain circumstances on narrow screens.

rules-list-toolbar-5

rules-list-toolbar-6

rules-list-toolbar-7

I've also made it possible to regenerate multiple rules at once, with the caveat that you only get to see the `rule-edit` screen where you can change the parameters if only one (managed) rule is selected.

In addition, functionality has been added that toggles the selection of all rules that are based on a certain template, if the "template badge" (orange) is clicked. That makes it easy to select all rules for a template that has been updated.

What remains is to handle the toolbar before the code in the "Source" tab. It was suggested that it is redundant, because the same buttons are available on the "Design" tab. That is true, except for the display of the "script type" (e.g "JRuby Scripting"). That's not available anywhere else, and the reason I included the status and enable/disable and run buttons is that I figured that I might was well use that vertical space for something in addition to showing the "script type". In `script-edit` (that previously served "the same purpose"), the "scripting type" is shown as a subtitle in the navbar on top. But, since `rule-edit` is tabbed, I didn't think that it would be suitable to put that info in "the tab header" itself. Thus, some vertical space had to be reserved for showing this, which resulted in this toolbar. Another (partial) reason is that `script-edit` has the status and the same buttons available, and I didn't want to "reduce functionality" compared to the previous solution.

We can find another way to display the information, strictly speaking, I only think the "script type" is what needs to be displayed. I do think that should be there though, so that the user can know what the "document" displayed is. If this information is deemed irrelevant, the whole toolbar can just go.

Alternatively, it can be made even thinner, have less space around it, be at the bottom etc. Any suggestions are welcome.

Any input on this @ghys and @rkoshak (or anybody else, of course)?


Nadahar:
We should probably also update [this documentation](https://next.openhab.org/docs/concepts/rules.html#rule-templates) to reflect that rules can now be regenerated from template. One thing to notice, is that rules that have been generated from templates in previous versions *can't* be regenerated.

The reason is that in the past, the link to the template was deleted when the rule was generated. It's therefore not possible to differentiate these from any other rules. This has changed in 5.0, where the link to the template is preserved, making regeneration possible.

rkoshak:
I don't have any opinion on the toolbar discussion.

I'm happy with the regeneration behavior as described and shown.

Nadahar:
@ghys I've found a way to show the script type that may be acceptable. This makes the toolbar redundant, so I've removed it:

rule-edit-source-type-1

WDYT?

ghys:
> @ghys I've found a way to show the script type that may be acceptable. This makes the toolbar redundant, so I've removed it

Your last screenshot, I'm not sure if it looks right? There's a big empty space where the bottom toolbar used to be, it cannot be right.
Anyways I'm glad you took the time to address my concerns and test out the UI on phone-like sizes.
There was a choice early on that the Main UI would be "catering to all screens" and you seem to have understood it. It's an additional burden but it's one that makes sense, about 50% of openHAB users have an iPhone so it's just common sense that we make sure our main UI looks good on iPhones.

Nadahar:
> Your last screenshot, I'm not sure if it looks right?

I forgot to adjust the height of the editor after removing the toolbar. Will fix 😉

Nadahar:
> It's an additional burden but it's one that makes sense, about 50% of openHAB users have an iPhone so it's just common sense that we make sure our main UI looks good on iPhones.

I'm aware that the UI should be responsive, but when testing things and addressing many different concerns, it's easy to forget to verify some aspects. That's one of the reasons that reviews/other sets of eyes are useful.

But, regarding iPhone in particular, isn't it so that the OH iPhone app still works with BasicUI only? I didn't think MainUI was even available there? Of course, you can use the web browser, but isn't in likely that most of the iPhone users use BasicUI as a consequence? On Android, on the other hand, MainUI works nicely.

Nadahar:
I've fixed the editor size:

rule-edit-source-type-1

Here are a couple of screenshots of this PR in dark mode and iOS theme:

rules-list-toolbar-8
rule-edit-source-type-2

edit: The reason it says "Slett" below the trashcan icon is that I've replaced the toolbar labels with translatable ones. It so happens that only the "delete" label is currently translated into Norwegian, so that one alone is translated.

Nadahar:
@ghys I think this PR is good to go now.

Nadahar:
@ghys @florian-h05 It would be nice if this could be merged so that it's not necessary to write documentation for "the intermediate stage" (that is in the code right now). All this does is to "polish" some of the changes made in #3197. Yeah, I made it possible to select multiple rules for regeneration from `rule-list`, which wasn't originally intended, and this introduced some logic changes, but it's all pretty minor stuff and any potential bugs to this logic should have a very limited impact. That said, I don't think the logic has any bugs 😉 

It's so much easier to be able to update the documentation to reflect what this does, then to have to first document #3197 and then later this one.

Nadahar:
Addressed Copilot comments - two of the "low confidence" once are wrong and thus ignored.

lolodomo:
Thank you @florian-h05 for reviewing just in time.

PR #3237, opened 2025-06-22
[BasicUI] Fix multiline buttons to optimize width when set to visible

### LINKED ISSUE #3233: [BasicUI] changing the visibility condition leads to an incorrect visualization update of a switch with mappings and visibility

- [ x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

After the update to OH 4.3.0 any change of a visibility condition leads to an incorrect visualization update of a switch with mappings and visibility -> mapping buttons are shown on top of each other instead of next to each other.

In order to get the expected visualization a manual reload of the page is required.

see https://community.openhab.org/t/sitemaps-mapping-buttons-are-shown-one-above-the-other-instead-of-next-to-each-other-after-visibility-condition-changed/164509

## Expected behavior

Mapping buttons should be shown next to each other after the change of the visibility condition automatically wihout the need for a manual reload of the page.

## Steps to reproduce

see https://community.openhab.org/t/sitemaps-mapping-buttons-are-shown-one-above-the-other-instead-of-next-to-each-other-after-visibility-condition-changed/164509

## Your environment

Hardware: Pi 5
OS: Raspberry Pi OS Lite (Bookworm)
Java Runtime Environment: OpenJDK RE (build 17.0.15)
openHAB version: 4.3.5
Browser: Firefox 139.0.4 (64-Bit) and Google Chrome 137.0.7151.120 (64-Bit)

This bug could be reproduced with a fresh openHABian install with nothing but the two test items and the test sitemap as described here: https://community.openhab.org/t/sitemaps-mapping-buttons-are-shown-one-above-the-other-instead-of-next-to-each-other-after-visibility-condition-changed/164509

jimtng:
Regression from #2388
Fix #3233

May need to be backported to 4.3.x

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -1425,6 +1425,9 @@
 			}
 
 			_t.minimizeWidth = function() {
+				if (!_t.visible) {
+					return;
+				}
 				// Minimize the width taken by the buttons without adding extra rows.
 				// Start from the maximum width the buttons can take,
 				// then shrink it down to the minimum without causing additional wrapping.
@@ -1454,6 +1457,13 @@
 		} else {
 			_t.minimizeWidth = function() {};
 		}
+
+
+		_t.parentSetVisible = _t.setVisible;
+		_t.setVisible = function(state) {
+			_t.parentSetVisible(state);
+			_t.minimizeWidth();
+		};
 
 		_t.destroy = function() {
 			_t.buttons.forEach(function(button) {


bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -1455,6 +1458,13 @@
 			_t.minimizeWidth = function() {};
 		}
 
+
+		_t.parentSetVisible = _t.setVisible;
```
lolodomo, line 1462: Good to know it is possible to do that.

lolodomo (approved):
Thank you

PR #3239, opened 2025-06-23
Location card: Fix measurement/status badges

### LINKED ISSUE #3238: Main UI not showing badges in v5 milestone
I am opening this Issue to capture the thread discussion [here](https://community.openhab.org/t/openhab-5-0-milestone-discussion/162686/139) and [here](https://community.openhab.org/t/openhab-5-0-milestone-discussion/162686/161) where some users are reporting that badges don't show any more.

I am not sure that we yet fully understand the issue; @mherwege thinks it may be due to the point tags that were moved to become property tags; however the OP showed pictures where CO2 badges no longer show up and this cannot be due to the point/property shift since CO2 has never been changed.

andrewfg:
Resolves #3238

I know nothing about programming in Vue so this a wild attempt to fix the badges, which definitely must be reviewed properly.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-matter.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-matter.vue
@@ -296,20 +296,16 @@ export default {
       // Get the mapped value for this option from the mapped child's metadata.config
       const mappedChild = this.getMappedChild(attributeName)
       if (!mappedChild) return optionValue
-      const attr = attributeName.toLowerCase()
-      if (!mappedChild.metadata.matter.config) return optionValue
-      if (!mappedChild.metadata.matter.config[attr]) return optionValue
-      // config[attr] is an object: { optionLabel: mappedValue, ... }
-      return mappedChild.metadata.matter.config[attr][optionLabel] || optionValue
+      const cfg = (mappedChild.metadata && mappedChild.metadata.matter && mappedChild.metadata.matter.config) || {}
+      return cfg[optionLabel] || optionValue
     },
     setChildMapping (attributeName, optionLabel, newValue) {
-      // Set the mapped value for this option in the mapped child's metadata.config
       const mappedChild = this.getMappedChild(attributeName)
       if (!mappedChild) return
-      const attr = attributeName.toLowerCase()
-      if (!mappedChild.metadata.matter.config) this.$set(mappedChild.metadata.matter, 'config', {})
-      if (!mappedChild.metadata.matter.config[attr]) this.$set(mappedChild.metadata.matter.config, attr, {})
-      this.$set(mappedChild.metadata.matter.config[attr], optionLabel, newValue)
+      if (!mappedChild.metadata.matter.config) {
+        this.$set(mappedChild.metadata.matter, 'config', {})
+      }
+      this.$set(mappedChild.metadata.matter.config, optionLabel, newValue)
       this.dirtyItem.add(mappedChild)
     }
   }


bundles/org.openhab.ui/web/src/components/cards/glance/location/measurement-badge.vue
```diff
@@ -61,13 +61,19 @@ export default {
...
         case 'co2':
-          direct = findPoints(this.element.properties, 'Point_Measurement', true, 'Property_CO2')
+          direct = findPoints(this.element.properties, 'Point_Measurement', true, 'Property_AirQuality_CO2')
           if (direct.length) return direct
-          return findPoints(allEquipmentPoints(this.element.equipment), 'Point_Measurement', true, 'Property_CO2')
+          return findPoints(allEquipmentPoints(this.element.equipment), 'Point_Measurement', true, 'Property_AirQuality_CO2')
         case 'luminance':
-          direct = findPoints(this.element.properties, 'Point_Measurement', true, 'Property_Light')
+          direct = [
+            ...findPoints(this.element.properties, 'Point_Measurement', true, 'Property_Light',
```
mherwege:
```suggestion
            ...findPoints(this.element.properties, 'Point_Measurement', true, 'Property_Light'),
```

mherwege:
I didn’t have time tocheck more in detail, but there are a few trailing spaces (indicated in the code tab) and one error I suggested to correct.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -72,16 +72,16 @@ export default {
...
       switch (this.type) {
         case 'battery':
-          direct = findPoints(this.element.properties, 'Point_Status_LowBattery', true)
+          direct = findPoints(this.element.properties, 'Point', true, 'Property_LowBattery')
           if (direct.length) return direct
-          return findPoints(allEquipmentPoints(this.element.equipment), 'Point_Status_LowBattery', true)
+          return findPoints(allEquipmentPoints(this.element.equipment), 'Point', true, 'Property_LowBattery')
         case 'lights':
-          return [...this.queryLightPoints, ...this.queryLightEquipment('Equipment_Lightbulb'), ...this.queryLightEquipment('Equipment_LightStripe')]
+          return [...this.queryLightPoints, ...this.queryLightEquipment('Equipment_LightSource')]
```
florian-h05, line 79: Does `Equipment_LightSource` include all different light equipments, i.e. if I tag with `LightBulb`, will core automatically add `Equipment_LightSource` as well?

florian-h05 (commented):
Apart from one question LGTM.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: No. But if I understand correctly the UI will show a light source badge.

florian-h05: So if I tag with Lightbulb instead of LightSource, there will be no badge?

andrewfg: Lightbulb is a child of LightSource so there should be a badge.

.. but perhaps you can check and confirm it?

andrewfg:
@mherwege please note that the following code (e.g.) with `'Point', true` should (probably) include all point types such as Point.Measurement, Point.Control, Point.Status, Point.Alarm -- **including Nota Bene Point.Point**

```
findPoints(allPoints, 'Point', true, 'Property_Light')
```

mherwege:
> @mherwege please note that the following code (e.g.) with `'Point', true` should (probably) include all point types such as Point.Measurement, Point.Control, Point.Status, Point.Alarm -- **including Nota Bene Point.Point**
> 
> ```
> findPoints(allPoints, 'Point', true, 'Property_Light')
> ```

Yes, agreed. Whereby I don't think there is a Point.Point. There just is Point without anything behind it. So this should be fine. It should show up in the REST API as Point, not Point_Point.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
florian-h05: I just checked and it seems this will work as expected 👍

andrewfg:
@mherwege I have a doubt about the `findPoints` method -- according to the code doc the fourth argument (e.g. 'Property_Opening' below) finds a match with a `relatesTo` property; my question is whether `findPoints` `relatesTo` code does find a positive matches against **children** of the property i.e. `OpenState` or `OpenLevel`??

```
points = findPoints(allPoints, 'Point', true, 'Property_Opening')
```

mherwege:
> @mherwege I have a doubt about the `findPoints` method -- according to the code doc the fourth argument (e.g. 'Property_Opening' below) finds a match with a `relatesTo` property; my question is whether `findPoints` `relatesTo` code does find a positive matches against **children** of the property i.e. `OpenState` or `OpenLevel`??
> 
> ```
> points = findPoints(allPoints, 'Point', true, 'Property_Opening')
> ```

You are right with your doubts. It looks like the properties were assumed to be flat so far. It doesn't look at the children.

I would think changing that method in `glance-helpers.js` to something like the below would solve that:
```
export function findPoints (arr, value, partial, property) {
  const points = arr.filter((p) => (partial) ? p.metadata.semantics.value.indexOf(value) === 0 : p.metadata.semantics.value === value)
  if (!property) return points
  return points.filter((p) => (partial) ? p.metadata.semantics.config.relatesTo.indexOf(property) : p.metadata.semantics.config.relatesTo === property)
}
```
You could also introduce a second boolean argument (after `property`) to be able to have different behaviour for the class and property.

andrewfg:
New PR opened https://github.com/openhab/openhab-webui/pull/3245

Issue #556, opened 2020-11-24
[Design/Spec] Location semantic cards: header


ghys:
This issue discusses the design and features of the header of a location card, when not customized by users using slots.
As a semantic card it inherits a number of common properties described in #555 (like background color/image).

## Title
By default, the title is the label of the item (or its name if it has none).

## Subtitle
If the location has a parent, the label or name of the parent is the default (but it should be possible to remove it, for instance if the cards are manually put in groups by the user (see #555).

## At-a-glance modules

The built-in at-a-glance modules (or "complications" as in watch screens) use the semantic model to gather information about the location and display it in a summarized manner with little icons and eventually a short label, "rewarding" the time taken to build a well-designed model by automatically providing information in the header of the card. The goal is not to be exhaustive, rather cover what should be the most popular scenarios. There will still of course be cases outside this anticipated well-known scenarios, and esoteric needs; therefore ideally users should be able to customize these modules, or add more (with slots). It should also be possible to hide automatically shown modules that doesn't work well or aren't desired.

A location card with all modules having something to display should resemble the following example ("worst case scenario"):

![image](https://user-images.githubusercontent.com/2004147/100102623-265a6e00-2e64-11eb-8e58-59b2d71e6506.png)

It is a little busy, but the average card should be much less so since a module isn't displayed when there is nothing to show.

### General principles for querying data from the model

- Normally the user will have put Equipment in the Locations. Most of these pieces of equipment will have Points. Gathering states from Points is always preferred as opposed to gathering states from Equipment item themselves, but we do it anyway if we couldn't find any matching Point. For instance, for blinds, look for OpenState points in equipment tagged Blinds (the preferred default), but if we can't find any, analyze the state of the Equipment items themselves - and consider a state of OPEN, or CLOSED appropriately.

- In order to leave out any ambiguity about which Points to consider, the admin has the option of relating Points directly to the Location without any Equipment in between. There are many ways to do it:
  * a Point can be made a member of both its parent Equipment, and also the Location it's in.
  * the item could be duplicated at the Location level and linked to the same channel so that it reflects the same data
  * a Group with an aggregation function (e.g. average) can be added as a Point to the Location to compute a state for the location from the state of the desired items.
<br />we should therefore prioritize the results by considering first only the Points which have a direct _hasLocation_ relation to the location; but if none match, we can consider those indirectly related through their equipment items.

[...]

Issue #239, opened 2020-04-26
[habpanel] Item ordering and filtering in template selection


eikowagenknecht:
Since the order of items is not alphabetically and sometimes quite confusing (see https://github.com/openhab/org.openhab.ui.habpanel/issues/340), a working filter function is essential.

But the ordering when filtering for an item name seems to be quite random. Eg. I filter for "g_EG_Licht", a group containing quite some lights. The filter function seems to include all items within that group as well (which is a nice idea but imho quite confusing) and the real group with the exact name I'm looking for is somewhere mixed in in the results list.

It gets worse when I search for "g_EG" which is a group, but also a partial match to the other group and thus the result contains >30 entries with the wanted one in the middle somewhere.

I'd suggest changing the ordering of the results so that entries that directly match with the item name appear on top of the list and only then the items that match via some hidden criteria (e.g. are contained in the griup that is searched for) are displayed.

paullangner:
I know that this is not the greatest solution, but if you click on the list of items, the list becomes searchable via the web-browser Search function

eikowagenknecht:
Closing my issues with openHAB that seem to have gone stale.

PR #3245, opened 2025-06-25
Location card: Various improvements and fixes

andrewfg:
This PR does the following..

1. The collection methods can now not only include exact Property / Equipment values but also (_**new**_) children of those values (optionally). This allows display of not only Lightbulb and Lightstripe, but also any LightSource or child thereof. And also of doors and windows that support either OpenState (digital) or OpenLevel (analog) positions; to name a couple.
2. Aligns the code in this PR with the [documentation](https://github.com/openhab/openhab-docs/pull/2514) and vice versa; taking into account [#556](https://github.com/openhab/openhab-webui/issues/556) and subsequent PRs related to that. This PR and #3239 added support for Motion detectors under presence, other displays under screens.
3. Cleans up the collection of lighting points to prevent duplicate entries in the location card; that may otherwise arise when (say) the user connects a Switch, a Dimmer, and a Color item to the same light channel.

Resolves https://github.com/openhab/openhab-webui/pull/3239#issuecomment-3004118138

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -4,12 +4,11 @@
  * @param {String} parent the potential parent tag
  * @returns true if tag is a child of parent
  */
-export function isChildOf(tag, parent) {
-return (!tag || tag.trim() === "") ? false
-      : tag.indexOf(parent) !== 0   ? false
+export function isChildOf (tag, parent) {
+  return (!tag || tag.trim() === '') ? false
+    : tag.indexOf(parent) !== 0 ? false
       : tag.length === parent.length ? true
-      : tag.charAt(parent.length) === '_' ? true
-      : false;
+        : tag.charAt(parent.length) === '_'
 }
 
 /**
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -1,11 +1,25 @@
 /**
+ * Check if the given semantic tag is a child of the given potential parent tag
+ * @param {String} tag the tag to be checked whether it is a child of parent
+ * @param {String} parent the potential parent tag
+ * @returns true if tag is a child of parent
+ */
+export function isChildOf(tag, parent) {
+    if (!tag || tag.trim() === "") return false;
+    if (tag.indexOf(parent) !== 0) return false;
+    if (tag.length === parent.length) return true;
+    if (tag.charAt(parent.length) === '_') return true;
+    return false;
+}
+
+/**
  * Retrieves equipment based on their semantic class
  * @param {Array} arr the array of equipment items & points to search
  * @param {String} value the semantic class (value) to find
  * @param {Boolean} partial match subclasses
  */
 export function findEquipment (arr, value, partial) {
-  return arr.filter((e) => (partial) ? e.item.metadata.semantics.value.indexOf(value) === 0 : e.item.metadata.semantics.value === value)
+  return arr.filter((e) => (partial) ? isChildOf(e.item.metadata.semantics.value, value) : e.item.metadata.semantics.value === value)
 }
 
 /**
@@ -25,7 +39,7 @@ export function allEquipmentPoints (equipment) {
  * @param {Boolean} children match child properties
  */
 export function findPoints (arr, value, partial, property, children) {
-  const points = arr.filter((p) => (partial) ? p.metadata.semantics.value.indexOf(value) === 0 : p.metadata.semantics.value === value)
+  const points = arr.filter((p) => (partial) ? isChildOf(p.metadata.semantics.value, value) : p.metadata.semantics.value === value)
   if (!property) return points
-  return points.filter((p) => (children) ? p.metadata.semantics.config.relatesTo.indexOf(property) === 0 : p.metadata.semantics.config.relatesTo === property)
+  return points.filter((p) => (children) ? isChildOf(p.metadata.semantics.config.relatesTo, property) : p.metadata.semantics.config.relatesTo === property)
 }
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -5,11 +5,11 @@
  * @returns true if tag is a child of parent
  */
 export function isChildOf(tag, parent) {
-    if (!tag || tag.trim() === "") return false;
-    if (tag.indexOf(parent) !== 0) return false;
-    if (tag.length === parent.length) return true;
-    if (tag.charAt(parent.length) === '_') return true;
-    return false;
+return (!tag || tag.trim() === "") ? false
+      : tag.indexOf(parent) !== 0   ? false
+      : tag.length === parent.length ? true
+      : tag.charAt(parent.length) === '_' ? true
+      : false;
 }
 
 /**
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -1,6 +1,6 @@
 /**
  * Check if the given semantic value is a child of the given potential parent semantic value
- * @param {String} semanticValue the vaue to be checked whether it is a child of the parent value
+ * @param {String} semanticValue the value to be checked whether it is a child of the parent value
  * @param {String} potentialSemanticParent the potential parent semantic value
  * @returns true if semanticValue is a child of potentialSemanticParent
  */
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -1,8 +1,8 @@
 /**
- * Check if the given semantic tag is a child of the given potential parent tag
- * @param {String} semanticValue the tag to be checked whether it is a child of parent
- * @param {String} potentialSemanticParent the potential parent tag
- * @returns true if tag is a child of parent
+ * Check if the given semantic value is a child of the given potential parent semantic value
+ * @param {String} semanticValue the vaue to be checked whether it is a child of the parent value
+ * @param {String} potentialSemanticParent the potential parent semantic value
+ * @returns true if semanticValue is a child of potentialSemanticParent
  */
 export function isChildOf (semanticValue, potentialSemanticParent) {
   return (!semanticValue || semanticValue.trim() === '') ? false
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -22,9 +22,10 @@ export function allEquipmentPoints (equipment) {
  * @param {String} value the semantic class (value) to find
  * @param {Boolean} partial match subclasses
  * @param {String} property return only points also related to this property
+ * @param {Boolean} children match child properties
  */
-export function findPoints (arr, value, partial, property) {
+export function findPoints (arr, value, partial, property, children) {
   const points = arr.filter((p) => (partial) ? p.metadata.semantics.value.indexOf(value) === 0 : p.metadata.semantics.value === value)
   if (!property) return points
-  return points.filter((p) => p.metadata.semantics.config.relatesTo === property)
+  return points.filter((p) => (children) ? p.metadata.semantics.config.relatesTo.indexOf(property) === 0 : p.metadata.semantics.config.relatesTo === property)
 }
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -1,14 +1,14 @@
 /**
  * Check if the given semantic tag is a child of the given potential parent tag
- * @param {String} semanticTagValue the tag to be checked whether it is a child of parent
- * @param {String} potentialSemanticTagParent the potential parent tag
+ * @param {String} semanticValue the tag to be checked whether it is a child of parent
+ * @param {String} potentialSemanticParent the potential parent tag
  * @returns true if tag is a child of parent
  */
-export function isChildOf (semanticTagValue, potentialSemanticTagParent) {
-  return (!semanticTagValue || semanticTagValue.trim() === '') ? false
-    : semanticTagValue.indexOf(potentialSemanticTagParent) !== 0 ? false
-      : semanticTagValue.length === potentialSemanticTagParent.length ? true
-        : semanticTagValue.charAt(potentialSemanticTagParent.length) === '_'
+export function isChildOf (semanticValue, potentialSemanticParent) {
+  return (!semanticValue || semanticValue.trim() === '') ? false
+    : semanticValue.indexOf(potentialSemanticParent) !== 0 ? false
+      : semanticValue.length === potentialSemanticParent.length ? true
+        : semanticValue.charAt(potentialSemanticParent.length) === '_'
 }
 
 /**
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
@@ -1,14 +1,14 @@
 /**
  * Check if the given semantic tag is a child of the given potential parent tag
- * @param {String} tag the tag to be checked whether it is a child of parent
- * @param {String} parent the potential parent tag
+ * @param {String} semanticTagValue the tag to be checked whether it is a child of parent
+ * @param {String} potentialSemanticTagParent the potential parent tag
  * @returns true if tag is a child of parent
  */
-export function isChildOf (tag, parent) {
-  return (!tag || tag.trim() === '') ? false
-    : tag.indexOf(parent) !== 0 ? false
-      : tag.length === parent.length ? true
-        : tag.charAt(parent.length) === '_'
+export function isChildOf (semanticTagValue, potentialSemanticTagParent) {
+  return (!semanticTagValue || semanticTagValue.trim() === '') ? false
+    : semanticTagValue.indexOf(potentialSemanticTagParent) !== 0 ? false
+      : semanticTagValue.length === potentialSemanticTagParent.length ? true
+        : semanticTagValue.charAt(potentialSemanticTagParent.length) === '_'
 }
 
 /**
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
@@ -236,7 +236,7 @@ export default {
       if (this.badgeOverrides && !this.exprAst) {
         const override = this.badgeOverrides[this.type]
         if (override && override.expression) {
-          this.exprAst = expr.parse(override.expression)
+          return expr.parse(override.expression)
         }
       }
       return this.exprAst
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
@@ -209,7 +209,6 @@ export default {
     },
     queryLightPoints () {
       // Look for all control points on the location with light property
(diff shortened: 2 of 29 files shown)


andrewfg:
Ping @mherwege

andrewfg:
There seems to be a dependency problem in the CI build ??

mherwege:
I still fear we may have missed some cases, see https://community.openhab.org/t/openhab-5-0-milestone-discussion/162686/228?u=mherwege. It would be best to carefully look through the code again to check.

My mistake, there is no issue in the linked post.

mherwege:
Everywhere we use `indexOf` to match with the tag and subtags, there is a risk.

Here is the code for equipment:
```
return arr.filter((e) => (partial) ? e.item.metadata.semantics.value.indexOf(value) === 0 : e.item.metadata.semantics.value === value)
```
When `partial` is true, this will match both `Equipment_Window` and `Equipment_WindowCovering` and their children, which is probably uninteded. The current use of the method avoid this, but some tag changes may still introduce this error.
The check should probably be on the parent tag and then on starting with parent tag + _.

Something like the following would be more secure:
```
return arr.filter((e) => (e.item.metadata.semantics.value === value) || ((partial) ? e.item.metadata.semantics.value.indexOf(value + '_') === 0 : false))
```

andrewfg:
Good point. I will add a helper method for that.

andrewfg:
@mherwege I just added a helper method

bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
```diff
@@ -1,11 +1,24 @@
+/**
+ * Check if the given semantic tag is a child of the given potential parent tag
+ * @param {String} tag the tag to be checked whether it is a child of parent
+ * @param {String} parent the potential parent tag
+ * @returns true if tag is a child of parent
+ */
+export function isChildOf (tag, parent) {
```
mherwege: I wouldn't call it `tag`. The tag only matches the last part of the string with underscores. I would call it just `value` or `semanticValue` to avoid confusion. In some parts of the UI code we work with tags directly.

andrewfg:
@mherwege I tried to update the documentation based on the code, but I then realised that I needed to update the code to match the documentation too. So can you please check the code and cross reference with the current version of https://github.com/openhab/openhab-docs/pull/2514 -- especially the Lights category which seems to includes LightSource child equipment and also not LightSource as well .. and I am not sure why..

bundles/org.openhab.ui/web/src/components/cards/glance/glance-helpers.js
```diff
@@ -1,11 +1,24 @@
+/**
+ * Check if the given semantic tag is a child of the given potential parent tag
+ * @param {String} semanticTagValue the tag to be checked whether it is a child of parent
+ * @param {String} potentialSemanticTagParent the potential parent tag
+ * @returns true if tag is a child of parent
+ */
+export function isChildOf (semanticTagValue, potentialSemanticTagParent) {
```
mherwege: This is not entirely what I had in mind. I would prefer not to use the word 'Tag' to avoid confusion.
E.g. for semantic class value `Equipment_LightSource_LightStripe`, the semantic tag is `LightStripe` and the semantic parent tag is `LightSource`. In other places in the UI code it is rather called semantic class, semantic value, or just value, but not tag. It has put me on the wrong leg a few times.

mherwege:
> especially the Lights category which seems to includes LightSource child equipment and also not LightSource as well .. and I am not sure why..

I admit I have difficulty understanding it just looking at the code. It got introded here: https://github.com/openhab/openhab-webui/pull/3101 and links to an issue.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -135,20 +135,22 @@ export default {
           allPoints = allEquipmentPoints(equipment)
           points = [
             ...findPoints(allPoints, 'Point_Status', false),
-            ...findPoints(allPoints, 'Point_Control', true)
+            ...findPoints(allPoints, 'Point_Control', true),
+            ...findPoints(allPoints, 'Point_Switch', false)
```
mherwege: Switch is a child of Control I believe, so I don't think you need it explicitly if you check for the existence of Point_Control.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -135,20 +135,22 @@ export default {
...
+            ...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Display_Television', true),
             ...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Screen', true)
           ]
           if (!equipment.length) return []
           allPoints = allEquipmentPoints(equipment)
           points = [
             ...findPoints(allPoints, 'Point_Status', false, 'Property_Power'),
-            ...findPoints(allPoints, 'Point_Control', true, 'Property_Power')
+            ...findPoints(allPoints, 'Point_Control', false, 'Property_Power'),
+            ...findPoints(allPoints, 'Point_Switch', false, 'Property_Power')
```
mherwege: I don't think you need it.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -135,20 +135,22 @@ export default {
...
-            ...findPoints(allPoints, 'Point_Control', true)
+            ...findPoints(allPoints, 'Point_Control', true),
+            ...findPoints(allPoints, 'Point_Switch', false)
           ]
           if (points.length) return points
           return equipment.filter((e) => e.points.length === 0).map((e) => e.item)
         case 'screens':
           equipment = [
-            ...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Display', true),
+            ...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Display_Television', true),
```
mherwege: I am in doubt on this. What is the recommendation for the use of the Screen tag?

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -163,7 +165,8 @@ export default {
           allPoints = allEquipmentPoints(equipment)
           points = [
             ...findPoints(allPoints, 'Point_Status', false, 'Property_Power'),
-            ...findPoints(allPoints, 'Point_Control', true, 'Property_Power')
+            ...findPoints(allPoints, 'Point_Control', false, 'Property_Power'),
+            ...findPoints(allPoints, 'Point_Switch', false, 'Property_Power')
```
mherwege: I don't think it is needed. I thought it was Point_Control_Switch.

andrewfg: Aha! I was overthinking it. I will revert.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: As below..

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: As below

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: I dunno. Someone in the review process insisted that a screen is something different than a display. And honestly I just rolled over and didn't fight it..

andrewfg:
@mherwege the one where I have the biggest doubt is lighting. Typically a light may have a Color channel which the binding would propose as Property_Color. And the user may link a Color control, a Dimmer control and a Switch to that channel. So by default all Items would have Property_Color and therefore excluded from the badge. So if the user wants the light to be included the badge they need to change just one of those Items to Property_Light (which one). Also all the Items will be (children of) Equipment_LightSource .. but probably that is not relevant. => Thoughts?

andrewfg:
More on lights..

I guess that lights will be Equipment.Lightsource children. And they will have one Point.Switch Item. On a color light the switch Item is linked to a Property.Color channel. On a dimmable light the switch item is linked to a Property.Brigtness channel. On an on/off light the switch item is linked to a Property.Light channel or Property.Power.

So I guess the proper (new) way to collect the lights into badges is to look for Equipment.LightSource plus Point.Switch .. and ignore the Property entirely. However this is quite different than had been done before. What do you think?

EDIT so the collection would be `((Equipment.LightSource && Point.Switch) || Property.Light)` ..

mherwege:
@andrewfg I found the original design issue for this: https://github.com/openhab/openhab-webui/issues/556

It is good reading and contains details on what the intentions where, also for lights. I didn't think through all the consequences yet, but I think it is worth revisiting the current state against the intentions in the original design (and the few changes done on it afterwards, also linked there).

andrewfg:
@mherwege that is a very useful document. I will digest it and get back later today.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: So I include Television and Screen under screens. And Projector under projectors.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: Done

andrewfg:
@mherwege in light of https://github.com/openhab/openhab-webui/issues/556 and its subsequent issues I revised the documentation in https://github.com/openhab/openhab-docs/pull/2514 and the code here. And I think I recaptured the original intentions and documented them as such. However some notes:

1. I tweaked the code for lights -- the reason is based on my observations above in this thread, where a color lamp might have a switch item, an dimmer item and a color item linked to them, and we should not include all three such items in the collection.
 
2. I reverted the code for HVAC from Property.Any to Property.Power -- the reason is that many HVAC devices also have Property.Mode (e.g. heat/cool) or fan speed so we should only collect the Items that expressly relate to them being powered on.

3. I reverted the Point.Control to implicitly include children (Point.Switch) according to your comments above.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -222,11 +225,11 @@ export default {
...
       return this.exprAst
     },
-    queryLightEquipment (equipmentType) {
-      let equipment = findEquipment(this.element.equipment, equipmentType, true)
+    queryLightEquipment () {
+      let equipment = findEquipment(this.element.equipment, 'Equipment_LightSource', true)
       if (!equipment.length) return []
       let allPoints = allEquipmentPoints(equipment)
-      let points = findPoints(allPoints, 'Point', true, 'Property_Light')
+      let points = findPoints(allPoints, 'Point_Control_Switch', false)
```
mherwege: In the previous code, as it does not look for a Switch, it will also find dimmer and color items. That is interpreted in line 203. Here it only looks for a switch.
I think we should look broader, but also avoid adding up multiple points from one equipment (e.g. having a Point_Control with a Dimmer item type and Point_Control_Switch with a Switch item type). In such case, the Switch item type should be used.

mherwege:
> 1. I tweaked the code for lights -- the reason is based on my observations above in this thread, where a color lamp might have a switch item, an dimmer item and a color item linked to them, and we should not include all three such items in the collection.

We wrote in parallel.

I am afraid this may cause quite a bit of questions for people that only have Dimmer items defined, suddenly not showing up anymore. I think we need to be more intelligent and first look for a Switch item, then a Dimmer item, then a Color item on the same Equipment.
I don't think we should force people to create extra items just to have their lights being counted again.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: > I think we should look broader, but also avoid adding up multiple points from one equipment (e.g. having a Point_Control with a Dimmer item type and Point_Control_Switch with a Switch item type). In such case, the Switch item type should be used.

I do basically agree with your statement :) .. however I don't know how to code it :( .. we would need to map each Item to the Thing that it comes from, and preferentially remove any duplications. => Perhaps you can kindly suggest some code?

andrewfg:
> We wrote in parallel.

Indeed. See my comment in response to your prior comment above.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: Or we could double count Projector both under screens and under projectors?

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
mherwege: > I do basically agree with your statement :) .. however I don't know how to code it :( .. we would need to map each Item to the Thing that it comes from, and preferentially remove any duplications. => Perhaps you can kindly suggest some code?

I need to look into it in more detail. I also don't think we should make it dependent on the thing (you could have 2 items essentially representing the same light, but linked to a different thing for some obscure reason). It should only be looking at the semantic relationships. What about prefering Point_Control_Switch over Point_Control? That may just do.

When you are done working on this, I will have a look to improve it.

mherwege:
In the following method, you changed from Point_Control to Point in the previous PR. It should probably be reverted to Point_Control as Point_Control_Switch would be covered as well.

```
    queryLightPoints () {
      let direct = findPoints(this.element.properties, 'Point', true, 'Property_Light')
      if (direct.length) return direct
      return findPoints(allEquipmentPoints(this.element.equipment), 'Point', true, 'Property_Light')
    }
```

mherwege:
@andrewfg I created a PR against your branch. It is not tested, but I believe it should largely solve the light issues. Please have a look.
Do you have a setup that allows you to test if it behaves?

andrewfg:
> Do you have a setup that allows you to test if it behaves?

Nope. I am just coding blind. ;)

mherwege:
@andrewfg I tested with the code changes I have done for lights, and it works for me. You would need to include the changes from my PR to your github repository on your PR branch.

andrewfg:
@mherwege see my comments to your PR

andrewfg:
@mherwege the CI build reported 'Unexpected side effect in "overrideExpression" computed property' at line `#239` and Copilot suggested the fix which I pushed in the last commit. => However please check it.

mherwege:
> @mherwege the CI build reported 'Unexpected side effect in "overrideExpression" computed property' at line `#239` and Copilot suggested the fix which I pushed in the last commit. => However please check it.

That's my mistake. As the method had no arguments, I just moved it to a computed property with the idea of avoiding recalculation if nothing changes. But the side effect is needed.
Your solution avoids the side effect, but does not set the property. So it loses part of the purpose.
So, it has to be moved back to a method like it was before, and all calls to it need the parantheses behind it to indicate a method is called (and not a computed property).

andrewfg:
> it has to be moved back to a method like it was before

Done.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -207,9 +208,29 @@ export default {
...
+        equipmentPoints = findPoints(e.points, 'Point_Control', true, 'Property_Light')
+        if (equipmentPoints.length) return equipmentPoints.slice(0, 1)
+        if (isLightSource) {
+          equipmentPoints = findPoints(e.points, 'Point_Control', false)
+          if (equipmentPoints.length) return equipmentPoints.slice(0, 1)
+        }
+        return []
+      }).flat())
+      // If there are no points, use the equipment items themselves
+      return equipment.filter((e) => e.points.length === 0).map((e) => e.item)
```
mherwege: Looking at this again, my last commit (trying to remove the Set duplication removal), also removed the return statement for points. That should go back in.
Also equipment with no points (but an item) will only get added when there are no points at all now.
```suggestion
      points.push(...equipment.filter((e) => e.points.length === 0).map((e) => e.item))
      return points
```

mherwege:
@andrewfg Thanks for this. Someone else needs to further review now.

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
andrewfg: HOWEVER at line `#233` I wonder if you should insert `if (points) return points` before trying to push the equipment Or ??

andrewfg:
> Someone else needs to further review now.

@mherwege can you please just check my last comment above?

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
mherwege: No, I don't think so. What gets added here are LightSource Equipment items, without any Points at all. The state of the LightSource item is simply used then. And that is a valid simplification in the model. I believe these should be added on top of the others that do have Points (where the treatment is looking at the Points).
My whole point about pushing it in the points array was to include them. Otherwise I could have just returned them.

mherwege:
> > Someone else needs to further review now.
> 
> @mherwege can you please just check my last comment above?

@andrewfg I did, and I have another interpretation.

andrewfg:
Ok. Shall I revert my last two commits then?

mherwege:
> Ok. Shall I revert my last two commits then?

That's my view on it indeed. You don't have to agree.
Before, it was not including them if there were LightSource equipments with Light points. But I think they should be included in all cases if it is a LightSource equipment.

andrewfg:
I think my point is to avoid the case of groups with points where both group and point mean that the badge shows two lamps rather than one.

mherwege:
> I think my point is to avoid the case of groups with points where both group and point mean that the badge shows two lamps rather than one.

Yes, but that shouldn't happen, because the last statement first selects on equipment without points (and the equipment could be a group or an item). If if it has no points, a group could still have non-semantic children and a state. So the equipment could still have a state, even if it is a group.

andrewfg:
> the last statement first selects on equipment without points

Ah yes. My bad.

andrewfg:
I will revert. But will also change the comment above..

andrewfg:
@mherwege many thanks for guiding me through this one; this is the first time that I coded in JavaScript inside Vue, so I apologise for my lack of knowledge.

andrewfg:
@florian-h05 , @ghys this took a long time (see my comment just above) and since @mherwege kindly contributed his own code to help me, it would be good if one of you could check this PR also yourself. TIA.

PS you may want to cross reference with https://github.com/openhab/openhab-docs/pull/2514

andrewfg:
@florian-h05 I noticed that you marked this as enhancement (which it certainly is) however it does also have some aspect of fixing regression arising from recent changes in tag parent / child relationships. So could we please try to get it in v5.0 release?

https://community.openhab.org/t/openhab-5-0-milestone-discussion/162686/224

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
```diff
@@ -126,22 +126,23 @@ export default {
...
+            ...findPoints(allPoints, 'Point_Status', false, 'Property_Power'),
+            ...findPoints(allPoints, 'Point_Control', true, 'Property_Power')
           ]
           if (points.length) return points
           return equipment.filter((e) => e.points.length === 0).map((e) => e.item)
         case 'screens':
           equipment = [
-            ...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Display', true),
+            ...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Display', false),
+            ...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Display_Television', false),
```
florian-h05, line 145: Why is this explicitly needed?
Couldn't we use 

```js
...findEquipment(this.element.equipment, 'Equipment_AudioVisual_Display', true),
```

?

florian-h05 (commented):
Code LGTM, one question though:

bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
mherwege: Projector is also a child of Display, and that has a separate badge. We didn’t want to break that.

florian-h05: Okay, fine then :+1:

florian-h05 (approved):

PR #3251, opened 2025-06-29
Add Maven Wrapper & Use for CI builds

florian-h05:
This improves DX by adding the ability to use ./mvnw, so you aren't required to manually install Maven anymore.
It also fixes (rare) issues with platform specific Maven builds (e.g. https://github.com/codehaus-plexus/plexus-compiler/issues/394).

Also use Maven Wrapper in GitHub Actions CI, which eliminates the need for the stCarols/setup-maven action.


florian-h05:
@wborn WDYT?

wborn:
Yes I think I also suggested some time ago somewhere to replace the stCarols/setup-maven Action with using the wrapper.

florian-h05:
Then I will merge this :rocket:

PR #3263, opened 2025-07-14
Transformations List: Fix incorrect counter title

jimtng:
Note, the transformations list page has other bugs related to selections, but that could get a revamp with select/deselect all too.


florian-h05 (approved):
LGTM, thanks!
That title did not make any sense ...

Issue #3280, opened 2025-07-20
[BasicUI] Preferences: no need to spell “Default is enabled”


dilyanpalauzov:
In BasicUI OH 4.3.5 under Preferences I see:

> Enable Iconify Icons 
> If enabled, the UI will render the iconify icons directly by downloading them from the Internet and caching them in the WEB browser. If disabled, the UI will ignore the iconify icons. Default is enabled. 

and the value of the preference.

There is no need to spell "Default is enabled" here, as the user sees the current value.

In addition in "Enable Iconify Icons" the word "Enable" is also not necessary.  Spelling the property name is enough, otherwise writing "Enable” can be done for consistency before all other binary options. 

>  Iconify Icons

lolodomo:
Maybe rather " Icons Display" and "Iconify Icons Display".

PR #3281, opened 2025-07-21
[BaiscUI] Adjust prefs label/description related to icons

lolodomo:
Clsoes #3280


bundles/org.openhab.ui.basic/src/main/resources/OH-INF/i18n/basic.properties
```diff
@@ -30,9 +30,9 @@ ui.config.basic.condensedLayout.label = Condensed Layout
 ui.config.basic.condensedLayout.description = When enabled, changes the layout so that more widgets can fit on the screen.
 ui.config.basic.defaultSitemap.label = Default Sitemap
 ui.config.basic.defaultSitemap.description = The sitemap to show if no parameter is passed.
-ui.config.basic.enableIconify.label = Enable Iconify Icons
-ui.config.basic.enableIconify.description = If enabled, the UI will render the iconify icons directly by downloading them from the Internet and caching them in the WEB browser. If disabled, the UI will ignore the iconify icons. Default is enabled.
-ui.config.basic.enableIcons.label = Enable Icons
+ui.config.basic.enableIconify.label = Iconify Icons Display
+ui.config.basic.enableIconify.description = If enabled, the UI will render the iconify icons directly by downloading them from the Internet and caching them in the WEB browser. If disabled, the UI will ignore the iconify icons.
+ui.config.basic.enableIcons.label = Icons Display
```
jimtng: "Icons Display" and "Iconify Icons Display" sounds weird to me. 

I think It should be:

"Show Icons"
"Support Iconify Icons"

However, I'm wondering why do we even need to let users disable Iconify icons? Why not just support it? Does MainUI or anywhere else have a similar option? I would suggest simply removing this (iconify) preference.

jimtng: @dilyanpalauzov wdyt?

dilyanpalauzov: I think that the user does not know what means “iconify icons”, so describing the label “Iconify Icons (Display or Support)” as “the UI will render the iconify icons directly by downloading them from the Internet” is a recursive description of the option.  It describes what means “iconify icons” by using the term “iconify icons”, and this is supposed to be different from the “Icons (ON/OFF)” preference?

It gets clear after reading https://community.openhab.org/t/new-feature-in-main-ui-iconify-support-for-icons/126085 .  So an “iconify icon” is an icon downloaded from the `https://iconify.design/` domain.  I suggest

> ui.config.basic.enableIconify.label=Utilize Iconify Icons
> ui.config.basic.enableIconify.description=BasicUI downloads icons from the `https://iconify.design/` domain.

As far as I remember some preferences - dark/light/auto mode and Audio Sink (WebAudio) are shared between BasicUI/MainUI.  The cited article mentions that Iconify Icons are available in MainUI.  But the currently discussed preference is present in  BasicUI preferences, and absent in MainUI preferences.  Is my understanding correct, that in BacisUI Iconify Icons can be disabled, while in MainUI iconify icons cannot be disabled (per browser/device)?

If caching in browser is implicit for all kinds of downloaded icons, it does not need to be explicitly spelled in the description for “Iconify Icons”. 

If the label contains “Support Iconify Icons”, Utilize Iconify Icons, Download Iconify Icons, or Fetch Iconify Icons is a nuance.

Instead of domain, also serivce or host can be used.


In `Clsoes #3280` there is a typo.

lolodomo: > However, I'm wondering why do we even need to let users disable Iconify icons? Why not just support it?

Because some users don't like to have a feature requiring an Internet connection. So they can disable it. I disable it myself as I do not use Iconify.

jimtng: If they really don't want to "require an internet connection" for their icons, then they can simply not use iconify icons to start with, I would think.

jimtng:
BTW need to fix typo in the PR title

bundles/org.openhab.ui.basic/src/main/resources/OH-INF/i18n/basic.properties
lolodomo: Changed

lolodomo: > If they really don't want to "require an internet connection" for their icons, then they can simply not use iconify icons to start with, I would think.

Yes
When this setting is disabled, one JavaScript is not included in pages.

lolodomo:
> BTW need to fix typo in the PR title

Done.
I am impressed by the number of typo I am able to do in just few words!!

lolodomo:
@jimtng @dilyanpalauzov : is it ok with "Show Icons" and "Support Iconify Icons"? Sorry but I don't want to loose more time on something that looks to me like a detail and keep this PR opened more time.
in case there is no consensus with you, I will simply close my PR.
Pinging @kaikreuzer for a potential review & merge.
If something has to be merged, we have to do it quickly to keep time to adjust translations before 5.1 is released.

bundles/org.openhab.ui.basic/src/main/resources/OH-INF/i18n/basic.properties
jimtng: Saying "Utilize Iconify Icons" would seem to imply that _all_ icons will somehow use iconify, and as far as I understand it, this is not true. So "Support" instead of "Utilize" seems to be the more apt term here.

dilyanpalauzov:
I am fine with the currently proposed change, as it removes “Default is enabled”.

lolodomo:
@kaikreuzer : please have a look and merge if that is fine for you too.

lolodomo:
@kaikreuzer : any chance to review this change before frozen date, it is only 3 lines updated.

PR #3288, opened 2025-07-24
Pin icons: Make "pinned" state icon solid and "unpinned" state outlined

jimtng:
The way the "pin" icons are currently displayed:

When an item is pinned, the pin icon is shown as red with slash. This makes me wonder: is this item currently pinned or not?

So instead of the icon showing what will happen when it's clicked, show the current state (is it currently pinned or unpinned). Clicking it will toggle it.

I think it's more intuitive this way.

Furthermore, when it's unpinned, show it as gray color, and when it's pinned, show it as blue, indicating that "yes it's pinned".

Before:
image

After:
image


florian-h05:
@jimtng Please rebase.

jimtng:
rebased

PR #3292, opened 2025-07-27
Enable Dependabot

holgerfriedrich:
We have Dependabot active in core and distro for quite a while, and recently introduced it in add-ons repo.

The first part of the config just takes care of keeping _Github Actions_ up to date.
The second part takes care of _Maven plugins_.

Maven standard config will both update the plugins (which it typically wanted) but also all dependences (which is largely not wanted). This is why this PR uses only specific names, e.g. everything that matches the string `plugin`.

I have defined some exclusions. Maybe we overlook some plugins which are not matching the patterns; maybe we need some more exclusions.

Note: This config is carried over from add-ons without change, maybe a few allow/ignore statements are not needed.

See also 
openhab/openhab-addons#19029
openhab/openhab-addons#19039

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,45 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


florian-h05 (approved):
LGTM, thanks!
Good to have Dependabot :+1:

PR #3304, opened 2025-07-28
Upgrade Karaf from 4.4.7 to 4.4.8

holgerfriedrich:
depends on openhab/openhab-core#4932


holgerfriedrich:
The settings for the webui repo are a bit strict. I can't even add the rebuild label, nor trigger the GHA job from the job view.

wborn:
My workaround for that is always to add a line in a commit, remove it in another. Squash and force push. 😉

Issue #3276, opened 2025-07-16
Semantic Tag editor can not delete a tag


rkoshak:
## The problem

If I have a new tag and enter one or more synonyms and then decide I want to delete one or more, I can't delete them. Clicking X next to one makes it go away but then clicking anywhere else, including clicking on "save" causes the synonym to come back. 

Hitting ctrl-s after clicking the X doesn't make it reappear, but it also doesn't appear to remove the synonym. Clicking to a new tag and back shows it's still there.

## Expected behavior

Clicking X next to a synonym and saving causes that synonym to be removed.

## Steps to reproduce

1. Create a new semantic tag
2. Add a synonym
3. Click the X next to the synonym
4. Click anywhere else on the page and see the synonym reappear

## Your environment

```yaml
runtimeInfo:
  version: 5.0.0.M4
  buildString: Milestone Build
locale: en-US
systemInfo:
  configFolder: /openhab/conf
  userdataFolder: /openhab/userdata
  logFolder: /openhab/userdata/logs
  javaVersion: 21.0.7
  javaVendor: Eclipse Adoptium
  javaVendorVersion: Temurin-21.0.7+6
  osName: Linux
  osVersion: 6.8.0-63-generic
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 146648624
  totalMemory: 763363328
  uptime: 14407
  startLevel: 70
addons:
  - automation-jsscripting
  - automation-pidcontroller
  - binding-amazonechocontrol
  - binding-astro
  - binding-chromecast
  - binding-exec
  - binding-http
  - binding-ipcamera
  - binding-mail
  - binding-mqtt
  - binding-nest
  - binding-network
  - binding-networkupstools
  - binding-openweathermap
  - binding-shelly
  - binding-vesync
  - binding-zigbee
  - binding-zwave
  - misc-openhabcloud
  - persistence-mapdb
  - persistence-rrd4j
  - transformation-basicprofiles
  - transformation-jinja
  - transformation-jsonpath
  - transformation-map
  - transformation-regex
  - ui-basic
  - ui-habot
  - voice-pipertts
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: true
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
[...]

Issue #3275, opened 2025-07-16
Semantic Tags Editor doesn't show up after a browser refresh


jimtng:
When I go to Developer Tools -> Semantic Tags, it would show the semantic tree as expected. However if I click my browser's Refresh button, the page content which holds the semantic tree doesn't appear.

<img width="736" height="331" alt="Image" src="https://github.com/user-attachments/assets/5a922835-3329-4fe8-8909-736b5bdb14c3" />

jimtng:
@mherwege do you also see this issue on yours?

mherwege:
@jimtng Sorry, I will not be able to test before release. It is not impossible there is a bug there.

rkoshak:
I'm seeing the same behavior, for what that's worth.

PR #3307, opened 2025-07-30
Semantic tags page: Fix page refresh & delete synonym bugs 

mherwege:
Fixes https://github.com/openhab/openhab-webui/issues/3275 (page refresh)
Fixes https://github.com/openhab/openhab-webui/issues/3276 (unable to delete synonym)


florian-h05 (changes requested):
Hmm, it seems the fix for the synonyms is not working properly, I can neither add not delete a semantic tag.
Somehow the save method doesn't see the tag with modified synonyms as modified tag.

mherwege:
@florian-h05 This should be fixed. The CI build problem seems to be unrelated.

PR #3309, opened 2025-08-01
Copy File Definition: Work around the security error in Safari

jimtng:
Reported in https://community.openhab.org/t/copy-dsl-yaml-definition-not-working-for-me/165233

This PR tests the clipboard copy functionality across browsers. If the operation fails, e.g. in Safari due to platform restrictions, an additional dialog is shown after the async fetch completes. This ensures that when the user clicks the "Copy" button, the clipboard action executes immediately using the pre-fetched data, complying with Safari's transient activation requirements.

The additional dialog is only shown in browsers that have such restriction.


jimtng:
Please backport to 5.0.x

florian-h05 (commented):
Can you please refactor the clibpoard copy code into a function residing in a new clipboard.js file in the js directory of the Main UI source?
We should share this fix across the UI.

jimtng:
Note the code in log-viewer didn't need this workaround, but refactoring it to use the common code will make the migration to vue3 _much_ easier

bundles/org.openhab.ui/web/src/js/clipboard.js
```diff
@@ -0,0 +1,34 @@
+import Vue from 'vue'
+import { f7 } from 'framework7-vue'
+
+export default function (copyOptions) {
```
florian-h05: Would be nice if you could add JSDoc here so IntelliSense can hint about the required properties of thr copyOptions object.

jimtng: jsdoc added

florian-h05 (approved):
LGTM, thanks!
I will merge once back at home and being able to backport.

Issue #3222, opened 2025-06-10
Search in developer bar doesn't show rules triggered


ErikDB87:
## The problem

As discussed in the "community" (https://community.openhab.org/t/search-in-developer-bar-doesnt-show-rules-triggered/164442):
My search for `gWindowsOpen` only finds the item itself. But there's a rule which has a trigger that revolves around the item. Wouldn't it make sense to have those pop up as well?

Because I want to do some reorganization, and in my future setup, `gWindowsOpen` would not be a logical trigger for this rule anymore. If I had not searched manually, that would have led to problems.

![Image](https://github.com/user-attachments/assets/19dd5432-88f8-4262-b9c2-6d3fe7db9ea5)

## Expected behavior

I expect the rule to be among the search results.

## Steps to reproduce

1. Choose an item which is involved in a rule trigger.
2. Search for that item in the developer bar.

## Your environment

```yaml
runtimeInfo:
  version: 4.3.5
  buildString: Release Build
locale: nl-BE
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 21.0.7
  javaVendor: Ubuntu
  osName: Linux
  osVersion: 6.8.0-60-generic
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 89225952
  totalMemory: 479199232
  uptime: 6071
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-astro
  - binding-buienradar
  - binding-chromecast
  - binding-enocean
  - binding-gpstracker
  - binding-hpprinter
  - binding-http
  - binding-icloud
  - binding-mail
  - binding-mqtt
  - binding-network
  - binding-networkupstools
  - binding-openweathermap
  - binding-semsportal
  - binding-sonos
  - binding-tradfri
  - binding-upnpcontrol
  - misc-openhabcloud
  - persistence-rrd4j
  - transformation-jsonpath
  - ui-basic
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
    pixelRatio: 1.25
    prefersColorScheme: light
  isSecureContext: false
  locationbarVisible: true
[...]

PR #3312, opened 2025-08-01
Developer Sidebar: Add rule.trigger.configuration.groupName to the search (OH 4.3.x)

jimtng:
A fix for #3222 for OH4.3

Note I haven't actually tested this on a 4.3.x installation.

A fix for this on OH5 is included in #3289, if that doesn't get merged, we'll have a regression unless this gets "forward ported". In fact, #3289 will go into 5.1, so I guess I'll also need to make this for 5.0.x


florian-h05:
I think this needs to go both to 5.0.x and 4.3.x. Change the head to 5.0.x, test it there and I can backport it to 4.3.x.

PR #3323, opened 2025-08-08
[Main UI] Modified vue/max-attributes-per-line eslint setting

jsjames:
This is ONLY a formatting change with the following eslint configuration:

    'vue/max-attributes-per-line': ["error", {
        'singleline': 3,
        'multiline' : 1
    }],

This will eliminate unnecessary comparisons with the pending vue3 branch.

LONG STORY

I had already used default settings in prettier which reformatted the code. While you can set the max-attributes-per-line, there is no easy way (from what I can tell) to return to the original code.  So, by aligning these settings in eslint between the two versions, it will minimize deltas.


florian-h05:
Thanks for the PR, I have created #3324 applying the new ESLint myself ... those thousands of lines are pretty difficult to review, easiest thing to do is just apply the linting yourself in such a case.

PR #3332, opened 2025-08-13
[WIP] Add refresh interval option to photos action in OH WebUI

Anpro88:
This is my first attempt at contributing to the OH WebUI.  
The goal is to add an optional refresh interval to the `photos` action so that images in the Framework7 Photo Browser can automatically reload after a defined number of seconds or milliseconds.

Changes include:
- Added `actionPhotosRefreshSec` (seconds) and `actionPhotosRefreshMs` (milliseconds) configuration parameters.
- Adjusted the `photos` action handling to trigger periodic refreshes when one of these values is set.

Status:
- This PR is still a work in progress (WIP).
- I currently have no test environment available, so I cannot verify the changes myself.
- Feedback from maintainers and other contributors is highly appreciated, especially regarding implementation style and any potential side effects.

Thanks in advance for your patience with my first PR attempt!

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.js
@@ -108,6 +108,14 @@ export const actionParams = (groupName, paramPrefix) => {
       .v((value, configuration, configDescription, parameters) => {
         return ['photos'].indexOf(configuration[paramPrefix + 'action']) >= 0
       }),
+    pi(paramPrefix + 'actionPhotosRefreshSec', 'Photo Refresh Interval (seconds)', 'Number of seconds between automatic refreshes of the photos')
+      .v((value, configuration, configDescription, parameters) => {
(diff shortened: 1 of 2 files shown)


ghys:
> This is my first attempt at contributing to the OH WebUI.

And you are most welcome!

> Added actionPhotosRefreshSec (seconds) and actionPhotosRefreshMs (milliseconds) configuration parameters.

Maybe from a UX perspective you could merge those and have the value be `5s` or `200ms` instead.

Otherwise, the code looks good!

Anpro88:
Thanks a lot for the feedback! 🙏
Just to clarify your suggestion:
Did you mean that instead of having both actionPhotosRefreshSec and actionPhotosRefreshMs, I should just provide one single parameter (e.g. always in seconds or always in milliseconds)?
Or were you actually suggesting to support a string format like 5s or 200ms?
I just want to make sure I understood you correctly before trying to update the PR.

Issue #3325, opened 2025-08-09
Location card light badge ignores second light in same equipment (works only for one)


obrador:
## The problem

Recently upgraded from OH3 to OH5, new installation and creating stuff from scratch.

In OH3, the Location card light badge would appear if **any** light in the room was on, even when there were multiple lights under the same Equipment.  
- If one light was on, the badge showed a light icon.  
- If more than one was on, it showed the light icon with a number (e.g., “2”).  

In OH5, when there are two or more lights (`Point: Control → Light`) under the **same Equipment**, the badge only reacts to one of them (apparently the first one).  
- If only the second light is on, **no badge appears at all**.  
- The badge never shows a count anymore for lights under the same Equipment.  

This is a regression from OH3 and makes the badge unreliable — it can give the impression that no lights are on when they actually are.

---

## Your suggestion

The badge should appear if **any** light under the Location is on, regardless of whether the lights are in the same Equipment or not.  
It should also count all lights that are on, as in OH3.

---

## Your environment

runtimeInfo:
  version: 5.0.0
  buildString: Release Build
locale: ca-ES
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 21.0.8
  javaVendor: Oracle Corporation
  osName: Linux
  osVersion: 6.1.0-37-amd64
  osArchitecture: amd64
  availableProcessors: 2
  freeMemory: 117186384
  totalMemory: 563085312
  uptime: 772979
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-amazonechocontrol
  - binding-astro
  - binding-mqtt
  - binding-telegram
  - persistence-influxdb
  - persistence-inmemory
  - persistence-mapdb
  - persistence-rrd4j
  - ui-basic
clientInfo:
  device:
    ios: false
    android: true
    androidChrome: true
    desktop: false
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: false
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    os: android
    osVersion: "13"
    webView: false
    webview: false
    standalone: false
    pixelRatio: 2.5
    prefersColorScheme: light
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 8
    language: en-US
[...]

PR #3344, opened 2025-08-25
Location card light badge and sub-equipment

mherwege:
Resolves https://github.com/openhab/openhab-webui/issues/3325#issuecomment-3219709449

This reverts part of the logic in https://github.com/openhab/openhab-webui/pull/3239

In the above PR, one equipment was expected to contain at most one light. It turns out that quite a few users created a semantic model that has one equipment containing multiple light points, and relied on the batch in the UI showing this.

This PR reverts that change, but it does come with a condition to be meaningful: one physical light should only be tagged once as a light. If it is tagged multiple times (e.g. a switch channel and a brightness channel tagged as points both tagged with property Light), it will double count.
This is a fair assumption as bindings should only create a single channel for a light (with the broadest functions, e.g. create a color channel, not a brightness channel, not a switch channel). We should guard the semantic tagging of channels in bindings to also respect make sure there is only one channel tagged, even if multiple channels would exist.

The badges will now also consider nested equipment and count properties in these nested equipment. The logic for all the badges has been synchronized with the logic for light equipment:
- Always consider both points on the location and the equipment
- Consider points on nested equipment
- Consider the item for equipment with the right tag without any points

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
+++ b/bundles/org.openhab.ui/web/src/components/cards/glance/location/status-badge.vue
@@ -204,7 +204,9 @@ export default {
       }
     },
     map () {
-      return this.query.map((item) => this.store[item.name].state)
+      // Make sure items are only counted once
+      const itemNames = new Set(this.query.map((item) => item.name))
+      return itemNames.map((itemName) => this.store[itemName].state)
     },
     reduce () {
       const ast = this.overrideExpression()


andrewfg:
See also

- https://github.com/openhab/openhab-docs/pull/2552
- https://github.com/openhab/openhab-addons/issues/19244

mherwege:
Build failure seems to be due to an npm rate limit, not related to the code I believe.

mherwege:
@andrewfg FYI, I have also enhance the badges to consider sub-equipment. So far, the badges only looked one level deep.

andrewfg:
@mherwege I see you have tweaked doors, windows, batteries, etc. to also allow nested equipments. That makes a lot of sense to me.

I am not a great expert in JavaScript but I understand that its array construct is more like a Java List than a Java Set insofar as the former may contain duplicates and the latter not. Now the location card code includes several alternate pathways whereby an item could be added to the card array e.g. via the location or via the (sub) equipment; and I haven't really thought this through but I do wonder if it is possible for the same item to be added twice to the final array via two code pathways? So perhaps somewhere your code should de-duplicate the final result? Or??

mherwege:
> Now the location card code includes several alternate pathways whereby an item could be added to the card array e.g. via the location or via the (sub) equipment; and I haven't really thought this through but I do wonder if it is possible for the same item to be added twice to the final array via two code pathways? So perhaps somewhere your code should de-duplicate the final result? Or??

I believe that could only happen if there is a loop in the group structure. Is that possible? But it would not harm to filter for non unique items in the final list indeed.

andrewfg:
```
var arr = [55, 44, 65];
var set = new Set(arr);
```

mherwege:
> ```
> var arr = [55, 44, 65];
> var set = new Set(arr);
> ```

This only works for primitives, not objects. Doing it on objects is more involved. But I solved it in another place.

andrewfg:
Based on the discussion in the other PRs I think we should go ahead with this. @mherwege do you agree? The other PRs will take more time. In particular retagging the 37 addons. But I think there is no reason for this PR to be delayed. => So LGTM

mherwege:
> @mherwege do you agree?

Agreed.

florian-h05:
@mherwege Please rebase

mherwege:
@florian-h05 rebase done

andrewfg:
@mherwege just for info; could you just quickly summarise the 6 commits you made yesterday? I don’t want to re review but it would be useful to know your logic..

mherwege:
@andrewfg These are the original commits, but rebased. This was needed as the move to Vue3 touched the same lines. So I rebased the branch on main. Leading to the 6 commits being adjusted.

jsjames:
@mherwege - I get the following error with this commit when going to the Locations tab.

status-badge.vue:209 Uncaught (in promise) TypeError: itemNames.map is not a function
    at Proxy.map (status-badge.vue:209:24)
    at refreshComputed (reactivity.esm-bundler.js:391:28)
    at get value (reactivity.esm-bundler.js:1655:5)
    at Object.get [as map] (runtime-core.esm-bundler.js:3591:22)
    at Object.get (runtime-core.esm-bundler.js:3156:16)
    at Proxy.reduce (status-badge.vue:226:23)
    at refreshComputed (reactivity.esm-bundler.js:391:28)
    at get value (reactivity.esm-bundler.js:1655:5)
    at Object.get [as reduce] (runtime-core.esm-bundler.js:3591:22)
    at Proxy._sfc_render (status-badge.vue:2:39)

mherwege:
> @mherwege - I get the following error with this commit when going to the Locations tab.
> 
> status-badge.vue:209 Uncaught (in promise) TypeError: itemNames.map is not a function at Proxy.map (status-badge.vue:209:24) at refreshComputed (reactivity.esm-bundler.js:391:28) at get value (reactivity.esm-bundler.js:1655:5) at Object.get [as map] (runtime-core.esm-bundler.js:3591:22) at Object.get (runtime-core.esm-bundler.js:3156:16) at Proxy.reduce (status-badge.vue:226:23) at refreshComputed (reactivity.esm-bundler.js:391:28) at get value (reactivity.esm-bundler.js:1655:5) at Object.get [as reduce] (runtime-core.esm-bundler.js:3591:22) at Proxy._sfc_render (status-badge.vue:2:39)

Sorry for that. I must have missed something in the rebase. I will only have the possibility to look into this next week I am afraid.

jsjames:
> > @mherwege - I get the following error with this commit when going to the Locations tab.
> > status-badge.vue:209 Uncaught (in promise) TypeError: itemNames.map is not a function at Proxy.map (status-badge.vue:209:24) at refreshComputed (reactivity.esm-bundler.js:391:28) at get value (reactivity.esm-bundler.js:1655:5) at Object.get [as map] (runtime-core.esm-bundler.js:3591:22) at Object.get (runtime-core.esm-bundler.js:3156:16) at Proxy.reduce (status-badge.vue:226:23) at refreshComputed (reactivity.esm-bundler.js:391:28) at get value (reactivity.esm-bundler.js:1655:5) at Object.get [as reduce] (runtime-core.esm-bundler.js:3591:22) at Proxy._sfc_render (status-badge.vue:2:39)
> 
> Sorry for that. I must have missed something in the rebase. I will only have the possibility to look into this next week I am afraid.

Should this be pulled out of the code until then, the Locations page is totally broken and nothing is displayed.

florian-h05:
I will take care

PR #3349, opened 2025-09-03
Adapt UI's to core sitemap registry implementation

mherwege:
Depends on https://github.com/openhab/openhab-core/pull/5004

This is required if the core PR would be merged, or BasicUI and CometVisu will not work anymore.

I have tested with BasicUI. I don't use CometVisu but have done similar changes as for BasicUI.

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
@@ -17,8 +17,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Objects;
 
-import org.eclipse.emf.common.util.ECollections;
-import org.eclipse.emf.common.util.EList;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.i18n.LocaleProvider;
@@ -67,7 +65,7 @@ public class ChartRenderer extends AbstractWidgetRenderer {
     }
 
     @Override
-    public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) throws RenderException {
+    public List<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) throws RenderException {
         Chart chart = (Chart) w;
 
         try {
@@ -165,7 +163,7 @@ public class ChartRenderer extends AbstractWidgetRenderer {
         } catch (ItemNotFoundException e) {
             logger.warn("Chart cannot be rendered as item '{}' does not exist.", chart.getItem());
         }
-        return ECollections.emptyEList();
+        return List.of();
     }
 
     private void buildRow(Chart w, @Nullable String lab, String cmd, String current, StringBuilder rowSB)
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ColortemppickerRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ColortemppickerRenderer.java
@@ -13,12 +13,11 @@
 package org.openhab.ui.basic.internal.render;
 
 import java.math.BigDecimal;
+import java.util.List;
 import java.util.Objects;
 
 import javax.measure.Unit;
 
-import org.eclipse.emf.common.util.ECollections;
-import org.eclipse.emf.common.util.EList;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.i18n.LocaleProvider;
@@ -77,7 +76,7 @@ public class ColortemppickerRenderer extends AbstractWidgetRenderer {
     }
 
     @Override
-    public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) throws RenderException {
+    public List<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) throws RenderException {
         Colortemperaturepicker ctp = (Colortemperaturepicker) w;
 
         BigDecimal currentK = null;
@@ -154,7 +153,7 @@ public class ColortemppickerRenderer extends AbstractWidgetRenderer {
         snippet = processColor(w, snippet);
 
         sb.append(snippet);
-        return ECollections.emptyEList();
+        return List.of();
     }
 
     private BigDecimal getMinimumInKelvin(Colortemperaturepicker widget, Unit<?> widgetUnit,
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/PageRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/PageRenderer.java
@@ -130,7 +130,7 @@ public class PageRenderer extends AbstractWidgetRenderer {
             throws RenderException {
         // put a single frame around all children widgets, if there are no explicit frames
         if (!children.isEmpty()) {
-            Widget firstChild = children.get(0);
+            Widget firstChild = children.getFirst();
             Parent parent = itemUIRegistry.getParent(firstChild);
             if (!(firstChild instanceof Frame || parent instanceof Frame || parent instanceof Sitemap)) {
                 String frameSnippet = getSnippet("frame");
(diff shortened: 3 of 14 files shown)


bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/PageRenderer.java
```diff
@@ -171,8 +168,7 @@ private void processChildren(StringBuilder sbPre, StringBuilder sbPost, EList<Wi
                 // multiple %children% sections found -> log an error and ignore all code starting from the second
                 // occurance
                 if (parts.length > 2) {
-                    String widgetType = w.eClass().getInstanceTypeName()
-                            .substring(w.eClass().getInstanceTypeName().lastIndexOf(".") + 1);
+                    String widgetType = w.getClass().getSimpleName();
```
lolodomo: Should `w.getWidgetType()` be used ?

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SliderRenderer.java
```diff
@@ -70,8 +74,8 @@ public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) th
             // Search the unit in the item state
             // Do not use itemUIRegistry.getState(w) as it will return a DecimalType for a slider widget
             // even if the item state is a QuantityType
-            String itemName = w.getItem();
-            State state = itemName != null ? itemUIRegistry.getItemState(itemName) : null;
+            String itemName = Objects.requireNonNull(w.getItem()); // Checked at creation there is an item
```
lolodomo: I don't understand why `w.getItem()` would be always not null

Same in other renderers.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
```diff
@@ -70,11 +72,10 @@ public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) th
...
             String itemParam = null;
-            boolean forceAsItem = false;
-            if (chart.getForceAsItem() != null) {
-                forceAsItem = chart.getForceAsItem();
-            }
-            Item item = itemUIRegistry.getItem(chart.getItem());
+            boolean forceAsItem = chart.forceAsItem();
+            Item item = null;
+            String itemName = Objects.requireNonNull(w.getItem()); // Checked at creation there is an item
+            item = itemUIRegistry.getItem(itemName);
```
lolodomo: You can restore the `Item item = itemUIRegistry...`

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
```diff
@@ -92,14 +93,12 @@ public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) th
...
-            boolean legend = item instanceof GroupItem && !forceAsItem;
-            if (chart.getLegend() != null) {
-                legend = chart.getLegend();
-                if (chart.getLegend()) {
-                    chartUrl += "&legend=true";
-                } else {
-                    chartUrl += "&legend=false";
-                }
+            boolean legend = chart.hasLegend();
+            legend = legend || (item instanceof GroupItem && !forceAsItem);
```
lolodomo: You changed the behaviour and your behaviour is I believe wrong as there is a case the legend can no more be disabled.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ColortemppickerRenderer.java
```diff
@@ -86,59 +87,57 @@ public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) th
...
-                            try {
-                                int[] rgb = ColorUtil
-                                        .hsbToRgb(ColorUtil.xyToHsb(ColorUtil.kelvinToXY(currentK.doubleValue())));
-                                logger.debug("Current {} K => RGB {} {} {}", currentK, rgb[0], rgb[1], rgb[2]);
-                                currentRGB = "#%02x%02x%02x".formatted(rgb[0], rgb[1], rgb[2]);
-                            } catch (IllegalArgumentException | IndexOutOfBoundsException e) {
-                                logger.debug("Can't get RGB for {} Kelvin, bypassing color gradient!", currentK);
-                            }
-                        }
+        String itemName = Objects.requireNonNull(w.getItem()); // Checked at creation there is an item
```
lolodomo, line 89: I can't understand why the item would always be defined by the user.

mherwege: Most widgets (not all) require an item to be defined. That is true in DSL and UI definitions. When trying to define the widget in a sitemap file or in the UI for these widgets, an error will be generated and the sitemap will not be read into the registry.
As this is not true for all the widgets, the item field cannot be defined as non null. But I believe the render code allows it when the item does not have to be defined, and shortcuts when it has to be defined. That's the reason for the special comment. If you see I missed a case, please let me know.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
mherwege: Correct. I missed that. It means the field has to remain a Boolean nullable. I will check.

mherwege: I changed it and changed it in the core PR to remain Nullable.

mherwege:
@lolodomo Thank you for the review. I did some updates on your remarks and clarified why I think the item property cannot be null for most widget types.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ColortemppickerRenderer.java
lolodomo: Ok, understood.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
```diff
@@ -70,11 +72,9 @@ public EList<Widget> renderWidget(Widget w, StringBuilder sb, String sitemap) th
 
         try {
             String itemParam = null;
-            boolean forceAsItem = false;
-            if (chart.getForceAsItem() != null) {
-                forceAsItem = chart.getForceAsItem();
-            }
-            Item item = itemUIRegistry.getItem(chart.getItem());
+            boolean forceAsItem = chart.forceAsItem();
+            String itemName = Objects.requireNonNull(w.getItem()); // Checked at creation there is an item
```
lolodomo, line 74: Please use itemName at line 79 and 81

lolodomo:
Only one minor comment and then that is good for me for the UI part.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/ChartRenderer.java
```diff
@@ -26,8 +26,7 @@
 import org.openhab.core.items.GroupItem;
 import org.openhab.core.items.Item;
 import org.openhab.core.items.ItemNotFoundException;
-import org.openhab.core.sitemap.Chart;
-import org.openhab.core.sitemap.Widget;
+import org.openhab.core.model.sitemap.sitemap.Widget;
```
lolodomo: Looks like you restored wrong import ?

mherwege: Looks like it. That's what happens if not compiling everything in the modified core before. Give me some time to get all in core adjusted and then this should fall in place as well.

mherwege:
@lolodomo This should be good again. Of course final compile validation can only happen after the core merge.

lolodomo:
Waiting for the merge of the PR in core and then a successful build of the PR.

mherwege:
The `WidgetRenderer.renderWidget` return type was changed in the PR, but not every implementing class had this change applied. I did that now.

lolodomo:
@kaikreuzer 
Kai, can you please trigger a new snapshot build ?

lolodomo:
> @kaikreuzer
> Kai, can you please trigger a new snapshot build ?

It is OK, it was included in build this morning.

But snapshot build seems to block on add-ons.

PR #3359, opened 2025-09-16
new logic block to and/or multiple operands

stefan-hoehn:
New Block that allows to AND or OR multiple operands based on an idea posted [here](https://community.openhab.org/t/blockly-and-and-or-blocks-with-multiple-inputs-mutuator/166107)

**Design**

image

**Generated code**

``
if (((items.getItem('MyItem').state == 'ON')) || ((items.getItem('MyItem').state == 'OPEN')) || ((items.getItem('MyItem').state == 'OPENED'))) {
  items.getItem('MyItem').sendCommand('CLOSE');
}
``

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-logic.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-logic.js
@@ -17,6 +17,7 @@ export default function (f7) {
       this.setOutput(true, null)
       this.setColour('%{BKY_LOGIC_HUE}')
       this.setTooltip('returns undefined as value')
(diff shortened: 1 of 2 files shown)


florian-h05:
> Thinking about the design a few days later I wonder if it would be nicer if it read "one-of" instead of OR and "all-of" instead of AND. Would that be more intuitive for a non-IT person as we always wanted Blockly to be "readable" when looking at the blocks... Feedback appreciated...

Can you do both? I.e. have the AND, OR selection and then display a label in the center of the block "all-of", "one-of"?

stefan-hoehn:
Like so?

image

florian-h05:
Exactly

stefan-hoehn:
added the dynamic label based on AND / OR => all of / any of

florian-h05 (approved):
LGTM, thanks!

PR #3366, opened 2025-09-25
Fix "Select All" in channel-list.vue

### LINKED ISSUE #3343: Select All does not select all channels while adding an equipment
## The problem

- I'm developing the Shelly binding and discover a new device
- After adding from Inbox the device gets online and create dynamically all relevant channels
- Those channels are organized in channel groups - so far so good
- Now I'm adding the equipment to the model
- I tick the advanced checkbox, scroll down and click Select All
- First: I need to do that twice
- Secondly: Not all channels are selected. I need to click multiple times on Select All until all channels get selected. I continue until Select All does not selects more channels. 
- Then I could add the equipment to the model and all channels are linked.

Problem was also confirmed by a @igi can reproduce the issue on Linux.

## Expected behavior

On click on Select All selects ALL channels in the view (standard + advanced if the checkbox was ticked)

## Steps to reproduce

- Take a binding, which organizes the channels into multiple channel groups
- Discover devices, try to add equipment, click Select All

## Your environment

- OH 5.1.0-SNAPSHOT, running on macOS
- DEV build of the Shelly Binding
- Safari or Chrome

Nadahar:
In `multipleLinksMode` mode, `channel-list.vue` hides trigger channels, but `toggleAllChecks()` didn't take that into account, which led to exceptions since trigger channels lack an `itemType`. By excluding trigger channels from `toggleAllChecks()`, the exceptions are avoided, and "Select All" works.

I believe this fixes #3343.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/channel-list.vue
+++ b/bundles/org.openhab.ui/web/src/components/thing/channel-list.vue
@@ -245,6 +245,7 @@ export default {
     },
     toggleAllChecks (checked) {
       this.thing.channels.forEach((c) => {
+        if (this.multipleLinksMode && c.kind === 'TRIGGER') return
         const channelType = this.channelTypesMap.get(c.channelTypeUID)
         if (!channelType) return
         if (channelType.advanced && !this.showAdvanced) return


Nadahar:
@florian-h05 Did you miss this? It's a tiny fix that solves an annoying issue.

PR #3376, opened 2025-10-08
[blockly] change event attributes, add return block

stefan-hoehn:
Depends on: https://github.com/openhab/openhab-addons/pull/19443

The event object/attributes in script actions and conditions is now the same as in JSRule and Rule Builder.
Backwards compatibility will be retained for some time, though it is advised to adjust, which is done in this PR according to: 
https://github.com/openhab/openhab-js/blob/51f164dc9f11a29e6223e24268848aa8f0be168f/src/rules/rules.js#L471

Additionally **a new return block** was added to the "Run & Process" category with return parameter.

image

image

1) The block without a return parameter is a nice improvement to easily leave a script without using non-elegant if/else statements. Bei adding the return statement the script is immediately ended. In this case the return parameter is not important. 
2) By adding a return parameter (in particular true or false) this can be used for condition, e.g. by using this condition:

image

which can implemented by using an inline blockly script that evaluates to true or false.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
@@ -267,9 +267,9 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
     }
 
   javascriptGenerator.forBlock['oh_context_info'] = function (block) {
-    const oldContectInfoName = block.getFieldValue('contextInfo')
+    const oldContextInfoName = block.getFieldValue('contextInfo')
     // map attributes to the correct event attribute names
-    const contextInfo = attributeMap[oldContectInfoName]
+    const contextInfo = attributeMap[oldContextInfoName]
 
     const type = block.getFieldValue('asType')
     if (contextInfo === 'eventAvailable') return ['(event !== undefined)', javascriptGenerator.ORDER_ATOMIC]
@@ -371,7 +371,7 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
 
   /*
   * Allows quick return from a rule. This is useful to stop further processing of a rule.
-  * Also allows returning a value from a rule if required which can be using blockly for a script condition
+  * Also allows returning a value from a rule if required which allows using blockly for a script condition
   * Blockly part
   */
   Blockly.Blocks['oh_rule_return'] = {
@@ -393,7 +393,7 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
   * Code part
   */
   javascriptGenerator.forBlock['oh_rule_return'] = function (block) {
-    // we need to add a wrapper function to allow returning from the rule
+    // we need to enable the wrapper to allow returning from the rule
     javascriptGenerator.provideFunction_('wrapper', ['"use wrapper;"'])
     const value = javascriptGenerator.valueToCode(block, 'value', javascriptGenerator.ORDER_ATOMIC)
     return `return ${value} \n`


rkoshak:
When you add the docs, make sure to make it clear that return can only be used when the wrapper is enabled in the JS Scripting configs (MainUI -> Settings -> JS Scripting.

florian-h05:
Our idea was to make blockly automatically set the directive to enable the wrapper when return is used.

stefan-hoehn:
you can actually see this in line 397 that creates "use wrapper" in the code.

rkoshak:
Is that something that can be done from a JS script? For example a rule template? 

I think I've found ways for my templates to work around that but knowing it's possible might prove helpful in the future.

florian-h05:
> Is that something that can be done from a JS script? For example a rule template?

I haven't written docs for that yet, but the `'use wrapper'` directive can be used in any JavaScript condition in the UI.
It just has to be first or second line, i.e. top of the script.
For actions, you don't need it, there the wrapper is always on (unless you explictly disable it, but that's mainly implemented for the case there is any unexpected issue).

rkoshak:
So I can enable the wrapper but not disable the wrapper? I'm not trying to argue about anything, just understand in case it's helpful to me in the rule templates and so I understand how it works in Blockly so I can help on the forum better. 

And don't you have actions and conditions backwards in your description. I thought the decision was to always enable it for Conditions and optionally enable it for Actions based on the add-on property. I could be remembering backwards, but that was what I thought. The reasoning was having the return statement in Conditions was worth keeping it always enabled, and doing so had minimal impact.

stefan-hoehn:
Not sure if it helps. As soon as Blockly adds one of these blocks, it adds "'use wrapper'" to the top of the script that it generates.

florian-h05:
You can also disable it, there are three ways of using the wrapper:

- `'use wrapper'`
- `'use wrapper=true'`
- `'use wrapper=false'`

The first two are equivalent.

 > And don't you have actions and conditions backwards in your description. I thought the decision was to always enable it for Conditions and optionally enable it for Actions based on the add-on property. I could be remembering backwards, but that was what I thought. The reasoning was having the return statement in Conditions was worth keeping it always enabled, and doing so had minimal impact.

I think you have to switch actions and conditions there. For actions, it is always enabled, since it has minimal to no impact, for conditions, there is the global add-on setting and the use wrapper directive to control the behaviour.

rkoshak:
> You can also disable it, there are three ways of using the wrapper:

Excellent! Thanks!

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
```diff
@@ -252,22 +252,39 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
...
+      'oldItemState': 'event.oldState',
+      'itemName': 'event.itemName',
+      'itemCommand': 'event.receivedCommand',
+      'channel': 'event.channelUID',
+      'event': 'event.receivedEvent'
+    }
+
   javascriptGenerator.forBlock['oh_context_info'] = function (block) {
-    const contextInfo = block.getFieldValue('contextInfo')
+    const oldContectInfoName = block.getFieldValue('contextInfo')
```
florian-h05:
```suggestion
    const oldContextInfoName = block.getFieldValue('contextInfo')
```

Typo

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
```diff
@@ -351,4 +368,34 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
     let enable = (enableType === 'Boolean') ? enableValue : (enableValue === '\'true\'' || enableValue === '\'enabled\'')
     return `rules.setEnabled(${ruleUID}, ${enable});\n`
   }
+
+  /*
+  * Allows quick return from a rule. This is useful to stop further processing of a rule.
+  * Also allows returning a value from a rule if required which can be using blockly for a script condition
```
florian-h05:
```suggestion
  * Also allows returning a value from a rule if required which allows using blockly for a script condition
```

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
```diff
@@ -351,4 +368,34 @@ export default function defineOHBlocks_Scripts (f7, transformationServices) {
...
+      this.setHelpUrl('https://www.openhab.org/docs/configuration/blockly/rules-blockly-run-and-process.html#returnRule')
+    }
+  }
+
+  /*
+  * Allows quick return from a rule.
+  * Code part
+  */
+  javascriptGenerator.forBlock['oh_rule_return'] = function (block) {
+    // we need to add a wrapper function to allow returning from the rule
```
florian-h05:
```suggestion
    // we need to enable the wrapper to allow returning from the rule
```

PR #3398, opened 2025-10-29
Upgrade to Blockly v11 and respective plugin versions

jsjames:
Moved to v11 to address issues with '@mit-app-inventor/blockly-plugin-workspace-multiselect' in port to vue3.

- adopted new required import syntax
- Addressed issue with 'workspace' being reactive (https://groups.google.com/g/blockly/c/myo3FXL3r_Y/m/NLgPZNc0CgAJVue3)
- @blockly/field-multilineinput now required to support multi-line
- @blockly/field-colour now required to support color blocks
- added ability to add comments (via context menu)

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/_tmp/blocks-exec.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/_tmp/blocks-exec.js
@@ -1,5 +1,6 @@
 import * as Blockly from 'blockly'
 import { javascriptGenerator } from 'blockly/javascript'
+import { valueToCode } from '@/assets/definitions/blockly/utils.js'
 
 export default function defineOHBlocks_Exec (f7) {
   Blockly.Blocks['oh_exec'] = {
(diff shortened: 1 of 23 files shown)


bundles/org.openhab.ui/web/src/components/config/controls/blockly-editor.vue
```diff
@@ -1250,7 +1260,7 @@ export default {
     this.$emit('mounted')
   },
   methods: {
-    load () {
+    async load () {
```
florian-h05: Why make this async? There is no await or something like that ...

florian-h05:
As testing Blockly is always a larger effort for @stefan-hoehn, I wonder whether we shouldn't try to upgrade to Blockly 12 now, so we don't need to test twice. WDYT @jsjames?

jsjames:
> As testing Blockly is always a larger effort for @stefan-hoehn, I wonder whether we shouldn't try to upgrade to Blockly 12 now, so we don't need to test twice. WDYT @jsjames?

I did try, but the plugin @mit-app-inventor/blockly-plugin-workspace-multiselect does not yet support it [](https://github.com/mit-cml/workspace-multiselect/pull/106).

Since there is an issue with our vue3 version where multi-select does not currently work (I recall it being a runtime issue in the @mit-app-inventor/blockly-plugin-workspace-multiselect ), this was the most expedient way to address.  The question is how critical is it to get the multi-select feature working?

florian-h05:
That’s quite unfortunate.
Looking at https://github.com/mit-cml/workspace-multiselect/pull/106 & https://github.com/google/blockly/pull/9261 it doesn’t seem the issue will be solved soon, so I guess we’ll stay with Blockly v11.

jsjames:
> That’s quite unfortunate. Looking at [mit-cml/workspace-multiselect#106](https://github.com/mit-cml/workspace-multiselect/pull/106) & [google/blockly#9261](https://github.com/google/blockly/pull/9261) it doesn’t seem the issue will be solved soon, so I guess we’ll stay with Blockly v11.

v11 does get rid of the remaining deprecated packages warnings in package.json!

PR #3399, opened 2025-10-29
Fixed import issues/their workarounds for production build

jsjames:
Moved import in troublesome locations to <script setup> vs. importing asynchronously

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
@@ -164,15 +164,17 @@
       opacity 0.55
 </style>
 
-<script>
-import { f7, theme } from 'framework7-vue'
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue
florian-h05, line 1: For me, the thinh details page doesn't work anymore with that change in dev server:

```
link-add.vue:196 Uncaught (in promise) ReferenceError: Cannot access 'ChannelList' before initialization
    at link-add.vue:196:5
```

Moving the Item import to the setup script and removing the components section from the Vue Options API definition fixes this.

jsjames: That works for me as well - submitted that change.

PR #3400, opened 2025-10-29
[basicUI] ColorpickerRenderer.java: remove comment about jquery.miniColors

dilyanpalauzov:
The deleted text was just copied from bundles/org.openhab.ui.classic/src/main/java/org/openhab/ui/classic/internal/render/ColorpickerRenderer.java

At commit 30dfe0d9eedf76687aae bundles/org.openhab.ui.classic/src/main/resources/snippets/colorpicker.html has required jquery.minicolors, and bundles/org.openhab.ui.basic/src/main/resources/snippets/colorpicker.html has not required it.

So the deleted text was never correct.


lolodomo (approved):

lolodomo:
Note that your contribution is not properly signed and should be rejected in principle.
As it is just a change in a comment, I make an exception to the rules.

PR #3407, opened 2025-10-30
Better clarity for rule-edit alert dialog when missing label or id (issue #3379)

### LINKED ISSUE #3379: If the **label** is not set when creating a rule, the error message insists on setting a **name**
When I want to “Create rule” and have not set label, I get an error message:

Image

* Either the field “Label” should be renamed to “Name”, or the error message must be “Set label to the rule”.  Here label and name mean the same thing, but are different words, and this is confusing.

Besides, I think the name should not be mandatory, openHAB can work with rules without names and the user can later set the name.  So it can be left up to the user, if the user wants to set or not a rule name.

I tried to see what happens, when file-based automation creates rules, if the rules have no names set.  The Groovy program below creates rules, without setting names.  In the UI some of the rules are shown having a number as name and some have no names.
```groovy
import org.openhab.core.automation.module.script.rulesupport.shared.simple.SimpleRule
import org.openhab.core.config.core.Configuration
import org.openhab.core.automation.*
import org.openhab.core.automation.util.TriggerBuilder;
scriptExtension.importPreset("RuleSupport")

def void addR() {
  automationManager.addRule(new SimpleRule() {
    { // name = "The name could be set here, but is not"
      triggers = [
        TriggerBuilder.create().withId("trig1").withTypeUID("core.ItemStateChangeTrigger")
        .withConfiguration(new Configuration([itemName: "r"])).build()
      ]
    }
    @Override Object execute(Action module, Map<String, ?> inputs) {}
  })
}

for (int i = 0; i < 3; i++) addR() // these rules have name not set

for (int i = 0; i < 5; i++)
  automationManager.addRule(new SimpleRule() {
    {   // these rules have name "2"
      triggers = [
        TriggerBuilder.create().withId("trig1").withTypeUID("core.ItemStateChangeTrigger")
        .withConfiguration(new Configuration([itemName: "r"])).build()
      ]
    }
    @Override Object execute(Action module, Map<String, ?> inputs) {}
  })

automationManager.addRule(new SimpleRule() {
    { // this rule has name "3"
      triggers = [
        TriggerBuilder.create().withId("trig1").withTypeUID("core.ItemStateChangeTrigger")
        .withConfiguration(new Configuration([itemName: "r"])).build()
      ]
    }
    @Override Object execute(Action module, Map<String, ?> inputs) {}
})

addR() //rule has no name
```

with openHAB 5.1.0 Build 4849.

jsjames:
Fixes #3379.

@florian-h05 - should we be moving these messages to i18n?

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
@@ -548,11 +548,11 @@ export default {
         }
       }
       if (!this.rule.uid) {
-        f7.dialog.alert('Please give an ID to the rule')
+        f7.dialog.alert('Please provide a unique rule ID. The ID must not be empty and should only contain letters, numbers, hyphens or underscores.', 'ID required').open()
         return Promise.reject()
       }
       if (!this.rule.name) {
-        f7.dialog.alert('Please give a name to the rule')
+        f7.dialog.alert('Please provide a rule label. The label is required and will be shown in the UI to identify this rule.', 'Label required').open()
         return Promise.reject()
       }
       const promise = (this.createMode)


florian-h05:
> Should we be moving these messages to i18n?

Administration area isn't localised at all, so I wouldn't to that.

PR #3410, opened 2025-10-31
Semantic Tags Editor: Don't allow dashes in tag names

jimtng:


florian-h05 (approved):

mherwege:
For my understanding, what’s the reason to disallow this? If it worked before, this is a breaking change.

florian-h05:
The docs for the YAML configuration file specifies the name to only include alphanumeric chars, see https://next.openhab.org/docs/configuration/yaml.html#tag-full-uid.
So I'd expect this was wrongly allowed in the UI.

jimtng:
- If you try adding a semantic tag now (before this PR) with `-` you'll get Error Bad Request from the server. See https://github.com/openhab/openhab-core/blob/b1670ec2fae5e523f440f220b582eef8e9d004ff/bundles/org.openhab.core.semantics/src/main/java/org/openhab/core/semantics/internal/SemanticTagRegistryImpl.java#L130
- By requiring semantic tags to conform to "identifier" syntax, JRuby can refer to it as an identifier, e.g. `Semantics::LivingRoom`. If we allowed `-`, `Semantics::My-Tag` wouldn't be a valid identifier as the `-` is parsed as a negative / minus operator.

PR #3417, opened 2025-11-01
oh-input: Select a default input type based on Item type

jsjames:
A default HTML input type will be used based on the openHAB item type. The user can still override by typing in a different type.

Further changes:

- placeholder of field in the item-metadata-widget will be the default type value (based on item type) to make clear default value is being used
- elaborated on type field comments to make more clear
- cleaned up duplicate html code in oh-input
- in app.style, color-scheme was not being set correctly hence standard html widgets were not rendered in dark mode
- (unrelated) set class of checkbox/label in things-list to get proper whitespace formatting

Fixes https://github.com/openhab/openhab-webui/discussions/3381#discussioncomment-14839964.


bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
```diff
@@ -79,6 +57,23 @@ import { OhInputDefinition } from '@/assets/definitions/widgets/system'
...
+  switch (itemType) {
+    case 'Number':
+    case 'Dimmer':
+      return 'number'
+    case 'DateTime':
+      return 'datetime-local'
+    case 'Date':
+      return 'date'
+    case 'Time':
+      return 'time'
```
digitaldan: `Date` and `Time` are not standard item types,  although you can have a UoM type `Number:Time` .... or is `itemType ` being calculated somewhere else to these additional values?

jsjames: Good catch.

digitaldan: So I would match  any type starting with  `Number` to a `number`, this will catch all the various UoM variations, of which there are many for `Number` 
https://www.openhab.org/docs/concepts/units-of-measurement.html

bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
```diff
@@ -79,6 +57,17 @@ import { OhInputDefinition } from '@/assets/definitions/widgets/system'
 
 import { useStatesStore } from '@/js/stores/useStatesStore'
 
+export function getHtmlInputType (itemType) {
```
florian-h05: It somehow feels wrong having this in here and importing it in other places. I think it should rather go into a dedicated file, but the question is where to put that.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
```diff
@@ -79,6 +57,17 @@ import { OhInputDefinition } from '@/assets/definitions/widgets/system'
...
+  if (!itemType) return 'text'
+  if (itemType.startsWith('Number')) {
+    return 'number'
+  } else if(itemType === 'DateTime') {
```
florian-h05:
```suggestion
  } else if (itemType === 'DateTime') {
```

bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-widget.vue
florian-h05, line 1: I can see what you want to achieve here, but IMO that feels very hacky and I don't think its a good idea to make the item-metadata-widget component work with widget description internals.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
```diff
@@ -1,8 +1,14 @@
 import { pi, pb, pt, pd } from '../helpers.js'
 
+export const ptType = pt('type', 'HTML Input Type',
+  'HTML input type for different OpenHAB Item types.\
```
florian-h05:
```suggestion
  'HTML input type for different openHAB Item types.\
```

florian-h05: I do think the previous description shouldn't be removed completely, as it provided helpful guidance if you want to set the type manually. The assumed default types should be combined with the existing description.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
florian-h05, line 1: Maybe this file is a place to put the function "get the default input type" to?

bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-widget.vue
jsjames: I don't disagree that this is "hacky", but I think the functionality is pretty useful as it provides the user with the default. Do you have a recommendation on how to fit this into the "description" internals in a more elegant manner?

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
jsjames: are you referrinmg to the link to the f7 page?  I included any f7 custom input types below where I think it makes sense to link to what is reasonable for different oh item types? can you elaborate on what you'd like to see?

bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
jsjames: I can move assuming we keep the feature in the item-metadata-widget. Again, if you have a suggestion for a cleaner implemnentation, it would be good.

bundles/org.openhab.ui/web/src/components/widgets/system/oh-input.vue
jsjames: prettier can't come soon enough!! :)

florian-h05: Yes. I am a bit hesitant with enabling it right now with all those open PRs, because this could require additional rebasing etc. Would like to get the big ones merged before enabling it.

bundles/org.openhab.ui/web/src/components/item/metadata/item-metadata-widget.vue
florian-h05: No, unfortunately I don't.
Let's keep it that way.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/input.js
florian-h05: Just understood what you did here, that's better than simply linking to the docs!

PR #3420, opened 2025-11-01
Refactor variable-mixin to be imported module vs. mixin

jsjames:
In order to start using composition api for future  components, we need to convert mixins to either composables or other module imports. This PR refactors variable-mixin to be function imports from module. This was straight forward as all functions in the mixin were standalone and only operating on parameters passed to the functions - no code changes to the functions themselves was required.


bundles/org.openhab.ui/web/src/js/openhab/variable.ts
florian-h05, line 1: I think this should rather go to src/components/widgets.

florian-h05 (approved):
Thanks!
I think some types could be refined further if we would analyze the usage of the code and get to know the types this way, but that isn't needed here.

PR #3425, opened 2025-11-02
oh-nav-content: Fix back navigation navigates to subpage

florian-h05:
The back button should always return to a previous page that is not a subpage of the current page. This fixes an issue where: Settings Menu -> Items -> Settings Menu and then clicking back on Settings Menu navigated back to Items.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/navigation/oh-nav-content.vue
+++ b/bundles/org.openhab.ui/web/src/components/navigation/oh-nav-content.vue
@@ -91,8 +91,8 @@ function back () {
   }
 
   if (previousPath === null) {
-    console.warn('No previous path found in history, cannot navigate back.')
-    return
+    console.warn('No previous path found in history, falling back to root path.')
+    previousPath = '/'
   }
   console.debug('Navigating back to previous path:', previousPath)
   f7router.navigate(previousPath, { force: true })


florian-h05:
@jsjames @digitaldan  WDYT?

jsjames:
> @jsjames @digitaldan WDYT?

sounds like a good solution.

bundles/org.openhab.ui/web/src/components/navigation/oh-nav-content.vue
```diff
@@ -80,13 +80,21 @@ function back () {
...
+  let previousPath : string | null = null
+  for (let i = f7router.history.length - 1; i >= 0; i--) {
+    const path = f7router.history[i]
+    if (!path.startsWith(currentPath)) {
+      previousPath = path
+      break
+    }
+  }
+
+  if (previousPath === null) {
```
digitaldan, line 93: I wonder if `previousPath` is null, we should default to going home or `/` ?

florian-h05: Good idea. Both of your suggested options will lead to the same page.
I will implement that later.

florian-h05: Done

PR #3450, opened 2025-11-11
Widget expressions: Add access to global i18n translations

florian-h05:
The naming of the function has to be discussed, this is just a PoC.
I would propose `translation`, so it is used like `=translation('Floor')`.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/widget-expression-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/widget-expression-mixin.js
@@ -126,6 +126,7 @@ export default {
             JSON,
             dayjs,
             user: useUserStore().user,
+            translations: i18n.global.t,
             t: i18n.global.t
           })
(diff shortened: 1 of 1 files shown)


florian-h05:
FYI @hmerk @digitaldan @JustinGeorgi

digitaldan:
I think `translation` works well, very clear what it does, an alternative might be `i18n()` , but unless your a dev, a user probably would not understand that without a little googling.

JustinGeorgi:
`translation` works but it is a little long. There's certainly a precedent for shortening longer names in here (props, vars, const, fn) so something like `trns` would also work. `local` and `lang` are also good short options.

If the longer `translation` sticks, though it might be a good idea to add a shortcut similar to @ and # (of course, we are running low on easy character options there).

florian-h05:
What about `translations('home.locations.tab')` and `t('home.locations.tab')`?

PR #3451, opened 2025-11-11
Fix F7Router issues with browser history

florian-h05:
According to https://framework7.io/docs/view#param-browserHistory, either browserHistoryRoot or browserHistorySeparator are required, so that the browser navigation works with F7Router

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/page/page-view.vue
+++ b/bundles/org.openhab.ui/web/src/pages/page/page-view.vue
@@ -151,7 +151,10 @@ export default {
       if (!this.context) return null
       const pageComponent =
         this.pageType === 'tabs'
-          ? this.tabContext(this.context.component.slots.default[this.currentTab]).component
+          ? (this.context.component.slots.default[this.currentTab]
+          // the tabbed page component can be null if this.currentTab is out of bounds
+            ? this.tabContext(this.context.component.slots.default[this.currentTab]).component
+            : null)
           : this.context.component
       if (!pageComponent || !pageComponent.config || !pageComponent.config.style) return null
       return pageComponent.config.style


jsjames:
Just curious, does this address many of the router issues we were having?

florian-h05:
It at least seems to address those related to browser back/forward navigation, I have to test this more.

florian-h05:
Another router issue to check:

- https://community.openhab.org/t/openhab-5-1-snapshot-discussion/165595/31?u=florian-h05

florian-h05:
@Nadahar @digitaldan @jsjames Could you please help testing this PR? I hope it fixes most router issues ... I will also deploy to my prod openHAB.

digitaldan:
Doing so now.......

digitaldan:
First issue, when selecting "Add Equipment to Model" from a Thing (Channel Tab), and then selecting "Expert Mode" the back navigation is set to developer tools instead of returning to "Add items from Thing" screen

image
 
EDIT: if i use the browser navigation to go back, it goes back to the right page, if that helps at all (probably not)

florian-h05:
That's a wrong back link being configured on the page. (If the backlink has a label, it is manually configured). I can remove that and leave back navigation to f7router there.

digitaldan:
I don't think this is related to this PR, but i have not seen it before deploying this, so i'll report it here first, i just got the app to "lock up", ie i can't navigate anything, nothing is clickable (but i can see sse updates).  Unfortunately i don't have a source map to make sense of this exception, i was on the overview page (or navigated to it, hard to remember).

```
index-yKkiFWBW.js:1070 TypeError: Cannot read properties of undefined (reading 'config')
    at Proxy.tabContext (index-yKkiFWBW.js:62416:25)
    at Proxy.pageStyle (index-yKkiFWBW.js:62344:50)
    at nt (index-yKkiFWBW.js:372:19)
    at get value (index-yKkiFWBW.js:930:14)
    at Object.get [as pageStyle] (index-yKkiFWBW.js:1658:23)
    at Proxy.PW (index-yKkiFWBW.js:62487:20)
    at ga (index-yKkiFWBW.js:2498:20)
    at Ke.s3 [as fn] (index-yKkiFWBW.js:2181:35)
    at Ke.run (index-yKkiFWBW.js:295:21)
    at re2 (index-yKkiFWBW.js:2195:77)
```

EDIT: clicking on `Proxy.tabContext (index-yKkiFWBW.js:62416:25)` shows this code if that helps with figuring out where its happening. 

```javascript
 tabContext(e2) {
        let t2 = {
          component: e2.config.page ? Ew().page(e2.config.page.replace("page:", "")) : e2.component,
          tab: e2,
          vars: this.vars,
          props: e2.config.pageConfig,
          store: mw().trackedItems
        };
        return e2.component === "oh-locations-tab" ? t2.slots = {
          locations: [
            e2
          ]
        } : e2.component === "oh-equipment-tab" ? t2.slots = {
          equipment: [
            e2
          ]
        } : e2.component === "oh-properties-tab" && (t2.slots = {
          properties: [
            e2
          ]
        }), t2;
      },
```

EDIT2: Thats `page-view.vue` line 247

jsjames:
> I don't think this is related to this PR, but i have not seen it before deploying this, so i'll report it here first, i just got the app to "lock up", ie i can't navigate anything, nothing is clickable (but i can see sse updates). Unfortunately i don't have a source map to make sense of this exception, i was on the overview page (or navigated to it, hard to remember).
> 
> ```
> index-yKkiFWBW.js:1070 TypeError: Cannot read properties of undefined (reading 'config')
>     at Proxy.tabContext (index-yKkiFWBW.js:62416:25)
>     at Proxy.pageStyle (index-yKkiFWBW.js:62344:50)
>     at nt (index-yKkiFWBW.js:372:19)
>     at get value (index-yKkiFWBW.js:930:14)
>     at Object.get [as pageStyle] (index-yKkiFWBW.js:1658:23)
>     at Proxy.PW (index-yKkiFWBW.js:62487:20)
>     at ga (index-yKkiFWBW.js:2498:20)
>     at Ke.s3 [as fn] (index-yKkiFWBW.js:2181:35)
>     at Ke.run (index-yKkiFWBW.js:295:21)
>     at re2 (index-yKkiFWBW.js:2195:77)
> ```
> 
> EDIT: clicking on `Proxy.tabContext (index-yKkiFWBW.js:62416:25)` shows this code if that helps with figuring out where its happening.
> 
> ```js
>  tabContext(e2) {
>         let t2 = {
>           component: e2.config.page ? Ew().page(e2.config.page.replace("page:", "")) : e2.component,
>           tab: e2,
>           vars: this.vars,
>           props: e2.config.pageConfig,
>           store: mw().trackedItems
>         };
>         return e2.component === "oh-locations-tab" ? t2.slots = {
>           locations: [
>             e2
>           ]
>         } : e2.component === "oh-equipment-tab" ? t2.slots = {
>           equipment: [
>             e2
>           ]
>         } : e2.component === "oh-properties-tab" && (t2.slots = {
>           properties: [
>             e2
>           ]
>         }), t2;
>       },
> ```
> 
> EDIT2: Thats `page-view.vue` line 247

There was an issue with the startup wizard which has similar manifestations (https://github.com/jsjames/openhab-webui/commit/79b7049764ba6967f87c10f56e823c06a669d687). When in this state, can you look at the classes on the main view?

florian-h05:
> First issue, when selecting "Add Equipment to Model" from a Thing (Channel Tab), and then selecting "Expert Mode" the back navigation is set to developer tools instead of returning to "Add items from Thing" screen

Just fixed that.

> I don't think this is related to this PR, but i have not seen it before deploying this, so i'll report it here first, i just got the app to "lock up", ie i can't navigate anything, nothing is clickable (but i can see sse updates). Unfortunately i don't have a source map to make sense of this exception, i was on the overview page (or navigated to it, hard to remember).

Fixed that as well.

digitaldan:
So far so good 👍

Nadahar:
I haven't forgotten this, but at the moment, my OH dev environment is in a "non-buildable state" due to other things I'm working on, which prevents me from testing the UI. I will test it when that situation changes.

Nadahar:
I've tested a little bit now, and I've found one issue (it's not new with Vue3, but still a bit annoying): If I go to e.g "equipment" on the overview page, and then click on one of the Items listed, they will sometimes bring you to a different page, like historic data/analyzer, e.g.: `analyzer/?items=TXNR807_Track_Number&chartType=&coordSystem=&aggregation=`. When clicking "Back" to close the graph, it will sometimes take you back to where you were, but sometimes reload the "Overview" page so that you have to navigate yourself back to where you were if you want to check one of the other Items in there.

edit: It doesn't always happen, I had to try quite a few times for this to reoccur now, but it finally did, and then it took me to `/locations/#card=home` - which rendered as a blank page with no content. Not sure why. If I reload the page, it correctly shows the "Home" box, which it didn't before the reload, probably because it was "expanded", which failed to render for some reason.

edit2: Now it happened again, but this time I was taken to `/overview/` (which did render correctly), so there seems to be a couple of issues going on here, where maybe only one of them is related to the router.

Nadahar:
Going to Thing -> Channel -> Link -> Item and back again seems to work now, which is a huge improvement for me 👍

Nadahar:
Backing out of adding rule modules also seems to work 🥳

Nadahar:
I haven't been able to find any other router issues than the one above at the moment, which doesn't mean that there are none, but things are at least drastically improved. One thing I can mention, that isn't a router issue, but is "related", is that localization of the "Back" button varies a lot. Some places it is translated, while I'd say "most places" it is not.

Nadahar:
I was just taken here: `/settings/items/items/` - so the issue where it loads the "parent" page that mismatches what's in the address bar still exists.

Now it went completely off the rails, after removing the extra `items` from the address bar and reloading, it loaded "Settings", and when I then clicked "Items" it took me to `/addons/items/`....

Nadahar:
Once again I was taken to `/settings/` while the address said "/settings/items/". I reloaded the page, and the same wrong page was loaded. However, when I reloaded it for a second time, it actually loaded `/settings/items/`. Why is this? Is the only a router issue, or some other "thing" going wrong? How can reloading the same address take me to different pages?

edit: This happened again, and I tried reloading the page over and over again, to get to the "correct" page. I had to give up this time, it kept loading the wrong page, but: While it's loading, it kind of looks to me like it's loading the correct page first, before being redirected to the wrong page (without the address being updated) for some reason. That might be the key to this "bug"..?

Nadahar:
Again, I got a case where it actually loaded the list of Items so that I could see them before suddenly "redirecting" to show the "Settings" page. I can't really find anything in the browser console that reveals what happens, but what I've found is that if I turn on throttling (so that loading is slowed down), it doesn't seem to happen.

I managed to "capture" a session where it happened (without throttling), so even though I can't figure out *why* it happens, I at least have the captured screenshots that "proves" that it starts out loading the correct page:

bilde

bilde

bilde

bilde

I hope this can help somewhat, it shows that it doesn't actually "route" to the wrong page, but that for some reason, during loading, it "switches" to the wrong page, while the router still "thinks" that you're on the page that was supposed to load. Everything from there on goes wrong, because address/history is "out of sync" with what actually happens.

Nadahar:
I actually found one more piece of the puzzle. It "Items" page that should have loaded, IS actually loaded, but for some reason it's hidden or "drawn over" by the "Settings" page. Both can be found in the Vue component overview as "views", so in a way, both pages have loaded:

bilde

When I browse the subtree of `ItemsListVList`, it seems like the content matches the page that should have shown, and that was shown very briefly while loading. So, "it's there", it's just not accessible. I don't know if I can "disable" a view from the browser console, but I'm betting that if I could (disable the "Settings view"), I would see the page that should have been there.

Finally, I "proved" it: I managed to find the right element in HTML, added `visibility: hidden` to it, and voilà: The "Items" page appeared one I hid the other one:

bilde

So, I guess now the only question is why this happens... and I have no clue about that, but hopefully somebody else will.

florian-h05:
>  One thing I can mention, that isn't a router issue, but is "related", is that localization of the "Back" button varies a lot. Some places it is translated, while I'd say "most places" it is not.

That depends on whether the translation is used ... I can fix that.

> I hope this can help somewhat, it shows that it doesn't actually "route" to the wrong page, but that for some reason, during loading, it "switches" to the wrong page, while the router still "thinks" that you're on the page that was supposed to load. Everything from there on goes wrong, because address/history is "out of sync" with what actually happens.

Framework7 sometimes sets the wrong classes for the current and last page (usually, those are in the DOM, so back navigation doesn't require rerendering). The previous page needs the `page-previous` class attached and that sometimes fails. Unfortunately I have neither recognised a pattern nor found the cause.

Nadahar:
> Framework7 sometimes sets the wrong classes for the current and last page (usually, those are in the DOM, so back navigation doesn't require rerendering). The previous page needs the `page-previous` class attached and that sometimes fails. Unfortunately I have neither recognised a pattern nor found the cause.

I don't understand enough of how this works to understand how that can result in "two pages existing in one DOM", but that's what happens anyway, and the "wrong one" is shown. The only thing that seems pretty clear from my experimentation yesterday, is that timing has something to do with it, because if I slow things down (using throttling), the error doesn't seem to occur.

florian-h05:
> That depends on whether the translation is used ... I can fix that.

But not now, I don't get it to work right now.

> The only thing that seems pretty clear from my experimentation yesterday, is that timing has something to do with it, because if I slow things down (using throttling), the error doesn't seem to occur.

That's sure, I for example haven't noticed that issue except in dev server.

florian-h05:
Let's merge this as it already is a big improvement and tackle the remaining issues later.

florian-h05:
Please create an issue for the timing issue.

Nadahar:
> That's sure, I for example haven't noticed that issue except in dev server.

My production server runs 4.2.3, so it's not very relevant to test with, so I use the dev server most of the time. I did however build MainUI with Maven using the latest `main` now, and ran without using the dev server - and still it didn't take too long until I was sent to `http://localhost:8080/settings/things/things/`. So, it doesn't seem like it's a "dev server issue".

Nadahar:
> Please create an issue for the timing issue.

#3515

PR #3452, opened 2025-11-11
Implement backwards compatibility for f7-swiper usage in widgets

florian-h05:
This automatically transforms any use of f7-swiper in widgets to oh-swiper, which handles the swiper children properly.
Existing swiper widgets don't need to be changed, even though F7 changed the swiper behaviour very much.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/generic-widget-component.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/generic-widget-component.vue
@@ -1,50 +1,56 @@
 <template>
-  <component v-if="componentType && componentType.startsWith('f7-') && visible"
-             :is="componentType"
-             v-bind="config">
-    <!-- eslint-disable-next-line vue/no-unused-vars -->
-    <template v-for="(slotComponents, slotName) in context.component.slots" :key="slotName" #[slotName]>
-      <ul v-if="componentType === 'f7-list'">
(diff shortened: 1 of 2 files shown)


jsjames:
@florian-h05 - the PR causes issues for me.  I get the following on my overview page:
```
[Vue warn]: Extraneous non-props attributes (class) were passed to component but could not be automatically inherited because component renders fragment or text or teleport root nodes. 
  at <GenericWidgetComponent ref_for=true class="" context= {component: Proxy(Object), rootcomponent: Proxy(Object), props: {…}, fn: undefined, const: undefined, …} > 
  at <MasonryGridItem key=9 > 
  at <Render > 
  at <MasonryGrid key=2 columns= {480: 1, 576: 2, 768: 3, 1023: 4, 1280: 3, 1400: 4, default: 5} > 
  at <OhMasonry class="" context= {component: Proxy(Object), rootcomponent: Proxy(Object), props: {…}, fn: undefined, const: undefined, …} > 
...
```

I traced the issue to removing the checks for "visible" in generic-widget-component - so this is likely needed to remain.

florian-h05:
Setting `inheritAttrs: false` for generic-widget-component might be the better solution.

jsjames:
> Setting `inheritAttrs: false` for generic-widget-component might be the better solution.

Yes - that did fix it - do you want me to submit a PR?

florian-h05:
Please do so :+1:

PR #3462, opened 2025-11-13
Refactor widget-expression-mixin to TS composable

florian-h05:


florian-h05:
@jsjames Please have a look.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
```diff
@@ -0,0 +1,229 @@
...
+    })
+    window.addEventListener('resize', updateViewAreaDimensions)
+  })
+
+  onUnmounted(() => {
+    window.removeEventListener('resize', updateViewAreaDimensions)
+  })
+
+  // computed
+  const screenWidth = computed(() => window.screen.width)
```
jsjames: I don't believe any that are dependent on window.screen need to be computed/reactive since window.screen itself is not reactive, correct?

bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  // data
+  const exprAst: ExpressionAstCache = {}
+  const viewAreaWidth = ref<number>(0)
+  const viewAreaHeight = ref<number>(0)
+
+  const instance = getCurrentInstance()
+  const global = instance?.appContext.config.globalProperties
+
+  // lifecycle
```
jsjames: Do we need to also check when a new page is brought into focus as different pages will have different view areas depending on navbar/etc.?

I haven't looked at the difference, but do you think we should be using a ResizeObserver vs. the resize event listeners?  Not sure if this is more "modern" or not?

jsjames:
This looks good and similar (but simpler) than what i was working on.  Just some minor questions/clarifications in my comments.

I think the next is getting the widget-mixin converted, then components could actually be implemented in the composition api. Are you going to tackle that one next? (I could too, but don't want to duplicate work).

florian-h05:
> Are you going to tackle that one next? (I could too, but don't want to duplicate work).

If you want to do that I will leave it to you. I have a few work and university related projects that require my attention as well 😉

When tackling the widget-mixin, that would also be a good opportunity to fix https://github.com/openhab/openhab-webui/discussions/3381#discussioncomment-14961096. I have looked into it a bit and it seems that `this.$el` behaves differently in Vue 3, with the best solution adding refs to the main containers of each widget to add the class to it.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
florian-h05: Correct, will change that.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
florian-h05: This composable is used by widget-mixin, which is used by each individual rendered widget, and as a image can be only rendered on a single page we don't need that.
But I have refactored to use ResizeObserver as this should be slightly more performant.

florian-h05:
@jsjames Let me know when you had another look and I can merge.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
jsjames: agree - i had to think through that one more!

jsjames:
> d I can merge.

I think there is still something not working correctly on initial loading of the page (1st load after navigating to the page). I added a debug log message in updateViewAreaDimensions and the majority of calls return 0, 0.

image

The issue is that page-current is not yet set on the active page before the mounted event is called. Instead of observing a resize on .page-current, we could observe .view-main - but this runs into the same issue that the .page-current is not initially found on the call to updateViewAreaDimensions.

(BTW, the same issue exists with the event listener method)

When I was looking into this before, I thought we'd have to watch for the onPageAfterIn event at the app level and notify that the initial size is available, but there might be a better way to do this.

florian-h05:
Not sure what was the case before the refactoring, but all my widgets that access the screen prop work fine.

jsjames:
> Not sure what was the case before the refactoring, but all my widgets that access the screen prop work fine.

Before the refactor, it would force a recalc on the nextTick if the querySelector('.page-current > .page-content') did not exist (that was the purpose of the recalculateScreenInfo reactive property).

I presume you use the viewArea* in your widgets?  If I just refresh my browser, it logs a good values, however starting from a fresh browser and navigating to the page, it logs 0, 0 for many of the calls.

florian-h05:
> Before the refactor, it would force a recalc on the nextTick if the querySelector('.page-current > .page-content') did not exist (that was the purpose of the recalculateScreenInfo reactive property).

Yes, I added that because we sometimes ended up with 0 values. Any ideas how to fix the 0 issue without that „ugly“ workaround?

jsjames:
> > Before the refactor, it would force a recalc on the nextTick if the querySelector('.page-current > .page-content') did not exist (that was the purpose of the recalculateScreenInfo reactive property).
> 
> Yes, I added that because we sometimes ended up with 0 values. Any ideas how to fix the 0 issue without that „ugly“ workaround?

Only thing i was thinking of was to watch for an onPageAfterIn at the top app level and triggering an update.

jsjames:
> > > Before the refactor, it would force a recalc on the nextTick if the querySelector('.page-current > .page-content') did not exist (that was the purpose of the recalculateScreenInfo reactive property).
> > 
> > 
> > Yes, I added that because we sometimes ended up with 0 values. Any ideas how to fix the 0 issue without that „ugly“ workaround?
> 
> Only thing i was thinking of was to watch for an onPageAfterIn at the top app level and triggering an update.

Was thinking about this more - here is my though ... write a simple composable that we include in each of the top level pages which
-  has the reactive variables for viewArea.
- handles the resize events when the page changes size
- uses the provide/inject mechanism to share with its children

Has the benefit of only a single reactive variable for each page vs. one for each widget.

florian-h05:
Or we check the calculation and maybe we can do it based in some data that is already available.

florian-h05:
> write a simple composable that we include in each of the top level pages which

I think we actually only need to include this composable in the page-view component and the page edit components.

florian-h05:
@jsjames I have now implemented a useViewArea composable to be used at the page level and inject the reactive ref for the view area dimensions into the useWidgetExpression composable.
The issue with 0, 0 values seems to be gone now.

jsjames:
> @jsjames I have now implemented a useViewArea composable to be used at the page level and inject the reactive ref for the view area dimensions into the useWidgetExpression composable. The issue with 0, 0 values seems to be gone now.

this looks good to me and I think its the cleanest way to do to this.  BTW, is it possible to put the "provide" in the composable - or does that need to be done in the setup for each module?

florian-h05:
>  BTW, is it possible to put the "provide" in the composable - or does that need to be done in the setup for each module?

This can be done in the composable, just commited it :+1:

PR #3464, opened 2025-11-14
Fix stylesheets not being applied from widget definitions

jsjames:
Added ref for the generic-widget-components so that a local stylesheet can be rendered correctly.

Fixes https://github.com/openhab/openhab-webui/discussions/3381#discussioncomment-14961096.


florian-h05 (approved):
LGTM, thanks!

PR #3465, opened 2025-11-14
Semantics store: Performance optimizations for setSemantics()

florian-h05:
The `setSemantics()` method was really slow, sometimes taking up to 0.8 seconds to finish according to the Chromium profiler on my machine (AMD Ryzen 9 6900HS, Fedora 42).

There were two main issues:

1. A large amount of writes to reactive refs, causing a lot of reactivity to run.
2. Use of JavaScript array convenience methods, which are usually slower than standard loops.

The issues have been fixed by:

1. Using local variables to write the data to and only assigning once to the reactive refs at the end of the method.
2. Putting everything into a single loop instead of using several convenience method calls, reducing from several iterations over the tags to a single iteration only.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useSemanticsStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useSemanticsStore.ts
@@ -24,28 +24,58 @@ export const useSemanticsStore = defineStore('semantics', () => {
 
   // Actions
   function setSemantics (tags: ModelTag[], i18n: I18n) {
-    Tags.value = tags
-    Tags.value.forEach((tag) => {
-      const tagParts = tag.uid.split('_')
-      tag.parent = tagParts.slice(0, -1).join('_')
-    })
-    Locations.value = tags.filter((t) => t.uid.startsWith('Location')).map((t) => t.name)
-    Equipment.value = tags.filter((t) => t.uid.startsWith('Equipment')).map((t) => t.name)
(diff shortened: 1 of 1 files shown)


florian-h05:
Tested against the semantic model of the demo server at demo.openhab.org.

Nadahar:
I can't really evaluate it "thoroughly enough" to submit a review, but this LGTM 👍 

I think the "reactivity fix" can make a big difference. Are there other "storage functions" that have been rewritten in a similar way? Maybe they could use the same "fix"?

florian-h05:
Skimming through the source code of the other stores, they do less to no processing on their data, and no one does process data and write to reactive refs at such an excessive level as the semantics store did.
So I haven't found other problematic code yet. 

If you find any other slowness and are sure its related to the code (and not the Vite dev server ;-)) please let me know.

Nadahar:
I just checked this out and did a new profiling, and this storage no longer "dominate", and the total loading time seem to have been reduced by almost 3 seconds in the "same test on the same computer".

florian-h05:
I think we are fine to merge, I have double checked and thought about the changes again, the code does the same, just in a different way. The model on the demo server looks the same (having used the demo server as backend for the devserver=).

florian-h05:
> I just checked this out and did a new profiling, and this storage no longer "dominate", and the total loading time seem to have been reduced by almost 3 seconds in the "same test on the same computer".

My profiling lead to the same result :+1:

PR #3468, opened 2025-11-14
Send source with item commands

ccutrer:
While this PR isn't strictly dependent on https://github.com/openhab/openhab-core/pull/5131, it won't do anything useful without it.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useStatesStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useStatesStore.ts
@@ -122,7 +122,7 @@ export const useStatesStore = defineStore('states', () => {
       command,
       'text/plain',
       'text/plain',
-      null
+      { 'X-OpenHAB-Source': 'org.openhab.ui'}
     )
   }
 


florian-h05:
I’d rather send the command to the JSON endpoint where the source can be transmitted in JSON instead of sending it in a header.
 WDYT?

ccutrer:
The mobile developers specifically requested that the web UIs put it in a header because that's easiest for them to intercept from their web views and inject their own package ids.

florian-h05:
Hmm okay. I personally don’t care wrt to Main UI.

bundles/org.openhab.ui/web/src/js/stores/useStatesStore.ts
```diff
@@ -122,7 +122,7 @@ export const useStatesStore = defineStore('states', () => {
       command,
       'text/plain',
       'text/plain',
-      null
+      { 'X-OpenHAB-Source': 'org.openhab.ui'}
```
florian-h05, line 125: If a user is logged in, it would be nice to attach the username.

ccutrer:
The username is attached by org.openhab.core.io.rest if you're logged in. When it adds itself as a "delegated" source. The same is done by org.openhab.io.openhabcloud when it flows that way.

florian-h05 (approved):

PR #3471, opened 2025-11-15
Config params: Accept either "step" or "stepsize"

Nadahar:
As discussed in #3457, I propose to make MainUI "agnostic" in regard to configuration description/parameter property `stepsize` and `step`.  The reason is that this is the only property that doesn't match the name of those used in [XML configuration files](https://www.openhab.org/docs/developer/addons/config-xml.html#xml-structure-for-configuration-descriptions), causing unnecessary confusion.

The idea is to also submit a Core PR if this is merged, to make Core agnostic too.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/parameter-datetime.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/parameter-datetime.vue
@@ -35,8 +35,7 @@ export default {
   },
   computed: {
     step () {
-      if (this.configDescription.stepsize !== undefined) return this.configDescription.stepsize
-      return 60
+      return this.configDescription.step || this.configDescription.stepsize || 60
     }
   },
   methods: {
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.ui/web/src/components/config/controls/parameter-datetime.vue
```diff
@@ -35,8 +35,7 @@ export default {
   },
   computed: {
     step () {
-      if (this.configDescription.stepsize !== undefined) return this.configDescription.stepsize
-      return 60
+      return this.configDescription.step || this.configDescription.stepsize || 60
```
florian-h05, line 38: Keep in mind || applies the „fallback“ if the left side is falsy, which includes 0. 
I don’t think this is a problem here though, because a step size of 0 doesn’t make sense.

Nadahar: Yeah, that one can easily bite you. I've used falsy throughout here though, my idea was that `0` or `undefined` must mean pretty much the same thing for steps...

Issue #3330, opened 2025-08-12
Non-semantic Groups in a semantically tagged Group not shown in Model viewer


rkoshak:
## The problem

If one sets a non-semantically tagged Group as a member of a Group with a semantic tag, that Group does not get shown at all in the model view.

See https://community.openhab.org/t/5-0-0-non-sematic-items-inside-sematic-groups-not-shown/165351

## Expected behavior

If "Show non-semantically tagged" is checked, the Group should appear in the model instead of being hidden. Ideally it should be shown under the semantically tagged Group it's a member of.

## Steps to reproduce

1. Create a semantically tagged Group
2. Create a non-semantically tagged Group or regular Item
3. Open the model and check "show non semantically tagged Items" and verify that the Item/Group from 2 is nowhere to be found in the model view.

## Your environment

```yaml
runtimeInfo:
  version: 5.0.0
  buildString: Release Build
locale: en-US
systemInfo:
  configFolder: /openhab/conf
  userdataFolder: /openhab/userdata
  logFolder: /openhab/userdata/logs
  javaVersion: 21.0.8
  javaVendor: Eclipse Adoptium
  javaVendorVersion: Temurin-21.0.8+9
  osName: Linux
  osVersion: 6.8.0-71-generic
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 166104832
  totalMemory: 2082471936
  uptime: 331324
  startLevel: 100
addons:
  - automation-jsscripting
  - automation-pidcontroller
  - binding-amazonechocontrol
  - binding-astro
  - binding-chromecast
  - binding-ephemeris
  - binding-exec
  - binding-http
  - binding-ipcamera
  - binding-mail
  - binding-matter
  - binding-mqtt
  - binding-nest
  - binding-network
  - binding-networkupstools
  - binding-openweathermap
  - binding-remoteopenhab
  - binding-shelly
  - binding-vesync
  - binding-zigbee
  - binding-zwave
  - misc-openhabcloud
  - persistence-mapdb
  - persistence-rrd4j
  - transformation-basicprofiles
  - transformation-jinja
  - transformation-jsonpath
  - transformation-map
  - transformation-regex
  - ui-basic
  - ui-habot
  - voice-pipertts
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: true
    macos: false
    windows: false
[...]

PR #3476, opened 2025-11-16
Model: Show non semantic children in model tree

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/3330

Non semantic items and groups will be made visible as children of semantic groups in the model tree when shown non-semantic is switched on. Semantic children will not be shown to avoid having semantic items visible multiple times in the model tree.

Some improvements where made to drag and drop:
- solved issue with tree not being updated after moving (not sure this resolves all issues, but it does resolve what I could reproduce)
- drag & drop failed when something was selected at the start of dragging: fixed by deselecting at start of drag and fix highlighting of selected node, make selection work as toggle on the model page
- scroll dragged node into view after dragging


florian-h05 (commented):
I found one JS error, breaking the ability to move non-semantic into the model, which I fixed.

Other then this LGTM, thanks!

PR #3482, opened 2025-11-18
Add script to generate TS definitions for oh- components

jsjames:
script to generate typescript definitions for oh- components

- built using the asset/definitions (same as docs)
- fixed some minor issues with current definition

Innocuous change which won't impact the current code in any way.

### DIFF
--- a/bundles/org.openhab.ui/web/build/generate-widget-component-ts.mjs
+++ b/bundles/org.openhab.ui/web/build/generate-widget-component-ts.mjs
@@ -144,7 +144,7 @@ function createCommonOptionsMap() {
 }
 
 function generateCommonTS(mapCommonOptions) {
-  let content = ''
+  let content = '// note: this file is generated and should not be edited by hand\n\n'
   const configCommon = config['_Common'] || {}
   mapCommonOptions.forEach((components, optionProp) => {
     if (components.length > 1) {
@@ -233,7 +233,7 @@ function generateComponentTS(mapCommonOptions) {
 
       content += configStr
 
-      let preamble = ''
+      let preamble = '// note: this file is generated and should not be edited by hand\n\n'
       let postamble = ''
       if (commonComponents.length > 0) {
         preamble += 'import {\n'
--- a/bundles/org.openhab.ui/web/src/types/components/widgets/chart/oh-aggregate-series.ts
+++ b/bundles/org.openhab.ui/web/src/types/components/widgets/chart/oh-aggregate-series.ts
@@ -1,3 +1,5 @@
+// note: this file is generated and should not be edited by hand
+
 import {
   OffsetUnit,
   Type,
--- a/bundles/org.openhab.ui/web/src/types/components/widgets/chart/oh-calendar-axis.ts
+++ b/bundles/org.openhab.ui/web/src/types/components/widgets/chart/oh-calendar-axis.ts
@@ -1,3 +1,5 @@
+// note: this file is generated and should not be edited by hand
+
 import {
   Orient,
   Action,
--- a/bundles/org.openhab.ui/web/src/types/components/widgets/chart/oh-calendar-series.ts
+++ b/bundles/org.openhab.ui/web/src/types/components/widgets/chart/oh-calendar-series.ts
@@ -1,3 +1,5 @@
+// note: this file is generated and should not be edited by hand
+
 import {
   OffsetUnit,
   AggregationFunction,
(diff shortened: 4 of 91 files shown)


jsjames:
@florian-h05, In adding this, I did find a few oddities in the assets/definition (i corrected a couple of them), these I didn't correct:

- oh-card-input has 2 different outline parameters, one as part of CardParameters, one in InputParameters - this renders 2 different listings in the docs.
- actionAnalyzerItems is listed as required - I don't think that is the case (for instance oh-time-series) - should this be changed to optional?

florian-h05:
> oh-card-input has 2 different outline parameters, one as part of CardParameters, one in InputParameters - this renders 2 different listings in the docs.

These actually are two different outlines, but I don't think both can be configured due to the same name??

> actionAnalyzerItems is listed as required - I don't think that is the case (for instance oh-time-series) - should this be changed to optional?

`actionAnalyzerItems` is for widget action `analyzer`, so I think it's correct then that this is required if it's visible.

florian-h05:
Some general questions: 

- Would it make sense to replicate the components directory structure in the types directory?
- Shouldn't type definitions use `.d.ts` as file extension?

jsjames:
> > oh-card-input has 2 different outline parameters, one as part of CardParameters, one in InputParameters - this renders 2 different listings in the docs.
> 
> These actually are two different outlines, but I don't think both can be configured due to the same name??
 
I just tested and if outline: true, then the outline appears for the card and the input element - so there is no way to distinguish between the two for an oh-input-card.

> > actionAnalyzerItems is listed as required - I don't think that is the case (for instance oh-time-series) - should this be changed to optional?
> 
> `actionAnalyzerItems` is for widget action `analyzer`, so I think it's correct then that this is required if it's visible.

jsjames:
> Some general questions:
> 
> * Would it make sense to replicate the components directory structure in the types directory?

This would make the generation script more complicated as it currently only uses the info from the assets/definitions which does not have any directory.  While I agree it would make sense to match the directory logically, i don't know if its worth the complication. thoughts?

> * Shouldn't type definitions use `.d.ts` as file extension?

Yes, I originally had these as .d.ts file, but since I am using typescript enums to build out the options (more on this below), these actually generate runtime objects - and as such the typescript transpiler will give an error when importing non-types from .d.ts files.

Regarding the use of enums, there are several fairly opinionated camps out there.  Some don't like them at all because they aren't strictly types and their "purist" argument is that typescript shouldn't modify/add to the javascript code. They also marginally increase the code size because of the created enum objects.

In this application though, I think the cleanliness of using enums makes a lot of sense as they can be iterated over in the UI. I use this quite a bit in the analyzer in generating the different options for the UI. Obviously if an enum is not use, it is tree-shaken, so the only impact in code-size is for the enum objects created for the ones used.

florian-h05:
> While I agree it would make sense to match the directory logically, i don't know if its worth the complication. thoughts?

I don’t think it’s worth the complication. Instead of this, at least the types should be in a subfolder making it clear those are types for widgets. Probably use the file name/path of the JS definition. Some type of hierarchy order would be helpful.

I agree on using enums, didn’t know this involves the requirement to use .ts instead of .d.ts. Thanks for explaining.

florian-h05:
> I just tested and if outline: true, then the outline appears for the card and the input element - so there is no way to distinguish between the two for an oh-input-card.

That’s rather unfortunate, but we don’t need to fix this here.

jsjames:
> > While I agree it would make sense to match the directory logically, i don't know if its worth the complication. thoughts?
> 
> I don’t think it’s worth the complication. Instead of this, at least the types should be in a subfolder making it clear those are types for widgets. Probably use the file name/path of the JS definition. Some type of hierarchy order would be helpful.
> 
> I agree on using enums, didn’t know this involves the requirement to use .ts instead of .d.ts. Thanks for explaining.

I implemented a directory structure now in the generate-component-ts in the latest update.

jsjames:
changes make sense.

@florian-h05 - I was thinking about something - is it going to be problematic since the name types are equal to the name of the components? For instance, if you are importing both the type definition AND the component, they will both be the same name (unless you import as a different name, but that could be cumbersome).  thoughts?

florian-h05:
Basically we would import now like the following:

```ts
import type { OhSipclient } from '@/types/components/widgets'

const config: OhSipclient.Config = ...
```

That of cource could clash with importing the component, but I wonder how often we need to import both (internal) types and widget components itself.

jsjames:
> Basically we would import now like the following:
> 
> ```ts
> import type { OhSipclient } from '@/types/components/widgets'
> 
> const config: OhSipclient.Config = ...
> ```
> 
> That of cource could clash with importing the component, but I wonder how often we need to import both (internal) types and widget components itself.

Not sure - let's keep as is, if it becomes problematic in the future we can adjust then.

PR #3487, opened 2025-11-18
Item picker: Implement filtering by type, groupType, tags

mherwege:
Related to https://github.com/openhab/openhab-webui/issues/3457#issuecomment-3544910500

This implements the possibility to configure the item picker to filter on item type, item group type and item tags. This allows creating widgets with item pickers to pick from a list of items filtered on type and tag.

A provided list of item types will match any of the types in the list. A provided list of item tags will match any tag in the list. Filtering on both types and tags will only yield items that have one of the types and one of the tags.

Filtering on item group type (groupType parameter name) will always force filtering on group.

When filtering Number items, checks are made for compatible units:
- if the filter is on Number without a dimension, any Number item regardless of the dimension will be returned
- if the filter is on Number with a dimension, only Number items with the same dimension or Number items without dimension will be returned

I added the ability to insert a filter toggle in the item-picker popup (left of the search bar), that allows you to show all items in the list. In normal cases, the relevant items should be filtered. But there have been instances (e.g. add item to channel when using profiles), where the logically matching item types would not apply.
The button is optional, but enabled already for linking items to channels.
It can be used through the props configuration by adding an extra filterCriteria (filterToggle), e.g.:
```
uid: parameterPlayground
tags: []
props:
  parameters:
    - context: item
      description: Only allow selection of Switch and Contact Items
      label: Switch or Contact Item
      name: switchContact
      required: false
      type: TEXT
      groupName: items
      filterCriteria:
        - value: Switch,Contact
          name: type
        - value: "true"
          name: filterToggle
```
This has been put in the filterCriteria as this will not require changes on the core side to be able to save this property configuration. Putting it one level up would require defining the parameter in core, and this is only relevant for the UI.

Filters for selecting a group when adding an item to a group, or selecting an item to add to a group have been refined to assure compatible group and item types. The picker popup for these case has the toggle button to show all (although I fail to see when this could be useful).

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -153,7 +153,7 @@
                      :items="items"
                      @input="(value) => this.item.groupNames = value"
                      :multiple="true"
-                     :filterGroupType="getCompatibleGroupTypes()"
+                     :filterGroupType="compatibleGroupTypes"
                      :filterToggle="true"
                      :set-value-text="false" />
       </f7-list-group>
@@ -273,6 +273,17 @@ export default {
       set (newPattern) {
         this.item.stateDescriptionPattern = newPattern
       }
+    },
+    compatibleGroupTypes () {
+      if (!this.itemType) return
+      let compatibleGroupTypes = []
+      compatibleGroupTypes.push(this.item.type)
+      if (this.itemType === 'Number') {
+        compatibleGroupTypes.push('Switch')
+      }
+      if (this.itemType === 'Color') { compatibleGroupTypes.push('Switch', 'Dimmer') }
+      if (this.itemType === 'Dimmer') { compatibleGroupTypes.push('Switch') }
+      return compatibleGroupTypes
     }
   },
   methods: {
@@ -370,17 +381,6 @@ export default {
         inputElement.dispatchEvent(new Event('input'))
       }
       this.item.label = event.target.value
-    },
-    getCompatibleGroupTypes () {
-      if (!this.itemType) return
-      let compatibleGroupTypes = []
-      compatibleGroupTypes.push(this.item.type)
-      if (this.itemType === 'Number') {
-        compatibleGroupTypes.push('Switch')
-      }
-      if (this.itemType === 'Color') { compatibleGroupTypes.push('Switch', 'Dimmer') }
-      if (this.itemType === 'Dimmer') { compatibleGroupTypes.push('Switch') }
-      return compatibleGroupTypes
     }
   },
   mounted () {
--- a/bundles/org.openhab.ui/web/src/components/item/group-members.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/group-members.vue
@@ -17,7 +17,7 @@
                        :value="pickedMemberNames"
                        label="Members"
                        :editableOnly="true"
-                       :filterType="getCompatibleItemTypes()"
+                       :filterType="compatibleItemTypes"
                        :filterToggle="true"
                        @input="(members) => pickedMemberNames = members" />
         </f7-list-group>
@@ -69,6 +69,19 @@ export default {
     },
     sortedGroupMembers () {
       return this.groupItem.members.toSorted((a, b) => (a.label || a.name).localeCompare(b.label || b.name))
+    },
+    compatibleItemTypes () {
+      const groupType = this.groupItem.groupType
+      if (groupType) {
+        let compatibleItemTypes = []
+        compatibleItemTypes.push(groupType)
+        if (groupType.startsWith('Number')) { compatibleItemTypes.push('Switch') }
+        if (groupType === 'Color') { compatibleItemTypes.push('Switch', 'Dimmer') }
+        if (groupType === 'Dimmer') { compatibleItemTypes.push('Switch') }
+        return compatibleItemTypes
+      } else {
+        return null
+      }
     }
   },
   methods: {
@@ -118,19 +131,6 @@ export default {
             })
         }
       )
-    },
-    getCompatibleItemTypes () {
-      const groupType = this.groupItem.groupType
-      if (groupType) {
-        let compatibleItemTypes = []
-        compatibleItemTypes.push(groupType)
-        if (groupType.startsWith('Number')) { compatibleItemTypes.push('Switch') }
-        if (groupType === 'Color') { compatibleItemTypes.push('Switch', 'Dimmer') }
-        if (groupType === 'Dimmer') { compatibleItemTypes.push('Switch') }
-        return compatibleItemTypes
-      } else {
-        return null
-      }
     }
   }
 }


bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
```diff
@@ -115,7 +116,11 @@ export default {
     this.smartSelectParams.closeOnSelect = !(this.multiple)
     if (this.setValueText) this.smartSelectParams.setValueText = this.setValueText
     if (!this.items || !this.items.length) {
-      this.$oh.api.get('/rest/items?staticDataOnly=true').then((items) => {
+      let url = '/rest/items?staticDataOnly=true'
+      if (this.filterTag) {
+        url = url + '&tags=' + this.filterTag.join('%2C')
```
Nadahar: Doesn't JS have some proper URL/URI class/object that can build URLs and handle URL encoding etc.?

mherwege: Yes, there is. I didn't see much use in the code base, but I changed it.

rkoshak:
I love the inclusion of dimension for number Items!

Nadahar:
I don't know the meaning of `:` in these strings, so I can't quite follow the logic.

But, I have an idea: What if there were some "button" in the picker that could make it ignore the filter? That would be a way to get things done when the filter would otherwise prevent it. Often, you can have situations where a filter is helpful most of the time, but prevents you from doing what you want in more specialized situations. Instead of trying to handle all these "special situations", maybe being able to just disable the filter at will could "solve" this problem?

mherwege:
> But, I have an idea: What if there were some "button" in the picker that could make it ignore the filter? That would be a way to get things done when the filter would otherwise prevent it.

Exactly what I try to achieve. Except, it is much harder than anticipated. The f7 list-item component with smartSelect does not allow you to insert a button in the popup. All my attempts have failed so far. I will probably have to code a custom popup for this to work. And then I need to make sure it keeps on working in all places the item picker is used.
I also want this button to be optionally visible only with a parameter, so I can avoid exposing it where not needed or wanted.
Some of the places the item picker is used already do a pre-filtering on the full list before the picker is even called. The filter toggle would then only be able to return to that pre-filtered list and still not the full list. But that may be OK.

Nadahar:
> Exactly what I try to achieve.

Seems like a good goal, but I don't envy you trying to fight F7 and "smartSelect"... I can only hope that you can achieve it. I don't know if there is some alternative way to achieve the same that doesn't require adding a button - but it's hard to imagine exactly what that could be.

mherwege:
I now added an (optional) button to revert the default filtering and get the full list back.

@Nadahar if you feel like testing it, I would appreciate it. Try the item picker in as many places as possible, as this is used all over the place, always with slightly different properties.

Nadahar:
> Try the item picker in as many places as possible, as this is used all over the place, always with slightly different properties.

I have very limited content in my "dev installation", I usually have to create/construct whatever I want to test. What are the criteria for the button to show? I've tried it on a widget this far, and it looks OK except that the "window" doesn't limit it's size if there are few Items - but that's maybe how it used to work?

edit: It's probably always been "full of nothing" if there are fewer Items, it's just that when the filtering works, you actually see it 👍 

image

Nadahar:
I think I found a place where the filtering works, when trying to link a channel. When I finally "saw" the button, it seems to work nicely. Maybe a mouseover would help explain what it's doing, and when it's "ON" or "OFF":

image

Are there any other places where the filter button will be visible?

mherwege:
At the moment, the only place where I have added it is indeed when trying to link a channel, as that was the pressing issue for @rkoshak. I only want to add it where it does make sense to avoid people expecting to be able to do that without knowing what they do.
However, you can add it as properties of a widget or rule template. I tested it with @rkoshak examples from the linked issue and added configuration parameters as shown in my first post on this PR. So it is fully accessible for templates.

> Maybe a mouseover would help explain what it's doing, and when it's "ON" or "OFF"

Yes, agreed I need to do something about that. I don't know what though. A tooltip will work and is the easiest to implement, but it only works on desktops and not on touch devices. I could make it a real button with some text on it, rather then an icon, but that increases the space required. So I tried to keep it as lean as possible.

mherwege:
I have added some tooltips for this (and the related pick from model icon). These will show when hovering on desktop or on long tap for touch devices.

mherwege:
> Are there any other places where the filter button will be visible?

The only other places in code where I see a filterType applied is in the persistence group configurations, and for picking groups an item belongs to.  In these cases it just filters on groups. No none group items are allowed, so it would not make sense to expose the toggle there. I believe it is now exposed in the only place it could make sense, and as an option to make visible in widget or rule template props.

Nadahar:
> I have added some tooltips

That helped a lot, but I think it would be sufficient with just "Item filter applied" or "No item filter applied"..?

Nadahar:
This is probably more taste than anything, but I think it's clearer that it's a button if I apply `raised` to it:

image

florian-h05:
> That helped a lot, but I think it would be sufficient with just "Item filter applied" or "No item filter applied"..?

I agree.

> This is probably more taste than anything, but I think it's clearer that it's a button if I apply raised to it:

Most buttons aren't raised in the UI, at least not those besides searchbars, see for example the expand model button.

Nadahar:
It was merged in the middle of my "evaluation", but the last thing I would have tweaked is to add `margin-inline-end: 10px` to the button (hopefully, it would work also for RTL when using inline end):

image

florian-h05:
It isn't merged yet ;-)

I wanted to but then saw your comments, so its only approved for now.

florian-h05:
> (hopefully, it would work also for RTL when using inline end):

Do we support RTL? I don't think so, the RTL CSS is not imported.

Nadahar:
> It isn't merged yet ;-)

I'm so used to approval and merging being "two clicks in a row" that I automatically expected one to follow the other 😄

Nadahar:
> Do we support RTL? I don't think so, the RTL CSS is not imported.

I don't know about that, I just always try to think of such things - it doesn't hurt if it works if RTL is applied in the future though.

mherwege:
> Most buttons aren't raised in the UI, at least not those besides searchbars, see for example the expand model button.

Should I change that styling (raised and margin-inline-length), or keep it as is, inline with other uses? I don't think the margin is needed if the button is not raised.

Nadahar:
> I don't think the margin is needed if the button is not raised.

I disagree - when you mouse over, the button gets colored, and then it looks "a bit bugged" when there's no margin there. The margin could probably be smaller if the button isn't raised.

Nadahar:
> Most buttons aren't raised in the UI, at least not those besides searchbars, see for example the expand model button.

Yeah, there's no secret that I don't like "the design" that is applied generally, so this might just be personal preference.

mherwege:
> picking groups an item belongs to

Thinking about it, this filter is currently not applied, but would be possible now: when adding an item to a group, you would only want to add it to a group without groupType or with a compatible groupType. It should fail on others. I think I can still add this in here.

mherwege:
I now added a little margin after the button (but didn't raise it).

There is also better filtering for adding items to groups or selecting groups for an item, making sure the groupType and item type are compatible. The filter toggle is available in these case. I don't actually think it is very useful to see all, so depending on feedback, I may remove it.

This is ready for review again.

bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -369,6 +370,17 @@ export default {
         inputElement.dispatchEvent(new Event('input'))
       }
       this.item.label = event.target.value
+    },
+    getCompatibleGroupTypes () {
+      if (!this.itemType) return
+      let compatibleGroupTypes = []
+      compatibleGroupTypes.push(this.item.type)
+      if (this.itemType === 'Number') {
```
florian-h05: Why do you use equals `Number` here and startsWith in the group-member?
I would expect to also allow Switch for UoM Items.

florian-h05: FYI I changed this to a computed value (better practise, method gets reevaluated on every rerender, computed value is cached).

mherwege: > Why do you use equals `Number` here and startsWith in the group-member? I would expect to also allow Switch for UoM Items.

itemType already has the dimension removed in this class.

PR #3350, opened 2025-09-04
Initial port to Vue3, vite, Framework7 v7.1.5, pinia stores, codemirror 6, vue-i18n v11, ...

jsjames:
Closes #2151.

Fully functioning release.
There are some things left to do, see #3378.

Please report issues in this GitHub discussion: https://github.com/openhab/openhab-webui/discussions/3381
We will then decided how critical the reported bug issue and depending on that, fix it right now or put it on the to do lost: https://github.com/openhab/openhab-webui/issues/3378

PR #3488, opened 2025-11-19
Fix global dayjs locale setup

jsjames:
Regression from #3350.

Addresses setting locale on initial loading of the app.

### DIFF
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -522,6 +522,9 @@ import { request } from 'framework7'
 import { f7, f7ready } from 'framework7-vue'
 import { mapStores } from 'pinia'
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/App.vue
```diff
@@ -826,7 +826,8 @@ export default {
             })
           this.updateTitle()
 
-          if (data[2]) dayjs.locale(data[2].key)
+          if (data[2]) dayjs.locale(data[2], null, false)
+          console.log('dayjs locale set to', dayjs.locale())
```
florian-h05: Please put the logging into the if clause.

florian-h05:
Apparently this is caused by the en locale not containing several keys: https://github.com/iamkun/dayjs/blob/dev/src/locale/en.js.

I have a fix for that.

Issue #3272, opened 2025-07-14
Item name generation does not take group into account.


lsiepel:
When creating a Chanel link with a new Item, the generated item name does not take the channel group into account. Leading to possible duplicates when the same channel name exists between different groups.

Relates to : https://github.com/openhab/openhab-addons/pull/18630#discussion_r2205684202

oh5m4

pacive:
Got hit by this as well when testing some updates to SMHI Binding. It has several channel groups with the same channels, corresponding to different time offsets (now, in 1 hour, in 2 hours, next day etc.). When using "Create equipment from Thing", this leads to the same Item getting connected to all channels with the same name. E.g., only one "Temperature" Item gets created, which gets connected to all selected temperature channels.

PR #3495, opened 2025-11-20
Link channel -> Item: Include channel group label in new item name

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/3272

When linking a new item to a channel, the proposed name does not consider the channel group. It therefore can lead to double item names.
With this code, if the channel is in a channel group, the channel group label is being included in the proposed name.

@hmerk You may want to have a look, as this changes the automatically generated item names and can have an impact on your widgets.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/channel-link.vue
+++ b/bundles/org.openhab.ui/web/src/components/thing/channel-link.vue
@@ -111,6 +111,7 @@ export default {
     channelId: String,
     channel: Object,
     thing: Object,
+    thingType: Object,
     opened: Boolean,
     extensible: Boolean,
     context: Object,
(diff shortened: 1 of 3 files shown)


florian-h05:
@hmerk Please let me know what you think.

hmerk:
Indeed, this may have impact on widgets expecting fixed Items names. I have not really come across many equipment using grouped channels, especially for lightning or heating, therefore I cannot really judge the impact of this PR.
But I would weigh the benefit from this PR more, as it prevents having double Item names if using e.g. create Equipment from Thing.

BTW, I will soon start with the next evolution of my Widgets. Perhaps I can find a way to move away from fixed Item names…

Issue #3480, opened 2025-11-17
[MainUI] “Show Advanced” setting does not show anytihng


dilyanpalauzov:
Under Settings → System Settings → “Add-on Management” I click on “Show Advanced”, disable UPnP, mDNS, IP-based, SDDP and USB Suggestion Finders (all of them) → Save.

I go then again over Settings → System Settings → “Add-on Management” and see that initially “Show advanced” is disabled, and all Suggestion Finders are off:

<img width="754" height="720" alt="Image" src="https://github.com/user-attachments/assets/a97bdb18-4485-4b1e-9b11-ec23ed17c3d4" />

I click on “Show advanced” - nothing happens - no more options are displayed:

<img width="754" height="720" alt="Image" src="https://github.com/user-attachments/assets/52a08ed3-38ef-4ea1-a1a8-b3f095a2a1f5" />

* When “Show advanced” wouldn’t do anything, the button should not be rendered.

dilyanpalauzov:
To be precise for the above behavior, when “Show advanced” is not selected, and I click on the “USB Suggestion Finder” toggle, the “USB Suggestion Finder” line disappears.  It appears again, only after I click on “Show advanced”.

* Settings shown, when “Show advanced” is disabled, should not disappear, when the value for the setting is changed.

The wording of the setting in case of USB is misleading:

* I thought it only disables UBS discovery of items for the inbox, but in fact it prevents loading the `nrjavaserial`, `org.openhab.core.io.transport.serial`, `org.openhab.core.io.transport.serial.rxtx` and `org.openhab.core.io.transport.serial.rxtx.rfc2217` bundles, so all USB serial devices are dysfunctional.

My understanding of the description “USB Suggestion Finder: Scan connected USB devices to suggest add-ons. Enabling/disabling may take up to 1 minute.” is that no scans for USB devices will be done, no suggestions will appear in the Inbox and no add-ons will be suggested for installation.  However installed USB/Serial things will work and it will be possible to install more USB/Serial devices.

With a binding under `openhab/shared/addons/.jar` with required `<feature>openhab-transport-serial</feature>` when I disable `USB Suggestion Finder`, Save, stop openhab, start openhab, openhab.log contains:
```
[ERROR] [Events.Framework                    ] - FrameworkEvent ERROR org.osgi.framework.BundleException: Could not resolve module: org.openhab.… [248]
  Unresolved requirement: Import-Package: org.openhab.core.io.transport.serial
[...]

PR #3503, opened 2025-11-21
Refactor config sheet advanced toggle

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/3480

As described in the issue, it is sometimes confusing advanced parameters with a non-default value are shown also when the show advanced toggle is off.

This refactors that behaviour:
- make the 'Show advanced' checkbox clearly toggle between showing the advanced parameters or not (and hide the option if there are no advanced options).
- put a badge next to the 'Show advanced' toggle that shows the number of advanced parameters with a non-default value (not shown if all have default values).
- the badges have tooltips clarifying their meaning.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/config-parameter.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-parameter.vue
@@ -13,16 +13,11 @@
                  :f7router
                  @input="updateValue" />
     </f7-list-group>
-    <f7-list-item v-else-if="readOnly && configDescription.context === 'password'"
-                  :is="passwords"
-                  :config-description="configDescription"
-                  :value="value"
-                  :parameters="parameters"
-                  :configuration="configuration"
-                  :title="configDescription.title" />
-    <f7-list-item v-else :after="value !== undefined && value !== null ? value.toString() : 'N/A'" />
+    <f7-list-item v-else
+                  :title="configDescription.label"
+                  :after="value !== undefined && value !== null ? (configDescription.context === 'password' ? '•'.repeat(20) : value.toString()) : 'N/A'" />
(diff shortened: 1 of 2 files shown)


mherwege:
I have a jest test failure I don't understand with this PR. It looks like plenty of lines are already commented in that test. I don't see how my changes would impact the test, but it doesn't succeed anymore.

florian-h05:
I fixed the test. The test checked whether the description was rendered by retrieving it using its class name, which you changed in this PR. Updating the class name in the test fixed the failure.

florian-h05:
> when showing advanced parameters, put a badge next to the advanced parameter description with a non-default value (value in the badge always 1), indicating the advanced parameter has a non-default value. The sum of these would be equal to the number in the badge next to the 'Show advanced' checkbox.

I find this a bit confusing, non-advanced parameters include no hint whether they are non-default or default.
I personally would leave this out.

mherwege:
> I find this a bit confusing, non-advanced parameters include no hint whether they are non-default or default.

I understand. This was an attempt to still visualize in some way there are parameters of interest in the advanced section, like the confusing behavior before. But if it is as confusing, I should just make it a straight show advanced or not toggle without trying to add any extras.

florian-h05:
My confusion is mainly related to the 1 for an individual parameter, in general a hint that advanced params are non-default is helpful.

mherwege:
I removed the special handling for non-default advanced properties (and also reverted the test change in doing that-.

I also noticed a problem with handling of passwords for read-only components. An f7-list-item was used with handling for a component, which did not work. I fixed this as well.

florian-h05 (approved):
LGTM, thanks!

PR #3506, opened 2025-11-23
Increase delay for "Communication failure" toast

florian-h05:
Fixes an issue where the popup always appeared when opening the iOS app.


digitaldan:
Oh, thank you ! 🙏

florian-h05:
Let me know if the delay still is not enough, I don't think it would harm putting it to 5 secs.

digitaldan:
I think it would be good to increase to 5 seconds.  The current delay definitely helps, but i'm still occasionally getting this pop up when opening the app.

florian-h05:
See #3522.

Issue #1555, opened 2022-11-26
For unset switch items sitemaps do not underline the unset status, unless mappings=[] is used


dilyanpalauzov:
## Which UI are you reporting an issue for?
Basic UI and OpenHAB-Android when displaying sitemaps.

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

In an `openhab/items/*.items` file I create an item with
> Switch test_switch "Let’s see"

In a sitemap I define:
>   Switch item=test_switch label="A1" mappings=[ON="X", OFF="Y"]
>   Switch item=test_switch label="B2"

That is, the same item is displayed twice in the sitemap file.  The value of the item was never set.

The mappings-line A1 shows clearly that the item has unset (undefined, null) value, as neither X nor Y are is selected.

The B2-line does not distinguish between OFF and UNSET value, both are presented to the user in the same way. 

## Expected behavior

I expect that even without `mappings=[ON=, OFF=]` the user can see whether a toggle has undefined state.

## Your environment

OpenHAB 3.3, openHAB-Android 2.23.4-beta.

lolodomo:
This was already discussed several years ago. Unfortunately, there was no solution validated to show NULL/UNDEF values.

lolodomo:
https://github.com/eclipse-archived/smarthome/issues/2962
https://github.com/eclipse-archived/smarthome/issues/3742

dilyanpalauzov:
The NULL/UNDEFINED state could be displayed as a toggle in the middle (neither left nor right).

Or implicit mappings=[ON="ON",OFF="OFF"] can be applied.  With mapping there is distinction between set and unset value.

Issue #3513, opened 2025-11-24
[BasicUI] Show Switches with UnDef and NULL state differently from ON and OFF state


dilyanpalauzov:
When openHAB boots, (switch) items have a state of `NULL`. [A Binding may set the state of an Item to UNDEF if it looses communications with a Thing (for example, a Z-wave doorbell with a dead battery). ](https://www.openhab.org/docs/configuration/items.html#state).  openHAB-Android show a Switch element in ON, OFF and NULL/UNDEF state in three distinct ways, the latter having a question mark on the toggle.  BasicUI with openHAB 5.1.0 build 4946 shows OFF and NULL in the same way.

* Display UNDEF and NULL values for a switch differently from ON and OFF values.

lolodomo:
Already discussed many times over years more generally for any widget but with no consensus on the solution.

kaikreuzer:
@lolodomo Could you reference those past discussions?
The solution in the Android app isn't bad, so I would not now a reason against doing the same in the Basic UI - but of course I could be missing some important arguments from past discussions, so please excuse me if that's the case.

lolodomo:
Duplicate of #1555.
Funny because I answered the same way.
Some links are present in #1555.

lolodomo:
Closing as duplicate

lolodomo:
I don't know if the proposed solution for switch widget is doable with the used framework.

mherwege:
> I don't know if the proposed solution for switch widget is doable with the used framework.

I think it is with a little bit of CSS. I created a PR for a question mark on the switch button.

PR #3514, opened 2025-11-25
[BasicUI] New rendering of switch widget for UNDEF or NULL

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/1555

This adds a question mark to the switch button in BasicUI when NULL or UNDEF.

See https://github.com/openhab/openhab-webui/issues/3513#issuecomment-3575496053

This is a first quick attempt at this, and only for switches. It did work for me.

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -2651,9 +2651,7 @@
 			} else {
 				_t.parentNode.classList.remove("unknown-state");
 			}
-			if (typeof componentHandler !== "undefined") {
-				componentHandler.upgradeElement(_t.parentNode);
-			}
+			componentHandler.upgradeElement(_t.parentNode);
 
 			if (_t.hasValue) {
 				_t.valueNode.innerHTML = value;
--- a/bundles/org.openhab.ui.basic/web-src/_layout.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_layout.scss
@@ -1012,6 +1012,7 @@ input::-webkit-calendar-picker-indicator {
 // Switch question mark when UNDEF or NULL
 .unknown-state .mdl-switch__thumb::after {
 	content: "?";
+	color: var(--switch-unknown-color);
 	font-size: 14px;
 	position: absolute;
 	top: 50%;
--- a/bundles/org.openhab.ui.basic/web-src/_theming.scss
+++ b/bundles/org.openhab.ui.basic/web-src/_theming.scss
@@ -7,6 +7,7 @@ body[data-theme="light"] {
 	--container-text-color: #616161;
 	--switch-off-track-bg: rgba(0,0,0,.26);
 	--switch-on-track-bg: #9fa8da;
+	--switch-unknown-color: #616161;
 	--border-color: #ccc;
 	--radio-unchecked-color: rgba(0,0,0,.5);
 	--input-undef-color: rgba(0,0,0,.26);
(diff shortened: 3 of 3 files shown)


lolodomo:
Can you show a screenshot please?

mherwege:
For me, it looks like this:

image

It is probably possible to do something similar for other controls that have no clear visualisation if something is null or undef. But that will require more work.

In dark mode, I feel like the ? is a bit light. I didn't look at how to change that yet:

image

kaikreuzer:
I quite like it!

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -2646,6 +2646,15 @@
 				}
 			}
 
+			if (itemState === "UNDEF" || itemState === "NULL") {
+				_t.parentNode.classList.add("unknown-state");
+			} else {
+				_t.parentNode.classList.remove("unknown-state");
+			}
+			if (typeof componentHandler !== "undefined") {
+				componentHandler.upgradeElement(_t.parentNode);
```
lolodomo: We generally have a componentHandler.downgradeElements when we have a componentHandler.upgradeElement.

mherwege: I don't think it is required here, but happy to be proven wrong. The code does not change the component, only applies or removes a class that forces a CSS to be applied. But to force the DOM to update, the upgradeElement call is required. There are no extra events being added. There are no extra HTML elements as child of parentNode being added. If the switch is turned ON or OFF, this will be called again and the question mark will be removed.
All examples in the code I could find of using downgradeElements is effectively removing events or child elements.

One way of getting more intelligent could be to first check if the unknown-state class already is on the element before adding and the other way around. This would save a call to upgradeElement. I am not sure it brings a lot though.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -2646,6 +2646,15 @@
 				}
 			}
 
+			if (itemState === "UNDEF" || itemState === "NULL") {
+				_t.parentNode.classList.add("unknown-state");
+			} else {
+				_t.parentNode.classList.remove("unknown-state");
+			}
+			if (typeof componentHandler !== "undefined") {
```
lolodomo: Is thatt test really required?

lolodomo:
> I quite like it!

So we are going to merge it ;)

bundles/org.openhab.ui.basic/web-src/smarthome.js
mherwege: You mean the check for undefined? It may not be, just safe programming from my side.

mherwege:
I slightly improved the rendering of the question mark in the dark theme and removed the undefined check.

kaikreuzer:
> So we are going to merge it ;)

Sure - as soon as you are happy with the code, feel free to merge!

mherwege:
@lolodomo I have similar logic for the slider now. Do you want me to include it in this PR, or keep it separate?

If I am not missing something, the only two other controls that show a state in the control itself are the color picker and color temperature picker. These would be other candidates to overlay the button with a question mark in my view. All the others don't show it directly in the control (can show it in the label though) or already have some representation (e.g. - in the input control).

digitaldan:
I quite like this idea as well, i might try to replicate this in our Main UI (maybe make it optional).

lolodomo:
> Do you want me to include it in this PR, or keep it separate?

Separate PR please so that I can merge this one quickly.

PR #3518, opened 2025-11-27
[BasicUI] New NULL or UNDEF rendering for Slider, Colorpicker and Colortemperaturepicker

mherwege:
This build on https://github.com/openhab/openhab-webui/pull/3514 for Switch controls and enables a similar question mark visualisation for Slider, Colorpicker and Colortemperaturepicker controls.

The Slider color scheme for the dark theme has also been adjusted slightly, as the slider was hardly visible in the dark theme so far.

Here are some screenshots:

image

image

Note that a slider in the dark theme used to look like:

image

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
@@ -33,6 +33,7 @@ import org.openhab.core.library.types.HSBType;
 import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.model.sitemap.sitemap.Widget;
 import org.openhab.core.types.State;
+import org.openhab.core.types.UnDefType;
 import org.openhab.core.ui.items.ItemUIRegistry;
 import org.openhab.core.util.ColorUtil;
 import org.openhab.ui.basic.internal.WebAppConfig;
@@ -141,7 +142,8 @@ public abstract class AbstractWidgetRenderer implements WidgetRenderer {
         String state = getState(w);
         snippet = snippet.replace("%state%", escapeHtml(state));
         snippet = snippet.replace("%state_in_url%", escapeURL(state));
-        if ("UNDEF".equals(state) || "NULL".equals(state)) {
+        State widgetState = itemUIRegistry.getState(w);
+        if (UnDefType.UNDEF.equals(widgetState) || UnDefType.NULL.equals(widgetState)) {
             snippet = snippet.replace("%unknown_state%", "unknown-state");
         } else {
             snippet = snippet.replace("%unknown_state%", "");


lolodomo:
The ? Is not very visible for the colour temperature widget. Maybe it would be more visible in white?

Thank you for the change of colour for slider widget in dark theme.

mherwege:
> The ? Is not very visible for the colour temperature widget. Maybe it would be more visible in white?

? In white, or the button background, like for the colorpicker. I know gray was chosen as it doesn’t match a temperature, but that may be less relevant when the question mark is there.

lolodomo:
> > The ? Is not very visible for the colour temperature widget. Maybe it would be more visible in white?
> 
> ? In white, or the button background, like for the colorpicker. I know gray was chosen as it doesn’t match a temperature, but that may be less relevant when the question mark is there.

I agree with you, we could probably move to white with the question mark like for the color widget.

mherwege:
> I agree with you, we could probably move to white with the question mark like for the color widget.

I changed it to white background.

lolodomo:
Can you update the screenshots?

mherwege:
> Can you update the screenshots?

Images have been updated.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
```diff
@@ -141,6 +141,11 @@ protected String preprocessSnippet(String originalSnippet, Widget w, boolean ign
         String state = getState(w);
         snippet = snippet.replace("%state%", escapeHtml(state));
         snippet = snippet.replace("%state_in_url%", escapeURL(state));
+        if ("UNDEF".equals(state) || "NULL".equals(state)) {
```
lolodomo: If we want to be rigorous, it would be better to check itemUIRegistry.getState(w) and not the resulting string as you could have a valid item state for a string item with "NULL" or "UNDEF"  state but which is not a NULL/UNDEF item state. 
I hope I am enough clear !

bundles/org.openhab.ui.basic/web-src/_layout.scss
```diff
@@ -1009,13 +1009,47 @@ input::-webkit-calendar-picker-indicator {
...
 	transform: translate(-50%, -50%);
 }
+.mdl-slider__container.unknown-state {
+	position: relative;
+	overflow: visible !important;
+}
+.mdl-slider__container.unknown-state::after {
+	content: "?";
+	color: var(--state-unknown-color);
+	font-size: 14px;
```
lolodomo, line 1029: Question: no need to set font-weight to bold like you do for color picker and color temperature picker ?

Same question for .unknown-state .mdl-switch__thumb::after

lolodomo:
@mherwege : did you check the rendering in these 4 use cases?

1. when opening the page and the item state is NULL/UNDEF
2. when opening the page and the item state is not NULL/UNDEF
3. when the page is opened with a NULL/UNDEF item state and there is a live update with a non NULL/UNDEF state
4. when the page is opened with a non NULL/UNDEF item state and there is a live update with a NULL/UNDEF state

lolodomo:
And please check the changes for the slider in several browsers  (at least Firefox , Chrome and Edge), I mean when sliding (your stuff that remove the question mark and restore it).

bundles/org.openhab.ui.basic/web-src/_layout.scss
mherwege: As you can see in the screenshots, I didn't need bold here. I used the font weight from the switch and only added bold to color picker and color temperature picker because it did look much less heavy then for the switch and slider without.

bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/AbstractWidgetRenderer.java
mherwege: Yes, good point. I will change it.

mherwege:
> @mherwege : did you check the rendering in these 4 use cases?
> 
> 1. when opening the page and the item state is NULL/UNDEF
> 2. when opening the page and the item state is not NULL/UNDEF
> 3. when the page is opened with a NULL/UNDEF item state and there is a live update with a non NULL/UNDEF state
> 4. when the page is opened with a non NULL/UNDEF item state and there is a live update with a NULL/UNDEF state

Yes, I had tested all 4 scenarios.

mherwege:
> And please check the changes for the slider in several browsers (at least Firefox , Chrome and Edge), I mean when sliding (your stuff that remove the question mark and restore it).

I developed with Firefox and did some tests with Chrome when developing. I now also tested with Edge. I have not seen issues with any of these browers. I suspect this may not work well with Internet Explorer, but I am not sure we should be too concerned about that anymore.

Issue #416, opened 2020-10-18
[MainUI] More actions in the discovery page


lolodomo:
In the discovery page, when you click on a discovered thing, you have only the option to create a thing.
I think we should have, if possible, the same options as in the inbox page, that is the actions "Add a thing", "Ignore" and "Remove".

PR #3519, opened 2025-11-28
Add thing: Add ignore & remove from inbox buttons

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/416

This exposes the same buttons as when clicking on a discovery result directly from the inbox.
Note that it is not possible to unignore an entry from that screen, as ignored entries are never shown there (but can still be unignored from the inbox).


florian-h05 (approved):
LGTM, thanks!
Great to resolve such old issues.

BTW, looking at the mixin and the mixins it used, those can be converted to TypeScript files, they do not depend (at least I haven't seen a dep while skimming through) on state to they can just be functions in a file, no need for mixin or composable.

PR #3521, opened 2025-11-28
Fix back navigation/initial load/reload displays wrong page

### LINKED ISSUE #3497: Settings highlights Model as selected, but shows Items
This is what I see in MainUI with openHAB 5.1.0 build 4925.  When I click on “Items” on the left nothing happens.  The browser console shows at the top “Uncaught (in promise) TypeError: ServiceWorker script at https://192.168.0.11:8443/service-worker.js for scope https://192.168.0.11:8443/ encountered an error during installation.” The title of the page is “Model - openHAB”.

Eventually I click on Pages. then on Items and then it works.  I do not know how I got in this state.

Image

### LINKED ISSUE #3515: Inconsistent navigation/routing problem
Coming from https://github.com/openhab/openhab-webui/discussions/3381#discussioncomment-14971027.

I don't know whether the router is involved in the problem or not, but basically, when navigating to "sub pages" of `/settings/`, navigating further in and then go "Back" to the "sub page", the "Settings" page is sometimes shown instead of the appropriate "sub page", but the router still thinks you're on the "sub page", so if you e.g. click "Items" again (because that's where you want to go), you end up with `/settings/items/items/`, which fails to load.

The issue seems to be generic, it does at the very least exist for Things and Items. It doesn't consistently occur, and "random navigation" is required to trigger the issue. I've found indications that it might be dependent on timing, by slowing down the browser (using throttling in the dev panel), the issue is drastically reduced or eliminated completely.

Below are some screenshots I captured while trying to browse to `/settings/items/` once this issue had manifested. Once it has, it will seemingly randomly reoccur or not if the page is reloaded (F5) in the browser. I found that it first loads the correct page, and then continues to load "the Settings page" as a second "view" in the DOM, which is then rendered over the page you wanted to see. The screenshots below shows how this evolves:

bilde

bilde

bilde

bilde

The "dual views" can be seen from Vue dev tools below:

bilde

The `SettingsMenu` element is the one causing the problem, hiding the other (correct) one.

By finding the correct element in the DOM and hiding it using `visibility: hidden`, I was able to "prove" that the correct page hides underneath:

bilde

florian-h05:
Fixes #3515. Fixes #3497.

When using async/asyncComponent routes with Framework7 Router,
preloading the previous page incorrectly sets the visible for the previous page instead of the current page.
This especially occurs when using back navigation or reloading a page, 
for example displaying the settings menu (from `/settings/` path) when (re-)loading `/settings/things/`.

Also:
- Fixes wrong signature for the route context used in `loadAsync`.
- Reorders two routes to put the more specific one in first place.
- Makes link-edit.vue page routable to fix issues with reloading it.
- Fixes counter in items-list-vlist not updated.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
@@ -297,6 +297,7 @@ export default {
         this.$refs.itemsList.$el.f7VirtualList.replaceAllItems(this.items)
         this.initSearchbar = true
         this.loading = false
+        this.updateListedItems()
         this.processFilter()
 
         if (!this.eventSource) this.startEventSource()


Nadahar:
My OH is currently not in a runnable state, but Eclipse is chewing through most of Core and some add-ons, and should be good to go once that completes. I can give this a spin then if you'd like me to verify that it solves what I have experienced.

florian-h05:
> I can give this a spin then if you'd like me to verify that it solves what I have experienced.

I will wait with merging then :+1:

Nadahar:
I'm not saying that this is the same issue, but I had one "strange routing" taking place just now. I had done "my usual" trip from Thing -> Channel -> Link -> Item and was on Item `http://localhost:8081/settings/items/BubbleUPnP_S62_Pro_Repeat`. When trying to go back from there, I was taken to `http://localhost:8081/settings/things/upnpcontrol:upnprenderer:fe18eda4-46b4-48e4-b3a6-0fc1deb6ea97/links/BubbleUPnP_S62_Pro_Repeat/repeat`, which was invalid.

The URL for the Thing was `http://localhost:8081/settings/things/upnpcontrol:upnprenderer:fe18eda4-46b4-48e4-b3a6-0fc1deb6ea97`. When clicking Channel -> Link, I'm taken to `http://localhost:8081/settings/things/upnpcontrol:upnprenderer:fe18eda4-46b4-48e4-b3a6-0fc1deb6ea97/links/BubbleUPnP_S62_Pro_Repeat/repeat`, which works when "going in". So, the address when going back is correct, but for some reason, the URL doesn't work when going back again. Are there some "props" from parent pages or something that's missing when going back via the history?

Nadahar:
This behavior seems to be consistent now (which is a fresh breath of air 😉 ), and you don't actually need to go back for the problem to manifest. If you go to Things -> Thing -> Channel -> Link and then just reloads the page, the URL that just loaded fine, isn't valid after a refresh.

florian-h05:
That's no router issue in the sense as before, but the issue we have no route registered for the link pages.

Nadahar:
> That's no router issue in the sense as before, but the issue we have no route registered for the link pages.

Is that hard to fix?

I haven't seen any trace of the previous behavior, and I also think it feels somewhat "snappier".

Nadahar:
I've also found another issue: If I go to Settings -> Items -> Some Item and try to go back, I get a blank page. If I then reload the page (F5) it loads, but now it only shows a list with a single Item. Reloading the page doesn't help, however, if I click the blue "Refresh" button at the bottom, it refreshed properly and all Items are loaded.

edit: This is some weirdness with the "VirtualList". I can see that when the page loads, `renderExternal (vl, vlData)` is called three times, the first time, `vlData.items` is empty, the two subsequent times, it is populated, and the content is the same from what I can see. When the page "works", `vlData.items` is fully populated, but when it doesn't, it only holds a single Item.

I don't quite understand how `f7VirtualList` is initialized, but the problem is probably related to that - or its filter. Could it be that when you click an Item, a filter is somehow applied to it, and is then never cleared when you "go back"?

Nadahar:
I see that the `Refresh` button that makes all Items appear, merely calls `load()`. So, the question then is, isn't `load()` called when you refresh the page, or why are the results different?

florian-h05:
Do these issues appear in production or the dev server build?

florian-h05:
> Is that hard to fix?

I have just pushed a commit that makes the link-edit page routable.

Nadahar:
> I have just pushed a commit that makes the link-edit page routable.

This seems to have solved reloading the "link" page, but now I get a white screen when "backing there" from the Item, just like I do in the "Items" page.

edit: I don't know if it's related, but in both cases where I get this "blank/white screen", the console reports:
```
EventSource failed loading: GET "http://localhost:8081/rest/events/states".
```

edit2: That doesn't seem (directly) related, I get the same error when pages that work loads as well. I'm wondering why I get this though, am I running out of SSE connections?

Nadahar:
> Do these issues appear in production or the dev server build?

I've done it all in dev server for now - do you think that matters?

florian-h05:
Yes, in production I cannot reproduce those issues.

Nadahar:
> Yes, in production I cannot reproduce those issues.

I just built and ran "preview", and neither can I. But why are they in dev? I have cleared my cache and "hard reloaded" to no avail.

Nadahar:
But you're right, it all seems to behave very well in production build now - which I guess is the most important 👍

Nadahar:
I think this fixes #3497 as well.

florian-h05:
> But why are they in dev? I have cleared my cache and "hard reloaded" to no avail.

I cannot really tell you the exact reason, but it very likely is related to bundled in prod vs loaded file-by-file in dev.
Previously we used Webpack, which also bundles in dev, but dev server and prod build times were really bad.
 
Vite is blazingly fast in both (at least the Rolldown (Rust-based vs Rollup JS-based) bundler preview version we use), and working on adding bundling for dev server as well, making the dev and prod environment the same :rocket:

Nadahar:
Back in dev mode, I actually got an error in the console that might be onto something, it complains that it failed to load:
```
:8081/settings/thing…wer/current-power:1
```
It seems like it somehow forgets to include the `http://localhost` part to the URL, which would explain why it doesn't load. Is it some problem with relative/absolute URIs?

florian-h05:
I think the console leaves that out ... when checking for issue with requests, always check the network tab.

PR #3532, opened 2025-12-01
Fix developer sidebar CSS leaking into global CSS

jsjames:
Fix #3530 leaking into global CSS.


florian-h05:
Ah the CSS leaks out … can you please rebase this PR?

jsjames:
> aks out … can you please rebase this PR?

Yeah, not sure if we should start to use the <style scoped> or not.  That would prevent the leaking.

florian-h05:
`<style scoped>` does not (at least in my experience) allow modifying CSS deep inside the components you use ...

florian-h05 (approved):

PR #3536, opened 2025-12-01
Remove manual service worker registration

GiviMAD:
The service worker generated by VitePWA is automatically registered, it does not need to be registered manually.
Therefore, remove the manual registration that wasn't working anyway from App.vue.


florian-h05:
I actually think we should remove the service worker reference from App.vue and check if it is still loaded ... AFAIK VitePWA can automatically inject the service worker without us needing to load it.

GiviMAD:
> I actually think we should remove the service worker reference from App.vue and check if it is still loaded ... AFAIK VitePWA can automatically inject the service worker without us needing to load it.

You are right, just removing the option in App.vue fixes the problem and the service worker gets registered anyway.

florian-h05 (approved):
LGTM, thanks!

Issue #3538, opened 2025-12-03
[MainUI] sheet and popover actions don't work in layout pages (5.1.0.M3)


dujal:
## The problem
I've updated my openHAB from 5.1.0.M2 to 5.1.0.M3 and found some issues in fixed grid layout page I use on wall mounted tablet.
Now actions of type 'sheet' and 'popover' don't work anymore. 'popup' actions works normally as expected.

There are also some visual layout problems, like outline border has some extra margin and card content overflow the fixed card area:

<img width="275" height="236" alt="Image" src="https://github.com/user-attachments/assets/0b2ed615-855d-4563-9b90-46e3f6dfcbc5" />

For reference, this is how the same looks like on demo server (OH 5.0.1):

<img width="263" height="227" alt="Image" src="https://github.com/user-attachments/assets/b5a74ab9-2fd0-4b53-9f55-6f49aee72432" />

## Expected behavior

Sheet or popover should be displayed after any card (e.g. label card) is clicked.

## Steps to reproduce
Test page to reproduce problem:
```yaml
config:
  fixedType: grid
  label: test cards
  layoutType: fixed
blocks: []
masonry: []
grid:
  - component: oh-grid-item
    config:
      h: 2
      w: 2
      x: 2
      y: 0
    slots:
      default:
        - component: oh-label-card
          config:
            item: VenkuTeplota
            title: Temp
            footer: some footer
            outline: true
            action: sheet
            actionModal: page:page_e917c70bbd
canvas: []
```

1.  open the provided test page
2. click on card

## Your environment

```yaml
runtimeInfo:
  version: 5.1.0.M3
  buildString: Milestone Build
locale: cs-CZ
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 21.0.9
  javaVendor: Eclipse Adoptium
  javaVendorVersion: Temurin-21.0.9+10
  osName: Linux
  osVersion: 6.12.47+rpt-rpi-2712
  osArchitecture: aarch64
  availableProcessors: 4
  freeMemory: 332204720
  totalMemory: 664797184
  uptime: 57677
  startLevel: 70
addons:
  - automation-jsscripting
  - automation-pidcontroller
  - binding-astro
  - binding-bluetooth
  - binding-gpstracker
  - binding-http
  - binding-ipcamera
  - binding-kodi
  - binding-lgthinq
  - binding-mqtt
  - binding-network
[...]

PR #3539, opened 2025-12-03
oh-grid-layout: Fix grid widget sizing issues

jsjames:
Refs #3538.

Fixes: For oh-grid-item the class "oh-grid-item-content" attribute was not automatically applied to the oh- widget in the generic-widget-component since there are multiple top-level components.

- Reverted css style for f7-cards to f7v5
- Bound attrs in generic-widget for oh- components


jsjames:
@florian-h05 - i only passed through attrs for the oh- components in generic-widget-component at this time. We may need to pass through for additional components in the future.

jsjames:
> @florian-h05 - i only passed through attrs for the oh- components in generic-widget-component at this time. We may need to pass through for additional components in the future.

@florian-h05 - After more digging - when we set inheritAttrs to false in #3461, this caused the class to not be inherited. I added the v-bind="$attrs" to the elements I believe are necessary, but would like your assessment as well.

Issue #1774, opened 2023-03-06
Create default input icon for sitemaps


mherwege:
Work is ongoing to add an input element to BasicUI (and potentially the mobile apps).

What is missing is a good default icon for it. This is an opportunity for someone to create such an icon. Here are the current icons: https://www.openhab.org/docs/configuration/iconsets/classic/
Look for the icons under Channels - Widgets. These represent the default icons for various sitemap widgets.

Related to https://github.com/openhab/openhab-webui/pull/1729#issuecomment-1455985544

PR #3554, opened 2025-12-06
Sitemap default icon for input widget

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/1774

As nobody picked up on this, I had a go at it.

Here are some screenshots:

image

image


dilyanpalauzov:
Using
```xml
<svg style="color-scheme:light dark" …>
… style="fill:currentColor" …
```
allows creating icons, which adapt their colour to black or white, depending on whether light or dark mode is used in BasicUI/browser’s preference.

If I remember correctly, the opposite colour is `fill:none` and to get the effect in BasicUI “Inline SVG” must be enabled.

In the input.svg icon all colours, expect the light blue for the inner filling, could be changed to `currentColor` and `none`.

lolodomo:
Take care also we have something "consistent" with other icons.
If I correctly remember, I removed all PNG icons and only SVG icons are remaining so it removes one problem.

mherwege:
It is an svg icon with the same size (I checked the text icon for comparison). I tried ti make it look somewhat consistent but I am not a graphics designer. And the quality and consistency of the icon can certainly be debated. It is just nobody else volunteers to do anything. If someone has a better icon, we can just replace the svg and it should be OK.

lolodomo:
I was not debating your design, my comment about consistency was in response to @dilyanpalauzov and his suggestion of adaptative colours.

PR #3555, opened 2025-12-07
oh-context: Properly inherit attributes

jsjames:
This addresses a 'Extraneous non-props attributes (class) were passed to component but could not be automatically inherited because component renders fragment or text or teleport root nodes' error.


florian-h05:
I think we should rather pass the attrs properly.

florian-h05 (approved):
Ah sorry, looked at the code now.

LGTM!

PR #3564, opened 2025-12-10
Item picker: Refine group filter and add semantic filter

mherwege:
This builds on https://github.com/openhab/openhab-webui/pull/3487

Discussion leading to this PR is here: https://community.openhab.org/t/parameter-playground/167338/5?u=mherwege

The PR has 3 changes:

1. Currently, when filtering on `groupType`, the results will also return all groups without a type as these are considered compatible. This may lead to a large number of groups being shown. The PR will now restrict the list to only compatible group types and only include groups without type if `type` is set to `Group` at the same time.
2. Give an option to ignore all items with a semantic tag from the filtered list: this makes it possible to easily exclude all semantic groups from the items. A new filter has been introduced, `excludeSemantic`, when set to `true` will filter out semantic items.
3. Introduce a new filter that will allow to only show all items with a specific semantic classification (and its children). A new filter `semantic` has been introduced. It can be set to a semantic tag or a list of semantic tags. E.g. if set to `WindowCovering` it will find all items with a `WindowCovering` tag, but also items with a `Blind` tag, which is a child of `WindowCovering`. This filtering works for Location, Equipment, Point and Property tags.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -153,7 +153,7 @@
                      :items="items"
                      @input="(value) => this.item.groupNames = value"
                      :multiple="true"
-                     :filterType="Group"
+                     filterType="Group"
                      :filterGroupType="compatibleGroupTypes"
                      :set-value-text="false" />
       </f7-list-group>
--- a/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
@@ -192,7 +192,7 @@ export default {
         }
       }
       if (this.filterGroupType?.length) {
-        const filterGroup = this.filterType === 'Group'
+        const filterGroup = this.filterType?.includes('Group')
         if (Array.isArray(this.filterGroupType)) {
           this.filteredItems = this.filterGroupItems(this.filteredItems, this.filterGroupType, filterGroup)
         } else {


bundles/org.openhab.ui/web/src/components/item/item-form.vue
```diff
@@ -153,8 +153,8 @@
                      :items="items"
                      @input="(value) => this.item.groupNames = value"
                      :multiple="true"
+                     :filterType="Group"
                      :filterGroupType="compatibleGroupTypes"
-                     :filterToggle="true"
```
florian-h05, line 157: Do you really want to remove the filter toggle?

florian-h05 (commented):
Looks good, thanks!

One question though:

bundles/org.openhab.ui/web/src/components/item/item-form.vue
mherwege: I don't see it serves a purpose in the context. This is for selecting groups for an item. The picker will show all groups that have a type compatible with the item or have no type (by adding filterType="Group"). Switching the toggle would mean also non-group items or group items with incompatible type are shown. I considered that more confusing then forcing the limit.
For me, the toggle mainly serves a purpose where there could be reasons to need to derive from the original filter, e.g. when picking an item for a channel when you want to apply a profile.

mherwege: Wait, I need to check again.

mherwege: I confirm my statement above. There was an error in the PR description which I have corrected now.
Groups without type will only be shown when type="Group".

mherwege:
There was still an error though, which I fixed now.

bundles/org.openhab.ui/web/src/components/item/item-form.vue
florian-h05: Fine then :+1:

Issue #3561, opened 2025-12-08
[MainUI] Thing DSL to thing fails for thing with dynamic channel groups


mherwege:
## The problem

I have a `systeminfo` thing defined through the UI. The thing has dynamic channels and channel groups, and in its initialization, the thing type `computer` is replaced with `computer-impl` to suppor this.
When switching to code view DSL and back, the parser will detect thing type `computer` instead of `computer-impl` and will generate an error that you cannot change the thing type.

## Expected behavior

Make switching between the Thing tab and code tab with DSL work for things with dynamic channels.

## Steps to reproduce

Create a thing for the systeminfo binding in the UI. Switch to Code tab and select DSL. Switch back to the Thing tab.

Switching to YAML code and back does not generate an error.

Switching to the Code tab, selecting DSL, switching back to the Thing tab, clicking OK on the error box, switching back to the Code tab (you see DSL now), switching to YAML, clicking OK on the error box, switching back to the Thing tab gives another error: Error parsing YAML: Server Error.

## Your environment

openHAB 5.1.0 Build #4978

## Remarks

This is may be a core issue, but I am not sure it can easily be resolved in core, as the used REST endpoint will just translate the input it gets into output and does not know about the changed thing type. So some extra processing may be required in the UI. Therefore I rather created it in the UI.

lolodomo:
Is the problem dynamic channels or dynamic change of thing type ?
Is the behaviour different between DSL and YAML?

lolodomo:
If I correctly remember, we don't have a thing type in DSL and YAML formats, the thing type is included in the thing UID. But we could try to retrieve it from registry in the REST API, it would return the dynamic thing type, at least when the binding is started.
I will check how the thing type is retrieved by the API in both directions.

mherwege:
> Is the problem dynamic channels or dynamic change of thing type ?

It is dynamic thing types, but that is required for dynamic channel groups.

> Is the behaviour different between DSL and YAML?

I gave some descriptions on it above. The message is different, but I think both struggle.

> But we could try to retrieve it from registry in the REST API, it would return the dynamic thing type, at least when the binding is started.

If I remember correctly, the idea was to be able to do the translation even before the thing was saved in the UI. So, it wouldn’t even exist in the registry. That’s why I was thinking we might have to solve it in the UI, reverting the thing type it gets back from the REST call.

lolodomo:
[...]

PR #3568, opened 2025-12-11
Thing details: Code tab fixes

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/3561

This PR solves 2 issues with thing code editing:

1. A thing with dynamic thing type, like for the systeminfo binding, always returned an error because the new dynamic thing type did not match the thing type from code parsing (both for YAML and DSL). In this PR, the check has been disabled. DSL and YAML do not carry the thingTypeUID separately, and the thingTypeUID after parsing is always derived from the thingUID. As it is not in the YAML and DSL formats, it cannot be modified in code at this time, so it makes sense to not check for it. This PR will make sure to keep the original thingTypeUID before converting to code.
2. DSL code generation on Windows has \r\n as line ending on Windows. The code editor (CodeMirror) will automatically update this to \n and emits a input event immediately with a content different from the original (only \n). As a consequence, any switch to a DSL code view on Windows immediately has the dirty flag set. This PR normalizes line endings before setting the code in the editor to avoid this.


florian-h05 (approved):
Thanks!

PR #3575, opened 2025-12-12
Migrate API usage from TypeScript code to typesafe @hey-api

jsjames:
For post 5.1 release

- Initial draft of api using fetch and @hey-api to autogenerate api code.
- modified a few instances of api calls (WIP), pending review of overall structure

### DIFF
--- a/bundles/org.openhab.ui/web/build/openhab-api-spec.json
+++ b/bundles/org.openhab.ui/web/build/openhab-api-spec.json
@@ -3001,7 +3001,7 @@
                 ]
             }
         },
-        "/items/{itemname}/metadata/{namespace}": {
+        "/items/{itemName}/metadata/{namespace}": {
             "put": {
                 "tags": [
                     "items"
@@ -3010,7 +3010,7 @@
                 "operationId": "addMetadataToItem",
                 "parameters": [
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3047,9 +3047,6 @@
                     "201": {
                         "description": "Created"
                     },
-                    "400": {
-                        "description": "Metadata value empty."
-                    },
                     "404": {
                         "description": "Item not found."
                     },
@@ -3073,7 +3070,7 @@
                 "operationId": "removeMetadataFromItem",
                 "parameters": [
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3112,7 +3109,7 @@
                 ]
             }
         },
-        "/items/{itemname}/tags/{tag}": {
+        "/items/{itemName}/tags/{tag}": {
             "put": {
                 "tags": [
                     "items"
@@ -3121,7 +3118,7 @@
                 "operationId": "addTagToItem",
                 "parameters": [
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3167,7 +3164,7 @@
                 "operationId": "removeTagFromItem",
                 "parameters": [
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3206,7 +3203,7 @@
                 ]
             }
         },
-        "/items/{itemname}": {
+        "/items/{itemName}": {
             "get": {
                 "tags": [
                     "items"
@@ -3250,7 +3247,7 @@
                         }
                     },
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3292,7 +3289,7 @@
                         }
                     },
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3361,7 +3358,7 @@
                         }
                     },
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3420,7 +3417,7 @@
                 "operationId": "removeItemFromRegistry",
                 "parameters": [
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3592,7 +3589,7 @@
                 ]
             }
         },
-        "/items/{itemname}/state": {
+        "/items/{itemName}/state": {
             "get": {
                 "tags": [
                     "items"
@@ -3601,7 +3598,7 @@
                 "operationId": "getItemState_1",
                 "parameters": [
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3651,7 +3648,7 @@
                         }
                     },
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -3702,7 +3699,7 @@
                 }
             }
         },
-        "/items/{itemname}/metadata/namespaces": {
+        "/items/{itemName}/metadata/namespaces": {
             "get": {
                 "tags": [
                     "items"
@@ -3711,7 +3708,7 @@
                 "operationId": "getItemNamespaces",
                 "parameters": [
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "item name",
                         "required": true,
@@ -4280,7 +4277,7 @@
                 ]
             }
         },
-        "/persistence/items/{itemname}": {
+        "/persistence/items/{itemName}": {
             "get": {
                 "tags": [
                     "persistence"
@@ -4297,7 +4294,7 @@
                         }
                     },
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "The item name",
                         "required": true,
@@ -4389,7 +4386,7 @@
                         }
                     },
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "The item name.",
                         "required": true,
@@ -4453,7 +4450,7 @@
                         }
                     },
                     {
-                        "name": "itemname",
+                        "name": "itemName",
                         "in": "path",
                         "description": "The item name.",
                         "required": true,
@@ -8067,16 +8064,16 @@
             "Module": {
                 "type": "object",
                 "properties": {
-                    "configuration": {
-                        "$ref": "#/components/schemas/Configuration"
-                    },
                     "typeUID": {
                         "type": "string"
                     },
-                    "description": {
-                        "type": "string"
+                    "configuration": {
+                        "$ref": "#/components/schemas/Configuration"
                     },
                     "label": {
+                        "type": "string"
+                    },
+                    "description": {
                         "type": "string"
                     },
                     "id": {
@@ -8110,8 +8107,12 @@
             "Template": {
                 "type": "object",
                 "properties": {
-                    "uid": {
-                        "type": "string"
+                    "tags": {
+                        "uniqueItems": true,
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
                     },
                     "visibility": {
                         "type": "string",
@@ -8121,17 +8122,13 @@
                             "EXPERT"
                         ]
                     },
-                    "tags": {
-                        "uniqueItems": true,
-                        "type": "array",
-                        "items": {
-                            "type": "string"
-                        }
-                    },
-                    "description": {
+                    "uid": {
                         "type": "string"
                     },
                     "label": {
+                        "type": "string"
+                    },
+                    "description": {
                         "type": "string"
                     }
                 }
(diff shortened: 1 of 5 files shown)


jsjames:
@florian-h05 - a few things to get your input:

- in the existing api, all the framework7 request promises were wrapped to reject the promise on a caught error. Is there any reason we wouldn't just throw the error?
- I haven't done much error checking yet in this implementation - so once we are okay with the structure, I can continue to make progress.
- I'd like to pull the openAPI spec from the source vs. demo.openhab.org - I have an enquiry on the forum for this - but if you might be able to point me in the right direction here.

any other feedback on overall structure? If not, I can continue to port calls over to this.

florian-h05:
First of all, great PR.
I've already used hey-api in a university project and having a fully typed API client without needing to write type defs on our own is great!

> in the existing api, all the framework7 request promises were wrapped to reject the promise on a caught error. Is there any reason we wouldn't just throw the error?

I think this was done to "simplify" error handling so you can simply check `err === 'Not found'` in the `.catch` block of a promise. But I always found this quite confusing, to let's throw the full error and then check `err.statusCode === 404` instead.

> I'd like to pull the openAPI spec from the source vs. demo.openhab.org - I have an enquiry on the forum for this - but if you might be able to point me in the right direction here.

I have to look into how to get the OpenAPI spec from the source.

florian-h05:
WRT to the general structure, I will have a look at this PR after the 5.1 release.

jsjames:
> WRT to the general structure, I will have a look at this PR after the 5.1 release.

FYI - @wborn helped me understand how the openapi spec is generated. I think this will be a bit iterative as there may be some enhancements to the openapi spec which will help to make the interface generation better (for instance removing the DTO from some of the object names and further specifying which fields cannot be blank.

For now, I've been modifying a local version of openhab-core and generating the spec and use that to build the interface. Eventually though, if we get all the changes done and submitted to core, we can pull the spec from there.

jsjames:
@florian-h05 - a few updates:

- I've submitted a PR to core to enhance/correct some ares of the openapi spec - which is generated at runtime using the swagger annotations. I have build a version of core and extracted the openapi spec that is currently in the build directory. So, assuming the PR will get approved, the spec will match what we feed into hey-api.
- I do expect we will find additional issues as we expand typescript in the ui that we might need to address in the core openapi spec.
- Lastly, the one thing that is not in the current openapi spec is whether a field is required or optional and as such, all generated typescript interfaces have all the fields as mandatory. As we find fields that are optional, we can annotate them in core so they will be automatically generated. I don't know of a more automated process to do this - so I think this will be manual.

jsjames:
@florian-h05 - this one is ready for your review. I have a parallel PR into core to enhance the open-api spec schema - however it does not need to be submitted before this one as I've built a version locally to generated the openapi spec currently in the build directory.

One enhancement is that the spec will now specifiy required vs. optional fields - unfortunately, there doesn't seem to be a systematic way to do this. It might be worth you looking at the generated api/types.gen.ts file to see if there are other enhancements. I can make those then in core if you see areas to update as we should keep the master spec the "source of truth".

All-in-all, I think its a good way to setup the api - and have modified several components/files with the new setup. The old one still exists and we can remove over time.

florian-h05:
Let me look through the DTOs in core to check what’s optional and what not / look me through the existing type definitions in the UI.
Is the OpenAPI Spec now available as a build artefact?

jsjames:
> Let me look through the DTOs in core to check what’s optional and what not / look me through the existing type definitions in the UI.

All the existing definitions should be complete as I used those as a reference.

> Is the OpenAPI Spec now available as a build artefact?

No, the OpenAPI spec is generated organically at runtime using the @Schema annotations. So, to generate the spec that is in the build directory, I had to build all the bundles, update them in an openhab environment and then pull the spec from /rest/spec.  The types.gen.ts file in the api directory would be a good one to review.

florian-h05:
Just to mention it here: https://github.com/openhab/openhab-webui/discussions/3381#discussioncomment-14718881 should be handled

jsjames:
> Just to mention it here: [#3381 (comment)](https://github.com/openhab/openhab-webui/discussions/3381#discussioncomment-14718881) should be handled

if you look at the generated api, it does return an empty object/array if a 204 response is returned. So, I think this is covered.

[EDIT] @florian-h05 - one thing to note, in order for hey-api to return an empty object, you have to force the parseAs as json - otherwise it will return a stream.  I personally think this is a bug in hey-api - but haven't submitted a question/issue to hey-api.  (i.e. see the use for firmware in the useThingEditStore). We would just need to make sure that all api request that could return a 204 are forced to parse as json.

jsjames:
@florian-h05 - okay - this is rebased with the oxc and i've built a new version of the openapi spec based on the changes from core.

One area to get your thoughts - hey-api can be configured to return just the data or the full response. I chose the former for simplicity (see the responseStyle parameter in createClient in the generate-rest-client script). However, then you can't get the response codes on failure. So, I created a new Error type which includes this (see ApiError in initApi) that when an error happens it gets thrown. You can configure the library to throwOnError, but then you don't get any response code so a custom error handler allows us to provide that.

jsjames:
@florian-h05 - I just rebased (again). Two things to note:

- I was getting typescript error between readonly and Readonly type definitions - from what i can tell readonly is deep whereas Readonly only refers to top-level (see useComponentStore). As an aside, it doesn't look like there is an easy way to get hey-api to flag whole structures as readonly on responses w/o adding a Scheme flag to EVERY element in the dto struct.
- In useThingEditStore, you had the type of the firmware status being an array, the openapi-spec is just a single FirmwareStatus element. If an array is actually returned, then, we have to tweak the @Schema for this.

florian-h05:
You may use DeepReadonly (exported from Vue), this marks deep readonly as applied by Vue's `readonly` method.

florian-h05:
It this ready for review from your POV, or is there something left to work on?

jsjames:
> It this ready for review from your POV, or is there something left to work on?

It’s ready.

bundles/org.openhab.ui/web/src/pages/settings/menu/settings-menu.vue
```diff
@@ -319,12 +321,15 @@ export default {
...
-        this.$oh.api.get('/rest/addons?serviceId=all').then((data) => {
+        api.getAddons({ serviceId: 'all' }).then((data) => {
           this.addonsInstalled = data
             .filter((a) => a.installed && !['application/vnd.openhab.ruletemplate', 'application/vnd.openhab.uicomponent;type=widget', 'application/vnd.openhab.uicomponent;type=blocks'].includes(a.contentType))
             .sort((s1, s2) => this.sortByLabel(s1, s2))
           this.persistenceAddonsInstalled = this.addonsInstalled.filter((a) => a.installed && a.type === 'persistence')
           this.addonsLoaded = true
+        }).catch((error) => {
+          console.error('Error loading addons:', error.message)
+          this.addonsLoaded = true
```
florian-h05, line 332: Shouldn't it rather be addonsLoaded = false?

bundles/org.openhab.ui/web/src/js/stores/useModelStore.ts
```diff
@@ -147,33 +147,28 @@ export const useModelStore = defineStore('model', () => {
     item.semanticLoopDetector = true
     item.modelPath = item.parent ? [...buildPathInModel(item.parent, items, filteredItems), item.parent] : []
     delete item.semanticLoopDetector
-    item.children = []
```
florian-h05, line 150: What was this code good for?

bundles/org.openhab.ui/web/src/App.vue
florian-h05, line 1: Have you tested what happens when trying to use Main UI with a server where basic auth is enabled through a reverse proxy?

jsjames:
> * In useThingEditStore, you had the type of the firmware status being an array, the openapi-spec is just a single FirmwareStatus element. If an array is actually returned, then, we have to tweak the @Schema for this.

@florian-h05 - did you get a chance to look at the type for response to the firmware status?

florian-h05:
No, not yet. I should probably do.

bundles/org.openhab.ui/web/src/App.vue
jsjames: No. I don’t have a setup like that to test. I can try to set something up unless you have a system already setup?

bundles/org.openhab.ui/web/src/js/stores/useModelStore.ts
jsjames: > What was this code good for?

This code is no longer necessary given #3733. These blank fields are now added when converting from the item to a model item after the api call. This could have been part of that pr.

bundles/org.openhab.ui/web/src/App.vue
florian-h05: No, unfortunately I don't have this at hand.

bundles/org.openhab.ui/web/src/pages/settings/menu/settings-menu.vue
jsjames: That just means the ui never exits the spinning waiting icon. Maybe a better solution is to display an error in the panel that the adding could not be loaded.

jsjames:
> No, not yet. I should probably do.

I just looked at the core code and it looks like the schema is correct and aligns with the core code.  I think we should be good here.

bundles/org.openhab.ui/web/src/pages/settings/menu/settings-menu.vue
florian-h05: Okay, but that’s for another PR.

bundles/org.openhab.ui/web/src/App.vue
jsjames: I did just build and test on my production system using myopenhab.org proxy and it works. Not sure if this covers the exact case you are thinking of?

jsjames:
> > No, not yet. I should probably do.
> 
> I just looked at the core code and it looks like the schema is correct and aligns with the core code. I think we should be good here.

I was expanding this to the thing-details page and it looks like there is an issue with how the PUT is described or I'm calling it wrong. Let me figure this out first before merging.

jsjames:
> > > No, not yet. I should probably do.
> > 
> > 
> > I just looked at the core code and it looks like the schema is correct and aligns with the core code. I think we should be good here.
> 
> I was expanding this to the thing-details page and it looks like there is an issue with how the PUT is described or I'm calling it wrong. Let me figure this out first before merging.

I'm at a loss on this one - I have the code here: [thing-detail-api](https://github.com/jsjames/openhab-webui/tree/thing-detail-api). SImply enabling/disabling a thing will not work. I believe hey-api is generating incorrect code for this and have filed an issue. If you get a sec, maybe you can spot an issue?

jsjames:
> I'm at a loss on this one - I have the code here: [thing-detail-api](https://github.com/jsjames/openhab-webui/tree/thing-detail-api). SImply enabling/disabling a thing will not work. I believe hey-api is generating incorrect code for this and have filed an issue. If you get a sec, maybe you can spot an issue?

So, in digging around in the hey-api code more, I think this is definitely a bug in generated api client. See my explanation here: https://github.com/hey-api/openapi-ts/issues/3250

Not sure if you have suggestions on how to fix?

jsjames:
> > I'm at a loss on this one - I have the code here: [thing-detail-api](https://github.com/jsjames/openhab-webui/tree/thing-detail-api). SImply enabling/disabling a thing will not work. I believe hey-api is generating incorrect code for this and have filed an issue. If you get a sec, maybe you can spot an issue?

This is definitely an error in the hey-api code - I have submitted a PR to openapi-ts: #3255 and included an interim patch for this PR. Things are now working for put/post.

jsjames:
> This is definitely an error in the hey-api code - I have submitted a PR to openapi-ts: #3255 and included an interim patch for this PR. Things are now working for put/post.

hey-api has already merged in this PR (they did make the fix in a different way). I have updated this PR with this new hey-api version and its working well. So, we can move forward with this merge.

florian-h05:
Glad they are still that fast in fixing bugs and releasing a new version ... I've also discovered a bug and proposed a fix in hey-api about a year ago when using it in a university project.

bundles/org.openhab.ui/web/src/App.vue
florian-h05: This is not the exact case, but the exact same environment when it comes to the HTTP layer, so this should be fine!

florian-h05:
@jsjames I have created #3812 which refactors some code to TypeScript and makes use of @hey-api there.
I would propose to migrate to TS Composition API step-by-step, whenever we like to rewrite something.

mrlubos:
> Glad they are still that fast in fixing bugs and releasing a new version ... I've also discovered a bug and proposed a fix in hey-api about a year ago when using it in a university project.

@florian-h05 I'm everywhere 🕵️‍♀️ let me know if you have any more feedback in the future!

florian-h05:
@mrlubos You indeed seem to be everywhere, for now I don't have further feedback. Many thanks for your great work at hey-api, my experience so far has been really good, no major issues!

jsjames:
> @jsjames I have created #3812 which refactors some code to TypeScript and makes use of @hey-api there. I would propose to migrate to TS Composition API step-by-step, whenever we like to rewrite something.

Okay - so we don't want to migrate existing non-TypeScript code to use the hey-api functions? We'll just migrate when we migrate that code to typescript? I did convert several over already to the new api.

Should we add anything to CONTRIBUTING.md regarding using the new hey-api protocol?

florian-h05:
I was thinking about this as well, theoretically we can migrate anything, but if something changes with the OpenAPI Spec and the API client changes, does Vue-tsc catch invalid usage of the API client in pure JS code?

jsjames:
> I was thinking about this as well, theoretically we can migrate anything, but if something changes with the OpenAPI Spec and the API client changes, does Vue-tsc catch invalid usage of the API client in pure JS code?

I don't think you get any typescript benefits in incorporating in plain js code. I see the benefits:

- more robust call format to the api - consistent parameters
- If we want to improve error handling in api calls - i think we should migrate to the new calls to be consistent and eliminate the need to recode error handling in the future
- new api handles 204 (so at a minimum we should migrate those)
- we are bound to find new issues with the openapi spec as we expand to other areas - i'd rather find those out sooner rather than later

Having said that, we risk breaking things by converting - so its a risk.

florian-h05:
Yeah, overall I agree everything should be migrated, I am just not sure about the exact schedule. But if you want to go ahead and convert all usages, fine for me :+1:

Issue #3569, opened 2025-12-11
Log Viewer: Implement keep alive mechanism for WS


florian-h05:
According to https://github.com/openhab/openhab-core/issues/5165, the log viewer needs some type of keep alive handling for the WebSocket:

> What exactly is needed there? What should the UI send?

I can't quite answer that, but usually when you make a heartbeat functionality, you want to use it for something. As it is now, MainUI sends them but never checks if they make it. Core just accepts them without doing anything as well. They might be intended to prevent the socket from timing out, but normally I'd expect that they should have some "meaning" in themselves.

One option I could see is that they could be used to indicate if the log viewer is "connected" - if failed heartbeats were caught, it could be used to set a status that's visible to the user, so that the user knows that the log isn't "online". Another option could be that if a heartbeat fails, an attempt is made to set up a new connection, perhaps on a timer, combined with some kind of online/offline indication.

As they are now, they seem quite pointless, and I also think that it's a problem that you can't really see if the log is "connected" or not - because it certainly has a tendency to "lose connection", and you have no way to know if no new entries arrive because the connection has been lost or because no loggable events have taken place. You can't reload the page "to make sure that you're connected" either, because if you do that, you lose all history except the last 100 entries (which is "nothing"). To me, the fact that you don't know it does receive log entries or not, is one of the things that make the log viewer a very poor replacement for Frontail (the lack of history is the biggest one though).

_Originally posted by @Nadahar in https://github.com/openhab/openhab-core/issues/5165#issuecomment-3629606606_

Nadahar:
I just found another reason why the log viewer fails/disconnects: Reauthentication or possibly expired token. I let #3576 run overnight to see how it behaved, and it disconnected after ~ 1 hour with:

```bash
POST http://localhost:8081/rest/auth/token net::ERR_CONNECTION_REFUSED - request.js:406
Uncaught (in promise) 0 - auth-mixin.js:90
```

So, something makes the auth "expire" after a while, which is a big problem for something that you might want to leave running.

florian-h05:
> So, something makes the auth "expire" after a while, which is a big problem for something that you might want to leave running.

This quite common for OAuth access tokens: https://www.oauth.com/oauth2-servers/access-tokens/access-token-lifetime/

Nadahar:
> This quite common for OAuth access tokens

I'm not really into OAuth, but I read that there's such a thing as refresh tokens. They should make sure that this situation doesn't occur. Don't OH use those?

[...]

PR #3576, opened 2025-12-12
Log viewer: Handle heartbeat & Improve connection robustness

Nadahar:
This is meant to illustrate some of what I suggested in #3569. I don't consider it complete, both because I'm not sure if what I did in `ws.ts` is correct/complete (no close callback implemented events, for example) and because I'm not sure if the visual solution is ideal.

I haven't attempted to make any reconnect timer mechanism either. But, it does seem to function. As a plus, it turns out that it also solves a problem I didn't know existed: The "heartbeat timer" is never stopped if the remote closes the connection. With this PR it does, so it doesn't keep trying to send the heartbeat over and over again after the socket is closed.

## AI-generated summary

This PR introduces several improvements to the log viewer's WebSocket connection, enhancing its stability and providing better user feedback.

**Key Changes:**

- **Heartbeat Handling:** The WebSocket connection now correctly handles heartbeats. Previously, the heartbeat timer would continue indefinitely even after the remote connection was closed, leading to unnecessary attempts to send heartbeats. This is now resolved by clearing the keep-alive timer on connection closure.
- **Automatic Reconnection with Exponential Backoff:** The log viewer now attempts to automatically reconnect to the WebSocket server if the connection is unexpectedly closed or fails to establish initially. This reconnection mechanism includes an exponential backoff strategy, gradually increasing the delay between retries to prevent overwhelming the server.
- **Enhanced Visual Feedback:**
    - The "Continue receiving logs" button now features a "connecting-flash" animation to visually indicate when the log viewer is actively trying to establish a connection.
    - New badge colors and tooltips (`periodRangeColor`, `periodRangeTooltip`) have been added to provide clear visual cues about the WebSocket connection status (connected, disconnected, connecting, paused) in the log period display.
- **Improved Error Handling:** The heartbeat sending mechanism now includes a `try-catch` block to gracefully handle potential errors during heartbeat transmission.
- **Refactored Connection Management:** The `WebSocketService` in `ws.ts` now exposes a `closeCallback` to allow consumers to react to connection closures, facilitating the new reconnection logic in the log viewer.
- **Minor Fixes:** Addressed potential errors in `scrollToBottom` when the table body is not yet rendered.

These changes significantly improve the user experience of the log viewer by making it more resilient to connection interruptions and providing transparent feedback on its operational status.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/ws.ts
+++ b/bundles/org.openhab.ui/web/src/js/openhab/ws.ts
@@ -224,12 +224,14 @@ const WebSocketService = {
    * @param topics array of event topics to filter by, if empty all events are received
    * @param messageCallback message callback to handle incoming messages
    * @param readyCallback ready callback
+   * @param closeCallback close callback
    * @param errorCallback error callback
    */
   events(
     topics: string[],
     messageCallback: MessageCallback,
     readyCallback?: ReadyCallback,
+    closeCallback?: CloseCallback,
     errorCallback?: ErrorCallback
   ): KeepaliveWebSocket {
     let socket: KeepaliveWebSocket
@@ -257,7 +259,7 @@ const WebSocketService = {
       socket.send(heartbeatMessage(socket.id))
     }
 
-    socket = this.connect('/ws/events', extendedMessageCallback, heartbeatCallback, extendedReadyCallback, undefined, errorCallback)
+    socket = this.connect('/ws/events', extendedMessageCallback, heartbeatCallback, extendedReadyCallback, closeCallback, errorCallback)
 
     return socket
   },
--- a/bundles/org.openhab.ui/web/src/js/openhab/ws.ts
+++ b/bundles/org.openhab.ui/web/src/js/openhab/ws.ts
@@ -160,8 +160,11 @@ function newWSConnection(
   // Handle WebSocket connection closed
   socket.onclose = (event: CloseEvent) => {
     socket.clearKeepalive()
-    if (closeCallback) closeCallback(event)
-    else console.debug('WebSocket connection closed', event)
+    if (closeCallback) {
+      closeCallback(event)
+    } else {
+      console.debug('WebSocket connection closed', event)
+    }
   }
 
   // Handle WebSocket message received
--- a/bundles/org.openhab.ui/web/src/js/openhab/ws.ts
+++ b/bundles/org.openhab.ui/web/src/js/openhab/ws.ts
@@ -110,6 +110,7 @@ type CloseCallback = (event: CloseEvent) => void
  * @param path the path to connect to, e.g. `/ws`
  * @param messageCallback the callback to handle incoming messages
  * @param readyCallback the callback to handle the connection being ready
+ * @param closeCallback the callback to handle the connection being closed
  * @param errorCallback the callback to handle errors
  * @param heartbeatCallback the callback to handle heartbeats
  * @param heartbeatInterval the interval in seconds for sending heartbeats
@@ -118,6 +119,7 @@ function newWSConnection(
   path: string,
   messageCallback: MessageCallback,
   readyCallback: ReadyCallback | undefined,
+  closeCallback: CloseCallback | undefined,
   errorCallback: ErrorCallback | undefined,
   heartbeatCallback: HeartbeatCallback | undefined,
   heartbeatInterval: number
@@ -153,6 +155,13 @@ function newWSConnection(
   socket.onopen = (event: Event) => {
     socket.setKeepalive(heartbeatInterval)
     if (readyCallback) readyCallback(event)
+  }
+
+  // Handle WebSocket connection closed
+  socket.onclose = (event: CloseEvent) => {
+    socket.clearKeepalive()
+    if (closeCallback) closeCallback(event)
+    else console.debug('WebSocket connection closed', event)
   }
 
   // Handle WebSocket message received
@@ -193,6 +202,7 @@ const WebSocketService = {
    * @param messageCallback message callback to handle incoming messages
    * @param heartbeatCallback heartbeat callback
    * @param readyCallback ready callback
+   * @param closeCallback close callback
    * @param errorCallback error callback
    * @param heartbeatInterval heartbeat interval in seconds (defaults to 5)
    */
@@ -201,10 +211,11 @@ const WebSocketService = {
     messageCallback: MessageCallback,
     heartbeatCallback: HeartbeatCallback,
     readyCallback?: ReadyCallback,
+    closeCallback?: CloseCallback,
     errorCallback?: ErrorCallback,
     heartbeatInterval: number = 5
   ): KeepaliveWebSocket {
-    return newWSConnection(path, messageCallback, readyCallback, errorCallback, heartbeatCallback, heartbeatInterval)
+    return newWSConnection(path, messageCallback, readyCallback, closeCallback, errorCallback, heartbeatCallback, heartbeatInterval)
   },
   /**
    * Connect to the event WebSocket, which provides direct access to the EventBus.
@@ -246,7 +257,7 @@ const WebSocketService = {
       socket.send(heartbeatMessage(socket.id))
     }
 
-    socket = this.connect('/ws/events', extendedMessageCallback, heartbeatCallback, extendedReadyCallback, errorCallback)
+    socket = this.connect('/ws/events', extendedMessageCallback, heartbeatCallback, extendedReadyCallback, undefined, errorCallback)
 
     return socket
   },
--- a/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/log-viewer.vue
@@ -840,7 +840,10 @@ export default {
       const LINE_HEIGHT = 31
 
       const tableContainer = this.$refs.tableContainer
-      const tableBody = this.$refs.dataTable.firstChild
+      const tableBody = this.$refs.dataTable?.firstChild
+      if (!tableBody) {
+        return
+      }
       const filteredItemsCount = this.filteredTableData.length
       const currentIndexAtTop = Math.floor(tableContainer.scrollTop / LINE_HEIGHT)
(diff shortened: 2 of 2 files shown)


Nadahar:
@florian-h05 @jsjames Have any of you ever had the time to look at this? The heartbeat "handling" is pretty straight forward, what I'm not so sure about is how to show it visually. Also, the current solution is hardly readable if you use the "light, filled" theme like I do. When the background turns red, I can't really read what it says - and I certainly can't tell that it's red, because the color is so similar to the orange surrounding header.

I could/can rebase and resolve it, but there's not much point if nobody is going to look at it - and the "clue" anyway is to decide how to indicate if the log viewer is connected or not.

florian-h05:
I've checked the existing code and it does sent a heartbeat to the server.
So this PR is more about visualising the state of the connection than keeping it alive I think.

Anyway: Adding support for setting a close callback is very much appreciated! If you rebase, I will have a look at the CSS fixes. We have to coordinate this with #4011 though.

Nadahar:
> I've checked the existing code and it does sent a heartbeat to the server.
> So this PR is more about visualising the state of the connection than keeping it alive I think.

This is so old now that I don't remember the details, but I see from the diff that the sending of the heartbeat was already there, but with some flaws (the time left alive after close in particular). Which is why I called it "handle heartbeat".

> We have to coordinate this with #4011 though.

I don't know #4011, but perhaps the visualization could just be done there, and this could be about the "handling" (closing, logging etc.)?

As I've said before, I found it hard to find a "good way" to visualize it, part of the problem is that the current view of the logviewer is already quite bad if you use light and filled theme (especially when the badge is red, it's hardly readable, and it's not at all possible to see that it's red/something wrong, it "merges" with the orange fill).

I'll try to see if I can rebase/revive this, although it's completely gone from memory by now.

Nadahar:
@florian-h05 I've rebased it, but as said before, the visual solution is bad. I can't quite find a way to make the badges properly readable.

I also wanted to have a "reconnect feature" where it tried to reconnect if it lost connection. I just don't know quite where to put it, and how to handle the timer "safely", and what the "rules" should be - how often/many times should it try, how should this be indicated to the user etc.

As I see it, reconnecting is important here, since you can't acquire history more than the last 100 entries. That means that you must let the log viewer stay open if you want to use it to "capture a problem", and it must stay connected so that the log events are actually captured. While I hope that some of the fixes I did in core (to the websocket handling) has improved the stability of the connection, there could still be other reasons (like network connectivity) why the connection might be lost, and I think it would be desirable if it could reconnect.

Nadahar:
I've added what I think is needed to support close callbacks for events as well. I couldn't find more than one place where `events()` is used, for the "debug stream events" - and that call doesn't specify any parameters, so no changes seem to be needed.

Is it correct that WS events haven't yet been taken into much use?

Nadahar:
@jsjames We might as well discuss this here, where I'm trying to deal with the behavior, not the layout.

> In porting to composition api and typescript, the scrolling functions were kept as is and I expect that the same bug/issue would exist in the original log-viewer. Having said that, a new bug could have been introduced - so if you see something please flag it.
>
> Longer term, I think it would be good to use this https://vueuse.org/core/useVirtualList/ out of the vueUse library. It handles all the virtual scrolling mechanism behind the scene, so I would anticipate it would be more robust than the current implementation.

I'm not sure, but I think the scroll has always been a bit off. You never really know if "autoscroll" is enabled or not, and sometimes, if autoscroll is enabled, and the stars are aligned (I have no idea what this factor is), it will "jitter" a pixel or two up and down repeatedly.

JavaScript isn't really my thing, and DOM manipulation certainly isn't my thing, so I'm having somewhat of a hard time to get all the nuances here (also, it's not my focus, I'm trying to make reconnecting work reliably), but I think using a ready-made solution for the autoscroll sounds like a good idea. This is difficult to get right across browsers/devices anyway, a dedicates solution is more likely to keep up with those nuances.

There is one bug in the log viewer that I suspect can be realated to the F7 upgrade, and that is that it looks like the top of the scrollable area extends under the top navbar. So, you can't actually see the top lines. I'm not sure if this is permanent or only in some situations, but I've seen it happening. AFAICR there was some manual padding to account for the header that has probably "falled out"/stopped working? That might be an easy fix.

Some of the bugs that I see isn't directly related to the scrolling though, I'm not sure exactly what happens, but sometimes when it tries to add new log entries, the DOM manipulation fails because it tries to modify `$ref`s that aren't set/are null.

The whole setup with dual view (table + text) kind of confuses me, it might be that this bug only occurs if you use "text view" (I exclusively do). Also, the bug is only visible in the browser console, but since what it's trying to do is aborted, one should think that is has some consequence.

Nadahar:
I know nothing of all these components, but it seems like there's no autoscroll and already similar jitter issues to what we have with "useVirtualList". Maybe [@tanstack/vue-virtual](https://tanstack.com/virtual/v3/docs/framework/vue/vue-virtual) is a better choice?

Nadahar:
@florian-h05 @jsjames There is something really weird going on with the dev server, which might explain why some things behave differently there:

bilde

Sometimes when I update the code, one or more "old version" if the code is still running in addition to the new one. See how the same log message is logged 3 times, from "different versions" of the code. It sometimes leads to strange behavior, where both the new and the old behavior is applied, usually leading to errors.

Nadahar:
I've made a couple of fixes and changed the visualization again. I've also created a reconnect logic that seems to work reasonably well from my testing.

Reconnecting starts with a 1-second retry delay, which increases slowly up to 5 minutes. I haven't made any final "give up" limit, so it will continue to try to reconnect every 5 minutes "forever". This can be stopped either by clicking the "Stop" button, and by regaining the connection.

I made a couple of GIF's showing the (re)connect animation yesterday:

Connecting
ConnectingDark

In addition to this, the connection state is reflected by the color of the "log range" badge, where green is connected, yellow is paused and red is disconnected. The tooltips of this badge also include the connection status.

I'll try to make a GIF that shows all of it.

Nadahar:
Here is the combined GIF:

LogViewerConnection

I think I've done all I set out to do, and then some. I included a fix that allows the log viewer to work without Karaf (because Karaf is needed to set/get the loggers), so that the log viewer also works when OH is run without Karaf (makes debugging/testing much easier). I also prevented the error that occurred most frequently in the table redrawing logic, but there are still errors from time to time in that code. I didn't intend to try to figure that out, so it should still ideally be fixed.

I think this makes this PR ready for review.

florian-h05:
> Is it correct that WS events haven't yet been taken into much use?

Correct, they haven't as we were waiting for openHAB Cloud to support WS proxying.

> Sometimes when I update the code, one or more "old version" if the code is still running in addition to the new one. See how the same log message is logged 3 times, from "different versions" of the code. It sometimes leads to strange behavior, where both the new and the old behavior is applied, usually leading to errors.

Webpack dev server also has shown that behaviour sometimes, depending on what code you work on a (hard) reload after a change seemed to be required in my experience.

florian-h05:
> Reconnecting starts with a 1-second retry delay, which increases slowly up to 5 minutes. I haven't made any final "give up" limit, so it will continue to try to reconnect every 5 minutes "forever". This can be stopped either by clicking the "Stop" button, and by regaining the connection.

The SSE event stream uses 1, 2, 4, 8, 10, 10, ... seconds, it does never increase above 10 seconds. I think that's a reasonable upper limit, 5 minutes is quite a long time IMO. WDYT?

Nadahar:
> The SSE event stream uses 1, 2, 4, 8, 10, 10, ... seconds, it does never increase above 10 seconds. I think that's a reasonable upper limit, 5 minutes is quite a long time IMO. WDYT?

I can very easily adjust it, the retry increases relatively slowly, so it takes a while for it to become very infrequent, but I can easily cap it at 10 seconds instead. There are no "right answers" with these things, I'm imagining an unattended system that lose connection because of a network outage or similar, and in that case, I didn't want to "stress the system" with too frequent retries. But yeah, 5 minutes is very conservative I guess.

Nadahar:
Please note that I've darkened the colors used on the badges (in the log viewer header only) slightly. They should have the same "tone" as the originals, and I don't think anybody will notice a difference unless they see them next to each other, but since the text in the badges is white, I think that this slight darkening improves readability significantly. As they were, they were hard to read at times.

florian-h05 (approved):
LGTM, many thanks!

I've pushed a minor formatting/code style change.

Nadahar:
@florian-h05 Btw., I've started looking at implementing the rule format conversion in the UI (from the model of Things and Items). I figured out that I need to regenerate the "REST API classes", but I am confused about exactly how to get the "correct" `*.spec` file. It also doesn't seem like this has been done since the sitemap API was modified.

Do you have any info about how this "should" be done?

florian-h05:
You need a current snapshot install, you can then navigate to `/rest/spec`, copy the JSON, paste it into openhab-api-spec.json (I then usually let IntelliJ format the JSON, but ideally we let oxfmt handle that), then `npm run generate-rest-client`.
For other Java applications that aren't that dynamic as openHAB (e.g. Quarkus apps), you can often already build an OpenAPI spec at build time, but for openHAB it is only available at runtime.

Nadahar:
> You need a current snapshot install,

I can't use that, because I need the changes I've made for rules to be in there. But, I've fetched it from a running instance of "my branch". The problem is that there are so many, seemingly unrelated, changes.

florian-h05:
> The problem is that there are so many, seemingly unrelated, changes.

In that case it would be good to update the main branch according to latest snapshot, that should reduce the diff I think. I'll handle that. You can then check if the diff is smaller.

florian-h05:
Unfortunately, recent changes have introduced a few bugs in the OpenAPI spec -- duplicate schema names:

- https://github.com/search?q=repo%3Aopenhab%2Fopenhab-core+%40Schema%28name+%3D+%22Condition%22%29&type=code
- https://github.com/search?q=repo%3Aopenhab%2Fopenhab-core+%40Schema%28name+%3D+%22Rule%22%29&type=code

Will comment that in the causing PR: https://github.com/openhab/openhab-core/pull/5459

Nadahar:
> Unfortunately, recent changes have introduced a few bugs in the OpenAPI spec -- duplicate schema names:

Ah, that might explain some of my confusion...

PR #3578, opened 2025-12-12
Analyzer/Charts: Fix week starting on Sunday starts on Monday

florian-h05:
The `week` chart type should range from Sunday to Sunday, but it has been rendered similarly to `isoWeek` from Monday to Monday.

This fixes that behaviour by properly calculating the startTime and removing the alignment logic from oh-time-axis, as startTime and endTime are already properly aligned by the chart-mixin.


florian-h05:
@jsjames Can you please have a look and confirm my thought process that we don't need to perform special alignment of min/max for oh-time-series as the chart-mixin provides already aligned startTime and endTime?

jsjames:
> @jsjames Can you please have a look and confirm my thought process that we don't need to perform special alignment of min/max for oh-time-series as the chart-mixin provides already aligned startTime and endTime?

Changes look good, i did submit a PR to your branch to make the week starting on sun vs. mon in the ui.

Issue #3204, opened 2025-05-28
oh-knob add responsiveness


mannemiethe:
## The problem

in the old version of oh-knob responsiveness was implemented and it was possible to set it with responsive = true,
After switching from the old component to vue-roundsliderui, the oh-knob responsiveness was not implemented.
## Your suggestion

Please add responsiveness. You can implement it using the resize event. See:
https://codesandbox.io/p/sandbox/vue-round-slider-responsiveness-4lbipj
https://vue.roundsliderui.com/#comments
## Your environment

openHAB 4.3.5
Release Build
Main UI Commit b1cd4467
```yaml

```

PR #3586, opened 2025-12-14
add responsiveness to oh-knob

mannemiethe:
to fix the issue: https://github.com/openhab/openhab-webui/issues/3204

ResponsiveOhKnob

## Add responsive sizing to oh-knob widget

This PR adds responsive sizing capabilities to the `oh-knob` widget, allowing it to automatically adjust its size based on its container dimensions.

### Changes
- Added `responsive` configuration option to enable dynamic sizing
- Implemented ResizeObserver to detect parent container size changes
- Added window resize and gridster-resized event listeners for layout changes
- When `responsive` is enabled and `size` is set (1-100 or "100%"), the knob scales proportionally to fit its container
- Falls back to measured pixel dimensions when no size is specified
- Properly cleans up listeners and observers on component destruction

### Usage
Set `responsive: true` in the widget configuration to enable automatic sizing. Optionally specify `size` as a percentage (1-100 or "100%") to control the knob's size relative to its container.

### Implementation
- Follows the same responsive pattern used in `oh-grid-item` and other layout widgets
- Compatible with grid layouts, canvas layouts, and standard containers
- Maintains backward compatibility - existing configurations work unchanged


florian-h05:
@mannemiethe Thank you for your PR. Due to the upcoming 5.1.0 release we are no in a feature freeze, meaning no new features can be merged before the 5.1.0 release is done. Your PR will be reviewed some time after the release.

mannemiethe:
@florian-h05: I understand. No Problem but can review it and merge it after the release?

florian-h05:
After the release we can review and merge again 👍

mannemiethe:
who can review and merge this PR?

jsjames:
@mannemiethe - I started to look at this already - but need a few days. The better place for handling resizing is at the the css level - where you could define the width/height and then the browser would handle everything for you - however the base round-slider component does not support this.

The round-slider component has a few other limitations regarding some of the the theme support features (#3738) where it overrides (for some reason) any defined css color class with defaults. I'm looking at how intrusive it will be to modify the component to support classes better and in doing so, there might be a good way to use css size properties to automatically support responsiveness.

Lastly, the round-slider is the only component requiring jquery in the UI. While jquery is not huge, it is not insignificant in size. I have an initial port of the round-slider component using dom7 which is working. Dom7 is what all the f7 components are built with and is syntactically similar to jquery, but there are a few deltas which i'm working through.

I'll hope to have something for testing in the next few days. Assuming this goes well, the responsiveness can be done at the browser level.

mannemiethe:
i mean until you can deliver this new component, it would be nice if this feature can be used.

mannemiethe:
@jsjames You cannot doit with css only you also need javascript for the knob. However i think it makes sense to first merge this PR until you have a new solution to remove jquery.

Issue #3583, opened 2025-12-14
(non) formatted values in charts


mstormi:
Apache echarts uses bare unformatted numbers to display in markPoints, legend and so on.
It should be using the item's display pattern (or rather the <item as Number>'s display pattern - UoM units may or may not mess up the display).

While this can be worked around using label: and formatter: that is very annoying and does not work everywhere.

Also it's using the default locale so e.g. a value that reads 1234,56 in DE locale will be displayed as 1,234.56.

jsjames:
Regarding the 2 items - these are new enhancements to be considered in the UI as neither has been implemented.

- localization of numbers. Unfortunately, echarts does not natively support localized number format. However, I have added a PR (#3587) which adds this capability with a custom number formatter.

- using displayState in echarts. This one needs more thought as the stateDescription is applied in the core and not the UI.  We'd likely have to read the stateDescription metadata and then implement similar formatting in the UI that is in the core.  @florian-h05, any additional houghts here?

florian-h05:
> using displayState in echarts. This one needs more thought as the stateDescription is applied in the core and not the UI. We'd likely have to read the stateDescription metadata and then implement similar formatting in the UI that is in the core. 

I think there was an issue about this, but I can't find it anymore. EDIT: https://github.com/openhab/openhab-core/issues/4697
You are right that this needs to be carefully thought about, and I am against trying to replicate the formatting core uses. AFAIK this is based on Java Formatter, and I neither want to reimplement nor maintain a port of the Java formatter and the UoM formatter in Main UI. That being said, if the REST API would optionally provide the state formatted, we could solve the issue way easier.

jsjames:
Given we addressed the localized formatting part of this issue in #3587 and the displayState is covered in #4697, I will close this issue out.

@mstormi, it would be good for you to test the localized formatting when you get a chance.

PR #3587, opened 2025-12-14
Charts: Add missing number format localization 

jsjames:
Refs #3583.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/misc/oh-chart-tooltip.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/misc/oh-chart-tooltip.js
@@ -16,6 +16,9 @@ export default {
     if (options.smartFormatter) {
       options.formatter = (params, ticket, callback) => {
         let tooltip = ''
+        // special tooltip for mark area:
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/components/widgets/system/oh-chart-component.vue
```diff
@@ -86,7 +86,9 @@ export default {
     let initOptions = echartsLocale ? {
       locale: echartsLocale
     } : null
-    return { echartsLocale, initOptions }
+
+    let numberFormatter = new Intl.NumberFormat(useRuntimeStore().locale)
```
florian-h05: Any reason to put this in oh-chart-component?
I'd prefer having chart-mixin not depend on data of its "users", but have it self-contained.

florian-h05:
Also, please always add reference to the issues in the PR description.

florian-h05 (approved):
LGTM, thanks! I have added using the numberFormatter when smartFormatter is enabled for oh-chart-tooltip.
I think it's reasonable to label this as bug fix so we can merge it now.

PR #3608, opened 2025-12-16
Fix back button issue

digitaldan:
When loading "oh.foo.com" we route in F7 to "oh.foo.com/overview/" but in some cases this leaves the f7 route history empty, and the back button will have nothing to go back to after the first navigation item is clicked.


florian-h05:
Does this fix https://github.com/openhab/openhab-webui/discussions/3381#discussioncomment-15150670?

digitaldan:
Correct.

PR #3624, opened 2025-12-20
[blockly] Fix oh_context_info block "as {type}" not working

florian-h05:
Regression from #3376.
Also simplifies the generated code (the members of the event object are now JS strings).


florian-h05:
@kaikreuzer @jsjames We should merge this PR in the code freeze as it fixes a critical blockly regression.

stefan-hoehn:
A pity we both overlooked the issue in the original PR. 😢
LGTM now.

PR #3625, opened 2025-12-20
Only show the clipboard copy confirmation when necessary

jimtng:
This is a regression from #3350, originally working in #3309


jimtng:
@openhab/webui-maintainers please merge this before 5.1 release

PR #3627, opened 2025-12-20
Charts: Fix oh-time-axis doesn't adjust begin/end for series with offset

florian-h05:
Regression from #3578.
Reported on the community.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/axis/oh-time-axis.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/axis/oh-time-axis.js
@@ -20,7 +20,10 @@ export default {
       axis.max = (v) => {
         if (isNaN(v.min)) return endTime.toDate().getTime()
         if (!isFinite(v.max)) return v.max
-        return startOf(chartType, v.min).add(1, chartType === 'isoWeek' ? 'week' : chartType).toDate().getTime()
+        return startOf(chartType, v.min)
+          .add(1, chartType === 'isoWeek' ? 'week' : chartType)
+          .toDate()
+          .getTime()
       }
     } else {
       axis.min = startTime.toDate().getTime()
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/chart-mixin.js
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/chart-mixin.js
@@ -56,9 +56,7 @@ export default {
     const config = this.context.component.config || {}
     const chartType = config.chartType
     const future = config.future === true ? 1 : (config.future ?? 0)
-    const endTime = (chartType)
-      ? this.addOrSubtractPeriod(startOf(chartType), 1 + future)
-      : this.addOrSubtractPeriod(dayjs(), future)
+    const endTime = chartType ? this.addOrSubtractPeriod(startOf(chartType), 1 + future) : this.addOrSubtractPeriod(dayjs(), future)
 
     return {
       speriod: config.period || DEFAULT_PERIOD,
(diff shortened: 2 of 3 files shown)


jsjames:
I can merge this to fix for time-series, but I think the issue will still persist for state-series.  Instead of relying on the timestamps in the data, we should calculate the min max based on the value of offset.

Let me know if you want to me to merge for the release.

florian-h05:
> I think the issue will still persist for state-series.

What do you mean?

jsjames:
> > I think the issue will still persist for state-series.
> 
> What do you mean?

I believe offsets of time won’t work with state-series if we rely on the callback functions to determine the start and stop since echarts does not understand the timestamp for a custom series. I haven’t tried this yet. I think a better soln would be to just calculate the min and max vs using the callback.

florian-h05:
I've also though of better solutions, though that would likely require some refactoring (which isn't bad, we can convert to TypeScript then).

florian-h05:
Since offset requires an additional time axis to be displayed, what about moving offset configuration to additionional axis? I think it makes more sense to have axis with offset and assigning series to them, instead of setting offset for series and having to add axis.

@ghys Please also chime in.

ghys:
I may be wrong here (it's been a while), but I think day.js can handle the day of the week by itself and it may be what's happening here (why the `week` period is not necessarily to be interpreted as "week starting on Sundays").

During the initialization of the app when the configured locale is retrieved from the server it seems to be passed over to day.js here:

https://github.com/openhab/openhab-webui/blob/2589928667e84ce187aade79a75d5c5faf7ebb7a/bundles/org.openhab.ui/web/src/App.vue#L836-L839

Then once the locale is set you can access things like the first day of the week in the local culture (https://day.js.org/docs/en/i18n/locale-data); including those where the week starts on Saturdays (https://www.reddit.com/r/MapPorn/comments/1283i3a/map_of_what_day_the_week_starts_in_different/)

I suspect the `week` period then becomes locally-aware, not necessarily "starting on Sundays" (that was before the locale was "injected").

Since you can combine English strings if you prefer to have the UI in English, with for example German culture (which includes the local first day of the week) by setting your locale to `en_DE` - and I believe most people do that (or should) - then the weekly charts should just use the locally-aware `week` period, and the `isoWeek` and other parameters should probably eventually be dropped. It would also simplify the UI.

jsjames:
@florian-h05, I reviewed and it looks good and restores the existing behavior. Since this is in draft, i'll let you change and can merge.

[EDIT] There are still issues even with the current implementation, for instance, if you have a time chart with fixed period of one month, and one of the series is only offset by 1 day, the chart is off.

Also, in testing, I don't think this is a big deal for the state-series as this will typically be on a different time-axis.

florian-h05:
> I may be wrong here (it's been a while), but I think day.js can handle the day of the week by itself and it may be what's happening here (why the week period is not necessarily to be interpreted as "week starting on Sundays").

I think that makes sense. The question is, do we want to force the begin of the week to be on Sunday or rename it to "Week (starting on Sunday/Monday depending on your locale)"? The latter would save us some work.

> isoWeek and other parameters should probably eventually be dropped. It would also simplify the UI.

isoWeek might still be useful if you want to foce the week to begin on Monday.

dujal:
May I ask if there is a plan to fix the regression at least for timeseries so it is included in one of future bugfix (5.1.x) releases, as you mentioned here https://community.openhab.org/t/openhab-5-1-release-discussion/167620/338 ?

My comparison charts (with offset) are still broken, so I'm really looking forward to this fix :)

Thanks

florian-h05:
Let’s merge this to fix the regression and take care about getting it working for oh-state-series.

jsjames:
> Let’s merge this to fix the regression and take care about getting it working for oh-state-series.

I haven't looked at this for awhile - but from my comment above, i don't think there's an issue with oh-state-series.

However, there is this issue:

> There are still issues even with the current implementation, for instance, if you have a time chart with fixed period of one month, and one of the series is only offset by 1 day, the chart is off.

Let's merge as it is so that at least work better.

jsjames:
@florian-h05 - needs to be rebased. I did it, but not sure best way to submit??

florian-h05:
@jsjames Normally the only way to submit a rebase is to force push. I just rebased through GitHub web.

PR #3637, opened 2025-12-21
Script/Code editor: Fix search panel closes on click

### LINKED ISSUE #3597: Editor search panel closes on click
In any of the editor panes (widgets, rules, configs) ctrl + f now brings up a search and replace panel at the bottom of the screen (which is cool). However, clicking anywhere on that panel closes it again instead of interacting with the elements on the panel. For example, you cannot select the regex option by clicking on it. The panel works for all of its functions if you navigate by tabbing, but not clicking.

(Build #4991)

Also related (though I don't know if this is a bug or not), ctrl + f is completely captured by this function. In many websites with custom search boxes like this a second ctrl + f will close the custom box and bring up the browser search function. That is not true with this search box.

florian-h05:
Fixes #3597.

This was caused by the close button being 100% width and overlaying other buttons.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
@@ -66,6 +66,9 @@
     // This affects the buttons in the search form
     .cm-button
       width auto
+    // Display the close button in the search form in upper right corner, not full width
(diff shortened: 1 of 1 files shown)


florian-h05:
@jsjames Please review and merge this for 5.1 (RC3 build this evening!!).

PR #3640, opened 2025-12-21
Add-on details: Fix invalid back navigation

florian-h05:


florian-h05:
Self-merging to make sure this is merged before the 20 PM CET deadline for RC3.

PR #3595, opened 2025-12-15
Fix stylesheet not working for layout grid items

florian-h05:
Also fixes #1425.

PR #3641, opened 2025-12-22
Fix stylesheet not applied for some widget components

florian-h05:
Regression from #3595.


florian-h05:
@jsjames Sorry, for self-merging, but we have to make sure this gets into the 5.1.0 release build and I don't know whether you are available before 3 pm CET.

Issue #3572, opened 2025-12-11
[MainUI] Group is lost when creating Point/Equipment from Thing channel


dujal:
## The problem
I use persistence configuration based on groups, so most of the time I create new Item from any Thing  by "Add Equipment to Model" or "Add Points to Model" I need to put the Item to additional persistence group Item.

I select the parent group in the wizzard dialog, I can see the Group is added to parent groups, but when I confirm the wizzard, Items are created without that Group (only with semantic parent Equipment Group). 

See the following video:

![Image](https://github.com/user-attachments/assets/553d1191-d570-44ea-84e7-aa15968b711f)

## Steps to reproduce

1. open existing Thing and switch to channel tab
2. click Add Equipment to Model or Add Points to Model
3. select one channel and add some Group to that channel 
4. Save the wizzard form
5. Check the created item has the parent Group

## Your environment

This problem is not related to any recent changes, it is there from OH 4.x.

## Browser console

```txt

```

florian-h05:
@mherwege Do you want to have a look?

mherwege:
> [@mherwege](https://github.com/mherwege) Do you want to have a look?

Yes, but towards the second half of the week most likely.

dujal:
@mherwege I'm going to try to create my first PR to openHAB repo for this bug.
I've already fixed the problem locally, so if you still have not worked on this, then please let it for me, thanks.

mherwege:
I didn’t get very far yet, so please provide a PR. We always welcome new contributors.

mherwege:
@dujal How is the progress on making a PR? I do want you to create one if possible, but I have a fix as well.

@florian-h05 I prefer letting @dujal create a PR for a fix, but if you want to get this in release, I can push a fix.

dujal:
@mherwege If you have the fix prepared already, then feel free to create the PR. I will have not much time to do so today.
I'll find something else to contribute.

PR #3642, opened 2025-12-22
Fix groups not saved when adding points or equipment to model

mherwege:
Fixes https://github.com/openhab/openhab-webui/issues/3572


florian-h05:
@mherwege Do you think we should squeeze this into 5.1.0?

mherwege:
Keep it for a first patch release I would say. I don't see much risk, but I would prefer to be on the safe side. There is a workaround: editing the item after creation, a bit more work but feasible.
It is up to you.

florian-h05:
I also tend to stay on the safe side, so this will go into the first patch release :+1:

florian-h05 (approved):
LGTM, thanks!

Issue #1159, opened 2021-09-08
Allow the "Show non-semantic" to be permanent


TBail:
## The problem
The checkmark "Show non-semantic" is reset everytime the page model is opened. It would be nice if there waere the option to show non-semantics on a permanent basis. This

## Your suggestion
There are two option to realize the desired behaviour. Firt: store the last state of the check mark and retrieve it on page reload. Second create a new area in the settings like "Main UI" and create a flag that enables the permanent display of non-semantics

## Your environment
```yaml
runtimeInfo:
  version: 3.2.0
  buildString: "Build #2485"
locale: de-DE
systemInfo:
  configFolder: /etc/openhab
  userdataFolder: /var/lib/openhab
  logFolder: /var/log/openhab
  javaVersion: 11.0.12
  javaVendor: Azul Systems, Inc.
  javaVendorVersion: Zulu11.50+19-CA
  osName: Linux
  osVersion: 5.10.0-8-amd64
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 245799304
  totalMemory: 407896064
bindings:
  - astro
  - automower
  - homematic
  - http
  - icalendar
  - mqtt
  - netatmo
  - tankerkoenig
  - unifi
  - wled
  - zwave
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: true
    cordova: false
    phonegap: false
    electron: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
    pixelRatio: 1.75
    prefersColorScheme: light
  isSecureContext: false
  locationbarVisible: true
  menubarVisible: true
  navigator:
    cookieEnabled: true
    deviceMemory: N/A
    hardwareConcurrency: 8
    language: de
    languages:
      - de
      - de-DE
      - en
      - en-GB
      - en-US
    onLine: true
    platform: Win32
  screen:
    width: 1564
    height: 1043
    colorDepth: 24
  support:
    touch: false
    pointerEvents: true
    observer: true
    passiveListener: true
    gestures: false
    intersectionObserver: true
  themeOptions:
    dark: light
    filled: true
    pageTransitionAnimation: default
    bars: filled
    homeNavbar: default
    homeBackground: default
    expandableCardAnimation: default
[...]

PR #3645, opened 2025-12-22
Keep model picker and sitemap editor visualisation toggle states across sessions

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/1159

In the model and model picker, persist:

- Show name
- Show tags
- Show non semantic

between sessions.

For the sitemap editor, equally store 'Show item name' between sessions.

Note I did not put the 'expanded' flag for the model and model picker in the local storage, so this will not be persisted between sessions.


florian-h05:
Backported because #3723 needs this or we have to resolve comflicts while cherry-picking.

PR #3650, opened 2025-12-23
Fix searchbar in binding selection when adding Thing

jimtng:


florian-h05:
Let's backport this as well.

jimtng:
> Let's backport this as well.

yes

florian-h05 (approved):

PR #3662, opened 2025-12-26
Fix Main UI load fails if implicit user role is disabled

florian-h05:
If the implicit use role setting is disabled (API security setting), Main UI load fails because it doesn't authorize early enough and the first REST requests fail.

Reported on the community:
https://community.openhab.org/t/5-1-0-m3-main-ui-shows-blank-screen-on-android/167432/49

### DIFF
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -713,9 +713,8 @@ export default {
           return request.json('/rest/')
         })
         .catch((err) => {
-          console.error('openHAB REST API connection failed with error:')
-          console.info(err)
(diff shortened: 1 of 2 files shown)


jsjames:
we should backport this one, correct?

florian-h05:
Yes, I've added it to the candidate list.
Let's test if a few days in the snapshots though, before backporting it.

jsjames (approved):

florian-h05:
Fix has been confirmed by reporting user: https://community.openhab.org/t/5-1-0-m3-main-ui-shows-blank-screen-on-android/167432/53?u=florian-h05

PR #3673, opened 2025-12-27
Thing page: Fix Z-Wave network viewer action, thing action caching, action config groups

### LINKED ISSUE #3671: Specific actions for Z-Wave controller disappeared from Main UI in OH 5.1
When opening the thing page of the Z-Wave controller thing in OH 5.0.3, we can all these actions:

Image

In OH 5.1, they all disappeared:

Image

### LINKED ISSUE #3672: Z-Wave network map is broken in OH 5.1
While it is still working in OH 5.0.3, the Main UI feature to show the Z-Wave network map is broken in OH 5.1.

digitaldan:
Fixes #3672
Fixes #3671

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -170,7 +170,7 @@
             <config-sheet
               ref="thingConfiguration"
               :parameter-groups="configDescriptions.parameterGroups"
-              :parameters="configDescriptions.parameters"
+              :parameters="filteredConfigParameters"
               :configuration="thing.configuration"
               :status="configStatusInfo"
               :set-empty-config-as-null="true"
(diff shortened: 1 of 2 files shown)


digitaldan:
This also fixes action "caching" when viewing things with no actions, and restores rendering action groups defined as configuration groups with context/name/lables of "actions" (pre thing actions)

jsjames:
Thanks Dan - this fixes #3671 as well, correct?

digitaldan:
it does, i would have broken it out as different Pr's, but the change is small and it was much easier in this PR since i had it open and was testing with it.

jsjames:
> it does, i would have broken it out as different Pr's, but the change is small and it was much easier in this PR since i had it open and was testing with it.

no problem - just wanted to mark it as such in the comment.

This looks good to me, but I'd like @florian-h05 to review as well since he is more familiar with this code.

bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
```diff
@@ -536,6 +559,9 @@ export default {
       this.codeDirty = codeDirty
     },
     load () {
+      this.ready = false
```
florian-h05, line 562: Why reset ready state here?

digitaldan: This is part of the fix so that actions from one thing do not persist when navigating to another thing since the component is reused.  By setting false here,  we show the loading placeholders and then when the data loads for the new thing, we set this back to true and show the right layout.

florian-h05 (approved):

lolodomo:
Thank you @digitaldan , I will test the fix when availability in a snapshot.

lolodomo:
> Thank you @digitaldan , I will test the fix when availability in a snapshot.

Snapshot 5049 just installed and problems are confirmed to be resolved. Bravo.

PR #3677, opened 2025-12-28
`oh-*-card` widgets: Fix style applied to both oh-card and child component

florian-h05:
- `oh-*-card`: The `style` config should only be applied to the parent/root element, but no children.
- Widget components: Do not v-bind `$attrs` to child element(s):
  `$attrs` automatically binds to the root element of the component, so do not v-bind to children.
  This for example applies style to the components and all its children, which isn't correct.
- `oh-block`: Do not v-bind `$attrs` to child element(s) & Remove wrapping div as f7-block already is a div.

Reported on the community: https://community.openhab.org/t/openhab-5-1-release-discussion/167620/247


florian-h05:
Looks good with both my setup as well as the demo server.

florian-h05:
@jsjames Please assign the milestone when merging

florian-h05:
> I presume this address the issues Mark_VG was having?

No, this fixes the issues reported here: https://community.openhab.org/t/openhab-5-1-release-discussion/167620/247

Issue #3676, opened 2025-12-28
Adaptations for iOS 26 mobile view


TAKeanice:
## The problem

The MainUI does not render too well on iOS 26. Let me show the screenshot of Safari on iOS first, then explain what I think could be improved. This ticket is based on the discussion over at openhab-ios: https://github.com/openhab/openhab-ios/discussions/1019#discussioncomment-15353164

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/49223b62-8e85-4659-b628-1a6ddfd5194a" />

* The tabbar looks out of place with the liquid glass design.
* Scrolling the content does not trigger minimizing the Safari´s address bar or, for an embedded web view, a potential toolbar or tabbar. Thus the screen real estate usage is suboptimal
* Scrolling the content does not hide / minimize the back button or navigationbar with the sandwich menu, thus wasting space on top

## Your suggestion

* To fix the tabbar consistency with liquid glass, https://wildfirestudios.ca/blog/redesigning-a-web-apps-components-for-ios-26-and-liquid-glass could maybe serve as an inspiration for a design adaptation
* Ideally, the tabbar of the MainUI could avoid a minimized app tabbar when in an embedded webview to enable the iOS app redesign. Scrolling the content should minimize the browser address bar / app tabbar or maybe hide the MainUI tabbar. openhab.org does this.
* The top bar should make way for content when scrolling down, either by hiding completely (though that is rather unusual on iOS) or rendering the back button or sandwich menu button as a circle. For inspiration, I added a screenshot of Github mobile below.

<img width="585" height="1266" alt="Image" src="https://github.com/user-attachments/assets/a700aca1-2ef2-49b3-a191-d9ef668619e3" /> 

<img width="585" height="1266" alt="Image" src="https://github.com/user-attachments/assets/43623ea3-e82c-458d-9bb5-159b452852fd" />

## Your environment

iOS26.x , Safari or native iOS app
```yaml

```

PR #3678, opened 2025-12-28
Hide nav bar background when nav bar is hidden

digitaldan:
This is related to #3676 , but this should be useful for all platforms i think (so not limiting to IOS)

This allows the use of the whole screen for viewing, while still keeping "safe" areas for the menu bar when visible. 

Apologies for the ugly red page, but was helpful for testing.

When menu shows

image

When scrolling

image

How it looked before this change when scrolling

image

Personally i think this is a much better experience.

### DIFF
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -218,8 +218,15 @@ html
   --f7-navbar-large-collapse-progress 1
   opacity 0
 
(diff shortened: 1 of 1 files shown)


florian-h05:
The navbar previously also became visible again, so this is unrelated here.
I’ll merge this then.

digitaldan:
Hmm, this is now not working for me, ~~let me look into why, my hunch is that this may have been co-pilot feedback, which i  tested, but maybe the browser was caching it.~~

Err nevermind, i did not co-pilot review this one.

PR #3681, opened 2025-12-29
Charts/Analyzer: Fix rotation issue

jsjames:
Fixes #3682.
Reported on the community: https://community.openhab.org/t/echarts-js-error-on-rotate/167700.
Regression from #3587.

PR #3683, opened 2025-12-29
chart-mixin: Follow-up for #3681

florian-h05:
Refs https://github.com/openhab/openhab-webui/pull/3681/changes#r2650773805.


jsjames (approved):
Good change - thanks!!

PR #3691, opened 2025-12-29
Map picker: Round lat/lng to 6 digits

### LINKED ISSUE #3689: [MainUI] Map picker should fill 4, at most 5, digits after the decimal comma for latitude and longitude
Under Settings → Regional Settings → Location when I click on the map, the field is filled with latitude/longitude containing up to 15 digits after the decimal comma.

Four digits after the decimal comma correspond to 10 meters precision, this is shown in the table at https://en.wikipedia.org/wiki/Decimal_degrees#Precision.  Five digits after the comma correspond to 1.1m precision and no device can measure with such precision.

[XKCD 2170: "Coordinate Precision"](https://xkcd.com/2170/)
---
Using OH 5.2.0 build 5050.

jsjames:
Resolves #3689

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/location-picker.vue
@@ -88,6 +88,8 @@ export default {
   methods: {
     mapClicked (evt) {
       this.marker = latLng(evt.latlng)
+      this.marker.lat = Math.round(this.marker.lat, 6)
+      this.marker.lng = Math.round(this.marker.lng, 6)
       this.$emit('input', this.marker)
     }
   }


florian-h05:
Just FYI GitHub doesn’t seem to recognise implements as a valid keyword before the issue number and so the issue won’t be closed automatically when merging.
Closes, Fixes, Resolves work.

florian-h05 (approved):

dilyanpalauzov:
[`Math.round()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round) accepts only one parameter.  That is, the second parameter is ignored and the result is integer.

florian-h05:
See #3692.

jsjames:
Thanks for fixing. I had to go back and see why this was working for me.

Long story short - i originally just used round which pulled in the lodash round implementation. However, there was a lint issue and I fixed (incorrectly) by assuming it was the Math round and didn't retest. Sorry!

PR #3697, opened 2026-01-01
oh-map/oh-plan: Fix Leaflet not rendering in prod build

florian-h05:
Regression from #3678.


florian-h05:
Can you please backport?

jsjames:
> Can you please backport?

yes - and thanks for the labeling as 5.2 as well!!

PR #3704, opened 2026-01-03
Thing details: Fix code tab emptied on save

### LINKED ISSUE #3699: Code tab empty after hit save button
While testing another issue i found this in build 5051 that was not there in 5.1.0 GA

The code tab when editing a thing definition empty after hit save button

https://github.com/user-attachments/assets/e05bbb61-58c6-47b7-800c-3b1830366f44

florian-h05:
Fixes #3699.
Regression from #3673.


florian-h05:
This should be backported.

PR #3708, opened 2026-01-03
oh-list-item: Merge `component.slots.after[0]` & `$slots.after`

florian-h05:
Restores Vue 2 behavior where slots were appended instead of being overwritten.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/standard/list/oh-list-item.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/standard/list/oh-list-item.vue
@@ -29,11 +29,13 @@
     <template #footer v-if="$slots.footer">
       <slot name="footer" />
     </template>
-    <template #after v-if="$slots.after">
-      <slot name="after" />
-    </template>
-    <template #after v-if="context.component.slots && context.component.slots.after && context.component.slots.after.length">
-      <generic-widget-component :context="childContext(context.component.slots.after[0])" />
+    <template #after v-if="$slots.after || context.component.slots?.after?.length">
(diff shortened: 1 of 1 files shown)


jsjames:
Thanks - i figured you would have a good solution to address. Just a quick question that I don't understand - why is the order in the actual rendered HTML page has the label followed by the switch, but the code looks like the switch should be first, followed by the label?

```
      <template v-if="context.component.slots?.after?.length">
        <generic-widget-component :context="childContext(context.component.slots.after[0])" />
      </template>
      <template v-if="$slots.after">
        <slot name="after" />
      </template>
```

florian-h05:
When passing a widget to the `after` slot of `oh-toggle-item` in a custom widget, the passed-in widget will be `context.component.slots.after[0]` in the `oh-list-item`, and the slot will be `$slots.after` in the `oh-list-item`.
This is because in `oh-toggle-item` the `oh-list-item` receives the widget config including the `after` widget, and the `#after` slot is filled with `oh-toggle` (hardcoded).

florian-h05:
This should of course be backported.
I don't know if you've already read it, but we are planning to do a patch release soon.

jsjames:
t

> This should of course be backported. I don't know if you've already read it, but we are planning to do a patch release soon.

I had not read that - was it posted on the forum?

florian-h05:
It was discussed in the release coordination thread. Only maintainers have access to it, but you should have. It however may be that you don't receive notifications, other maintainers had such problems in the past.
See https://community.openhab.org/t/release-coordination/136871/1511?u=florian-h05.

Issue #3700, opened 2026-01-02
Actions from differnt bindings show up on other bindings


lsiepel:
Might be related to navigating and/or clearing the DOM. Really don't know what is going on.

One user noticed this: 
_I’ve noticed that since installing the Z-Wave JS binding the “Actions” section appears against some things created by different bindings…_

<img width="690" height="499" alt="Image" src="https://github.com/user-attachments/assets/6e31ec65-d5e3-4cb5-ab2f-cd3ea68f93ed" />

The above is a screenshot from a Thing created using the Amazon Echo Control Binding

While looking for the cause, another user discovered it seemed unrelated to the zwavejs binding. As similar issue happened with fineoffsetweather binding containing actions from the SMTP binding:

<img width="526" height="500" alt="Image" src="https://github.com/user-attachments/assets/37160f11-3a33-4fa6-bbac-fb3a55d61cc2" />

_after refreshing the Thing page in MainUI, the wrong Actions are gone_

andrewfg:
I think I saw it too. See my comment on the forum thread. In my case it was a crossover between an Action on the Homekit binding wrongly appearing for a Thing on the Hue binding.

PS in case this may be of relevance, my case was using Chrome on Windows 11 to view a Thing on my OH test instance running on the same machine; albeit the browser was loading the thing url via a specific dotted ipV4 address (i.e. NOT looped back using localhost or 127.0.0.1)

msw1970:
So, a refresh of the Thing page in Main UI, does indeed resolve the problem, but only temporariliy. I have refreshed the page and when I then go to a Thing that doesn’t have Thing Actions (such as the Amazon EchoControl Account thing) no Thing actions are displayed. If I then go to a Thing created by the Astro binding and then back to the Amazon Account thing, the Amazon account thing has actions from the Astro binding.

Refresh the MainUI Thing page and the Actions are gone from the Amazon Account thing. I then go to a Thing created from the Network Binding and then back to the Amazon Account thing it now has Actions from the Network binding!!

JustinGeorgi:
I'm on a recent 5.2 snapshot and I cannot replicate this behavior on either a chromium browser or firefox.

Build #5051
Main UI Commit 6d6017be

lolodomo:
It was fixed by @digitaldan .
Problem of cache.

lolodomo:
Closed by #3673 
This will be in first patch of 5.1 and already in 5.2 snapshots.

PR #3710, opened 2026-01-04
Add failure handing to selected REST endpoints

Nadahar:
This is related to https://github.com/openhab/openhab-core/pull/5236. I've been trying to test and prepare for the changes to the REST API that are being discussed there.

I have some reservations on how I've chosen to handle errors here. I don't really know what is the desired reaction, I've made them "Back" out where I thought that made sense, and take other actions where I thought that made more sense. But, all didn't work like I wanted it to, in particular the pages where you have to select a Thing/Item and the failure reveals itself after the selection. I tried to get them to "reset" the selected element, but it didn't quite work as I wanted it to. Neither of them actually cleared the selected text, and neither of them stopped the "loading spinner" animation. But, neither of these situations are very likely to occur in the first place, and there's a lot of missing error handling in MainUI in general, so this is hardly worse than what's already there.

If somebody has some suggestions for improved failure handling, please let me know and I'll try to adapt.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
@@ -57,67 +57,27 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
   const hasLinkedItems = computed(() => thing.value?.channels?.find((c) => c.linkedItems?.length))
 
   // methods
-  async function loadThingActions(thingUID: string): Promise<void> {
-    // getAvailableActionsForThing will return 204 if no actions are available for this thing
-    const data = await api.getAvailableActionsForThing({ thingUID })
-
-    if (!data || Array.isArray(data) === false) {
-      return
-    }
-
-    thingActions.value = data
-      .filter((a) => a.visibility === 'VISIBLE' || a.visibility === 'EXPERT')
-      .filter((a) => a.inputConfigDescriptions !== undefined)
-      .sort((a, b) => (a.label || '').localeCompare(b.label || ''))
-  }
-
-  async function loadConfigDescriptions(thingUID: string): Promise<void> {
-    // getConfigDescriptionByUri will return 404 if no specific config description is available for this thing
-    // if an error or no specific description is available, try to use the config description from the Thing Type
-    try {
-      const data = await api.getConfigDescriptionByUri({ uri: 'thing:' + thingUID })
-      configDescriptions.value = data || null
-    } finally {
-      if (configDescriptions.value) return
-      if (thingType.value && thingType.value.configParameters && thingType.value.parameterGroups) {
-        console.debug('No specific config description available for this thing, using config description from thing type instead.')
-        configDescriptions.value = {
-          parameterGroups: thingType.value.parameterGroups,
-          parameters: thingType.value.configParameters
-        }
-      } else {
-        console.debug('No config description available for this thing.')
-        configDescriptions.value = null
+  function applyThingTypeConfigDescriptions(): boolean {
+    if (thingType.value && thingType.value.configParameters && thingType.value.parameterGroups) {
+      console.debug('No specific config description available for this thing, using config description from thing type instead.')
+      configDescriptions.value = {
+        parameterGroups: thingType.value.parameterGroups,
+        parameters: thingType.value.configParameters
       }
+      return true
     }
-  }
-
-  async function loadFirmwares(thingUID: string): Promise<void> {
-    // getThingFirmwareStatus will return 204 if no firmware info is available for this thing
-    await api.getThingFirmwareStatus({ thingUID }).then((data) => {
-      if (data && Object.keys(data).length > 0) {
-        firmwares.value = data
-      }
-    })
+    return false
   }
 
   function load(thingUID: string, loadingFinishedCallback: (success: boolean) => void) {
     if (loading.value) return
     loading.value = true
 
-    // reset data to initial state
-    thing.value = null
-    thingType.value = null
-    channelTypes.value = null
-    thingActions.value = null
-    configDescriptions.value = null
-    configStatusInfo.value = null
-    firmwares.value = null
-
     api
       .getThingById({ thingUID })
       .then((data) => {
         if (!data) {
+          thing.value = null
           console.warn('Thing not found')
           loadingFinishedCallback(false)
           loading.value = false
@@ -129,50 +89,84 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
         Promise.allSettled([
           api.getThingTypeById({ thingTypeUID: thing.value.thingTypeUID }),
           api.getChannelTypes({ prefixes: 'system,' + thing.value.thingTypeUID.split(':')[0] }),
-          loadThingActions(thingUID)
-        ]).then((results) => {
-          if (results[0].status === 'fulfilled') {
-            thingType.value = results[0].value || null
+          api.getAvailableActionsForThing({ thingUID }),
+          api.getConfigDescriptionByUri({ uri: 'thing:' + thingUID }),
+          api.getThingFirmwareStatus({ thingUID }),
+          api.getThingConfigStatus({ thingUID })
+        ]).then(([thingTypeResult, channelTypesResult, actionsResult, configDescResult, firmwareResult, configStatusResult]) => {
+          if (thingTypeResult.status === 'fulfilled') {
+            thingType.value = thingTypeResult.value && Object.keys(thingTypeResult.value).length > 0 ? thingTypeResult.value : null
           } else {
+            thingType.value = null
             const message = `Cannot load thing-type '${data.thingTypeUID}': `
-            console.error(message + getErrorMessage(results[0].reason, true))
-            f7.dialog.alert(message + getErrorMessage(results[0].reason))
+            console.error(message + getErrorMessage(thingTypeResult.reason, true))
+            f7.dialog.alert(message + getErrorMessage(thingTypeResult.reason))
           }
 
-          if (results[1].status === 'fulfilled') {
-            channelTypes.value = results[1].value || []
+          if (channelTypesResult.status === 'fulfilled') {
+            channelTypes.value = channelTypesResult.value || []
           } else {
+            channelTypes.value = null
             const message = `Cannot load channel-types for thing-type '${data.thingTypeUID}': `
-            console.error(message + getErrorMessage(results[1].reason, true))
-            f7.dialog.alert(message + getErrorMessage(results[1].reason))
+            console.error(message + getErrorMessage(channelTypesResult.reason, true))
+            f7.dialog.alert(message + getErrorMessage(channelTypesResult.reason))
           }
 
-          if (results[2].status !== 'fulfilled') {
-            const message = getErrorMessage(results[2].reason, true)
+          if (actionsResult.status === 'fulfilled') {
+            if (!actionsResult.value || Array.isArray(actionsResult.value) === false) {
+              thingActions.value = []
+            } else {
+              thingActions.value = actionsResult.value
+                .filter((a) => a.visibility === 'VISIBLE' || a.visibility === 'EXPERT')
+                .filter((a) => a.inputConfigDescriptions !== undefined)
+                .sort((a, b) => (a.label || '').localeCompare(b.label || ''))
+            }
+          } else {
+            thingActions.value = []
+            const message = getErrorMessage(actionsResult.reason, true)
             console.error(`Cannot load '${thingUID}' thing actions: ${message}`)
           }
 
-          Promise.allSettled([loadConfigDescriptions(thingUID), loadFirmwares(thingUID), api.getThingConfigStatus({ thingUID })]).then(
-            (results) => {
-              if (results[0].status !== 'fulfilled') {
-                console.error(`Cannot load '${thingUID}' config descriptions: ` + getErrorMessage(results[0].reason, true))
+          if (configDescResult.status === 'fulfilled') {
+            if (configDescResult.value && Object.keys(configDescResult.value).length > 0) {
+              configDescriptions.value = configDescResult.value
+            } else {
+              // no specific config description available for this thing, try to use the thing type config description
+              if (!applyThingTypeConfigDescriptions()) {
+                console.debug(`No config description available for Thing '${thingUID}'`)
               }
-              if (results[1].status !== 'fulfilled') {
-                console.error(`Cannot load '${thingUID}' firmwares: ` + getErrorMessage(results[1].reason, true))
-              }
-              if (results[2].status === 'fulfilled') {
-                configStatusInfo.value = results[2].value ?? null
-              } else {
-                console.error(`Cannot load '${thingUID}' config status: ` + getErrorMessage(results[2].reason, true))
-              }
-              savedThing.value = cloneDeep(thing.value)
-              loadingFinishedCallback(true)
-              loading.value = false
             }
-          )
+          } else {
+            if (!applyThingTypeConfigDescriptions()) {
+              configDescriptions.value = null
+            }
+            console.error(`Cannot load '${thingUID}' config descriptions: ` + getErrorMessage(configDescResult.reason, true))
+          }
+          if (firmwareResult.status === 'fulfilled') {
+            firmwares.value = firmwareResult.value && Object.keys(firmwareResult.value).length > 0 ? firmwareResult.value : null
+          } else {
+            firmwares.value = null
+            console.error(`Cannot load '${thingUID}' firmwares: ` + getErrorMessage(firmwareResult.reason, true))
+          }
+          if (configStatusResult.status === 'fulfilled') {
+            configStatusInfo.value = configStatusResult.value ?? null
+          } else {
+            configStatusInfo.value = null
+            console.error(`Cannot load '${thingUID}' config status: ` + getErrorMessage(configStatusResult.reason, true))
+          }
+          savedThing.value = cloneDeep(thing.value)
+          loadingFinishedCallback(true)
+          loading.value = false
         })
       })
       .catch((err) => {
+        thing.value = null
+        thingType.value = null
+        channelTypes.value = null
+        thingActions.value = []
+        configDescriptions.value = null
+        configStatusInfo.value = null
+        firmwares.value = null
         let message = `Cannot load Thing '${thingUID}': ${err}`
         console.error(message)
         f7.dialog.alert(message)


bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
florian-h05, line 1: Ideally, the thing-details page would work without a Thing type loaded and allow displaying Thing configuration.

florian-h05 (commented):
Generally, please do not alert the `err`, but only the message, or leave it out completely.
Always, please log the error together with a meaningful message to the console (console.error).

Nadahar:
> Generally, please do not alert the `err`, but only the message, or leave it out completely.

I initially used `err.message`, but when testing it, it was always undefined. Only when using `err` could I actually get anything useful out of it. It's the curse of untyped. I guess that "err" isn't always of the same type then, sometimes just a string and sometimes an object?

Even if the error doesn't tell the user that much, I think it's important that there is something there, because that's what they're going to report, and any grain of information can make figuring out what's causing that much faster.

I could also have put more effort into producing a more informative error message, but keep in mind: Most of these are very hard to actually experience, I had to modify the MainUI code to force invalid requests when testing. I don't know how I would provoke these to happen "in real life". That somehow limits how much effort it seems worth putting into the error messages themselves. Also, the lack of error handling in general in MainUI doesn't give a lot of good examples of how to handle things. That, and the fact that I don't know what the "err" object/string I get actually is, and if it will even be stable/the same every time.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
Nadahar: I didn't look that deeply into what's actually happening in these pages. What's the thing-type used for there, do you remember? I'm sure this can be tweaked to behave better, it just takes a lot of time to trace down how everything is connected/what the impact of various missing information is.

florian-h05:
> I initially used err.message, but when testing it, it was always undefined. Only when using err could I actually get anything useful out of it. It's the curse of untyped. I guess that "err" isn't always of the same type then, sometimes just a string and sometimes an object?

I guess the implicit `err.toString()` call seems to make a message display.

> I don't know how I would provoke these to happen "in real life".

Create a Thing, then uninstall the binding. Try to open the Thing in the UI and Thing types will be missing.

Nadahar:
> Create a Thing, then uninstall the binding. Try to open the Thing in the UI and Thing types will be missing.

You can't install/uninstall bindings in the "demo app" that I'm using. You have to configure the POM and then resolve the whole project (can be slow and error-prone) to add/remove bindings, so I generally do that as little as possible.

But I can try to build a Karaf build and do it I guess.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
Nadahar: The way it works with the PR now, is that you first get the alert:

bilde


..and then this screen is shown:

bilde

Isn't that basically what you want?

By the way, I changed `err` to `err.toString()`, and it produces the exact same message. I think `err` *is* a string here. Is that some difference between TS and JS perhaps?

edit: Actually, that isn't true - it gets stuck in the loading animation. But, when I changed to `loadingFinishedCallback(true)` in the catch block, it behaves as shown above. I just forgot that I had changed it.

edit2: The previous edit isn't true either - this seems volatile, behavior isn't consistent - so there must be some factor that decides that I haven't found yet.

Nadahar:
Actually, Google claims that you must do this in every catch - because the type is unknown:
```ts
try {
  await somePromiseThatMightReject();
} catch (error) {
  if (error instanceof Error) {
    // TypeScript knows 'error' is an Error object here
    console.error("Error message:", error.message);
  } else {
    // Handle cases where a non-Error value was thrown (e.g., a string or number)
    console.error("An unknown error occurred:", error);
  }
}
```

I don't know if `error.toString()` will achieve the same with less code, as I don't have an `Error` instance to experiment with.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
florian-h05: > By the way, I changed err to err.toString(), and it produces the exact same message. I think err is a string here. Is that some difference between TS and JS perhaps?

No, but I've just remembered that the err is "reduced" to its message in api.js

florian-h05:
> Actually, Google claims that you must do this in every catch - because the type is unknown:

Since we do

https://github.com/openhab/openhab-webui/blob/a60f99279f1166e3ef2a39bc4bd45cb227ce0fc8/bundles/org.openhab.ui/web/src/js/openhab/api.js#L9

in api.js, we always receive a primitive type, not the real Error.

Nadahar:
@florian-h05 I've done quite a few changes to the Thing page to try to make it as robust as possible. I'm constantly a bit out of my comfort zone here, messing with things I'm not sure that I understand well enough, so I'd appreciate if it's looked over by somebody who knows this better.

If the Thing itself is not found, it's still stuck in the loading animation, but at least the user gets an alert that the Thing isn't found. All "lower level" details shouldn't prevent the page from loading, it should load, but with missing information. I've kept alerts for missing `thing-types` and `channel-types`, everything else is just logged to the browser console.

jsjames:
we should be close with getting the hey-api client rest interface implemented. I have migrated the useThingEditStore to use the new api calls. I would suggest we get that one merged and then rebase any fixed from this PR that aren't covered.

Nadahar:
Just let me know what it's time, and I'll try to figure it out - with some help probably.

jsjames:
@Nadahar - the hey-api interface as now been merged and we can work on this PR.

- For any api request you want to add error handling to, we should adopt the new hey-api interface (see @/api/sdk.gen.ts) for the functions.
- We should make sure any api calls which can return a 204 are updated to the hey-api interface as it will automatically return an empty object/array so we don't need to do any special handling.

Nadahar:
> * For any api request you want to add error handling to, we should adopt the new hey-api interface (see @/api/sdk.gen.ts) for the functions.

I'm not sure if I can figure out how to use it correctly, do you have an example of something that has been "converted" to use it?

> * We should make sure any api calls which can return a 204 are updated to the hey-api interface as it will automatically return an empty object/array so we don't need to do any special handling.

Actually, the way Core is today, "all" API endpoints can return 204 since that's what happen if an `IOException` occurs during processing. Not everything *can* throw an `IOException` I suppose, but it's a very common exception in Java with many subclasses (which then also qualifies for this behavior).

Are both the "old" and the "new" API library available in parallel so that things are going to move gradually, or is the "old" API library expected to "stick around for eternity"?

jsjames:
> I'm not sure if I can figure out how to use it correctly, do you have an example of something that has been "converted" to use it?

there are quite a few files already migrated over, just search for files with an import from "@/api". A few more complicated ones are useThingEditStore and useComponentStore.

> Are both the "old" and the "new" API library available in parallel so that things are going to move gradually, or is the "old" API library expected to "stick around for eternity"?

Both will be around for awhile, but the hope is to migrate all calls to the new interface at some point.

Nadahar:
@jsjames I've tried to rebase and resolve conflicts (without doing anything to the API library at this time), but there had been made quite a lot of changes (and even more reformatting) to the involved code, so I'm not sure if I did it correctly. I could need some help looking if the resulting code looks "sane".

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -126,29 +126,60 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
+
+          Promise.allSettled([
             loadConfigDescriptions(thingUID),
             loadFirmwares(thingUID),
             api.getThingConfigStatus({ thingUID }).then((data) => {
               configStatusInfo.value = data ?? null
             })
-          ]).then(() => {
+          ]).then((results) => {
+            if (results[0].status !== 'fulfilled') {
```
Nadahar: `loadConfigDescriptions()` and `loadFirmwares()` seem to have changed since I wrote this. Can they ever not be fulfilled as they are now?

Is it a good design to just "swallow" the errors in the methods like this? It makes it very hard for the error to actually reach the user. I don't know how many places they are called from, but I think giving feedback in the UI is part of the point - at least it is for me. Because, when they UI just don't behave as you expect, but there's no error, most people don't understand where to start to figure out what must be done to fix it. When you get a concrete error message, regardless of if you can understand it or not, you have something concrete to work with, and the "road to solution" is usually much shorter and less frustrating.

edit: It seems like there's no practical difference, since "my code" also only log these to the console log - as the methods themselves does. But, it begs the question if the `allSettled()` is needed here at all, if there can never be a failure returned..?

edit2: The advantage of logging it "my way" is that you know the ThingUID and can include that in the logged message.

Nadahar:
> there are quite a few files already migrated over, just search for files with an import from "@/api". A few more complicated ones are useThingEditStore and useComponentStore.

It seems to me like the API calls I've dealt with have already been "upgraded"?

edit: It also seems like they don't throw on error, which might make when I've done pointless... but again, I don't think hiding errors is the best route.

Nadahar:
Initial testing confirms that my errors no longer work. Instead, you just see a layout with most things missing without an explanation. How do I get these API calls to throw?

Nadahar:
I can't get it to work with the new API library, even doing this:
```vue
          api.getThingTypeById({ thingTypeUID: thing.value!.thingTypeUID }, {throwOnError: true}),
          api.getChannelTypes({ prefixes: 'system,' + thing.value.thingTypeUID.split(':')[0] }, {throwOnError: true}),
```
...doesn't help. The promise is still "fulfilled" when a 404 is returned.

I thought that this would handle 204 specifically, but are you instead saying that it will now just return an empty object on *all failures*? Because, that sounds quite terrible to me... how can you ever handle errors if they are all hidden?

jsjames:
> o seems like they don't throw on error, which might make when I've done pointless... but again, I don't think hiding errors is the best route.

They should throw on any error (>400 response code). Is there one that isn't throwing?

jsjames:
> I thought that this would handle 204 specifically, but are y

I don't expect it to throw on a 204 - this isn't an error. For instance, if the thing doesn't have any firmware, a 204 isn't an error, it just indicates the thing does not have any firmware - hence a 204 - No Content. So, returning an empty object is a good response - I don't think any other reporting is required - in fact in my system, none of my devices have firmware and so I wouldn't want by logs filing up with messages.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -126,29 +126,60 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
+            console.error(message)
+            f7.dialog.alert(message)
+          }
+
+          if (results[2].status !== 'fulfilled') {
+            console.warn(`Cannot load '${thingUID}' thing actions: ${results[2].reason}`)
+          }
+
+          Promise.allSettled([
             loadConfigDescriptions(thingUID),
```
jsjames: In looking at the new loadConfigDescriptions, in addition to checking for undefined, we might want to also check for an empty object and use the thing-type description instead (similar to the catch block).

Nadahar:
> They should throw on any error (>400 response code). Is there one that isn't throwing?

Yes, it's not throwing when I try to open a Thing with a missing thing type (`useThingEditStore.js`):

```js
        Promise.allSettled([
          api.getThingTypeById({ thingTypeUID: thing.value!.thingTypeUID }, {throwOnError: true}),
          api.getChannelTypes({ prefixes: 'system,' + thing.value.thingTypeUID.split(':')[0] }, {throwOnError: true}),
          loadThingActions(thingUID)
        ]).then((results) => {
          if (results[0].status === 'fulfilled') {
            thingType.value = results[0].value
          } else {
            let message = `Cannot load thing-type '${data.thingTypeUID}': ${results[0].reason}`
            console.error(message)
            f7.dialog.alert(message)
          }

          if (results[1].status === 'fulfilled') {
            channelTypes.value = results[1].value
          } else {
            let message = `Cannot load channel-types for thing-type '${data.thingTypeUID}': ${results[1].reason}`
            console.error(message)
            f7.dialog.alert(message)
          }

          if (results[2].status !== 'fulfilled') {
            console.warn(`Cannot load '${thingUID}' thing actions: ${results[2].reason}`)
          }
```

It something weird going on, because it never gets to the "catch" clause in `client.gen.ts`:
```ts
    try {
      response = await _fetch(request);
    } catch (error) {
      // Handle fetch exceptions (AbortError, network errors, etc.)
      let finalError = error;

      for (const fn of interceptors.error.fns) {
        if (fn) {
          finalError = (await fn(
            error,
            undefined as any,
            request,
            opts,
          )) as unknown;
        }
      }
```
Instead, a 404 is logged in the console when `await_fetch()` is called, but a breakpoint placed in the catch never triggers - which seems to be why nothing is thrown either.

Nadahar:
> I don't expect it to throw on a 204 - this isn't an error. For instance, if the thing doesn't have any firmware, a 204 isn't an error, it just indicates the thing does not have any firmware - hence a 204 - No Content. So, returning an empty object is a good response - I don't think any other reporting is required - in fact in my system, none of my devices have firmware and so I wouldn't want by logs filing up with messages.

I think you misunderstood me - I agree that it shouldn't throw on 204 - that was the problem with the "old lib". But, I can't get it to throw on 404 either, so I was wondering if this new lib didn't throw anything at all. Also, `throwOnError` defaults to false, and when the "return type" is `data`, it looks to me like the code will ignore errors and just return `undefined` if one happens. But, I'm not sure that I understand all that happens, my JS is bad, my TS is non-existent.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
Nadahar: I'm not sure what the consequence of using the thing-type description is. What is that usually populated with?

Nadahar:
I need some orientation to make sense of this - the code that is generated, what is it generated from? Where is "the real source"? I thought that TS was "compiled" into JS, and thought that was what this "generation" perhaps was, but that can't be, since the "generated" code is TS.

I also can't figure out where the source code for the "client" is, so it's hard to follow calls to e.g. `client.get()`. VSCode can't figure it out either, so it's not Ctrl+clickable.

Nadahar:
Look at this code in `client.get.ts` (in the catch clause that is never triggered):

```ts
      if (opts.throwOnError) {
        throw finalError;
      }

      // Return error response
      return opts.responseStyle === 'data'
        ? undefined
        : {
            error: finalError,
            request,
            response: undefined as any,
          };
```

`throwOnError` is false by default, and as far as I can see, we use `data` as `responseStyle`. As far as I can understand from this code, that means that errors will just "vanish" - since they won't be thrown, and they won't be returned in the response.

jsjames:
> > They should throw on any error (>400 response code). Is there one that isn't throwing?
> 
> Yes, it's not throwing when I try to open a Thing with a missing thing type (`useThingEditStore.js`):
> 
> ```js
>         Promise.allSettled([
>           api.getThingTypeById({ thingTypeUID: thing.value!.thingTypeUID }, {throwOnError: true}),
>           api.getChannelTypes({ prefixes: 'system,' + thing.value.thingTypeUID.split(':')[0] }, {throwOnError: true}),
>           loadThingActions(thingUID)
>         ]).then((results) => {
>           if (results[0].status === 'fulfilled') {
>             thingType.value = results[0].value
>           } else {
>             let message = `Cannot load thing-type '${data.thingTypeUID}': ${results[0].reason}`
>             console.error(message)
>             f7.dialog.alert(message)
>           }
> 
>           if (results[1].status === 'fulfilled') {
>             channelTypes.value = results[1].value
>           } else {
>             let message = `Cannot load channel-types for thing-type '${data.thingTypeUID}': ${results[1].reason}`
>             console.error(message)
>             f7.dialog.alert(message)
>           }
> 
>           if (results[2].status !== 'fulfilled') {
>             console.warn(`Cannot load '${thingUID}' thing actions: ${results[2].reason}`)
>           }
> ```
> 
> It something weird going on, because it never gets to the "catch" clause in `client.gen.ts`:
> 
> ```ts
>     try {
>       response = await _fetch(request);
>     } catch (error) {
>       // Handle fetch exceptions (AbortError, network errors, etc.)
>       let finalError = error;
> 
>       for (const fn of interceptors.error.fns) {
>         if (fn) {
>           finalError = (await fn(
>             error,
>             undefined as any,
>             request,
>             opts,
>           )) as unknown;
>         }
>       }
> ```
> 
> Instead, a 404 is logged in the console when `await_fetch()` is called, but a breakpoint placed in the catch never triggers - which seems to be why nothing is thrown either.

Can you put a break the section of code in hey-api.ts. The throw on error doesn't occur directly in client.gen.ts, it is thrown in this intercepter.

```
client.interceptors.response.use((response) => {
  if (response.status >= 400) {
    const message =
      response.status >= 500
        ? 'Server error'
        : response.status === 404
          ? 'Not Found'
          : response.status === 400
            ? 'Bad request'
            : 'Client error'
    throw new ApiError(message, response)
  }
  return response
})
```

jsjames:
> so I was wondering if this new lib didn't throw anything at all. Also, `throwOnError` defaults to false, and when the "return type" is `data`, it looks to me like the code will ignore errors and just return `undefined` if one happens. But, I'm not sure that I understand all that happens, my JS is bad, my TS is non-existent.

I decided not to use the built-in throwOnError signal because there is no way to get the actual response. Instead, all responses should be sent to the interceptor and that throws an error (ApiError) if the response >= 400 and also  attaches the response to the error.

A response of >=400 should throw an error (in the interceptor), so we need to understand why that is not happening in this case.

Nadahar:
That breakpoint is triggered. One thing I notice is that `response.statusText` isn't used - but very generic messages are used instead. The API sometimes gives a somewhat more meaningful "reason" that I assume end up in `statusText`, so maybe it would be better to sent `statusText` instead of the generic message, if it has content?

When following from that breakpoint, the code continues on line 133 in `client.get.ts`, so it skips the entire catch clause.

Nadahar:
On the top of the "client method", I see this:
```ts
  const request: Client['request'] = async (options) => {
    // @ts-expect-error
```
What does `@ts-expect-error` do? Can that be related to why the catch doesn't trigger?

jsjames:
> Look at this code in `client.get.ts` (in the catch clause that is never triggered):
> 
> ```ts
>       if (opts.throwOnError) {
>         throw finalError;
>       }
> 
>       // Return error response
>       return opts.responseStyle === 'data'
>         ? undefined
>         : {
>             error: finalError,
>             request,
>             response: undefined as any,
>           };
> ```
> 
> `throwOnError` is false by default, and as far as I can see, we use `data` as `responseStyle`. As far as I can understand from this code, that means that errors will just "vanish" - since they won't be thrown, and they won't be returned in the response.

I think you are correct here - in looking at the code we likely need to set throwOnError to true even if we create our own error.  Let me give you some instructions on how to do this.

[EDIT] Actually, that is not right. The code here in that same function get executed which calls our interceptor - that interceptor will still throw the error regardless of whether throwOnError is set or not.

```
    for (const fn of interceptors.response.fns) {
      if (fn) {
        response = await fn(response, request, opts);
      }
    }
```

Nadahar:
> Can you add throwOnError the setConfig function in hey-api.ts.
> 
> ```
> client.setConfig({
>   parseAs: 'json',
>   throwOnError: true
> })
> ```

I've done it, but I can't see that it makes a difference. Since the catch clause is never invoked, `throwOnError` is never evaluated. Also, I've already managed to set it to true by making the call like this:
```ts
api.getThingTypeById({ thingTypeUID: thing.value!.thingTypeUID }, {throwOnError: true})
```
I've also verified in the debugger that it is ´true´, but I still never get a failure of the promise.

jsjames:
> On the top of the "client method", I see this:
> 
> ```ts
>   const request: Client['request'] = async (options) => {
>     // @ts-expect-error
> ```
> 
> What does `@ts-expect-error` do? Can that be related to why the catch doesn't trigger?

No, that just quiets a typescript error - it doesn't have anything to do with the actual execution of the code.

jsjames:
Because it is a error thrown from a promise, we have to use the promise catch vs. the try/catch block.

If you look at the loadFirmware function, this is how the catch block should be coded.

jsjames:
> Because it is a error thrown from a promise, we have to use the promise catch vs. the try/catch block.
> 
> If you look at the loadFirmware function, this is how the catch block should be coded.

Try this to see of an error is caught:

```
        Promise.all([
          api.getThingTypeById({ thingTypeUID: thing.value!.thingTypeUID }).catch(e) { console.log(e) },
          api.getChannelTypes({ prefixes: 'system,' + thing.value.thingTypeUID.split(':')[0] }),
          loadThingActions(thingUID)
        ]).then((data2) => 
```

Nadahar:
> Because it is a error thrown from a promise, we have to use the promise catch vs. the try/catch block.
> 
> If you look at the loadFirmware function, this is how the catch block should be coded.

The code that doesn't catch it isn't "mine", it's in `client.gen.ts`:

```ts
    try {
      response = await _fetch(request);
    } catch (error) {
      // Handle fetch exceptions (AbortError, network errors, etc.)
      let finalError = error;

      for (const fn of interceptors.error.fns) {
        if (fn) {
          finalError = (await fn(
            error,
            undefined as any,
            request,
            opts,
          )) as unknown;
        }
      }
```
So, `await _fetch(request)` is never caught by the catch following immediately after.

jsjames:
> > Because it is a error thrown from a promise, we have to use the promise catch vs. the try/catch block.
> > If you look at the loadFirmware function, this is how the catch block should be coded.
> 
> The code that doesn't catch it isn't "mine", it's in `client.gen.ts`:
> 
> ```ts
>     try {
>       response = await _fetch(request);
>     } catch (error) {
>       // Handle fetch exceptions (AbortError, network errors, etc.)
>       let finalError = error;
> 
>       for (const fn of interceptors.error.fns) {
>         if (fn) {
>           finalError = (await fn(
>             error,
>             undefined as any,
>             request,
>             opts,
>           )) as unknown;
>         }
>       }
> ```
> 
> So, `await _fetch(request)` is never caught by the catch following immediately after.

Correct, _fetch doesn't throw the error and so that isn't caught there. It is thrown in the interceptor which is called later in that function.

Nadahar:
> > Because it is a error thrown from a promise, we have to use the promise catch vs. the try/catch block.
> > If you look at the loadFirmware function, this is how the catch block should be coded.
> 
> Try this to see of an error is caught:
> 
> ```
>         Promise.all([
>           api.getThingTypeById({ thingTypeUID: thing.value!.thingTypeUID }).catch(e) { console.log(e) },
>           api.getChannelTypes({ prefixes: 'system,' + thing.value.thingTypeUID.split(':')[0] }),
>           loadThingActions(thingUID)
>         ]).then((data2) => 
> ```

As I expected, it makes no difference. The returned "data2" is an array where `data.[0]` (the one that failed) is simply empty. It's the same as I see with `allSettled`. 

I'm no expert in this "promise" thing, but it seems like something is missing - something that makes the promise fail if an exception is thrown.

Nadahar:
It's probably easier if you debug it yourself. You strictly don't need "my code" to see the error, but it might make it easier to see, so you could just check out [Nadahar:rest-response](https://github.com/Nadahar/openhab-webui/tree/rest-response).

What I've done to provoke the error is to set up some Thing using a binding, then uninstall that binding. That makes the thing-type non-existent, which is the situation I'm trying to test at this time.

Nadahar:
I don't know how familiar you are with Git, but what I typically do is to add another remote to the existing local repo I have (`openhab-webui` in this case). The new remote could be named "nadahar" and point to https://github.com/Nadahar/openhab-webui.git

```shell
git remote add nadahar https://github.com/Nadahar/openhab-webui.git
```

Once that's in place, you just do a `git fetch nadahar` and then `git checkout -b rest-response nadahar/rest-response`.

jsjames:
> It's probably easier if you debug it yourself. 

Sounds good, I'll look at this tomorrow. Thanks for the suggestion on how to create the issue.

Nadahar:
To me, it seems like there would have to be a catch somewhere that caught errors and then failed the promise. I don't know if that's supposed to "happen automatically", but it doesn't seem like it's working. The errors seem to just reach some "top level" that logs them, but the promise is unaffected and is still fulfilled.

Nadahar:
I don't think the methods like `api.getThingTypeById()` are promises at all. I think they are just normal functions, and that the "async" part is on some other level? I don't really get it, but I think there's a fundamental problem with treating them like promises. There's no sign of promise handing in the methods themselves, and it would explain why they never fail, I assume.

jsjames:
> I don't think the methods like `api.getThingTypeById()` are promises at all. I think they are just normal functions, and that the "async" part is on some other level? I don't really get it, but I think there's a fundamental problem with treating them like promises. There's no sign of promise handing in the methods themselves, and it would explain why they never fail, I assume.

It is a promise - the await on line 101 in client.gen.ts makes the call a promise.

Nadahar:
> It is a promise - the await on line 101 in client.gen.ts makes the call a promise.

Ok, but where is the "promise object"? Because the content of "response" doesn't seem like a promise?

Nadahar:
The problem must be inside "_fetch" then, and how it deals with rejection. I haven't been able to figure out where the code for "_fetch" is.

jsjames:
> Ok, but where is the "promise object"? Because the content of "response" doesn't seem like a promise?

The promise is implicit since its an async function.

Nadahar:
> The promise is implicit since its an async function.

Yeah, there are things I don't know here - but what I do know, is that nothing ever reaches the catch block. From what I can understand, `await` is supposed to make sure that a rejected promise is turned into an error? But, does that also mean that an error thrown inside the async code also ends up there? This is unclear to me.

jsjames:
> > The promise is implicit since its an async function.
> 
> Yeah, there are things I don't know here - but what I do know, is that nothing ever reaches the catch block. From what I can understand, `await` is supposed to make sure that a rejected promise is turned into an error? But, does that also mean that an error thrown inside the async code also ends up there? This is unclear to me.

I don't expect fetch to actually throw an error here. there is no "javascript" error just because a response code of >=400 is returned. That get's turned into an error in the interceptor. I'm looking at the code now and will provide some suggestions today.

jsjames:
@Nadahar - Just submitted a more robust thing-detail / useThingEditStore. Instead of showing a bunch of errors, I think we should make sure the thing-detail page can properly show the user when there is an issue. In this case, if the thingType never is set - it will show this . "Real" errors that aren't just a response of 404 indicated something doesn't exist are logged.

image

Given the useThingEditStore and thing-details is fairly involved and the code is already typescript, I recommend we break those specific changes to its own PR. I can start porting the other files to typescript which will also improve robustness.

Nadahar:
Exactly how you show the problem hasn't been my focus, it has been that the user should be presented with some information that something is wrong instead of just being stuck in "eternal loading".

That said, I still think it's vital that it should be possible to handle failures in some predictable way. Using promises was how it used to be, it doesn't have to be that, but it should be a consistent way across API calls. I don't know how you have solved it here, but you seem to have found a way to present it to the user?

I generally don't find it useful to hide errors too much, because it just makes it harder for users to figure out what's wrong. Very "generic" errors are bad in that sense, because they aren't specific to the particular problem. Therefore, I find it useful to give details even if the user can't understand them - because it makes the chance that they will find something relevant when they search for the error that much greater.

Nadahar:
> Just submitted a more robust thing-detail / useThingEditStore.

Where did you submit it?

jsjames:
> > Just submitted a more robust thing-detail / useThingEditStore.
> 
> Where did you submit it?

i should have submitted to your rest-response branch - [edit] i just resubmitted.

Nadahar:
> [edit] i just resubmitted.

Now it's there.

Nadahar:
I'm not sure that we're on the same page. You have removed all the error handing, which was the point of this PR. Instead, you have made a "special case" for if no thing-type is found and given it a generic error message that might be wrong, depending on the circumstances.

The error handling I had made was meant to cover must more than just 404 - it was meant to handle all error conditions, and to give as "precise feedback" as possible with the little information available.

To me, moving error handing far away from the presentation page is also not the way to do it. API calls can be made from various places, perhaps it's only one place today, but that might change in the future. So, you don't want the calls to react to the error, you want the error to be returned to the caller, because then the caller can handle the error in the way that is appropriate for that particular page/location.

jsjames:
> I'm not sure that we're on the same page. You have removed all the error handing, which was the point of this PR. Instead, you have made a "special case" for if no thing-type is found and given it a generic error message that might be wrong, depending on the circumstances.
> 
> The error handling I had made was meant to cover must more than just 404 - it was meant to handle all error conditions, and to give as "precise feedback" as possible with the little information available.
> 
> To me, moving error handing far away from the presentation page is also not the way to do it. API calls can be made from various places, perhaps it's only one place today, but that might change in the future. So, you don't want the calls to react to the error, you want the error to be returned to the caller, because then the caller can handle the error in the way that is appropriate for that particular page/location.

We can still add specific error messages if desired, but we need to make sure the UI doesn't freeze with the spinning wheel of death. The changes I made will make sure that never happens.

All unexpected errors that occur should still be logged with this catch. Do you ever see executed - it should be fairly verbose as the err contains the full response/api call.

```
     .catch((err) => {
        let message = `Cannot load Thing '${thingUID}': ${err}`
        console.error(message)
        f7.dialog.alert(message)
        loadingFinishedCallback(false)
        loading.value = false
      })
```

For normal conditions like a thing doesn't have any actions or firmware, I don't see why those need to be logged. In fact, that is probably 80% of the things in my system don't have some actions or firmware.

Nadahar:
> For normal conditions like a thing doesn't have any actions or firmware, I don't see why those need to be logged. In fact, that is probably 80% of the things in my system don't have some actions or firmware.

No actions or firmware aren't "errors" and they should return 204. That's not what I'm after here. What I've been trying to address is 1) Make sure that nothing breaks if https://github.com/openhab/openhab-core/pull/5236 is merged, which will made some methods return 404 instead of 204. 2) Make sure that when actual errors happen, like we experienced when people upgraded to OH 5.1, that it should be possible to figure out what goes wrong - instead of just having a "broken UI" as a result, where people think that the UI is the problem and have no idea what the real reason is.

Regarding the console, I agree that it's good to log things there, but one must realize that most users have no idea what it is or how to make sense of it. For actual errors, I don't think merely logging to the console is good enough.

> All unexpected errors that occur should still be logged with this catch. Do you ever see executed - it should be fairly verbose as the err contains the full response/api call.

No, that's what I'm trying to say. Something is broken here, the promises are never "rejected" so no `catch()` clauses for the promises are activated. Also, only if the Thing itself has a problem, will it reach that particular catch (if the promises worked as they should). The others were caught in the above block, before you removed the error handing. The point with this is to give useful feedback - feedback about the issue at hand. A "collective" error handler for many different problems is a *bad* error handler, because it doesn't know which context in which the error occurred - so it can't give the user a useful response.

jsjames:
Do you have another situation that I can test where a 404 error is returned?

Errors are being thrown in the api calls - otherwise the catch in the loadThingActions and loadConfigDescriptions functions wouldn't output the "No actions available for this thing". Do you see that for a thing which doesn't have actions?

Nadahar:
> Do you have another situation that I can test where a 404 error is returned?

Not that is easily available. What I do is that I modify the Java code in the REST API to make it return what I want to test, and then run MainUI against that code. But, to do that, you need to have a core dev environment. I assume that you don't have that?

But, I did this to test this PR as the code was, I tested all kind of situations "that I could think of". Nothing seems to work at the moment, I just can't quite understand where the errors end up.

I don't understand this:

```ts
    for (const fn of interceptors.error.fns) {
      if (fn) {
        finalError = (await fn(error, response, request, opts)) as string;
      }
    }

    finalError = finalError || ({} as string);

    if (opts.throwOnError) {
      throw finalError;
    }
```

Why cast an error to a string? What is left of the error then? A lot of information must be lost when doing that, and I don't quite get the idea. Also, does it really work to "throw a string"?

Nadahar:
> Do you see that for a thing which doesn't have actions?

Yes, I see those in the console, but I never get any error "at the level above", in the `allSettled()` calls.

Nadahar:
Actually, I just commented out the section for 404 in `loadThingActions()`, and that actually results in a rejected promise then. But for some reason, the same doesn't happen for `getThingTypeById()`. But, I also see a difference between them... one is a function that is defined to return a promise, the other calls the API directly.

jsjames:
> or to a string? What is left of the error then? A lo

that error is never thrown because throwOnError is false. Instead, the 

> l above", in the `allSettled()` calls.

You won't as the error is caught and not rethrown in the catch block. Only if the error is not an "expected" error (i.e. the statusText !== 'Not Found') will it throw the error. 

```
      if (e.response?.statusText === 'Not Found' || e.response?.status === 404) {
        console.log('No actions available for this Thing')
        return
      }
      console.error('Error loading thing actions: ' + e)
      throw e
 ```

**Comment by Nadahar** [comment]
> You won't as the error is caught and not rethrown in the catch block. Only if the error is not an "expected" error (i.e. the statusText !== 'Not Found') will it throw the error.

That much I understand, it's why I said I commented out the code:

```ts
  async function loadThingActions(thingUID: string): Promise<void> {
    // getAvailableActionsForThing will return 404 if no actions are available for this thing (204 would be better response)
    const data = await api.getAvailableActionsForThing({ thingUID }).catch((e) => {
      // if (e.response?.statusText === 'Not Found' || e.response?.status === 404) {
      //   console.log('No actions available for this Thing')
      //   return
      // }
      throw e
    })
```
This makes that promise reject, but the ones I want to reject are the other two. I've experimenting, maybe I'm onto something.

Nadahar:
This is so weird. The Thing I've entered has no handler, so both the first and the third promise should fail. Yet, this is the result:

image

The first promise, the call to `api.getThingTypeById()` is "fulfilled" with value `undefined`, while the call to `loadThingActions()` works like it should (after I've commented out the code as shown above).

Nadahar:
Aargh..... sorry! It's my fault! I didn't check out the correct core code, so it doesn't actually return 404 on the first one - but 204. That explains it 😥 

edit: That's the result of working on too many different things at once. I am in the middle of something else in core, so I didn't switch branch - but I forgot that the return status for this particular endpoint was among those that changed. 🥸

jsjames:
> Aargh..... sorry! It's my fault! I forgot to check out the correct core code, so it doesn't actually return 404 on the first one - but 204. That explains it 😥

Yeah, that is why I check for that condition before assigning to thingType.value. That is one of the ones that could return a 404 as well IMO.

Nadahar:
> That is one of the ones that could return a 404 as well IMO.

If https://github.com/openhab/openhab-core/pull/5236 is merged, it will be 404.

Nadahar:
> Yeah, that is why I check for that condition before assigning to thingType.value. That is one of the ones that could return a 404 as well IMO.

What fooled me was that I saw 404 in the browser debugger. But I now realize that it was from the action call, not the thing-type call. I thought it was the other one, so I was in "no doubt" that a 404 was in fact returned... and of course it didn't help that a 404 was in fact returned the last time I tested this (with the correct core branch).

Nadahar:
@jsjames https://github.com/openhab/openhab-core/pull/5236 is still a draft because it's waiting for this PR. I don't want to do those changes to the API without making sure that MainUI can handle it.

jsjames:
> @jsjames [openhab/openhab-core#5236](https://github.com/openhab/openhab-core/pull/5236) is still a draft because it's waiting for this PR. I don't want to do those changes to the API without making sure that MainUI can handle it.

In looking at the changes in the useEditThingStore, it should work and we can remove the special casing for the 204 response. The 404 error code will throw an error which will be logged to the console. Then, the values in the useEditThingStore will still be the value at initialization - which the thing-details page should handle and provide the appropriate output. Now, if we want to output an alert or some other warning in the UI, that needs to be added.  But the code I shared should work.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -52,68 +53,79 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
       })
   }
 
   function load(thingUID: string, loadingFinishedCallback: (success: boolean) => void) {
     if (loading.value) return
     loading.value = true
 
+    // reset data to initial state
+    thing.value = null
+    thingType.value = null
```
Nadahar: I'm unsure if this is the right thing to do when reactivity is in play. If a page is reloaded, everything will first be set to the default and then the actual value again - triggering two reactive operations, while otherwise, assuming that the value hasn't changed, nothing would be triggered.

jsjames: It will trigger a reactivity - but it only occurs when a new thing is loaded - which happens on the initial loading of the thing-details page. It will be an insignificant performance penalty - but it makes sure everything is in a clean state. There were quite a few issues people saw of the store not getting reset to a clean state and there were carry-overs.

Its one of the cases where there is a very minor performance impact, but the coding is clean and will reduce potential issues.

Nadahar: Are you sure that `load()` is never called at other times, if the item is updated through an event for example?

I understand that all fields must be set to their default if they can't be retrieved, I'm just saying that it might be better to set the default once the failure is a fact, instead of upfront, might be a better idea.

I can show what I mean in a commit.

jsjames: That was where there were a lot of issues seen in the original code - you have to make sure everything gets reset in catch blocks. it was problematic.

If you are really concerned about reactivity issue - you can use a dummy state and then copy everything from the dummy state to the reactive state as the last step before returning from load.

jsjames: > Are you sure that `load()` is never called at other times, if the item is updated through an event for example?

Yes, its only called from thing-details.

Nadahar: > That was where there were a lot of issues seen in the original code - you have to make sure everything gets reset in catch blocks. it was problematic.

It's not problematic when you use `allSettled()` instead of `all()`. That's kind of the point of what I'm trying to do, make error handling consistent. I think it's important that it's done as similar as possible for all calls, and everything handled in one place. That way, it's neither confusing or easy to miss anything.

> Yes, its only called from thing-details.

That sounds like a bug in itself. If an event is received that the Thing has changed via SSE, the page *should* update..?

jsjames: > That was where there were a lot of issues seen in the original code - you have to make sure everything gets reset in catch blocks. it was problematic.
> 
> If you are really concerned about reactivity issue - you can use a dummy state and then copy everything from the dummy state to the reactive state as the last step before returning from load.

see:
https://github.com/openhab/openhab-webui/issues/3774
https://github.com/openhab/openhab-webui/issues/3700

I think there were a few other cases - those have all been individually addressed - but this way its much more robust.

jsjames: > That sounds like a bug in itself. If an event is received that the Thing has changed via SSE, the page _should_ update..?

Maybe an enhancement - but the UI does not adapt to changes in the structure real-time while on a page. It only really adapts to state changes.

Nadahar:
It took some time, but I've restored the error handling and, in the last commit, reorganized the initialization and error handing so that it is uniform.

While some will claim that "long methods are messy" and all that, I call that BS. It should not be the only thing to consider. When the error handling is spread over various methods, handled in slightly different ways, it becomes very hard to get an overview and make sure that nothing is forgotten. I think doing it this way makes the handling very clear and easy to assess/evaluate, minimizing the risk that error situations are left unhandled, or that things remain uninitialized. At the same time, this respects reactivity in that in doesn't cause unnecessary changed.

Nadahar:
I have, by modifying the core API, tested error conditions for each API call involved here, and as the code is now, it's all handled "without surprises". Errors are logged in the console, and the most problematic ones are shown in a modal to make the user aware that there's a problem. The page itself does what it can with the data it has.

Nadahar:
@jsjames Were the changes in `package-lock.json` supposed to be here? It looks like something that is unrelated to this PR.

bundles/org.openhab.ui/web/src/api/client/client.gen.ts
```diff
@@ -229,15 +229,18 @@ export const createClient = (config: Config = {}): Client => {
     }
 
     const error = jsonError ?? textError;
-    let finalError = error;
+    let finalError: unknown = error;
```
jsjames: Did this file get modified by hand, or did this get generated?  When I generate, the output is different?

jsjames:
Am I supposed to get an alert if thingType is invalid?  I get the following in the log, but now alert:

image

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -126,29 +96,90 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
-          ]).then(() => {
-            savedThing.value = cloneDeep(thing.value)
-            loadingFinishedCallback(true)
-            loading.value = false
-          })
+          api.getAvailableActionsForThing({ thingUID }),
+          api.getConfigDescriptionByUri({ uri: 'thing:' + thingUID }),
+          api.getThingFirmwareStatus({ thingUID }),
+          api.getThingConfigStatus({ thingUID })
+        ]).then((results) => {
```
jsjames: please destructure the array to make the code more readable:

```
        ]).then(([thingTypeResult, channelTypesResult, actionsResult, configDescResult, firmwareResult, configStatusResult]) => {
          if (thingTypeResult.status === 'fulfilled') {
            thingType.value = thingTypeResult.value || null
          } else {
            thingType.value = null
            const message = `Cannot load thing-type '${data.thingTypeUID}': `
            console.error(message + unwrapErrorMessage(thingTypeResult.reason, true))
            f7.dialog.alert(message + unwrapErrorMessage(thingTypeResult.reason, false))
          }
```

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -52,62 +53,30 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
-        if (data && Object.keys(data).length === 0) {
-          console.debug(`Firmware info not available for Thing ${thingUID}`)
-        }
-        firmwares.value = data!
-      })
-      .catch((err) => {
-        firmwares.value = null
-        console.debug(`Firmware info not available for Thing ${thingUID}: ` + err.message)
-      })
+  function unwrapErrorMessage(err: any, debug: boolean): string {
```
jsjames: If this is a general purpose method to de-structure api call errors - we should put in a general location.

jsjames: Also, do we need this complexity - why not just write the debugmessage to console and use the message for the dialog?

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -126,29 +96,90 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
-            loadingFinishedCallback(true)
-            loading.value = false
-          })
+          api.getAvailableActionsForThing({ thingUID }),
+          api.getConfigDescriptionByUri({ uri: 'thing:' + thingUID }),
+          api.getThingFirmwareStatus({ thingUID }),
+          api.getThingConfigStatus({ thingUID })
+        ]).then((results) => {
+          if (results[0].status === 'fulfilled') {
+            thingType.value = results[0].value || null
```
jsjames: I know your core PR will make this return a 404, but I'd recommend we keep the blank object check here as well and if we get a blank object, set the value to null. That will make the UI code less dependent on the specific core code its run against.

jsjames:
> @jsjames Were the changes in `package-lock.json` supposed to be here? It looks like something that is unrelated to this PR.

No - there shouldn't be any.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
Nadahar: I don't know TS, so I just have to find ways that work. It should probably be in another location, yes. The idea with the method was that you could feed it "anything", an `ApiError`, a regular `Error` or just a string - and that it should be able to return a suitable string from it. The reason I added the "debug" setting, was that I saw that you had modified the message in `ApiError` to build what I'd call a "debug string" - that included the status code and URL.

I think a better solution would be to not do `debugString` in `ApiError` itself, and rather "build it on demand" in this method, for example. Either way, the main reason for this method was that it's not certain that what we receive is an `ApiError` from what I've seen - so I wanted to handle that "complexity" once and be done with it.

bundles/org.openhab.ui/web/src/api/client/client.gen.ts
Nadahar: I modified it by hand, again, I don't know typescript and I don't know how you generate these files. I needed to make it `unknown` to get out of type errors other places, because `finalError` is sometimes assigned one type and other times another, and it seems that TS couldn't work this out properly, resulting in errors if I didn't make it "unknown".

Just tell me how it should look, and I'll change it.

Nadahar:
> Am I supposed to get an alert if thingType is invalid? I get the following in the log, but now alert:
> image

That's strange, because this is what I see:

image

image

The first two console entries are logged by some other code, probably hey-api or something, I've never found what generates them. That leads to every failure being "logged twice", where the "Failed to load resource" being so generic that it has little value.

I've only made it create an alert if the thing itself returns >=400, thing-type returns >=400 or channel-types returns >=400. The other types are just logged to the console. This was, as far as I remember, because @florian-h05 didn't want alerts "for everything", so I tried to find a balance in what to alert for and not depending on "what level of degradation" that failure will lead to in the UI. But, the way this is made here, it's very easy to change what is and isn't alerted for, it's all done in one place in the code.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
Nadahar: I didn't intend to make anything "less resilient", and I tried to keep the existing handling. What you want is to check that it's not an "empty object", and if it is, make it `null` instead?

bundles/org.openhab.ui/web/src/api/client/client.gen.ts
jsjames: yeah - these can't be modified - where was the typescript error?

you can run "npm run generate-rest-client" and it will re-generate the file (or discard your changes).

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
jsjames: yeah. the issue is that if we keep as an empty object, typescript thinks it has the required parameters and won't highlight the issue in other code.

bundles/org.openhab.ui/web/src/api/client/client.gen.ts
Nadahar: > yeah - these can't be modified - where was the typescript error?

I don't remember now, but that's easy to find by removing it again I guess.

But, what's the point in generating the files, which we then modify? I don't get this... how to you influence what it generates? I have made a couple of other changes to that file as well - not that I'm sure if those are event "correct", but what I don't understand is: what is the basis for the generation? Since it has to be generated, the generation must be based on something:

https://github.com/openhab/openhab-webui/pull/3710/changes#diff-3d0ca9e7ad75fc78b4751149c6220babf0a1e78c1d2fa2064ef692f0ff4783f1

jsjames: we don't modify the files that are generated. The hey-api tool automatically generates those files based on the openAPI spec (see the generate-rest-client script in the build directory).

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
jsjames: what if you added a function to the ApiError called toDebugString(), then when you log to the console, just do a quick check if that function exists:

console.error('....' + (e.toDebugString) ? e.toDebugString() : e.toString))

bundles/org.openhab.ui/web/src/api/client/client.gen.ts
Nadahar: Yes, I understand, that means that I must find other ways to solve the same issues. It's very bad that it flattens errors to strings:
```ts
        finalError = (await fn(error, response, request, opts)) as string;
```
How do we make it not to that?

jsjames: Just to make sure we are on the same page, those are just errors regarding the commented exceptions like AbortError, network errors, etc - those aren't related to the 404 ApiErrors.

If I regenerate the file and then run npm run typescript:check, I don't see any issues.

jsjames: We don't have an error interceptor - so that function doesn't get run.

I haven't studied the code enough to know why they type it as a string, however, typescript doesn't actual modify the actual javascript code that is run - its just a "build" feature to sanity check the code.  So, that conversion might cause a typescript issue, but won't effect the actual code run.

Nadahar: No, it's line 236 in client.gen.ts, long after the "catch block" for AbortError etc.

But, this is all quite messy to me, it's not clear what the structure is. For the modifications I made, I found the location using the debugger. Looking at it now, I don't get how that particular line is relevant, as it seems to be when triggering the error interceptors (and I'm not sure that there are any?).

I'll have to go back and look at the details of this again, but isn't this a strange thing to have in a generated file?
```ts
    // TODO: we probably want to return error and improve types
```

It sounds kind of half-finished to me.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
Nadahar: It's not clear to me how that would work. The "error" can be either a string, an Error or an ApiError as far as I've been able to figure out. What will `Error.toString()` result in? The message only? What about a string?

That said, I find it "cleaner" to have a function that handles this instead of having to do such checks "manually" every time - that was why I created it. Because, if you later find that it should do it slightly differently under some circumstance, you just modify the function, you don't have to touch all the places in the code and apply the same "tweak".

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -126,29 +96,90 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
-            loadingFinishedCallback(true)
-            loading.value = false
-          })
+          api.getAvailableActionsForThing({ thingUID }),
+          api.getConfigDescriptionByUri({ uri: 'thing:' + thingUID }),
+          api.getThingFirmwareStatus({ thingUID }),
+          api.getThingConfigStatus({ thingUID })
+        ]).then(([thingTypeResult, channelTypesResult, actionsResult, configDescResult, firmwareResult, configStatusResult]) => {
+          if (thingTypeResult.status === 'fulfilled') {
+            thingType.value = thingTypeResult.value && Object.keys(thingTypeResult.value).length > 0 ? thingTypeResult.value : null
```
Nadahar, line 98: @jsjames I hope this was what you meant.

Nadahar:
> No - there shouldn't be any.

I've reverted those changes - I've also reverted the modifications to `client.gen.ts`. I was struggling a long time when using the "wrong" core code that returned 204 when I thought it was returning 404, so some of those changes might have been a result of conclusions made in error because I thought the situation was different from what it was. That said, I still think the generated client looks somewhat "dodgy", but there seems to be no direct conflicts with what I'm trying to do here, so I'll just drop that.

I've applied a couple of fixes, but it's still not clear to me what the best solution to making sensible error messages from the various types of errors is, so I haven't done anything to that yet.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
Nadahar: I finally think I understand what you mean - look at 8e61a23edcf5b847eecfb32203ef86b4c3d3b298

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
jsjames: every object has a method toString, so even if called on a string, it works.
The default implementation of toString on a javascript error is to output the error and message - which is appropriate for the dialog given the change to the message ApiError.

You should still add a toDebugString() to the ApiError and generate the debug output string there.

Then, if you still want a function to determine what get's logged, put a function like getDetailedErrorMessage() in hey-api.ts and import that function where you want to use it.

Nadahar: @jsjames  Check out the latest commit and see if this is what you want.

jsjames: Yeah - that looks pretty clean to me.

Nadahar:
@jsjames I have to rebase this PR at some point, I just didn't want to do it in the middle of the reviews. I guess I can wait for Copilot's review too.

jsjames:
> @jsjames I have to rebase this PR at some point, I just didn't want to do it in the middle of the reviews. I guess I can wait for Copilot's review too.

There aren't many (if any) changes to main that would impact this PR - so the results should be still be good.

Nadahar:
> There aren't many (if any) changes to main that would impact this PR - so the results should be still be good.

Yes, the results, but GitHub can't handle the code comments when you rebase, some just get "outdated", some seem to disappear entirely. So, I think it's best to resolve all before doing a rebase.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -126,29 +86,90 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
+            if (!actionsResult.value || Array.isArray(actionsResult.value) === false) {
+              thingActions.value = []
+            } else {
+              thingActions.value = actionsResult.value
+                .filter((a) => a.visibility === 'VISIBLE' || a.visibility === 'EXPERT')
+                .filter((a) => a.inputConfigDescriptions !== undefined)
+                .sort((a, b) => (a.label || '').localeCompare(b.label || ''))
+            }
+          } else {
+            thingActions.value = null
```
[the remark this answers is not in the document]
Nadahar: @jsjames Do you have any view on this? I actually don't quite know when it's right to use an empty array and when it's right to use `null`, so I've only tried to keep whatever logic that was there. As such, I can't quite tell if this is a valid complaint or not.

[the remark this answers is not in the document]
Nadahar: The argument seems to make sense to me, the alternative is to add "protection" in the template I guess. Actually, when looking at `advancedThingActionsCount()` I think it should have some more protection in any case.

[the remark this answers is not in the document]
jsjames: Yes, I'd go with the suggestion. I'd also change line 24 (and the catch block) and set the initial value to [] vs. null as that cleans up the code elsewhere because it will always be a valid array.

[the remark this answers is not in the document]
Nadahar: @jsjames Look at my latest commit - I think it should make leaving `thingActions` as `null` safe..?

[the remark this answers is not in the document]
Nadahar: > Yes, I'd go with the suggestion. I'd also change line 24 (and the catch block) and set the initial value to [] vs. null as that cleans up the code elsewhere because it will always be a valid array.

I get that, but it would have to be set to an empty array in another place as well (in the Thing catch block). But, my real worry here is that this isn't really documented anyway, so, since all the other things dealt with here defaults to `null`, making it default to an empty array becomes "an outlier" that might cause a bug in the future. 

I can make it an empty array, but I still think the protections I made in 703583959dbb32572fb3fd0ed9511373b6d0e8ae makes sense, don't you?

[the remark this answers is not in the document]
jsjames: > I can make it an empty array, but I still think the protections I made in [7035839](https://github.com/openhab/openhab-webui/commit/703583959dbb32572fb3fd0ed9511373b6d0e8ae) makes sense, don't you?

If "null" is removed as a possible type for thingActions, typescript would catch that (when we convert the other modules to ts).  But, if you want to leave in you can since we haven't converted to ts yet across the board.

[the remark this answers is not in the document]
Nadahar: I still don't really know what TS "does", it seems to put a "fake layer" of type safety on top of JS, but it's not really true, so it's still rather confusing to me.

I think the more protection the better, and once everything is "protected" by typescript, I guess it's easy to go around removing such protections (by e.g. searching for `Array.isArray`).

Since I have both done what this suggests and added protection, I guess this point comment is resolved.

Nadahar:
I did the rebase, and think this should be ready now.

jsjames:
@florian-h05 - this LGTM now. Let me know if you want a final look before I merge.

jsjames:
> @florian-h05 - this LGTM now. Let me know if you want a final look before I merge.

@florian-h05 - i was merging stuff and inadvertently thought you had reviewed - it would be good for you to review and we can open up a new PR if you have additional feedback.

florian-h05:
I will try to have a look

Nadahar:
> I will try to have a look

@florian-h05 Is your silence that you haven't had time to look at it, or that you have no objections? I'm asking because I'm waiting to see if I should clean up the branch/PR locally or not.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -126,29 +86,90 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
+              thingActions.value = []
+            } else {
+              thingActions.value = actionsResult.value
+                .filter((a) => a.visibility === 'VISIBLE' || a.visibility === 'EXPERT')
+                .filter((a) => a.inputConfigDescriptions !== undefined)
+                .sort((a, b) => (a.label || '').localeCompare(b.label || ''))
+            }
+          } else {
+            thingActions.value = []
+            const message = getErrorMessage(actionsResult.reason, true)
```
jsjames, line 126: @Nadahar - we should add back in the check for 404 and only log as info vs. error.

```
      if (e.response?.statusText === 'Not Found' || e.response?.status === 404) {
        console.log('No actions available for this Thing')
        return
      {
```

Nadahar: That depends a bit on what the 404 is about as I see it, if you're trying to retrieve a Thing that doesn't exist, it's definitely an "error" as I see it - but other situations can be different. https://github.com/openhab/openhab-core/pull/5236 changes "no Thing actions" to 204, and if there are other such "no error situations", it might be better to have them return 204. It's hard to set one "level" that fits all scenarios...

Nadahar: Do you regularly see errors logged now? If so, from which endpoint is it?

jsjames: If a valid thing doesn't have actions, we definitely don't want to report as error  - I'm definitely seeing that error for valid things in my log. Let's put it back in since that is the way florian had the code originally. Long-term, if core is changed to report a 204 when there are no actions, then it can be removed.

Nadahar: I'm not expecting that to be "long term", the PR should hopefully be quick - it's simple. I just want this PR to be "confirmed finished" by @florian-h05 before I make it ready for review. But, you basically confirm that the problem is only with the Thing actions?

jsjames: Yes, if you look at the original code, that was the only case where the 404 was treated differently.

Nadahar: > Yes, if you look at the original code, that was the only case where the 404 was treated differently.

Ok, but if so, then https://github.com/openhab/openhab-core/pull/5236 should solve it. We're not about to have a milestone release, are we? Because, if not, this will only be a "problem" in a few snapshots.

I don't know if @florian-h05 is very busy or away, but I think I should just make the core PR ready for review. It's hard to see how any of those changes should be a problem, and it has already been approved except for the change to the thing actions - which is clearly needed since a "workaround" was needed in MainUI.

Nadahar: I've opened https://github.com/openhab/openhab-core/pull/5236 for review, so let's see if it can be solved swiftly.

Nadahar: https://github.com/openhab/openhab-core/pull/5236 has been merged, so there should be no need for a workaround.

florian-h05:
Sorry for now following up on this ... been to FOSDEM in Brussels last weekend representing openHAB and spent a few days in Brussels after. 

I have created #3861 with a few changes:
- Always log an error, as this logs the object to the console, it might give more detail than only the message.
- Remove the `... = null` statements in the error handling => null is set as the initial value when the vars are declared.

Nadahar:
> * Remove the `... = null` statements in the error handling => null is set as the initial value when the vars are declared.

It's supposed to be there, in case `load()` is called more than once - in the future, when it will refresh itself based on events from core for example. It's "costs" so little to set it to `null` in the error handler, and it makes sure that *if* the code is run multiple times for some reason, and one of them fails, that the "old" variable isn't kept.

Nadahar:
> * Always log an error, as this logs the object to the console, it might give more detail than only the message.

The only reason I didn't was that you previously said to only log the message - at least that's what I understood. I agree that "more is better" in the console 👍

florian-h05:
> It's "costs" so little to set it to null in the error handler

I restored your nulling.

PR #3656, opened 2025-12-24
Script editor: Allow return outside functions/global return

florian-h05:
JS Scripting supports global return thanks to the wrapper, but ESLint displays a parser error.

Reported on the community:
https://community.openhab.org/t/openhab-5-1-release-discussion/167620/53

PR #3712, opened 2026-01-04
Script editor: Fix linter config

florian-h05:
Regression from #3656.


florian-h05:
Should be backported as well.

PR #3713, opened 2026-01-04
Script editor: Enable ESLint rules to catch bugs 

florian-h05:


jsjames:
I didn't see this was still in draft. It might be worth @jimtng reviewing as well since he updated the code-mirror code.

jsjames:
I presume this one should not be backported?

florian-h05:
It was draft because I needed to rebase once a different PR had been merged.
This shouldn't be backported.

I don't think Jim is needed for review, but if you think so, let's wait. This only enables a few rules to catch common errors, the openhab-js globals should be complete (I am openhab-js maintainer as well ;-))

Issue #3378, opened 2025-10-11
To-Dos wrt Vue 3 and Framework7 v7 upgrade


florian-h05:
After https://github.com/openhab/openhab-webui/pull/3350 is merged, the following tasks need to be tackled:

### High Priority

- [x] Take care of `TODO-V3.1` comments
- [x] **(BUG)** Duplicate entries in model treeview on drag & drop sometimes, see https://github.com/openhab/openhab-webui/issues/3378#issuecomment-3457221006 => fixed by #3476??  -- Broke out as a separate issue that can be addressed over time when the model UI code gets updated to typescript/composition api.
- [x] Refactor to use Vue router instead of F7 Router as F7 Router has serious bugs with navigation??
- [x] **(BUG)** Blockly multi drag/drop not working - recommended solution is upgrading the library, but it requires blockley v11 -> will be fixed by https://github.com/openhab/openhab-webui/pull/3398
- [x] Write down which manual changes are required for swiper
- [x] **(BUG)** ECharts localisation not working
- [x] **(BUG)** Model drag & drop not working
- [x] Store menu pinned state.
- [x] **(BUG)** on initial loading of setup-wizard, the UI is unresponsive. Corrects itself on reloading
- [x] Test floorplan pages
- [x] Test map views with markers
- [x] Test embedded SVG state tracking
- [x] **(BUG)** slider does not appear for dimmer items on item-detail page on ios devices.
- [x] Test masonry layout
- [x] **(BUG)** back button from settings/items, settings/things - the URL in the address bar does not change.
- [x] **(BUG)** Item state preview is missing in production build
- [x] **(BUG)** Add link and edit link not working on Items details page
- [x] Test Web Audio Support
- [x] **(BUG)** on item-detail->oh-label-card->oh-trend - the width setting is only correct AFTER the page is display - doesn't seem to be reactive
- [x] ~oh-swiper: Automatically map parameters object properties onto config object to ease transition~ => manual changes are needed anyway

### Low Priority

- [x] Run Prettier & enable prettier format checking in CI
- [x] Upgrade marked: 13 -> 14 breaks markdown list rendering: #3718
- [x] Clean up tabs - there's a lot of unecessary code that is handled automatically by Vue, e.g. `@click=switchTab`
- [x] Clean-up command emits/listening for generic-widget-component, I haven't found a place 'command' is emitted from
- [x] Refactor `$oh.api` from using Framework7 Requests to:
    a. Use the fetch API directly #3575
[...]

PR #3717, opened 2026-01-04
Marked update to 17.0.1

jsjames:
Update to v17.0.1 - I don't see the commented issue in #3378 about it breaking list rendering in v14. This must have been fixed in one of the interim versions between v14 and v17.


jsjames:
Incorrect pull request.

Issue #3660, opened 2025-12-25
Very strange behavior of the semantic model tree (collapse on edit)


yndtrud:
## The problem

The bug is back!!!! It was introduced to the public in version 4.X and fixed in 5.0.0. 
It was absent from 5.0.1–5.0.3 and returned again in 5.1.0 in an even more aggressive manner!

When I add, delete and change elements or can simply add/delete/change element metadata, the tree very often collapse. Previously (3.X.X and 5.0.1–5.0.3), it was always stable. Now it has become extremely inconvenient to work with it.

Of course, there is a button that expands the entire model, but this is also inconvenient, since I have a rather deep and rich model and I have to open (search) for elements in a multi-level model for a very long time. In addition, this does not save from the fact that the model tree can close under certain circumstances.

## Expected behavior

The tree should not close, twitch or blink.

## Steps to reproduce

Just use OH as usual.

## Your environment

```yaml
runtimeInfo:
  version: 5.1.0
  buildString: Release Build
systemInfo:
  javaVersion: 21.0.9
  javaVendor: Eclipse Adoptium
  javaVendorVersion: Temurin-21.0.9+10
  osName: Windows Server 2025
  osVersion: "10.0"
  osArchitecture: amd64
  availableProcessors: 4
  freeMemory: 84251320
  totalMemory: 536870912
  uptime: 188068
  startLevel: 100
addons:
  - automation-jsscripting
  - binding-astro
  - binding-asuswrt
  - binding-exec
  - binding-gpstracker
  - binding-ipcamera
  - binding-lgthinq
  - binding-logreader
  - binding-mqtt
  - binding-network
  - binding-openweathermap
  - binding-serial
  - binding-smsmodem
  - binding-telegram
  - misc-metrics
  - persistence-jdbc-postgresql
  - persistence-mapdb
  - persistence-rrd4j
  - transformation-basicprofiles
  - transformation-jsonpath
  - transformation-map
  - transformation-regex
  - transformation-scale
  - ui-basic
  - ui-habpanel
  - voice-rustpotterks
  - voice-voicerss
  - voice-voskstt
clientInfo:
  device:
    ios: false
    android: false
    androidChrome: false
    desktop: true
    iphone: false
    ipod: false
    ipad: false
    edge: false
    ie: false
    firefox: false
    macos: false
    windows: true
    cordova: false
    electron: false
    capacitor: false
    nwjs: false
    webView: false
    webview: false
    standalone: false
    os: windows
[...]

PR #3732, opened 2026-01-06
Model: Fix expansion state reset on returning into model

mherwege:
Closes https://github.com/openhab/openhab-webui/issues/3660

This moves the expansion state and the currently selected item to the runtime store, so it can be retrieved and rendered again when returning to the model after branching out.

This is reset when going into the model from the settings menu.


florian-h05:
Backported this as well as #3645 so I don't have to resolve conflicts.

PR #3736, opened 2026-01-07
Appy Prettier & Enable in CI

florian-h05:


florian-h05:
Prettier is doing really weird stuff, e.g. in health-orphanlinks:

```xml
          <f7-list-button color="red" @click="purgeAllManaged()">
            Purge all managed links (will purge {{ purgeableLinksCount }} managed link{{ plural(purgeableLinksCount)






            }})
          </f7-list-button>
```

bundles/org.openhab.ui/web/src/types/components/widgets/chart/oh-category-axis.ts
```diff
@@ -1,8 +1,6 @@
 // note: this file is generated and should not be edited by hand
 
-import {
-  NameLocation
-} from '../common.ts'
+import { NameLocation } from '../common.ts'
```
jsjames, line 3: We should add a .prettierignore for all "generated" code.

jsjames:
I don't see an option in prettier to include a space before the parenthesis in the function declaration.  I had thought this was important to maintain? thoughts?

florian-h05:
I also think so it's important. There are other reports of people having same issue with prettier online, probably we should try out some alternatives.

florian-h05:
I have created #3742 using oxfmt (from the Oxidation Compiler project, as drop-in replacement for Prettier and super fast), seems it's doing a better job with the current config. I think the space before the function parentheses can be left out, more importantly oxfmt doesn't mess up with the template as Prettier did.

PR #3741, opened 2026-01-09
oh-list: Fix various issues when editing

florian-h05:
Reported on the community: https://community.openhab.org/t/openhab-5-1-release-discussion/167620/363


florian-h05:
This should be backported

PR #3745, opened 2026-01-09
oh-list: Fix overflow on narrow screens

florian-h05:
Regression from #3741.


jsjames:
backport? given its a regression?

florian-h05:
I've backported it given its a regression from a backported bug fix.

PR #3750, opened 2026-01-10
Fix stylesheet config not applied to oh-block, oh-grid-col, oh-grid-row

florian-h05:


florian-h05:
This should be backported.

florian-h05:
Backported a few seconds ago :+1:

PR #3687, opened 2025-12-29
oh-map/oh-plan: Fix various Leaflet related issues

florian-h05:
Fixes #3679.

Fixes the following issues:
- map-edit/plan-edit previews behind toolbar.
- Clean-up Leaflet `Icon.Default` code from oh-map-page/oh-plan-page as it has no effect.
- oh-plan-marker: Fix default icon not displayed.
- oh-plan-page: Fix failure in plan-edit preview.
- oh-plan-page: Fix widget expressions for icons not evaluated in design tab.

Code enhancements:
- oh-map-marker: Use default icon from Leaflet dist instead of base64 encoded.
- map-edit/plan-edit: Use a single oh-*-page component for the preview, not one for each tab.

PR #3752, opened 2026-01-11
oh-map-/oh-plan-marker: Fix incorrect marker/tooltip positioning

florian-h05:
Regression from #3687.

This ensures icon size is always set explicitly.
I decided to remove the icon shadow as I didn't figure out how to place the shadow correctly and displaying shadow for the default icon, but not for custom icons is inconsistent.


florian-h05:
Has to be backported as the regression has been backported as well.

PR #3753, opened 2026-01-11
oh-list: Fix list rendering if not editmode

florian-h05:
Regression from #3741.


florian-h05:
Fixes:

image

florian-h05:
Has to be backported as the regression has been backported as well.

PR #3758, opened 2026-01-15
Widgets: Fix default slots overwritten with empty array

florian-h05:
Regression from #3741.


florian-h05:
Has to be backported as #3741 was backported.

jsjames (approved):

PR #3760, opened 2026-01-15
oh-canvas-item: Fix 0 values for x, y, w, h config parameters ignored

florian-h05:


jsjames:
backport? since this is a real bug

florian-h05:
Yes, also low risk of regression

PR #3761, opened 2026-01-15
f7-list-group css: Fix style regression from grouped list indentation fix

florian-h05:
Fixes an issue where padding was missing as the CSS variables were set to 0 for elements in: f7-list > f7-list-group > f7-list-item

It is enough to remove the left padding for the list-group itself, we don't want to remove it for its list items as well.

Regression from #3350.


jsjames:
backport?

florian-h05:
Yes please

PR #3762, opened 2026-01-15
oh-swiper: Only unset width if slidesPerView set to auto

florian-h05:
Regression from #3653.

When setting a fixed size per slide, it is not applied anymore as !important overwrites it.
When setting slidesPerView to auto, however width is calculated improperly, so only apply the fix from #3653 if slidesPerView is actually set to auto.
Reported on the community: https://community.openhab.org/t/universal-toggle/115500/29


florian-h05:
This should be backported as the PR causing the regression has also been backported.

PR #3763, opened 2026-01-15
Pinia stores: Make exported state & getters in readonly where appropiate

florian-h05:


jsjames:
I submitted a PR to your branch. I have the same simplification for the useComponentStore regarding the loadWidgetsAndPages simplification - so it looks good.

PR #3765, opened 2026-01-15
Item details: Display persistence status

mherwege:
This PR adds a section on persistence in item detail screen. It will:

1. Show the different persistence services and indicate if the current item is configured to be persisted and the strategies used.
2. Provides a link to the specific service configuration if editable.
3. Show a general link to persistence service configurations.
4. Shows a badge indicating if the item is actually in the persistence store (with a count of persisted values). This badge will only show if the persistence service has implemented the getItemInfo() method, and not just returns an empty list. This is a contraint of the persistence service implementations, to be revisited in the the persistence addons.

Here is a screenshot:
image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
+++ b/bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
@@ -9,7 +9,7 @@
           :title="p.label"
           :subtitle="strategies[p.id]?.join(', ') || 'Not persisted'"
           :badge="persistedBadges[p.id]"
-          badge-color="green"
+          :badge-color="persistedBadges[p.id] === '0' ? 'red' : 'green'"
           :class="!p.editable ? 'item-persistence-badge-non-editable' : ''"
           :link="p.editable ? '/settings/persistence/' + p.id : null">
           <template #media>
@@ -73,11 +73,7 @@ const persistedBadges = computed(() => {
       badges[service.id] = null
     } else if (persisted === 'not_persisted') {
       // Item has not been persisted
-      // badges[service.id] = '0'
-      // TODO: Currently, the API call does not distinguish between API call not supported for service and item not persisted.
-      // To avoid showing misleading information, don't show badge if we don't get anything back (should be covered by first case if REST API is fixed.)
-      // This can be reverted to showing 0 if the API is fixed.
-      badges[service.id] = null
+      badges[service.id] = '0'
     } else if (persisted.count != null && Number.isInteger(Number(persisted.count))) {
       // Show numeric count when available
       badges[service.id] = String(persisted.count)
@@ -109,21 +105,24 @@ const loadService = async (service: api.PersistenceService): Promise<Persistence
   try {
     serviceConfig = (await api.getPersistenceServiceConfiguration({ serviceId: service.id })) ?? null
   } catch (err: unknown) {
-    if (err instanceof ApiError && (err.response.statusText === 'Not Found' || err.response.status === 404)) {
+    if (err instanceof ApiError && err.response.status === 404) {
       // No configuration yet, continue with an empty one
     } else {
       console.debug('Error loading persistence config for', service.id, err)
     }
   }
 
-  let itemsPersisted: Array<api.PersistenceItemInfo> | 'not_persisted' | 'unsupported' = 'unsupported'
+  let itemsPersisted: api.PersistenceItemInfo | 'not_persisted' | 'unsupported' = 'unsupported'
   try {
-    itemsPersisted = await api.getItemsForPersistenceService({ serviceId: service.id }) ?? 'unsupported'
+    itemsPersisted = (await api.getItemsForPersistenceService({ serviceId: service.id, itemname: props.item.name }))![0]!
   } catch (err: unknown) {
-    if (err instanceof ApiError && (err.response.status === 400)) {
-      // Not supported for service, leave itemsPersisted null
+    if (err instanceof ApiError && err.response.status === 404) {
+      itemsPersisted = 'not_persisted'
+    } else if (err instanceof ApiError && err.response.status === 405) {
+      // Persistence service not queryable or getting Item info not allowed
+      itemsPersisted = 'unsupported'
     } else {
-      console.debug('Error loading persistence items for', service.id, err)
+      console.error('Error loading persistence items for', service.id, err)
     }
   }
 
@@ -132,7 +131,7 @@ const loadService = async (service: api.PersistenceService): Promise<Persistence
     configs: serviceConfig?.configs ?? [],
     aliases: serviceConfig?.aliases ?? {},
     editable: serviceConfig?.editable === undefined ? true : serviceConfig?.editable,
-    persisted: Array.isArray(itemsPersisted) ? (itemsPersisted.find((item) => item.name === props.item.name) ?? 'not_persisted') : itemsPersisted
+    persisted: itemsPersisted
   } satisfies PersistenceService
 }
 


bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
```diff
@@ -0,0 +1,128 @@
...
+      persistenceConfig?.configs?.forEach((config) => {
+        const items = config.items
+        if (items && config.strategies?.length) {
+          let match = false
+          // First find all positive matches for the item
+          if (items.includes('*') || items.includes(this.item.name) || items.filter((configItem) => !configItem.startsWith('!') && configItem.endsWith('*')).map((configItem) => configItem.slice(0, -1)).find((groupName) => this.item.groupNames.includes(groupName))) {
+            match = true
+          }
+          // Remove negative matches
+          if (items.includes('!' + this.item.name) || items.filter((configItem) => configItem.startsWith('!') && configItem.endsWith('*')).map((configItem) => configItem.slice(1, -1)).find((groupName) => this.item.groupNames.includes(groupName))) {
```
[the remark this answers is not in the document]
mherwege: I believe there still is a lot of code duplication in your proposal and the names and arguments of the match methods do not clearly convey the meaning. I have created an alternative.

florian-h05:
During testing, I discovered that persistence services such as InMemory return an array of `null` if they aren't able to provide information about the persisted items. Will https://github.com/openhab/openhab-core/pull/5336 fix that?
I had to add filtering for this in this PR, otherwise the UI would fail due to a null access.

mherwege:
> During testing, I discovered that persistence services such as InMemory return an array of `null` if they aren't able to provide information about the persisted items. Will [openhab/openhab-core#5336](https://github.com/openhab/openhab-core/pull/5336) fix that? I had to add filtering for this in this PR, otherwise the UI would fail due to a null access.

I am surprised by this. InMemory does return information about the persisted items and does return an empty set if there are no items peristed. From the InMemory addon code, I don't see how it returns a null value in that set:

```
    public Set<PersistenceItemInfo> getItemInfo() {
        return persistMap.entrySet().stream().map(this::toItemInfo).collect(Collectors.toSet());
    }
```

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
```diff
@@ -33,124 +33,135 @@
...
-        groupPatterns = items.filter((configItem) => !configItem.startsWith('!') && configItem.endsWith('*')).map((configItem) => configItem.slice(0, -1))
-      } else {
-        itemPattern = '!' + itemName
-        groupPatterns = items.filter((configItem) => configItem.startsWith('!') && configItem.endsWith('*')).map((configItem) => configItem.slice(1, -1))
+  return {
+    ...service,
+    configs: serviceConfig?.configs ?? [],
+    aliases: serviceConfig?.aliases ?? {},
+    editable: serviceConfig?.editable === undefined ? true : serviceConfig?.editable,
+    persisted: itemsPersisted.find((item) => item.name === props.item.name) ?? null
```
mherwege: In my code, if the item was not found, I used an empty object, not null. That's what caused the null in your changes.

mherwege:
With https://github.com/openhab/openhab-core/pull/5336 and https://github.com/openhab/openhab-addons/pull/20200 there would indeed be much better information available for this.
It will:
- clearly distinguish between nothing persisted and not implemented.
- allow calling the API for a single item, so you do not have to query all every time.

This would change the API as it would return a 404 (not found) if nothing was found, and a 405 (not allowed) if the method was not implemented for the service, making a clear distinction. So this code would have to be updated accordingly as well.
The single item query would actually always return something if the item has been persisted because there is a default implementation that just queries the last persisted state. So for a single item request, a 405 would not happen.

The core and addon PR's where a consequence of what I tried to achieve here, which I found limiting. This should work without these other PR's. But if the other PR's are accepted and merged, this should be revisited to see if adjustments are needed to improve it.

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
florian-h05: I don't think the issue was caused by my code change. The issue is that `itemsPersisted` included `null` entries and I don't see how

```js
itemsPersisted ? (itemsPersisted.find((item) => item.name === this.item.name) || {}) : null
```

avoided the null access when calling `item.name`.

florian-h05:
> I am surprised by this. InMemory does return information about the persisted items and does return an empty set if there are no items peristed. From the InMemory addon code, I don't see how it returns a null value in that set:

I neither, but it definitely does for me. I'll debug this further.

florian-h05:
Ready for merging?

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
mherwege: Sorry, but I don't see it. Look at the Java code above from inMemory DB (https://github.com/openhab/openhab-addons/blob/d525f7760a829c206f46e4c98970e10e19c95a6b/bundles/org.openhab.persistence.inmemory/src/main/java/org/openhab/persistence/inmemory/internal/InMemoryPersistenceService.java#L114). The `toItemInfo` (https://github.com/openhab/openhab-addons/blob/d525f7760a829c206f46e4c98970e10e19c95a6b/bundles/org.openhab.persistence.inmemory/src/main/java/org/openhab/persistence/inmemory/internal/InMemoryPersistenceService.java#L195) method being called is defined as:

```
    private PersistenceItemInfo toItemInfo(Map.Entry<String, PersistItem> itemEntry) {
        Lock lock = itemEntry.getValue().lock();
        lock.lock();
        try {
            String name = itemEntry.getKey();
            Integer count = itemEntry.getValue().database().size();
            Instant earliest = itemEntry.getValue().database().first().timestamp().toInstant();
            Instant latest = itemEntry.getValue().database.last().timestamp.toInstant();
            return new PersistenceItemInfo() {

                @Override
                public String getName() {
                    return name;
                }

                @Override
                public @Nullable Integer getCount() {
                    return count;
                }

                @Override
                public @Nullable Date getEarliest() {
                    return Date.from(earliest);
                }

                @Override
                public @Nullable Date getLatest() {
                    return Date.from(latest);
                }
            };
        } finally {
            lock.unlock();
        }
    }
```
For each entry, this will always return a name and it cannot be null. If it would, this code would not compile. It can only return an empty set if there are no entries, but that is not the same as a set with null entries.

florian-h05: The issue is in serialisation of the anonymous implementations of PersistenceItemInfo used in InMemory (and the PersistenceResource when resolving aliases). This serialisation fails and returns the null value.

florian-h05: The serialisation in general is error-prone because serialisation depends on the field names, which may differ between the different PersistenceItemInfo implementations from different persistence services.
I wiill fix this.

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
```diff
@@ -0,0 +1,170 @@
...
+  services.value.forEach((service) => strategies[service.id] = serviceStrategies(service))
+  return strategies
+})
+
+const persistedBadges = computed(() => {
+  const badges: Record<string, string | null> = {}
+  services.value.forEach((service) => {
+    const persisted = service.persisted
+    if (!persisted) {
+      // Query for persistence items not supported
```
mherwege, line 72: The behaviour here is different from my code. In my code, when the item is not found in the persistence service {} is returned, which was not falsy. So with the current REST API this would always pass. This should only be null when it is not supported and without the changes in the REST API we do not know, so we should not bail out here immediately.

mherwege:
> Ready for merging?

Not quite. I think your refactoring still changes the logic. The end result will be the same, but only because the new REST API hasn't been fully implemented yet (and this code adapted to that). I prefer keeping the difference between null and {} so the changes will be easier afterwards. Minimal changes would all be covered in the TODO.

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
```diff
@@ -0,0 +1,170 @@
...
+    } else {
+      console.debug('Error loading persistence config for', service.id, err)
+    }
+  }
+
+  let itemsPersisted: Array<api.PersistenceItemInfo> = []
+  try {
+    itemsPersisted = await api.getItemsForPersistenceService({ serviceId: service.id }) ?? []
+  } catch (err: unknown) {
+    if (err instanceof ApiError && (err.response.statusText === 'Not Found' || err.response.status === 404)) {
```
mherwege: At the moment, no error code is generated from the API at all.
In my proposal in the core PR, this would become 405 instead of 404.
404 would be used for a single item PeristenceItemInfo query with item not found, or for a wrong serviceId parameter.

florian-h05:
https://github.com/openhab/openhab-core/pull/5387 fixes the issue with InMemory persistence and potentially also other services.

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
florian-h05: Sorry, messed up the response code when refactoring, will fix that.

mherwege:
> [openhab/openhab-core#5387](https://github.com/openhab/openhab-core/pull/5387) fixes the issue with InMemory persistence and potentially also other services.

Great. Note I already had a very similar fix for this included in https://github.com/openhab/openhab-core/pull/5336.

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
mherwege: Mine would have been wrong anyway as well. I had 400. The REST API does not generate that now, or in the future.

florian-h05:
@mherwege Please have a look at my latest changes, these should fully restore the previous logic.

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
```diff
@@ -0,0 +1,172 @@
...
+const persistedBadges = computed(() => {
+  const badges: Record<string, string | null> = {}
+  services.value.forEach((service) => {
+    const persisted = service.persisted
+    if (persisted === 'unsupported') {
+      // Query for persistence items not supported
+      badges[service.id] = null
+    } else if (persisted === 'not_persisted') {
+      // Item has not been persisted
+      // badges[service.id] = '0'
```
mherwege: If the core PR is merged, the TODO can be removed and this should be:
```
badges[service.id] = '0'
```

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      console.debug('Error loading persistence config for', service.id, err)
+    }
+  }
+
+  let itemsPersisted: Array<api.PersistenceItemInfo> | 'not_persisted' | 'unsupported' = 'unsupported'
+  try {
+    itemsPersisted = await api.getItemsForPersistenceService({ serviceId: service.id }) ?? 'unsupported'
+  } catch (err: unknown) {
+    if (err instanceof ApiError && (err.response.status === 400)) {
```
mherwege: This should be 405 after the core changes.

bundles/org.openhab.ui/web/src/components/persistence/item-persistence-details.vue
```diff
@@ -0,0 +1,172 @@
...
+    if (err instanceof ApiError && (err.response.statusText === 'Not Found' || err.response.status === 404)) {
+      // No configuration yet, continue with an empty one
+    } else {
+      console.debug('Error loading persistence config for', service.id, err)
+    }
+  }
+
+  let itemsPersisted: Array<api.PersistenceItemInfo> | 'not_persisted' | 'unsupported' = 'unsupported'
+  try {
+    itemsPersisted = await api.getItemsForPersistenceService({ serviceId: service.id }) ?? 'unsupported'
```
mherwege: After the core change, it would be possible to use:
```
itemPersisted = await api.getItemsForPersistenceService({ serviceId: service.id, itemName: item.name }) ?? 'unsupported'
```
If the item is not in the persistence store, a 404 would be returned.

This would have 2 advantages:
1. The size of the JSON returned is smaller.
2. It also returns something if the general getItemInfo is not supported (there is a default implementation that just checks if it is persisted).

mherwege:
@florian-h05 What is your view on this? Do we wait for the core and addons PR's to be merged to finalize this? The remaining remarks for me all have to do with the changes as done in the core PR, that will make this more consistent and give better information.
I think the addon PR should also be reviewed and merged at the same time as that one will make sure the UnsupportedOperationException is thrown, instead of still returning an empty list for getItemInfo when not supported. That's ultimately what will make the difference.

In our code base, there is one more call to `getItemInfo`: https://github.com/openhab/openhab-webui/blob/95d5395181fcf069e07333b9ce43704fef60ad44/bundles/org.openhab.ui.cometvisu/src/main/java/org/openhab/ui/cometvisu/internal/backend/rest/DataProviderResource.java#L207

This can never have worked, as RRD4J never implemented this and always returned an empty Set. But I don't know if it will still compile whitout adjusting it for the exception that can be thrown.

florian-h05:
> Do we wait for the core and addons PR's to be merged to finalize this?

Let's wait a few days and hope the core PR gets merged soon.

florian-h05:
> This can never have worked, as RRD4J never implemented this and always returned an empty Set. But I don't know if it will still compile whitout adjusting it for the exception that can be thro

We'll see. If compilation breaks, either you or me can fix it quickly ;-)

florian-h05:
Your core PR has been merged. Do you want to adjust this PR or should I do it (though I‘ll be away from keyboard for a few days)?

mherwege:
I can adjust it, but it may take me a few days as well. It is a bit busy at the moment.

florian-h05:
@mherwege Good news! I had some spare time to finish this PR here. FYI, I had to rebase to get the API client changes in from the main branch. Please have a look at my changes from https://github.com/openhab/openhab-webui/pull/3765/commits/55145cbe82421d0cee86b1a2c50840ae564e2477 and confirm they are fine, so I can merge after you had a look.

mherwege:
@florian-h05 I didn't do further testing, but the code looks good to me. Thanks.

PR #3775, opened 2026-01-18
Prevent padding metadata value with space when value is missing

jimtng:
Refs https://github.com/openhab/openhab-core/pull/5283


florian-h05:
@jimtng What‘s the status here? The linked core PR has been merged,

jimtng:
I meant to do some more tests or at least look through, but I really haven't had time. We can merge this and if there's any problems later, we can address it then? wdyt?

florian-h05:
I think this should work fine, so let’s merge it.
Can you please undraft it?

PR #3776, opened 2026-01-18
useThingEditStore: Fix firmware data no reset if not available

### LINKED ISSUE #3774: Firmware versions are incorrectly displayed after changing things
## The problem

The firmware versions and properties are not loaded correctly when changing things. If I go to a thing that has firmware versions, then change to a different thing that has no firmware versions, the previous data is still showing.

Image

## Expected behavior

The firmware status for the currently displayed thing should be displayed.

jsjames:
Fixes #3774

@florian-h05 - I also reset the thingActions on any failure - which seemed the right place.

Note, I don't have any things in my system which returns firmware, so I wasn't able to test, but pretty sure this is the fix.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
@@ -75,9 +75,9 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
         .filter((a) => a.inputConfigDescriptions !== undefined)
         .sort((a: ThingAction, b: ThingAction) => a.label.localeCompare(b.label))
     } catch (e: any) {
+      thingActions.value = []
       if (e === 'Not Found' || e === 404) {
         console.log('No actions available for this Thing')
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -77,9 +77,9 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
     } catch (e: any) {
       if (e === 'Not Found' || e === 404) {
         console.log('No actions available for this Thing')
-        thingActions.value = []
```
florian-h05, line 80: I think this change is not correct. If a 404 is received, then the thingActions aren't reset anymore because of the return below.
You have to keep this statement, but you can add the one below.

jsjames: correct. Just submitted a change.

florian-h05 (approved):

PR #3781, opened 2026-01-19
widget-mixin: Fix JS error in childWidgetContext if widget not found

florian-h05:


florian-h05:
This should be backported.

PR #3788, opened 2026-01-21
useUIOptionsStore: Make darkMode reactive

### LINKED ISSUE #3757: [MainUI] In settings rename Dark mode to Color mode; align terminology with Basic UI
In openHAB 5.1.1. under `/about` the option group between Theme and Navigation bars style is called “Dark mode”. In fact Dark mode is one of the available options, and it weights equal to the “Light mode”.  The same setting in Gnome is called Appearance/Style, in Firefox Appearance/Color Scheme, in Chromium Appearance/Mode.  Basic UI calls it “Theme”.

I propose renaming the group to Color mode, Color scheme, Theme, Mode; and aligning the terminology for this preference between Basic UI and Main UI.

Image

Image

### LINKED ISSUE #3787: [Main UI] Code DSL tab does not switch automatically from dark to light mode
In openHAB 5.1.1 I load for a thing the Code tab and show there DSL, while my desktop and browser are in dark mode.  Then I change the desktop, and implicitly the browser, to light mode.  The code tab stays in dark mode (black background) while the remailing UI changes to white background.

If I chick somewhere else (`GET /settings/model`) and then reload the Code tab, then the latter shows in white background.

Image

jsjames:
Fixes #3787, Resolves #3757

- Made darkMode a computed property so it is now reactive across the app, also now has both a get/set function
- Changed all references to darkMode vs getDarkMode() 
- Changed "Dark Mode" title on About page to "Appearance Mode"
- Centralized all f7 darkMode operations into useUIOptionsStore

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
+++ b/bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
@@ -1,7 +1,7 @@
 {
   "about.theme": "Theme",
   "about.theme.auto": "Auto",
-  "about.darkMode": "Appearance Mode",
+  "about.darkMode": "Appearance",
   "about.darkMode.auto": "Auto",
   "about.darkMode.light": "Light",
   "about.darkMode.dark": "Dark",


jsjames:
@digitaldan - I did test this to make sure it works with the ios app, but if you get a chance to test as well, that would be great! Thx

bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
```diff
@@ -1,7 +1,7 @@
 {
   "about.theme": "Theme",
   "about.theme.auto": "Auto",
-  "about.darkMode": "Dark mode",
+  "about.darkMode": "Appearance Mode",
```
florian-h05: Appearance mode sounds a bit awkward. Either keep "Dark Mode" or make it "Appearance" only.

bundles/org.openhab.ui/web/src/js/stores/useUIOptionsStore.ts
```diff
@@ -58,35 +61,46 @@ export const useUIOptionsStore = defineStore('uiOptions', () => {
...
+      storedDarkMode.value = value
+      if (value === 'auto') {
+        f7.enableAutoDarkMode()
+        localStorage.removeItem('openhab.ui:theme.dark')
+      } else {
+        f7.disableAutoDarkMode()
+        localStorage.setItem('openhab.ui:theme.dark', value)
+      }
+
+      bars.value = 'light' // Reset bars to light when dark mode changes
```
florian-h05, line 78: I am not sure people will like this behaviour ...

jsjames: are you referring to the reset of bars to 'light'. This is how the current code is implemented and I'm pretty sure how it has been all along.

[edit] i'm okay removing though if you think that is the right thing to do.

florian-h05: Ah sorry, didn’t notice this was the current behaviour, then it’s fine. 
I’m just cautious with changing things.

PR #3815, opened 2026-01-26
oh-plan/map-page: Fix stylesheets for leaflet pages

jsjames:
Missing scopeCss dIv at top level.

Reference: https://community.openhab.org/t/page-floorplan-change-label-tooltip-backgound/168119/2


florian-h05 (approved):
Thanks!

Do you think we should backport this? I would think this is a regression from #3350.

jsjames:
> Thanks!
> 
> Do you think we should backport this? I would think this is a regression from #3350.

Yeah, its low-risk and enables/re-enables a capability.  I just thought it wasn't ever implemented - but your probably right it was a regression.

PR #3817, opened 2026-01-26
oh-masonry: Fix style leak & Clean-up unused class from home

florian-h05:
Fixes issues where dropdown menu items where displayed behind menu buttons.


florian-h05:
Should be backported.

PR #3838, opened 2026-02-03
Fix auto dark mode selection in Android app

### LINKED ISSUE #3835: Main UI always opens in dark mode inside openHAB app
### Describe the bug
When opening the openHAB Main UI from within the openHAB mobile app, the UI is always displayed in dark mode, regardless of the Main UI theme setting or browser/system preference.

### Expected behavior
The Main UI should respect either:
the Main UI theme setting, or
the system/browser theme (similar to opening Main UI in a web browser)
Alternatively, there should be an option to override the theme behavior for Main UI inside the app.

### Steps to reproduce
Open the openHAB mobile app
Navigate to Main UI
Observe that Main UI is always displayed in dark mode

### Additional context
When opening Main UI in a regular web browser, the theme behaves as expected and follows the browser or system setting. The issue only occurs when Main UI is opened inside the openHAB app.

maniac103:
The app callback 'preferDarkMode()' doesn't return a boolean, but a string ('light' or 'dark'). Since either string evaluates truthy, the UI always selected dark mode.

Fixes #3835

### DIFF
--- a/bundles/org.openhab.ui/web/src/types/OHApp.d.ts
+++ b/bundles/org.openhab.ui/web/src/types/OHApp.d.ts
@@ -1,8 +1,8 @@
 // Android app: https://github.com/openhab/openhab-android/blob/c7b942c/mobile/src/main/java/org/openhab/habdroid/ui/activity/AbstractWebViewFragment.kt#L401
 // iOS app: https://github.com/openhab/openhab-ios/blob/da9a7e2/openHAB/OpenHABWebViewController.swift#L37
 interface OHApp {
-  preferTheme?: () => string
-  preferDarkMode?: () => string
+  preferTheme?: () => 'md' | 'ios' | 'aurora'
+  preferDarkMode?: () => 'light' | 'dark'
   exitToApp?: () => void
   goFullscreen?: () => void
   pinToHome?: () => void


florian-h05:
Is this also the case for the iOS app?

florian-h05:
If so, this change is fine, but we need to also adjust the type declaration then.

maniac103:
> Is this also the case for the iOS app?

AFAIK the iOS app doesn't use that callback.

> If so, this change is fine, but we need to also adjust the type declaration then.

The [type declaration](https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/types/OHApp.d.ts#L5) is already correct (and says return value is `string`).

jsjames:
> > Is this also the case for the iOS app?
> 
> AFAIK the iOS app doesn't use that callback.
> 
> > If so, this change is fine, but we need to also adjust the type declaration then.
> 
> The [type declaration](https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/types/OHApp.d.ts#L5) is already correct (and says return value is `string`).

Just curious - do either of you know why the android app had to implement the callback vs. just relying on the framework7 auto mode?  Same thing applies to the preferTheme?

maniac103:
> do either of you know why the android app had to implement the callback vs. just relying on the framework7 auto mode?

I don't remember the details, but what's safe to say is that dark mode on Android evolved over quite some time, and I imagine there are API levels (= Android versions) and/or WebView versions where relying on the framework didn't work. See https://github.com/openhab/openhab-android/issues/2299

> Same thing applies to the preferTheme?

Would F7 auto-select 'md' on Android? Maybe @mueller-ma knows, as he [implemented](https://github.com/openhab/openhab-android/pull/2328) both callbacks.

PR #3858, opened 2026-02-06
useUIOptionsStore: Fix `themeOptions().dark` is Ref instead of string

florian-h05:
Regression from #3788.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useUIOptionsStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useUIOptionsStore.ts
@@ -175,7 +175,7 @@ export const useUIOptionsStore = defineStore('uiOptions', () => {
 
   function themeOptions() {
     return {
-      dark: darkMode,
+      dark: darkMode.value,
       autoDarkMode: isAutoDarkMode(),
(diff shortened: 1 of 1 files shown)


florian-h05:
BTW: We should consider backporting #3788, #3825 and this one due to https://github.com/openhab/openhab-android/issues/3957. WDYT @jsjames ?

jsjames:
> BTW: We should consider backporting #3788, #3825 and this one due to [openhab/openhab-android#3957](https://github.com/openhab/openhab-android/issues/3957). WDYT @jsjames ?

darkMode was not reactive in 5.1.x - this was added after the fact. So, I don't see how this could be the cause of the issue?

jsjames:
> > BTW: We should consider backporting #3788, #3825 and this one due to [openhab/openhab-android#3957](https://github.com/openhab/openhab-android/issues/3957). WDYT @jsjames ?
> 
> darkMode was not reactive in 5.1.x - this was added after the fact. So, I don't see how this could be the cause of the issue?

I went back and re-read your thread - because it was method call and not a computed function resulted in it being called each time.  Yes, in that case, we can backport these PRs.

PR #3859, opened 2026-02-07
oh-canvas-item: Make content use 100% height and align vertical centered

florian-h05:
Related to https://community.openhab.org/t/vertical-alignment-of-standard-library-widgets-in-fixed-canvas-layout/168211.


florian-h05:
Before:

image

image

After:

Note how the NULL value is properly aligned and the widget takes up the full frame.

image

image

florian-h05:
@JustinGeorgi @hmerk @stefan-hoehn WDYT? I think the change makes canvas bevave correctly, but I am neither actively using it nor want to break existing canvas.

JustinGeorgi:
I have never used the canvas for anything significant, so I'm in the same situation. I think this looks like a good change but, I can't say if it will be problematic.

hmerk:
Used canvas layout for the demo case only, so not really having much experience with it.

peterkappes:
Here is my comment from the community post to continue discussion here:

> First of all, thank you very much for the in-depth investigation of the problem and for proposing a way to fix it. Would this also “fix” the slider widget?
>
> Perhaps the following questions can help point us in the right direction:
>
> - What is the reason that, in a fixed grid, the item content is vertically centered?
>
> - What is the goal of the “preserve classic style” option?
>
> In my opinion, the widgets from the standard library are not really usable at all when working with a fixed canvas. They are not consistent with the row/column or fixed grid layout.

jsjames:
@florian-h05 - just confirming this is no longer draft (as stated in the initial comment).

florian-h05:
Let's undraft and merge this, we are far away from 5.2.0 so we have enough time to test this and revert in case it was a bad idea.

ghys:
thanks, this will potentially and probably break some things but sometimes it has to be expected by widget authors after all... this is genuinely better so the backwards compatibility can take a back seat, as long as it's clearly announced and documented IMO.

PR #3861, opened 2026-02-07
Follow-up changes for #3710

florian-h05:
Review changes for #3710.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
@@ -98,18 +98,18 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
             thingType.value = thingTypeResult.value && Object.keys(thingTypeResult.value).length > 0 ? thingTypeResult.value : null
           } else {
             thingType.value = null
-            const message = `Cannot load thing-type '${data.thingTypeUID}': `
-            console.error(message + getErrorMessage(thingTypeResult.reason, true))
-            f7.dialog.alert(message + getErrorMessage(thingTypeResult.reason))
+            const message = `Cannot load thing-type '${data.thingTypeUID}'`
+            console.error(message, thingTypeResult.reason)
+            f7.dialog.alert(message + ': ' + getErrorMessage(thingTypeResult.reason))
           }
 
           if (channelTypesResult.status === 'fulfilled') {
-            channelTypes.value = channelTypesResult.value || []
+            channelTypes.value = channelTypesResult.value ?? []
           } else {
             channelTypes.value = null
-            const message = `Cannot load channel-types for thing-type '${data.thingTypeUID}': `
-            console.error(message + getErrorMessage(channelTypesResult.reason, true))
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -98,18 +97,18 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
...
-            console.error(message + getErrorMessage(thingTypeResult.reason, true))
-            f7.dialog.alert(message + getErrorMessage(thingTypeResult.reason))
+            const message = `Cannot load thing-type '${data.thingTypeUID}'`
+            console.error(message, thingTypeResult.reason)
+            f7.dialog.alert(message + ': ' + getErrorMessage(thingTypeResult.reason))
           }
 
           if (channelTypesResult.status === 'fulfilled') {
-            channelTypes.value = channelTypesResult.value || []
+            channelTypes.value = channelTypesResult.value ?? []
```
Nadahar, line 107: Is there any practical difference? I thought that was exactly what `||` did in JS, even though I find it "hacky" (to me, it just means "true" - not the value). So, why this "new" operator?

florian-h05: ?? is called nullish-coalescing and only evaluates to the right if the left hand value is null or undefined.
|| evaluates to the right if the left hand is "falsy", which also included 0, empty array [] etc.

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
```diff
@@ -123,8 +123,7 @@ export const useThingEditStore = defineStore('thingEditStore', () => {
             }
           } else {
             thingActions.value = []
-            const message = getErrorMessage(actionsResult.reason, true)
-            console.error(`Cannot load '${thingUID}' thing actions: ${message}`)
+            console.error(`Cannot load '${thingUID}' thing actions`, actionsResult.reason)
```
Nadahar, line 126: How does this work? Will `reason` just be appended after the string? Doesn't it need a placeholder? I just created the variable so that I didn't have to figure out the formatting syntax, I didn't "like it" either, but it was a "safe" way. Is there something "special" with `console.error` that makes it appear on a separate line?

bundles/org.openhab.ui/web/src/pages/settings/things/link/link-edit.vue
```diff
@@ -211,6 +211,7 @@ export default {
             })
           })
         }).catch((err) => {
+          console.error('Error loading profile type or channel type', err)
```
Nadahar, line 214: 👍 Somehow, those slipped by unnoticed...

bundles/org.openhab.ui/web/src/js/stores/useThingEditStore.ts
florian-h05: The console usually shows the string and then logs whatever the reason is after it - if it's a string, it will log the string, if it's an object, you can "explore" it.

Nadahar (approved):
LGTM

jsjames (approved):

PR #3862, opened 2026-02-08
Refactor widget-mixin to TS composable

### LINKED ISSUE #3919: Custom widgets with no slots object fail as child widgets
I can confirm the bug described in [this forum thread](https://community.openhab.org/t/list-items-not-displaying-anymore/168594) on a recent snapshot (Build #5165 Main UI Commit 9682acda).

If a widget does not have a `slots` object (which is reasonable for list item widgets):

```yaml
component: oh-toggle-item
config:
  color: '=(@@props.item == "ON") ? "yellow" : "gray"'
  icon: '=(@@props.item == "ON") ? "iconify:ph:lightbulb-duotone" :
    "iconify:ph:lightbulb"'
  iconColor: '=(@@props.item == "ON") ? "orange" : "gray"'
  item: =props.item
  title: =props.title
```
then that widget fails to render when included as a child widget in model cards do to a null error in `childWidgetContext`

https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui/web/src/components/widgets/widget-mixin.js#L153

Currently, adding an empty slots object to the widget config is a valid workaround, but not beginner user friendly.

jsjames:
This PR updates the widget-mixin to a composable - as a step necessary to use composition api with widget components.

fixes #3919 

In composition API, components must be registered globally if referenced by name - hence all widgets are registered globally. Note, there is no performance impact here.

Cleanup of importing components in each of the widgets will happen in later PR now that they are globally registered.

Adds typescript definitions for widget definitions

Removed special title overflow in oh-list-item for divider as css text-overflow does the same operation and is now part of the f7 styles.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
+++ b/bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
@@ -55,7 +55,8 @@ export function useWidgetContext(context: WidgetContext) {
     if (!context.component) return {}
     let evalConfig: Record<string, unknown> = {}
     // Fallback to modelConfig for oh- components to allow configuring them in modals
-    const sourceConfig: Record<string, unknown> = context.component.config || (componentType.value?.startsWith('oh-') ? context.modalConfig : {})
+    const sourceConfig: Record<string, unknown> =
+      context.component.config || (componentType.value?.startsWith('oh-') ? context.modalConfig : {})
     if (sourceConfig) {
       if (typeof sourceConfig !== 'object') return {}
       for (const [key, value] of Object.entries(sourceConfig)) {
@@ -66,7 +67,7 @@ export function useWidgetContext(context: WidgetContext) {
     return evalConfig
   })
 
-  const props = computed<Record<string, unknown>>(()=> {
+  const props = computed<Record<string, unknown>>(() => {
     if (!context.component) return {}
     if ('props' in context.component && context.component.props.parameters) {
       let defaultValues: Record<string, unknown> = {}


bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  // methods
+  function evaluateExpression(key: string, value: any, _context?: WidgetContext, _props?: Record<string, any>): any {
+    return _evaluateExpression(key, value, _context || context, _props || props.value)
+  }
+
+  function childContext(component: UiComponent): WidgetContext {
+    return {
+      component,
+      // rootcomponent: context?.root || context?.component,
```
jsjames: @florian-h05 - I don't see anywhere that rootcomponent is used? Is this legacy and can we remove?

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+    if (context?.vars) {
+      for (const varKey in context.vars) {
+        widgetVars.value[varKey] = context.vars[varKey]
+      }
+    }
+    if (context?.component.slots) Object.assign(widget.slots, context.component.slots)
+    const widgetContext = {
+      component: widget,
+      // root: widget,
```
jsjames: @florian-h05 same here, i don't see where root is used.

jsjames:
just noticed a but where the navbar at top doesn't display -> converted to draft until fixed

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
```diff
@@ -0,0 +1,209 @@
...
+      component,
+      // rootcomponent: context?.root || context?.component,
+      props: props.value,
+      fn: context?.fn,
+      const: context?.const,
+      vars: context?.vars,
+      varScope: varScope.value || context?.varScope,
+      ctxVars: context?.ctxVars,
+      loop: context?.loop,
+      store: context?.store,
```
jsjames: @florian-h05 - also it looks like store is always useStatesStore().trackedItems. Not sure if this is legacy code, but if it is always useStatesStore, we can remove from WidgetContext and just use the store directly in all the components?

jsjames:
@florian-h05 - any feedback on my questions?

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
florian-h05: I also see no way how this could be accessed from widget expressions, so I think we can really remove this.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
florian-h05: I also see no usage (except for rootcomponent above).

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
florian-h05: Hmm, I don't know why it was done that way. But since it all delegates to the same store, I see no technical reason to pass around the store like that, however I find it quite comfortable to be able to access it without having to import it everywhere.

jsjames: okay - let's keep as is.

jsjames:
Moved to draft as I will rebase this to the #3911 which added the typeChecked as it will be cleaner.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
```diff
@@ -0,0 +1,209 @@
...
+    return evalConfig
+  })
+
+  const props = computed(() => {
+    if (!context?.component) return {}
+    if (context.component.props?.parameters) {
+      let defaultValues: Record<string, any> = {}
+      context.component.props.parameters.forEach((p: any) => {
+        if (p.default !== undefined) {
+          defaultValues[p.name] = p.default
```
jsjames, line 77: @florian-h05 - could you review this one? the original code had p.default and that isn't a property on ConfigDescriptionParameter - so I don't believe this ever worked correctly??

florian-h05: `default` is commonly specified for in widgets for their props

jsjames:
this one is ready for review.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (!widget) {
+      console.warn('widget not found, cannot render: ' + componentType.value)
+      return null
+    }
+    if (context.vars) {
+      for (const varKey in context.vars) {
+        widgetVars.value[varKey] = context.vars[varKey]
+      }
+    }
+    if (context.component && 'slots' in context.component) Object.assign(widget.slots, context.component.slots)
```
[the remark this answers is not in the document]
florian-h05: @jsjames Have you checked that comment?

bundles/org.openhab.ui/web/src/types/scope-css.d.ts
florian-h05, line 1: Why are there such unrelated changes in this PR? Does it need rebasing?

jsjames: useWidgetContext imports scope-css - so it is to support proper typescript checking.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetContext.ts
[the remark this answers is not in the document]
jsjames: just submitted a change.

florian-h05:
> Cleanup of importing components in each of the widgets will happen in later PR now that they are globally registered.

@jsjames: Can you please look into this? The current static global import is causing Vite to bundle a lot of large dependencies (like ECharts, Leaflet) directly into the JS entrypoint.

EDIT: Trying it myself

florian-h05:
See #3969.

florian-h05:
For comparison, the bundle visualization from rollup-plugin-visualizer before and after the current changes:

[before-stats.html](https://github.com/user-attachments/files/25804401/before-stats.html)
[after-stats.html](https://github.com/user-attachments/files/25804402/after-stats.html)

PR #3864, opened 2026-02-09
oh-list: Fix missing hairlines between items

florian-h05:
Regression from #3741.


florian-h05:
I think this should be backported

PR #3867, opened 2026-02-11
Bidirectional arrow for zwave controllers in map

apella12:
With support for 700 zwave controllers (as of OH4.1), the find neighbors routine for controllers was dropped to avoid an endless loop. To keep the controller on the map, the ID and neighbor for the controller was set to itself (from Zwave binding). 
```
        // Set controller properties for Network Map since it will not be healed
        getThing().setProperty(ZWaveBindingConstants.PROPERTY_NEIGHBOURS, sucNode.toString());
        getThing().setProperty(ZWaveBindingConstants.PROPERTY_NODEID, sucNode.toString());
```

This PR applies the assumption that if a node "sees" the controller it must be bidirectional. First time with this repository and code.
Test environment data
ZWavemap2026-02-11 155641
controller prop 2026-02-11 155550

Before version
ZWavemapold2026-02-11 155641

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/network/providers/zwave-provider.ts
+++ b/bundles/org.openhab.ui/web/src/components/thing/network/providers/zwave-provider.ts
@@ -10,16 +10,16 @@ import type {
   NetworkNode,
   NetworkLink,
   NetworkLegend,
-} from "../types";
+} from '../types';
 
 /**
  * Color constants for Z-Wave node types
  */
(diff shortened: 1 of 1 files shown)


jsjames:
please run `npm run format' to fix the string quotations - strings should be single quotes.

apella12:
Can't get npm run format to run locally. Not that important to me. Thought I would give it a try.  Thanks for your time

PR #3869, opened 2026-02-12
zwave: Fix zwave network map display in 5.1.x

apella12:
reference: https://community.openhab.org/t/z-wave-network-map-empty-after-hardware-upgrade/167672/9

Fix zwave network map display in 5.1.x Tested locally on 5.1.2

Add AI explanation; Ah, this is a classic Vue reactivity + conditional‑rendering interaction bug, and the symptoms you’re describing line up perfectly with one specific root cause:
🎯 The  item (v-if="thingType?.UID?.startsWith('zwave:')"disappears when the list is filtered. 
🔧 The fix: Give the v-if item a stable key. This forces Vue to treat the “View Network Map” item as a distinct, non-reusable DOM node, so its click handler stays intact.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -201,6 +201,7 @@
               </f7-block-title>
               <f7-list class="margin-top" media-list>
                 <f7-list-item
+                  key="zwave-network-map"
                   v-if="thingType?.UID?.startsWith('zwave:')"
                   title="View Network Map"
                   link=""


jsjames:
@apella12 - just for completeness, can you add a description to the problem this solves. The fix looks fine to me, but it would be good to have what the actual problem was. Thx

apella12:
Problem is the map is not produced.
https://community.openhab.org/t/z-wave-network-map-empty-after-hardware-upgrade/167672/9 and https://community.openhab.org/t/openhab-5-1-release-discussion/167620/499

jsjames (approved):

jsjames:
BTW, is this also an issue in the main branch?

apella12:
I did not have a problem with a 5.2 snapshot. Can't exactly say why, but the section in 5.2 changed to include a matter thread map, so maybe that forced Vue to preserve the action.

PR #3892, opened 2026-02-18
Z-Wave network map: Use bidirectional edges for controller

apella12:
With support for 700 zwave controllers (as of OH4.1), the find neighbors routine for controllers was dropped to avoid an endless loop. 
```
        // Set controller properties for Network Map since it will not be healed
        getThing().setProperty(ZWaveBindingConstants.PROPERTY_NEIGHBOURS, sucNode.toString());
        getThing().setProperty(ZWaveBindingConstants.PROPERTY_NODEID, sucNode.toString());
```
To keep the controller on the map, the ID and neighbor for the controller was set to itself. This PR applies the assumption that if a node "sees" the controller it must be bidirectional, otherwise the map implies the controller has no connection to any node. 
Test data
ZWavemap2026-02-11 155641
controller prop 2026-02-11 155550

Before version
ZWavemapold2026-02-11 155641

Second try had formatting issue, hopefully solved. This is an enhancement

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/network/providers/zwave-provider.ts
+++ b/bundles/org.openhab.ui/web/src/components/thing/network/providers/zwave-provider.ts
@@ -31,9 +31,24 @@ export class ZWaveNetworkProvider implements NetworkGraphProvider {
 
   private static readonly LEGEND: NetworkLegend = {
     nodeRoles: [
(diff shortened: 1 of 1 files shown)


jsjames (approved):

apella12:
Thanks! As a novice and using windows with github desktop this repository was challenging

PR #3894, opened 2026-02-18
useStatesStore: Fix error in expression tester with =items formula

### LINKED ISSUE #3876: Expression-tester crashes with items object return value
This looks like a regressions from #3749 

If the final result of the expression in the expression tester is just the `items` object then the expression tester fails to render with an error of:

```
index-CVHPjz-J.js:1217 TypeError: _a.toString is not a function
    at Proxy.result (developer-dock-Cqc192gQ.js:2013:100)
    at ot (index-CVHPjz-J.js:380:19)
    at get value (index-CVHPjz-J.js:994:14)
    at Object.get [as result] (index-CVHPjz-J.js:2436:23)
    at developer-dock-Cqc192gQ.js:2049:12
    at r2 (index-CVHPjz-J.js:1311:14)
    at R (index-CVHPjz-J.js:2270:23)
    at Proxy.vg (index-CVHPjz-J.js:12863:7)
    at go (index-CVHPjz-J.js:2713:20)
    at Xe.s3 [as fn] (index-CVHPjz-J.js:3083:20)
```

This is true with both the simplest of forms:

```
=items
```

and more complex expressions:

```
=(1 == 2)?("True"):(items)
```

jsjames:
Fixes #3876

Adds `toString()` function to `trackedItems` proxy.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useStatesStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useStatesStore.ts
@@ -63,6 +63,7 @@ export const useStatesStore = defineStore('states', () => {
     get(obj: Record<string, ItemState>, prop: string | symbol): ItemState {
       if (prop === '_keys') return Object.keys(itemStates.value) as any
       if (prop === '__ob__') return (obj as any).__ob__
+      if (prop === 'toString') return (() => '[object TrackedItems]') as any
 
       // to avoid the Vue devtools requesting invalid items in development
       if (INVALID_PROPS.has(prop.toString())) return {} as any


florian-h05 (approved):

BrettLHolmes:
@jsjames  by the looks of things this was merged into 5.1.3 -

Was this the same issue with the developer tools and the Widgets Expression Tester?

 if so I loaded 5.1.3 this morning and the issue is still there or did I assume wrong and its not part of 5.1.3

florian-h05:
The widget expression tester is part of the developer tools, what distinction do you make here? Please reload Main UI and clear cache through the about page.

BrettLHolmes:
> The widget expression tester is part of the developer tools, what distinction do you make here? Please reload Main UI and clear cache through the about page.

Hi @florian-h05  when i use the Developer sidebar to show the current status of an item i get [object Object] rarther than the current status

Here are two screenshots from a  5.1.2/5.1.3 and a 5.0.3 system
image
image

PS i did reload the App from the About Menu and its the same.

jsjames:
That is an unrelated issue to what this PR fixed. This PR fixed a exception that occurred when just referring to "=items". I would submit a new issue regarding this that we can track and assess.

BrettLHolmes:
> That is an unrelated issue to what this PR fixed. This PR fixed a exception that occurred when just referring to "=items". I would submit a new issue regarding this that we can track and assess.

Will do thanks

florian-h05:
I've also noticed this, but it is totally unrelated to what has been fixed here. I will get it back working though.

florian-h05:
@BrettLHolmes Please don't create an issue if you haven't done so yet, I have a PR: #3920

PR #3911, opened 2026-02-20
ESLint: Enable recommendedTypeChecking lint rules

jsjames:
This PR adds in tseslint.configs.recommendedTypeChecked to improve our typescript code. The largest required update is that the "any" type can no longer be used as a crutch as it eliminates any typescript checking.

Most of the files have been updated - there are a few where specific eslint checks have been disabled at the top of the file. Those can be addressed over time.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/thing/network/providers/thread-provider.ts
+++ b/bundles/org.openhab.ui/web/src/components/thing/network/providers/thread-provider.ts
@@ -76,6 +76,7 @@ interface Route {
   rloc16: number
   extAddress: string
 }
+
 interface UnknownDevice {
   extAddress: string
   seenBy: string[]
@@ -95,7 +96,7 @@ interface ProcessedNode extends NetworkNode {
   ownExtAddress: string | null
   isRouter: boolean
   isBorderRouter: boolean
-  routingRole: number
+  routingRole: RoutingRole
   neighbors: UnknownDevice[]
   routes: Route[]
 }


jsjames:
@digitaldan - could you review the changes made to the thread-provider code. the changes are typescript enhancements to address some of the errors in linting.

bundles/org.openhab.ui/web/src/js/openhab/sse.ts
```diff
@@ -104,14 +106,14 @@ function newSSEConnection(
     })
 
     es.onmessage = (event: MessageEvent) => {
-      let evt: any
+      let evt: unknown
       try {
-        evt = JSON.parse(event.data)
+        evt = JSON.parse(event.data as string)
+        messageCallback(evt)
```
florian-h05: I would rather keep the messageCallback call outside of the try-catch block.

bundles/org.openhab.ui/web/src/js/openhab/auth.ts
```diff
@@ -120,10 +120,11 @@ export function setAccessToken(token: string, api: any): Promise<void> {
...
         return Promise.resolve()
       })
-      .catch((err: any) => {
-        if (err === 'Unauthorized' || err === 401) requireToken = true
+      .catch((err: unknown) => {
+        if (err instanceof ApiError && (err.response.statusText === 'Unauthorized' || err.response.status === 401)) {
+          requireToken = true
+        }
         accessToken = token
-        return Promise.resolve()
```
florian-h05, line 126: Isn't it required to return a Promise.resolve or reject?

jsjames: I should have changed it to an async function since it returns a promise and then the resolve is not required as the async function autowraps the response in a promise. However, if we want to explicitly return a Promise.resolve, we could - but would recommend we standardize on one method for the code base. In most of the code I have touched recently, I was removing the explicit Promise.resolve in async functions.

Regarding the reject, we should adopt either a Promise.reject or throw - which in my understanding operate the same for async functions.

jsjames: There needs to be some promise clean-up in the code. I'd recommend we add the rule:

@typescript-eslint/promise-function-async - which will force all functions that return a promise to be async - there about ~10 of those in the current codebase.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-promise-resolve-reject.md - This one will standardize that async functions don't include the Promise.resolve/reject in place of just a return or throw.

I can submit as a separate PR if you support.

jsjames:
I added the @typescript-eslint/promise-function-async rule and fixed those instances where a functions returned a promise and were not async.  I'd recommend (and this is what I implemented) is that we rely on the autowrapping that occurs for async functions - i think that will be cleaner.

As an aside, I think we should enable this rule as well:
"@typescript-eslint/explicit-function-return-type": "error"

This will make the core more readable/documented as it will be clear what the return value is for functions.  Thoughts?

bundles/org.openhab.ui/web/src/js/openhab/auth.ts
florian-h05: I'd vote to go for async and return result/throw error instead of returning Promise.resolve or Promise.reject, IMO this is much cleaner code.
I support adding/enabling the rules you proposed above, you can do this in this PR here if you prefer.

florian-h05:
> This will make the core more readable/documented as it will be clear what the return value is for functions. Thoughts?

For my part, I don't need it, as IntelliJ does a good job in displaying the return type if none is explicitly written down, and IMO it keeps the code "less verbose" to not write it explicitly.

bundles/org.openhab.ui/web/src/js/openhab/auth.ts
florian-h05: > https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-promise-resolve-reject.md - This one will standardize that async functions don't include the Promise.resolve/reject in place of just a return or throw.

I've just added this - only needed to touch two lines of code to adjust our codebase.

digitaldan:
Sorry, just seeing this today, I just tried this out and it looks good to me 👍

Issue #3916, opened 2026-02-22
Add support in MainUI for currently unsupported `context` values.


andrewfg:
Please create a new issue where the required changes in Main UI (e.g. missing support for some context) are listed according to your table. 

 _Originally posted by @florian-h05 in [#3909](https://github.com/openhab/openhab-webui/issues/3909#issuecomment-3941468112)_

---

Main UI currently does not support the `context` values listed below. This Issue is to keep track of them and to propose priorities for implementing them.

#### Priority 1:
- [ ] `url` - in the schema but not supported in Main UI
- [ ] `network-address` - in the schema and there is existing commented out code in Main UI
- [ ] `mac-address` - NOT  in the schema but as part of the fix for `network-address` ipv4 and ipv6 address patterns, we could easily add a pattern for mac address too

#### Priority 2: 
- [ ] `service`
- [ ] `tag`
- [ ] `week`
- [ ] `group`
- [ ] `password-create`
- [ ] `channelType` - probably we should change this to `channel-type` before implementing

PR #3918, opened 2026-02-23
Config sheet: Add support for missing context values

andrewfg:
Refs #3916.

This PR does the following:

- fixes support for `network-address` context
- adds support for `url` context
- adds support for `mac-address` context
- adds support for `week` context

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/validation-pattern.ts
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/validation-pattern.ts
@@ -1,13 +1,15 @@
 const hex2 = `[0-9A-Fa-f]{2}`
 const hex4 = `[0-9A-Fa-f]{4}`
+const hex12 = `[0-9A-Fa-f]{12}`
 /**
  * MAC address handler.
  * Supports:
  * - `AA:BB:CC:DD:EE:FF`
  * - `AA-BB-CC-DD-EE-FF`
  * - `AABB.CCDD.EEFF`
+ * - `AABBCCDDEEFF`
  */
-export const MacAddress: string = `(?:(?:${hex2}[:\\-]){5}${hex2}|(?:${hex4}\\.){2}${hex4})`
+export const MacAddress: string = `(?:(?:${hex2}[:\\-]){5}${hex2}|(?:${hex4}\\.){2}${hex4}|${hex12})`
 export const MacAddressCompiled: RegExp = new RegExp(`^${MacAddress}$`)
 
 // const ipv4Addr = `(?:(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])`


florian-h05:
I think this also addresses #3457 by supporting all context types supported by core. EDIT: it doesn’t, but your issue #3916 better summarises what’s missing.

jsjames:
HTML 5 has a URL input type already which will support validation. To support URL, a better/simpler option would be just to return 'url' from the computed property controlType.

```
    controlType () {
      if (this.configDescription.context === 'password' && !this.showPassword) return 'password'
      if (this.configDescription.context === 'email') return 'email'
      if (this.configDescription.context === 'telephone') return 'tel'
      if (this.configDescription.context === 'color') return 'color'

```

jsjames:
> HTML 5 has a URL input type already which will support validation. To support URL, a better/simpler option would be just to return 'url' from the computed property controlType.
> 
> ```
>     controlType () {
>       if (this.configDescription.context === 'password' && !this.showPassword) return 'password'
>       if (this.configDescription.context === 'email') return 'email'
>       if (this.configDescription.context === 'telephone') return 'tel'
>       if (this.configDescription.context === 'color') return 'color'
> ```

There is also a 'week' type which could be added and most browsers will show a nice calendar view for input. However, I don't know if the 'value' is compatible to what openhab expects.

```
Value Format: The value submitted to the server is a string in the format "YYYY-WW" (e.g., "2025-W25" for the 25th week of 2025)
```

andrewfg:
> HTML 5 has a URL input type already which will support validation

Hmm. It seems the validation is delegated to code within the browser. And it seems that Chrome on Windows (at least) has extremely loose validation insofar as it accept `foo:\//bar¬"£$:::@` as perfectly valid. So @jsjames my question here is whether we want to force a stricter validation via our own `url` pattern, or are we happy to have it browser dependent?

> There is also a 'week' type which could be added and most browsers will show a nice calendar view for input. 
> I don't know if the 'value' is compatible to what openhab expects.

It is a moot question since AFAICT there is nothing in OH-Core or OH-Addons that currently uses it. => So I will just enable it anyway so that it can be later sorted out whenever actually used.

---

EDIT apropos 'week' -- OH has its own input forms for all chronological fields except week => so perhaps we should ask if it would be better to forgo the OH versions and delegate to the browser versions. => WDYT?

```
      } else if (configDescription.type === 'TEXT' && configDescription.context === 'month') {
        return ParameterMonth
      } else if (configDescription.type === 'TEXT' && configDescription.context === 'dayOfWeek') {
        return ParameterDayOfWeek
      } else if (configDescription.type === 'TEXT' && configDescription.context === 'time') {
        return ParameterTime
      } else if (configDescription.type === 'TEXT' && configDescription.context === 'date') {
        return ParameterDate
      } else if (configDescription.type === 'TEXT' && configDescription.context === 'datetime') {
        return ParameterDatetime
```

andrewfg:
@jsjames / @florian-h05 I just committed a version that validates the following properly now..

- network-address
- mac-address
- url
- week

The first three use F7 custom validation patterns. Whereas the last one defers to the browser. As you can see the latter looks very nice. (So there is still an open question if we should harmonise all chrono- related contexts to use the browser validator or the in built OH WebUI standard controls. => I am easy so please tell me if you have preferences.

andrewfg:
I have a doubt concerning network-address; currently the regex validator allows ipv4 or ipv6 or host names (e.g. localhost); but it fails if one would add a :port suffix. I think that is correct from a semantic point of view. And that bindings that require a port id would do it via a separate numeric config param. However as network-address had previously not had any validation before, it is imaginable that some bindings may have been misusing network-address for an address plus port combination. I will do a hand check of addons that use network-address to see if there may be any that misuse it with a port id. But I wonder if we should not allow port ids in network-address just in case. => WDYT?

---

EDIT:

> I will do a hand check of addons that use network-address to see if there may be any that misuse it with a port id.

- there are 6 bindings that use the `network-address` context specifically to configure a URL parameter
- there are 128 bindings that use the `network-address` context to configure an address AND also do not have a 'port' configuration parameter (thus implying that the address may possibly contain a :port_id ..)

=> So my conclusion is that we should set the validation regex for the `network-address` context to allow a) simple addresses, b) addresses with ports, and c) urls. => Can you confirm? (Effectively this makes `network-address` and `url` synonyms of the same validation pattern)

jsjames:
> So @jsjames my question here is whether we want to force a stricter validation via our own url pattern, or are we happy to have it browser dependent?

Can we do both? Include a controlType of 'url' for urls AND do the more extensive validation? I don't know if the various web browsers offer any input advantage if the controlType is set to url which could be a benefit.

Regarding the network-address, I agree it would be good to keep more general. If the field were just "IP address", i would recommend removing the port, but network address implies something different.

andrewfg:
> Can we do both? Include a controlType of 'url' for urls AND do the more extensive validation?

Yes. That is what I have done.

> Regarding the network-address, I agree it would be good to keep more general.

Ok. I have made it so that `network-address` does the same validation as `url`.

> If the field were just "IP address", i would recommend removing the port

Indeed, so in the latest commits today I added a new `ip-address` that validates only the pure numeric address without and prefixes or suffixes..

---

EDIT: so we have:
1. `url` which validates any URL including http(s), host name, ipv4, ipv6, port, path, query
2. `network-address` identical to `url` (so we should mark this is deprecated for future use)
3. `ip-address` pure numeric ip address only.

Nadahar:
> If the field were just "IP address", i would recommend removing the port, but network address implies something different.

Technically there's a difference between an address and a port. The only reason that "we associate them" is that URLs allow specifying the port as well, the port is completely unrelated to the address.

> 2\. `network-address` identical to `url` (so we should mark this is deprecated for future use)

Maybe it would be worth investigating the existing use. Are you certain that port is used with this context today? "Network-address" as either an IP or a hostname is a very relevant field to have because that is what you generally can feed directly to the socket/connection you're trying to open. Port isn't a part of this, but restricting it to only IP address would be quite annoying for those that have name resolution on their LAN.

A URL is a completely different beast that can, in addition to address, include protocol, port, username, password, a path, parameters and fragments (and probably more that I don't remember). If all you want is the address, it's very inconvenient to start out with a URL, having to parse that and discard (or create errors for) all the unsupported information. The problem, when it comes to validation, is that an address is hard to restrict, because there are many different name resolution regimes that allow different characters. DNS isn't the only name resolution system. So, I'd say that a "network address" field would have to be pretty much allowed to be anything. Which might make having a context for it kind of pointless? How is it different from just a string..?

andrewfg:
We have had some interesting discussions in this thread (including the wrong assumption that I need lessons of what is a url). However these discussion **_have lost track of what this PR is really about_**.

So to refresh your memories, this PR is about adding functionality to MainUI concerning TWO THINGS ONLY:

1. Shall the OH-WebUI MainUI Thing configuration pages handle existing known context values better? And if so how?
2. Shall the OH-WebUI MainUI Thing configuration pages handle any new context values?  And if so how?

Let us discuss THESE TWO THINGS ONLY PLEASE!!

The details of this PR are as in the table below, which shows the current functionality of MainUI for each context, and the proposed future functionality.

Notes:
- There are 417 bindings that request a context from MainUI.
- Of those 417 there are 248 that request the `network-address` context. So they presumably do expect some kind of validation, where none is actually provided by MainUI.
- Of those 248 about 142 are probably looking for validation of a dotted ipv4 address, and the rest are probably asking for more.
- This PR is ONLY about the functionality provided in MainUI. It has nothing to do with whether we apply a stricter schema validation on the Thing's configuration-description xml. _**This PR can be merged whether the schema PR in OH-Core gets merged or not.**_

The @openhab/core-maintainers need only to decide and tell me which functionality you want to have in column 4 of this table.

| Context Value   | Current Use   | Main UI Existing Functionality    | Main UI Proposed Functionality             |
|-----------------|---------------|-----------------------------------|--------------------------------------------|
| network-address | 248 bindings  | known but not handled -- see 0.   | validate inputs as a url -- see 1.         |
| url             | 0             | unknown and not handled -- see 0. | validate inputs as a url -- see 1.         |
| mac-address     | 0             | unknown and not handled -- see 0. | validate inputs as a mac address -- see 4. |

---

0. Input not handled: MainUI treats this as a plain text input field
    - allows all characters, white space, special characters, foreign characters, accented characters, and symbols etc.

1. Valid url inputs are is as defined in RFC 1738 and RFC 3986 comprising
    - optional prefix 'http(s)://' protocol of both upper and lower case 
    - required host name -- see 2)
    - optional suffix for port id like ':n' where n is an integer between 0 and 65535
    - optional suffix(es) for path+query like '/path?query=value' comprising alphanumerics of both cases plus a set of additional allowed special characters

2. Valid host name fields are as defined in the respective RFCs above comprising
    - either alphanumeric string of both cases plus dot or dash e.g. www.disney-WORLD-456.co.uk 
    - or a dot delimited numeric octal ipv4 address per RFC 1166 e.g. 192.168.1.1 
    - or a colon delimited hex ipv6 address per RFC 4291 / 5952 e.g. [2001:db8::ff00:42:8329] enclosed in square brackets

3. Valid mac address inputs comprise
    - either colon delimited hex pairs in canonical form e.g. 11:22:33:DD:EE:FF or lowercase
    - or dash delimited hex pairs in canonical form e.g. 11-22-33-DD-EE-FF or lowercase
    - or dot delimited hex triplets in canonical form e.g. 1122.33DD.EEFF or lowercase ("Cisco style")

bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
```diff
@@ -76,16 +79,54 @@ export default {
       if (this.configDescription.context === 'email') return 'email'
       if (this.configDescription.context === 'telephone') return 'tel'
       if (this.configDescription.context === 'color') return 'color'
+      if (this.configDescription.context === 'week') return 'week'
```
florian-h05: `week` context here returns values such as: `2026-W06`
According to the ConfigDescriptionParameter Javadoc it should be a INTEGER:

```
     * <li><b>week</b>: A week of year [0-52]</li>
```

andrewfg: Ok. We are lucky in that week is not currently used by any addons. So we have the choice a) to change to doc to match the code, or b) change the code to match the doc. => Do you have a preference?

Nadahar:
> (including the wrong assumption that I need lessons of what is a url)

I'm sorry if that offended you, but I don't know how else to address the fact that you seem to think that network address should be validated like a URL.

> * Of those 417 there are 248 that request the `network-address` context. So they presumably do expect some kind of validation, where none is actually provided by MainUI.

I think that's part of the problem here. Authors use a "fitting context" if they find one, not because they expect validation, but because they think "it's the right thing" to do. I've been in that situation myself, looking through the contexts, wondering if anything fits. The idea is, that if you assign the right context, it might get some kind of validation, it might be displayed differently, or it might be "tagged"/included in some search. If you have a field that fits one of the contexts, it is reasonable to use that without knowing all the details of potential consequences.

What level of knowledge about name resolution and how that all works should we expect from binding authors? My experience is that network knowledge isn't that great overall, among people. And that's fine, if you only use IP addresses, you don't need to know anything about name resolution. But does that mean that we should prevent it working for those that do use it? Especially, when it's all provided "for free" for us, because the OS does all the work?

> * Of those 248 about 142 are probably looking for validation of a dotted ipv4 address, and the rest are probably asking for more.

How can you conclude that? As I've told you, if you have name resolution configured, it will work *in place* of an IP address. You can just ask Java to connect to "foo", the OS will query name resolution for "foo" and if it has an entry, it will replace it with the IP address registered. If it has no such entry, it will return an error aka "unknown host". So, how would a binding author "know" that he/she wanted a dotted IPv4 address?

> * This PR is ONLY about the functionality provided in MainUI. It has nothing to do with whether we apply a stricter schema validation on the Thing's configuration-description xml. _**This PR can be merged whether the schema PR in OH-Core gets merged or not.**_

Yes, but I don't think it would be good if MainUI refuses to accept perfectly valid values so that people that use name resolution will have to either use file-based Thing configurations, or set the configuration using the API directly. I think it would be a very bad thing to needlessly restrict valid values.

> The @openhab/core-maintainers need only to decide and tell me which functionality you want to have in column 4 of this table.

Since I'm not a maintainer, I guess I'm forbidden to answer this, but I'll do it anyway. I think `url` and `mac-address` are fine, I think `network-address` is wrong (it doesn't even allow the very common case of network names containing spaces), and I'm very skeptical about introducing `ip-address`. The reason I'm skeptical is that I think it has very few legitimate use cases (it can only be when you know for a fact that name resolution can't exist), and I think that, for the same reason as explained above, many binding authors might pick that "because it fits what they're seeking". They think that they want an IP address, knows nothing about name resolution and the fact that a name will work just as well, and then needlessly "disables" name resolution by picking this context.

> allows all characters, white space, special characters, foreign characters, accented characters, and symbols etc.

The current behavior is correct for `network-address`. There are no standards or rules that cover "name resolution". Each name resolution type has its own rules, but there are none that apply to all.

andrewfg:
@Nadahar look I am getting tired of this. Instead of you being continuously negative, please give me your positive suggestion of exactly what validation you are actually proposing.

Nadahar:
> look I am getting tired of this. Instead of you being continuously negative, please give me your positive suggestion of exactly what validation you are actually proposing.

I'm not "being negative". I'm saying that I don't think it's possible to do validation for `network-address` because it can be literally anything and still be a valid value.

Far back I asked if there's any point to having a `network-address` context at all, given that you can't really validate it, you can't create a "lookup" for it, so I don't quite see what "added functionality" can exist for that field.

Regarding `ip-address` I have no problem if people would know when to use it (almost never), but I'm pretty sure that's not what will happen if it's added.

florian-h05:
> I think network-address is wrong (it doesn't even allow the very common case of network names containing spaces), and I'm very skeptical about introducing ip-address. The reason I'm skeptical is that I think it has very few legitimate use cases

Which „network name“ contains spaces? IP addressed don’t do and hostnames and domains also can’t include spaces.
Wrt to IP address: I see limited use cases for this, but there definitely are, e.g. configuring multicast addresses. I’m fine with adding that, and if the docs make clear when to use what and still someone chooses incorrectly, then the choice has to be corrected, but not the options.

bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
florian-h05: I would understand week as a calendar week independent from the year, but with values 1 to 53. Calendars don’t count like computer science ;-)

andrewfg:
> I don't think it's possible to do validation for network-address because it can be literally anything and still be a valid value.

If that were the case (which I personally do not believe) then you are essentially saying that the network-address context has literally zero semantic meaning. And if that were indeed the case then the only logical consequence would be to completely excise that context from all of the 248 bindings that currently (wish to) use it. There is no value in perpetuating a delusion; better to clarify it immediately!

(HOWEVER I think that most of the 248 developers, including myself, and probably most of the OH maintainers, would disagree with the above hypothesis; and would instead ask to get some actual value from the context rather than delete it..)

Nadahar:
> Which „network name“ contains spaces? IP addressed don’t do and hostnames and domains also can’t include spaces.

Hostnames can. If you for example, like me, use WINS, which "builds on" the NETBIOS naming, spaces are perfectly valid. The same is true for some Apple protocol as far as I know, and there are name services using LDAP that I think can contain pretty much anything. I don't know all of the name services, but the idea that a hostname "can't contain space" is wrong.

The problem is that DNS has become so "common" that many people only know DNS. DNS does not allow spaces, or most non-English alphabet letters. mDNS uses DNS' definition, so in theory follows the same rules, but in reality it does not, since Apple's Bonjour, "the origins" of mDNS, does allow space - so any mDNS implementation that actually want to work with real-world devices should handle spaces, despite that the standard formally doesn't allow it.

You also have things like [IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name) that basically allows the use of all Unicode characters, not only the English alphabet.

Nadahar:
> If that were the case (which I personally do not believe) then you are essentially saying that the network-address context has literally zero semantic meaning. And if that were indeed the case then the only logical consequence would be to completely excise that context from all of the 248 bindings that currently (wish to) use it. There is no value in perpetuating a delusion; better to clarify it immediately!
> 
> (HOWEVER I think that most of the 248 developers, including myself, and probably most of the OH maintainers, would disagree with the above hypothesis; and would instead ask to get some actual value from the context rather than delete it..)

Yes, I'm saying that it has literally zero semantic meaning. The fact that it has been here all these years, and done exactly nothing, supports this. The reason it "works" now, is that its use has no consequence.

Nadahar:
In a way, I find the whole discussion a bit bizarre. You can't just type anything into a network-address field, you must type something that actually resolves to a device on your network. Most of the "errors" people are likely to make, will pass syntactic validation and still fail (because they use an unresolvable name or non-existing IP). There will have to be "feedback" in the form of an error message of some kind when you enter anything invalid anyway, I just don't get what syntactic validation, that will only end up preventing potentially valid values, adds to the picture.

andrewfg:
Here are the requirements for a valid WINS/NetBIOS name:
Length: Maximum of 15 characters for the computer name.
Total Name: The NetBIOS name is technically 16 characters, but the 16th character is a hidden suffix (00-FF hex) indicating the resource type, reserved by the system.
Allowed Characters: Letters (A-Z), numbers (0-9), and hyphens (-).
First Character: Must be alphanumeric.
Disallowed Characters: Spaces, periods (.), backslashes (), slashes (/), colons (:), asterisks (*), question marks (?), quotation marks ("), less than (<), greater than (>), vertical bars (|), commas (,), tildes (~), exclamation points (!), at signs (@), number signs (#), dollar signs ($), percent signs (%), carets (^), ampersands (&), and apostrophes (').
Special Considerations:
Underscores (_): While sometimes tolerated, they are technically invalid in NetBIOS names, and using them can cause issues with DNS interoperability.
NetBIOS Scope ID: If a scope ID is configured, it is appended to the name, but this is rarely used in modern networking.
Numeric Only: A name consisting only of numbers is generally invalid due to potential conflicts with IP address parsing.

andrewfg:
I think it is possible to modify the F7 logic so that instead of simply rejecting the inputs, we could indicate that the input fails validation and also present a "use anyway" button. @florian-h05 => WDYT?

---

image

I have not quite got the formatting correct. => Perhaps @florian-h05 you can help me with this?

florian-h05:
> I think it is possible to modify the F7 logic so that instead of simply rejecting the inputs, we could indicate that the input fails validation and also present a "use anyway" button.

Sounds like a good compromise, I'll have a look and help you with the formatting in case it's needed.
There should be a `error-message` slot, but it doesn't work.

bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
```diff
@@ -102,6 +106,11 @@ export default {
     pattern () {
       const ctx = this.configDescription.context;
 
+      if (ctx === 'week') {
+        // week number 1-52
+        return `(?:0?[1-9]|[1-4][0-9]|5[0-2])`
```
florian-h05: I think week is a INTEGER/NUMBER type then?
Which would mean the config-parameter would have to map it to parameter-number with proper step and min/max configuration.

florian-h05:
BTW, in case someone needs a way to test this, here is a widget:

```yaml
uid: widget_3a0d5fc54a
tags: []
props:
  parameters:
    - context: mac-address
      label: MAC Address
      name: mac
      required: true
      type: TEXT
    - context: url
      label: URL
      name: url
      required: true
      type: TEXT
    - context: network-address
      label: Network Address
      name: net
      required: true
      type: TEXT
    - context: email
      label: Mail
      name: mail
      required: true
      type: TEXT
    - context: ip-address
      label: IP
      name: ip
      required: true
      type: TEXT
    - context: week
      label: Week
      name: week
      required: true
      type: TEXT
  parameterGroups: []
timestamp: Feb 27, 2026, 12:19:35 PM
component: f7-card
config: {}
```

bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
andrewfg: Ok. I will move it.

Nadahar:
> Here are the requirements for a valid WINS/NetBIOS name:

Regardless of what you find of definitions, names with spaces works. They tried to "restrict this" as they moved to AD where name resolution is based on DNS - aka "DNS interoperability". But, this is beside the point. WINS/NetBIOS is just one name resolution service. When you combine what "all of them" can allow, you end up with "everything allowed".

I simply cannot understand what you hope to achieve with this "validation", it's not the syntax that decides if a network address is "valid", but whether or not it exists on the (local) network. "Soft validation" still serves no purpose, it's just a workaround to allow using valid names - but how does having a "warning" for a valid name help the user?

Nadahar:
As far as I can tell, if the network address doesn't "conform" to DNS rules, users will be met with "Value does not match expected network address format.". I really think an explanation is needed for what is "the expected format" and why. Otherwise, the user gets the impression that he/she does something wrong, even though they use a perfectly valid value.

andrewfg:
florian-h05 there were a few issues updating the custom error message when focus shifted, and modal dialogs showed, however I have cleaned it up, and it all seems to be working now; and the regex for validating the url is now independent from the one for network-address.  => IMHO it is all ready to go!

---

PS @florian-h05 you have this marked as "Awaiting Other PR" but in reality it can be merged without dependencies. We do have a discussion about applying a schema, but whether it is applied or whether the contexts are simply documented, it still makes full sense to implement this PR.

---

image

bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
andrewfg: fixed.

andrewfg:
Just for the avoidance of doubt, I have removed support for the contested ip-address context.

andrewfg:
@florian-h05 the respective PRs in OH-Core, OH-Website, and OH-Doc are now merged.

florian-h05:
Your changes since my last reviewed seem to have introduced a bug, I am getting this error when opening the config dialog of my widget above:

```
widget-edit.vue:11 [Vue warn]: Unhandled error during execution of mounted hook 
  at <ParameterText read-only=false config-description= 
{context: 'mac-address', label: 'MAC Address', name: 'mac', required: true, type: 'TEXT'}
 value=undefined  ... > 
  at <F7ListGroup key=0 > 
  at <F7List ref="parameter" class="config-parameter" no-hairlines-md=true  ... > 
  at <ConfigParameter key="mac" config-description= 
{context: 'mac-address', label: 'MAC Address', name: 'mac', required: true, type: 'TEXT'}
 value=undefined  ... > 
  at <AsyncComponentWrapper key="mac" config-description= 
{context: 'mac-address', label: 'MAC Address', name: 'mac', required: true, type: 'TEXT'}
 value=undefined  ... > 
  at <F7Col > 
  at <F7Row key=0 > 
  at <F7Block width="100" class="parameter-group no-margin no-padding" > 
  at <F7Col > 
  at <F7Block key=0 class="config-sheet no-margin" ref="sheet" > 
  at <ConfigSheet parameterGroups= 
[]
 parameters= 
(6) [{…}, {…}, {…}, {…}, {…}, {…}]
 configuration= 
{}
 > 
  at <F7Col > 
  at <F7Block key=0 class="no-padding" > 
  at <F7PageContent ptr=false ptrDistance=undefined ptrPreloader=true  ... > 
  at <F7Page key=0 > 
  at <F7Popup ref="widgetProps" close-on-escape="" class="widgetprops-popup"  ... > 
  at <F7PageContent ptr=false ptrDistance=undefined ptrPreloader=true  ... > 
  at <F7Page onPage:afterin=fn<bound onPageAfterIn> onPage:beforeout=fn<bound onPageBeforeOut> > 
  at <WidgetEdit key="1772708827574_1" uid="widget_3a0d5fc54a" f7router= 
Router {eventsParents: Array(1), eventsListeners: {…}, params: {…}, isAppRouter: false, app: _Framework7, …}
  ... > 
  at <F7View url="/" main=true class="safe-areas"  ... > 
  at <App>

parameter-text.vue:172 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'on')
    at Proxy.mounted (parameter-text.vue:172:13)
    at async Proxy.validateSoft (parameter-text.vue:203:9)
```

bundles/org.openhab.ui/web/src/components/config/config-parameter.vue
```diff
@@ -116,7 +125,7 @@ export default {
         return ParameterRule
       } else if (configDescription.type === 'TEXT' && configDescription.context === 'channel') {
         return ParameterTriggerChannel
-      } else if (configDescription.type === 'TEXT' && configDescription.context === 'persistenceService') {
+      } else if (configDescription.type === 'TEXT' && configDescription.context && configDescription.context.indexOf('persistence') == 0) {
```
florian-h05:
```suggestion
      } else if (configDescription.type === 'TEXT' && configDescription.context && configDescription.context.indexOf('persistence') === 0) {
```

andrewfg:
> changes since my last reviewed seem to have introduced a bug

OK. I will look into that. Personally I was testing with the Main UI thing properties screen rather than using your widget.

Nadahar:
Nobody has ever explained what the "soft validation" of the `network-address` achieves. In which situation does it *help* users? I know that it will be an annoyance for some, I just can't understand who it will make life better for.

Nadahar:
I just got an idea: If the point here is to find *some meaning* with using the `network-address` context, instead of the "soft validation" it could add a small (i) icon to the field, that if clicked could open a pop-up explaining that "this field must either contain an IP address or a resolvable name that allows OH to reach the device" or similar. That would make this context "do something", without causing annoyance.

andrewfg:
> changes since my last reviewed seem to have introduced a bug

@florian-h05 I just committed a couple of edits that I think resolve your issue. To be honest I did see some other error message in the Chrome browser console but I think it is related to a quite different control element, so I could not debug it.

Please note that we have changed week from text to integer, and removed ip-address, so you need to modify your widget accordingly.

```yaml
uid: widget_3a0d5fc54a
tags: []
props:
  parameters:
    - context: mac-address
      label: MAC Address
      name: mac
      required: true
      type: TEXT
    - context: url
      label: URL
      name: url
      required: true
      type: TEXT
    - context: network-address
      label: Network Address
      name: net
      required: true
      type: TEXT
    - context: email
      label: Mail
      name: mail
      required: true
      type: TEXT
    - context: week
      label: Week
      name: week
      required: true
      type: INTEGER
  parameterGroups: []
timestamp: Feb 27, 2026, 12:19:35 PM
component: f7-card
config: {}
```

Also @florian-h05 I think I discovered a general bug in the widget code, which I think goes beyond the scope of this PR. Namely that even if the inputs have validation errors (e.g. you entered week == 123) the `Save` button does still persist the values. The case is different in the Main UI Thing properties page where the `Save` button does actually check the validity of inputs.

andrewfg:
>  add a small (i) icon to the field

The underlying native field only allows a text, so the overridden field has to be text only too. So it cannot have an actual icon. But instead I used a text with an icon like character in it `⚠ This is not a standard URL, IP address, or host name.` with an `onclick` handler that shows a popover.

bundles/org.openhab.ui/web/src/components/config/controls/validation-pattern.ts
```diff
@@ -0,0 +1,36 @@
+const hex2 = `[0-9A-Fa-f]{2}`
+const hex4 = `[0-9A-Fa-f]{4}`
+/**
+ * MAC address handler.
+ * Supports:
+ * - `AA:BB:CC:DD:EE:FF`
+ * - `AA-BB-CC-DD-EE-FF`
+ * - `AABB.CCDD.EEFF`
+ */
+export const MacAddress: string = `(?:(?:${hex2}[:\\-]){5}${hex2}|(?:${hex4}\\.){2}${hex4})`
```
lsiepel: Specific reason not to support undelimited mac adress?
AABBCCDDEEFF

andrewfg: No reason. I will add it. And then we can let the PR in addons refer to this context.

andrewfg:
> LGTM now, thanks!

Many thanks @florian-h05 and apologies that it took so long..

PR #3920, opened 2026-02-23
useStatesStore: Set `toString` override for ItemState & Fix overflow in developer sidebar

florian-h05:


florian-h05:
Can you backport this please?

PR #3923, opened 2026-02-23
Refactor add-on store components to TypeScript Composition API

florian-h05:


bundles/org.openhab.ui/web/src/App.vue
```diff
@@ -505,7 +505,7 @@ import { i18n, loadLocaleMessages } from '@/js/i18n'
 
 import { useI18n } from 'vue-i18n'
 
-import { AddonIcons, AddonTitles } from '@/assets/addon-store'
+import { AddonIcons, AddonTitles } from '@/assets/addon-store.ts'
```
jsjames: just curious as to why you are adding the .ts extension - is that something you want to do to show which addons have been converted?

bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
```diff
@@ -122,41 +122,36 @@
...
+
+// props
+const props = defineProps<{ addon: api.Addon, headline?: string, installActionText?: string, lazyLogo?: boolean }>()
+
+// emits
+const emit = defineEmits(['addon-button-click'])
+
+// computed
+const autoHeadline = computed<string>(() => {
+  // @ts-expect-error: like_count is not typed
```
jsjames: instead of bypassing the error - would it be better to just to the type conversion? `const likeCount = props.addon.properties.like_count as number | undefined`

bundles/org.openhab.ui/web/src/components/addons/addon-info-table.vue
```diff
@@ -37,168 +37,175 @@
...
-          afterIcon: 'question_circle_fill',
-          linkUrl: this.addon.link
-        })
-      }
+  if (props.addon.properties && props.addon.properties.updated_at) {
+    info.push({
+      id: 'updated',
+      title: 'Updated At',
+      // @ts-expect-error: updated_at is not typed
+      value: dayjs(props.addon.properties.updated_at).utc('z').local().format('LLL')
```
jsjames: i didn't investigate, but is the typescript definition for Dayjs.utc(keepLocalTime?: boolean | undefined) correct? If so, is 'z' the right parameter?

jsjames:
I submitted a PR to your repository with a few other cleanup suggestions.

florian-h05:
Thanks for the suggestions, I just merged them :+1:

bundles/org.openhab.ui/web/src/App.vue
jsjames: @florian-h05 - this one isn't a big deal - but curious why you decided to include the ts extension? Is that something you want to adopt globally?

PR #3934, opened 2026-02-27
doc/oh-map-page: typography

dilyanpalauzov:


florian-h05:
This change will get overridden the next time the component docs are regenerated. The edited code is inside `<!-- GENERATED props -->` comments.

jsjames:
> This change will get overridden the next time the component docs are regenerated. The edited code is inside `<!-- GENERATED props -->` comments.

Thanks - i missed that.  I will submit a PR to fix in the correct location.

PR #3956, opened 2026-03-02
oh-canvas: Fix broken oh-canvas-item & Refactor to TS

florian-h05:
Regression from #3862.

#3862 broke the oh-canvas-item due to lifecylce issues with the <setup script> and the setup() method in the Options API script.
This fixes that issue by refactoring oh-canvas-item to use only the Composition API.
Also refactor oh-canvas-layer.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
@@ -90,7 +90,7 @@
     </vue-draggable-resizable>
     <div
       v-else
-      ref="root"
+      :ref="root"
       :style="{
         left: x + 'px',
         top: y + 'px',
@@ -186,7 +186,7 @@
 </style>
 
 <script setup lang="ts">
-import { ref, computed, watch, onMounted, onBeforeUnmount, useTemplateRef } from 'vue'
+import { ref, computed, watch, onMounted, onBeforeUnmount, type VNodeRef } from 'vue'
 import 'vue-draggable-resizable/style.css'
 import VueDraggableResizable from 'vue-draggable-resizable'
 
@@ -220,7 +220,7 @@ const emit = defineEmits<OhCanvasItemEmits>();
 const { config, visible, childContext, defaultSlots } = useWidgetContext(props.context)
 
 // data (state)
-const root = useTemplateRef('root')
+const root = ref<VNodeRef | null>(null)
 const reloadKey = ref(0)
 const dragging = ref(false)
 const resizing = ref(false)


jsjames:
I haven’t looked at this yet, but do you think this issue will affect other components or was there something unique about the oh-canvas-item?

florian-h05:
I think there was something unique:

It had a setup script that imported the components, and a normal script with a setup section. So both Composition and Options API were used in parallel. IIRC there are some other components that use a setup script for imports but none of them was a widget.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-layer.vue
```diff
@@ -20,64 +24,62 @@
...
-              selected: false,
-              id: Math.random().toString(36).substring(2)
-            })
-          } else {
-            console.log('Wrong component type in canvas layer: ' + item.component)
-          }
-        })
-      }
-      this.layer = layer
+defineExpose({ widget: OhCanvasLayerDefinition })
```
jsjames: Is the defineExpose the correct macro to use here vs. defineOptions? They may both work - I had been using defineOptions, but if this is the better choice we can adopt defineExpose.

florian-h05: I think defineOptions might be more suitable, will change that.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
```diff
@@ -185,204 +185,226 @@
...
+  (e: 'oci-dragged', id: string, x: number, y: number): void
+}
+
+const emit = defineEmits<OhCanvasItemEmits>();
+
+// composables
+const { config, visible, childContext, defaultSlots } = useWidgetContext(props.context)
+
+// data (state)
+const root = ref<VNodeRef | null>(null)
```
jsjames, line 223: I've been using useTemplateRef to get access to component refs as recommend in vue 3.5+. Not critical change though.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-layer.vue
florian-h05: Changed

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
```diff
@@ -185,204 +185,226 @@
...
+const resizing = ref(false)
+const active = ref(false)
+// run mode only:
+const ready = ref(false)
+const resizeObserver = ref<ResizeObserver | null>(null)
+
+// computed
+const x = computed({
+  get: () => (config.value?.x as number) ?? 20,
+  set: (val) => { props.context.component.config.x = val }
```
jsjames, line 237: since this modifies a prop, I think its better to use provide/inject vs. manipulating the prop. Maybe that's not possible though if the context is unique per instantiation?

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
florian-h05: Changed

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
florian-h05: This would behave significantly different from the previous code then, and I don't understand how provide/inject would help here.

jsjames: Understood - and see that it would complicate to either use provide/inject or emits.

Let's add a comment like:
// Direct mutation is intentional - parent context expects config changes

We should just try to limit this in the codebase where we can.

florian-h05: I agree that mutating props is bad behaviour, but don't know a better "path" yet.

florian-h05: I am currently looking into it.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
florian-h05: Reverted that change as it broke auto size

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
florian-h05: So, doing this properly would be way more complicated, so let's leave it

Issue #3866, opened 2026-02-11
[MainUI] Charts data misalignment on oh-category-axis when using series offset


michaeljoos72:
## The problem

When using oh-aggregate-series with an offset amount (e.g., to compare current year data with the previous year), the data points for the offset series are not aligned correctly with the X-axis categories.

In my use case, using offset amount: 1 and offset unit: year results in data being shifted horizontally—for example, values belonging to March are being displayed at the February position on the axis.

<img width="1950" height="1113" alt="Image" src="https://github.com/user-attachments/assets/df13ca8f-4098-497a-b999-8a377991a4e2" />

```
config:
  chartType: month
  icon: iconify:game-icons:high-grass
  label: Rasen (GTS)
  options:
    color:
      - green
      - lightgreen
  order: "35"
  period: Y
  sidebar: true
slots:
  dataZoom:
    - component: oh-chart-datazoom
      config:
        show: true
        type: inside
  grid:
    - component: oh-chart-grid
      config:
        bottom: 10%
        containLabel: false
        show: false
        top: "120"
  legend:
    - component: oh-chart-legend
      config:
        bottom: "10"
        orient: horizontal
        show: true
  series:
    - component: oh-aggregate-series
      config:
        aggregationFunction: last
        dimension1: date
        gridIndex: 0
        item: Gruenland_Temp_Summe
        lineStyle:
          width: 2
        markArea:
          data:
            - - yAxis: 200
              - itemStyle:
                  color: rgba(50, 205, 50, 0.08)
                yAxis: 300
            - - name: Vegetationsbeginn
                yAxis: 150
              - itemStyle:
                  color: rgba(193, 255, 193, 0.1)
                yAxis: 200
          silent: true
        markPoint:
          data:
            - name: max
              type: max
        name: Grünland-Temperatursumme
        service: rrd4j
        smooth: false
        transpose: false
        type: line
        xAxisIndex: 0
        yAxisIndex: 0
    - component: oh-aggregate-series
      config:
        aggregationFunction: last
        dimension1: date
        gridIndex: 0
        item: Gruenland_Temp_Summe
        lineStyle:
          opacity: 0.2
          type: dotted
        name: Grünland-Temperatursumme (Vorjahr)
        offsetAmount: 1
        offsetUnit: year
        service: rrd4j
        smooth: false
        transpose: false
        type: line
        xAxisIndex: 0
        yAxisIndex: 0
  title:
    - component: oh-chart-title
      config:
        left: 10%
[...]

PR #3958, opened 2026-03-02
Charts: Fix add/subtract period doesn't account for different month lenghts

florian-h05:
Regression from #3337.


florian-h05:
Should be backported to 5.1.x

cweitkamp:
Does it solve #3866 too? Because it removes the `duration()` method?

florian-h05:
I haven’t tested this.

I’m am planning to rework the whole offset handling soon (currently refactoring the charting code to TypeScript) and hence don’t look into issues with the current implementation. The current implementation IMO has a few drawbacks both in configuration and implementation, oh-State-series also doesn’t work properly with offsets.

PR #3962, opened 2026-03-04
Charts: Improve tooltip for markArea using Item's stateDescription & Hide markArea series from timeseries tooltip

florian-h05:
Depends on #3977.

Improve the oh-chart-tooltip smartFormatter:

- Compact time range for mark area to a single line.
- Display mapped Item state for mark area Item (if available) based on Item state description.
- Hide mark areas from tooltip when hovering over time series.
- Append unit to timeseries value.


jsjames (approved):
LGTM

PR #3963, opened 2026-03-05
useWidgetExpression: Add type param for value = return type

florian-h05:


florian-h05:
Self-merging as this only changes doc and type defs and I need the type for my refactoring.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
```diff
@@ -66,6 +66,13 @@ interface ScreenInfo {
   appHeight: number
 }
 
+export type EvaluateExpressionFn = <T = unknown>(
```
jsjames, line 69: just curious on this (not suggesting change) - was it required to define the type - wouldn't the type have been inferred based on the function? Maybe it is required for template functions?

florian-h05: I need the type for declaring the function as a parameter somewhere else.

PR #3966, opened 2026-03-05
Remove unused template7 dependency

florian-h05:


florian-h05:
Out of interest: Why do you always merge the main branch into a PR? There should be no need for this, and generally rebasing is much better as it doesn’t appear in the commit history.

jsjames:
> Out of interest: Why do you always merge the main branch into a PR? There should be no need for this, and generally rebasing is much better as it doesn’t appear in the commit history.

I assumed when this comes up that it is best to update the branch and so I alway just hit the button to update.  So, do you not update it at all, or do you manually check it out rebase and push the change?

image

florian-h05:
Normally there is no need to update to be in sync with the main branch. The only reason is if the main branch contains changes for openHAB core changes, then a rebase might be required to get those changes into a PR and have Java compilation succeed again. This mainly happens on Karaf upgrades, but was very rare in the last time.

PR #3973, opened 2026-03-07
`$oh.api`: Correct type definitions

florian-h05:


florian-h05:
Self-merging this minor change to type defs

PR #3974, opened 2026-03-07
Refactor chart component definitions to WidgetDefinition & Parameter

florian-h05:


florian-h05:
Regenerating both the component docs and the component type defs shows this does not introduce any functional changes, so I'll self-merge this soon.

PR #3975, opened 2026-03-08
Reduce unwanted JavaScript editor autocompletion in comments and outside code expressions.

jimtng:
Address https://community.openhab.org/t/annoying-action-in-code-editor/168720


ErikDB87:
I noticed `action` is also suggested after `[`. Maybe also something to remove (if that's not already the case)?

jimtng:
> I noticed `action` is also suggested after `[`. Maybe also something to remove (if that's not already the case)?

No the autocompletion it's not disabled after `[`. Should it be?

ErikDB87:
> > I noticed `action` is also suggested after `[`. Maybe also something to remove (if that's not already the case)?
> 
> No the autocompletion it's not disabled after `[`. Should it be?

I have no idea what this suggested `action` does, so when I type `[`, I want to create an array, which doesn't involve `action`.

jimtng:
I've flipped the logic now so that the implicit autocompletion is triggered only in specific cases. This should reduce unexpected autocompletions

PR #3976, opened 2026-03-09
Display channel properties in channel-general-settings 

andrewfg:
Currently the Main UI channel general settings view does not display any channel properties.

This PR adds such a display (see screen shot below).

---

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/channel/channel-general-settings.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/channel/channel-general-settings.vue
@@ -1,5 +1,5 @@
 <template>
-  <f7-block class="padding-vertical no-padding-horizontal">
+  <f7-block class="channel-general-settings padding-vertical no-padding-horizontal">
     <f7-col>
       <f7-list class="no-margin" inline-labels no-hairlines-md>
         <f7-list-input
@@ -81,14 +81,14 @@
 </template>
 
 <style lang="stylus">
-.list
-  .item-subtitle
-    overflow-wrap break-word
-    white-space inherit
-
-.property-key
-  display inline-block
-  padding-left 12px
+.channel-general-settings
+  .list
+    .item-subtitle
+      overflow-wrap break-word
+      white-space inherit
+  .property-key
+    display inline-block
+    padding-left 12px
 </style>
 
 <script>


jlaur:
@andrewfg - are channel properties documented somewhere? I'm still struggling to understand their purpose/value, and thus if the concept should be kept and added to the UI, or dropped.

andrewfg:
> are channel properties documented somewhere

Apparently not. Copilot says _"the openHAB documentation confirms that channels are part of the Thing model and can carry metadata, but it does not provide a dedicated section for channel properties"_.

However the rest of Copilot's output is very much on point (see below). => I will open a PR to add this top the documentation. _**EDIT: done!**_

In my case in the HomeKit binding I am using them for _"read‑only attributes that should not appear as configuration parameters"_

---

Channel properties in openHAB are a small but important part of the Thing/Channel model, and they behave differently from both Thing properties and Channel configuration parameters.

What channel properties are:
Channel properties are metadata attached to a channel, represented as a simple key–value map.
They are exposed through the REST API as part of the Thing structure and are typically used by bindings to store non-user-editable, binding-defined attributes.
Examples include:
- Identifiers derived from the device (service, scheme, source, index)
- Internal metadata needed by the handler
- Read‑only attributes that should not appear as configuration parameters
The openHAB documentation confirms that channels are part of the Thing model and can carry metadata, but it does not provide a dedicated section for channel properties. They appear in the REST response for a Thing, as noted in community discussions.

How channel properties differ from other channel metadata
1) Channel configuration parameters
These are user-editable and appear in the UI as “Configure Channel”. They are defined in the binding’s XML (channel-type → config-description). They are persisted separately and can be changed by the user.
2) Channel properties
These are binding-defined, usually read-only, and not part of the configuration description. They are returned in the Thing JSON under each channel. They are not editable in Main UI unless the UI explicitly exposes them.
3) Thing properties
These belong to the Thing itself, not to channels.

andrewfg:
@jlaur I added a screen shot at the top of this PR ..

andrewfg:
See also https://github.com/openhab/openhab-docs/pull/2671

andrewfg:
This PR is ready to go!

florian-h05:
Out of interest: Which binding makes use of channel properties?

andrewfg:
> Out of interest: Which binding makes use of channel properties?

@florian-h05 see this [comment](https://github.com/openhab/openhab-core/issues/5278#issuecomment-4019773015) .. in general it seems there are ca. 36 bindings that use them; .. and speaking for myself I authored the code for the HomeKit binding that uses them.

PR #3978, opened 2026-03-10
oh-input: Fix value computation for variables

florian-h05:


florian-h05:
This has to be backported to 5.1.x

PR #3969, opened 2026-03-06
Dynamically import `oh-` components through a registry & Improve chunking

florian-h05:
- Removes the static import for all `oh-` components from #3862 by introducing a registry that provides async access to all components.
- Dynamic import eslint-linter-browserify.
- Fix incorrect lodash import (improves chunking).
- oh-sipclient: Code improvement for JsSIP dynamic import.
- oh-video: Import different implementations statically, instead import video.js dynamically.

PR #3990, opened 2026-03-13
oh-masonry: Fix missing oh-placeholder-widget

jsjames:
Regression from #3969.


florian-h05:
Regression from #3969?

jsjames:
yes - i confirmed the other places where this is used it is imported.

florian-h05 (approved):

PR #3992, opened 2026-03-15
Update OpenAPI REST client

jsjames:
Refs https://github.com/openhab/openhab-core/pull/5404.

Includes corrections to the openapi spec to address several of the persistence endpoints.


florian-h05:
Did you take the API description from your core PR or the current snapshot? I'm getting a slightly different OpenAPI JSON from snapshot 5.2.0 Build 5237.

jsjames:
it was from my core PR and I updated all the affected bundles in a install on my end and then pulled it from /rest/spec endpoint. Not sure if there were other PRs that changed the spec?

florian-h05:
I haven't checked the PRs, but I can see a diff in the OpenAPI spec. Should I place the latest JSON in your PR and update?

jsjames:
> I haven't checked the PRs, but I can see a diff in the OpenAPI spec. Should I place the latest JSON in your PR and update?

Yes - that would be good - i'll look at the deltas as well to make sure they don't effect anything.

PR #3997, opened 2026-03-15
parameter-text: Fix extraneous props and use of f7-popover-inner

jsjames:
Regression from #3918 

Addresses:
```
[Vue warn]: Failed to resolve component: f7-popover-inner
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement. 
  at <ParameterText read-only=false config-description= {context: 'ip-address', label: 'IP', name: 'ip', required: true, type: 'TEXT'} value=undefined  ... > 
  at <F7ListGroup key=0 > 
  at <F7List ref="parameter" class="config-parameter" no-hairlines-md=true  ... > 
```
and

```
[Vue warn]: Extraneous non-props attributes (parameters, configuration, title, f7router) were passed to component but could not be automatically inherited because component renders fragment or text or teleport root nodes. 
  at <ParameterText read-only=false config-description= {context: 'week', label: 'Week', name: 'week', required: true, type: 'TEXT'} value=undefined  ... > 
  at <F7ListGroup key=0 >
```

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
@@ -1,5 +1,5 @@
 <template>
-  <ul v-if="multiple" ref="inputs">
+  <ul v-if="multiple" ref="inputs" class="parameter-text">
     <f7-block-header class="no-margin">
       <div class="margin-horizontal item-label" style="padding-top: var(--f7-list-item-padding-vertical); color: var(--f7-text-color)">
         {{ configDescription.label }}
@@ -30,7 +30,7 @@
       @focus="gotFocus"
       :placeholder="configDescription.placeholder" />
   </ul>
-  <ul v-else>
+  <ul v-else class="parameter-text">
     <f7-list-input
       ref="input"
       :floating-label="theme.md"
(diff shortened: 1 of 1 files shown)


jsjames:
@andrewfg - can you please verify this still works as intended - these were console warnings based on the #3918 PR.

I just removed the reference to f7-popover-inner on the network-address context - it seems to still render correctly.

andrewfg:
> please verify this still works as intended

Just two things..

1. I tested it on the configuration page of a real thing. And the popover does appear. However I recall that in the prior version when moving the mouse over the text that should trigger the popover the mouseover cursor would change to a hand cursor. And now the mouseover cursor is a plain caret like the background. So in other words there is no clue that there may be a popover to pop.

2. In the original PR @florian-h05 had posted a [custom widget](https://github.com/openhab/openhab-webui/pull/3918#issuecomment-4006335793) for testing. Which was useful since it runs the page in a detached context than on the configuration page of a real thing. However I can't figure out how I installed that widget on my test system, so I was not able to test it in such detached context.

florian-h05:
> And now the mouseover cursor is a plain caret like the background. So in other words there is no clue that there may be a popover to pop.

Will fix this with some CSS.

I've tested the changes with my custom widget and it looks good. You can install the widget through Developer Tools => Widgets => `+` Button => Paste YAML

PR #4004, opened 2026-03-17
Enable oxfmt for embedded languages

jsjames:
Setting this to auto did cause failures with multi-line events as oxfmt separated into separate lines and removed the embedding semicolon. To address, i added a flag to not format those lines.

Not sure if we should submit bug to oxfmt and/or force multiline events to be encapsulated in a function call?


florian-h05 (approved):
LGTM, thanks!

florian-h05:
> Not sure if we should submit bug to oxfmt and/or force multiline events to be encapsulated in a function call?

Ideally we do both.

PR #4010, opened 2026-03-20
Fix oh-chart/other components not loading in production build

jsjames:
`oh-chart`, `oh-knob` and possibly other components as well weren't loading in the prod build.
Additionally, widget edit also wasn't working anymore. 
Reported on the community: https://community.openhab.org/t/openhab-5-2-snapshot-discussion/167691/13

This fixes the issue by globally registering asyncComponents for widgets only, affecting import resolution.
Pages and tabs are explicitly imported where needed, but in most cases async as well.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/home/overview-tab.vue
+++ b/bundles/org.openhab.ui/web/src/pages/home/overview-tab.vue
@@ -17,12 +17,7 @@
     </f7-block>
 
     <template v-else>
-      <oh-layout-page
-        v-if="overviewPage"
-        v-show="!inChatSession"
-        :context="overviewPageContext"
-        :class="{ notready: !ready }"
-        :f7router />
+      <oh-layout-page v-if="overviewPage" v-show="!inChatSession" :context="overviewPageContext" :class="{ notready: !ready }" :f7router />
       <div v-else-if="!inChatSession" class="empty-overview">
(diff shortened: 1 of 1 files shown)


jsjames:
@florian-h05 - I think this is a better fix - I learned that you can globally register a async component and it will only load when it is called. This simplifies the registry quite a bit and my initial bundle analysis - it addresses the other issue we were seeing of bloat.

florian-h05:
TBH, I am not sure I like global imports. IMO, it makes the code less maintainable when `oh-*` components can be just used without having an import where they are used.

digitaldan:
So my UI looks very broken with this PR, it seems like some images are not working...... but my plan markers do show ! 

image

image

florian-h05:
Icons aren't loading for you ... what version of that PR have you tried? The first commit only or the second as well?

florian-h05:
@jsjames I pushed some changes, please have a look at them.

jsjames:
@digitaldan - I presume this will work for you know, but it would be good to verify!!

jsjames:
> Fixes the issue 👍

LGTM

florian-h05:
I also expect it for Dan to work, have seen similar issues from time to time where assets weren’t loaded (request did not complete), always fixable with a reload.

digitaldan:
Thanks !!!! Yes, this PR fixes my issue and markers load as well.

PR #3993, opened 2026-03-15
oh-plan-page: Fix floor plan marker visibility

digitaldan:
Regression from #3862

PR #4012, opened 2026-03-20
[mainui] Fix floor plan markers not rendering in production build

digitaldan:
This is a follow up to #3993  .   That fix worked when running under the Vue dev server, but when i deployed the UI bundle, the markers remained hidden, which was really baffling.  I assumed at first this was a caching issue or i was not loading the right bundle, but after lots of troubleshooting, it become clear it was a vue issue.  I have to admit, i then used claude quite a bit to resolve this as i was a bit out of my league.  I would suggest we try to get this in our milestone release. 

Here's what it came up with, which does now work when running under the dev server and deployed.

The static import of oh-plan-marker.vue caused Rolldown's code-splitting (via import.meta.glob in the component registry) to place the marker in a separate chunk where its vue-leaflet imports (LMarker, LTooltip, LIcon) failed to resolve, rendering as raw HTML elements instead of Vue components.

Fix by resolving the marker through the component registry using defineAsyncComponent, matching the pattern used by all other widgets.

Also fix CSS specificity for .leaflet-div-icon override, which was being overridden by leaflet.css loading after the plan page's CSS chunk.

Additionally simplify zoom-based marker visibility by using a per-marker v-if with an isMarkerVisible method instead of imperatively filtering into a markers data array.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/plan/oh-plan-page.vue
@@ -27,7 +27,7 @@
       <l-feature-group v-if="context.component.slots" ref="featureGroup">
         <template v-for="(marker, idx) in defaultSlots" :key="idx">
           <component
-            :is="markerComponent(marker)"
+            :is="markerComponent"
             v-if="isMarkerVisible(marker)"
             :context="childContext(marker)"
             @update="onMarkerUpdate" />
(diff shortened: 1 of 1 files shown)


jsjames:
@digitaldan - can you try #4010 to see if that fixes the issue for you?  I think that should be a global fix for the build issues we are seeing (also in this thread: https://community.openhab.org/t/openhab-5-2-snapshot-discussion/167691/13?u=jsjames)

digitaldan:
Ah, so i'm not going crazy ;-) Yes i will give that a try  and let you know today.

jsjames:
Closed - covered by #4010

PR #4013, opened 2026-03-20
Fix oh-chart & oh-knob not loading in production build

florian-h05:
oh-chart-page was displaying an empty page and oh-knob was not rendering content in some occasions due to import issues in the production build.
These changes fix them and also merge oh-chart-component & oh-chart components.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-video.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-video.vue
@@ -23,7 +23,6 @@
 </template>
 
 <script>
-import { defineAsyncComponent } from 'vue'
 import { f7 } from 'framework7-vue'
 
 import { OhVideoDefinition } from '@/assets/definitions/widgets/system'


florian-h05:
@digitaldan Can you give this PR a try? It tries to fix the same as #4010, but in a different way (and includes a RFC).

jsjames:
Do you think the issue is with statically and dynamically linked components?  If so, do we need to update all imports to be defineAsync?  Any idea why this was just happening with oh-charts and Dan's issue with the markers?

florian-h05:
I think the issue might be as described in Dan's marker PR:
It's a combination of dynamic and static imports that cause bad code splitting.

jsjames:
> I think the issue might be as described in Dan's marker PR: It's a combination of dynamic and static imports that cause bad code splitting.do we need a more systematic way though to ensure this doesn’t happen?  Should we make all imports of widgets be defineasync?

florian-h05:
I am currently looking into this, and now have a possible solutuon based on your #4010 with some changes

florian-h05:
Closing in favor of #4010

PR #4017, opened 2026-03-20
oh-stepper: Fix `enableInput` always shows 0 & Refactor to TS

### LINKED ISSUE #3951: oh-stepper-item widgets always shows 0
I have a strange UI display problem with many `oh-stepper-item` widgets showing a 0 value.
When you change them, the true value appears and is correctly used as the starting point.
Reload also makes all values appear. But when I navigate elsewhere and back to the page it's all 0 again. FWIW this is on a tabbed page.
I have some stepper widgets on the same page that work think it happens only for those widgets with `enableInput: true`.

florian-h05:
Fixes #3951.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-stepper.vue
@@ -1,103 +1,109 @@
 <template>
   <f7-stepper
     ref="stepper"
-    v-bind="config"
+    v-bind="stepperConfig"
+    :value="value"
+    :input="config.enableInput === true"
+    :format-value="formatValue"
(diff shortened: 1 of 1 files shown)


florian-h05:
Just FYI, draft PRs should only by the author marked as ready for review. Never undraft PRs from other people.
Now that we have postponed M2 it doesn't matter but my plan was to hold back that PR until after M2 to reduce regression risks.

jsjames:
> Just FYI, draft PRs should only by the author marked as ready for review. Never undraft PRs from other people. Now that we have postponed M2 it doesn't matter but my plan was to hold back that PR until after M2 to reduce regression risks.

Sorry about that - I didn't see that it was a draft.

florian-h05:
You marked this PR as ready for review before merging ;-)

PR #4020, opened 2026-03-20
Charts: Don't enable boundary & itemState for forecast charts

florian-h05:
This fixes an issue where forecasted charts with bar time-series only displayed very narrow bars because of to short minimum time intervals caused by current Item state and/or boundary manipulation.


jsjames (approved):
LGTM - expect for formatting

jsjames:
btw, is there a way for me to fix the formatting issue w/o having to create a PR to your repository?  Can I fix and just push to origin?

florian-h05:
Yes, you can just fix and push to my branch

PR #4021, opened 2026-03-21
[basicui] Fix stale state when tab returns to foreground

jimtng:
When a browser tab is backgrounded, item state changes received via SSE are displayed correctly while the stream is alive, but the page becomes stale if the SSE connection drops (the 10-second reconnect setInterval is heavily throttled by browsers for background tabs). More importantly, even when the SSE connection stays alive throughout, the page never back-fills the changes that occurred while it was hidden, so the user always had to manually refresh to re-sync.

Fix using the Page Visibility API:

- On 'hidden':   close the SSE connection immediately (frees the server- side subscription socket) and cancel any in-flight reconnect interval.
- On 'visible':  POST a fresh subscription, then use the existing connectionRestored() path to reload the current sitemap page HTML and open a new SSE stream.  This gives an unconditional full state re-sync the moment the user returns to the tab, regardless of whether the SSE was alive or dead while the page was hidden.

ChangeListenerEventsource.pause() is overridden to actually close the EventSource (the base AbstractChangeListener.pause() only set a flag). ChangeListenerLongpolling.pause() already aborted its XHR correctly.

### DIFF
--- a/bundles/org.openhab.ui.basic/web-src/smarthome.js
+++ b/bundles/org.openhab.ui.basic/web-src/smarthome.js
@@ -3816,6 +3816,17 @@
 			_t.closeConnection();
 			_t.paused = true;
 		};
+		// Resume from hidden state: re-establish subscription via POST so that
+		// connectionRestored -> startSubscriber flow is used (reload + SSE reopen)
+		_t.resume = function() {
+			_t.paused = false;
+			ajax({
+				url: _t.subscribeRequestURL,
+				type: "POST",
+				callback: _t.connectionRestored,
+				error: _t.subscriberError
+			});
+		};
 	}
 
 	function ChangeListenerLongpolling() {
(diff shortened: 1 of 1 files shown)


jimtng:
ping @lolodomo hoping this would make it into 5.2

lolodomo:
It is on my TODO list to review that.
My current personal priority is all the interesting stuff about sitemap started by @mherwege I am participating too.

bundles/org.openhab.ui.basic/web-src/smarthome.js
```diff
@@ -4050,6 +4057,44 @@
...
+			// A new subscribe POST gives us a fresh subscription ID.  The
+			// connectionRestored callback then reloads the page HTML (full state
+			// re-sync) and opens the new SSE stream – exactly the same flow used
+			// after a normal connection-error recovery.
+			ajax({
+				url: _t.subscribeRequestURL,
+				type: "POST",
+				callback: _t.connectionRestored,
+				error: _t.subscriberError
+			});
```
lolodomo: This part seems to not be adapted to long polling mode.
Why not rather call _t.resume() and override _t.resume in event source mode ?

lolodomo:
I will test before merging.

PR #4069, opened 2026-03-31
Things list: Add dynamic blue dot to status badges if status detail is set

andrewfg:
The purpose of this PR is to allow a dynamic information badge ("blue dot") to be applied to Things in the overview page in order to mark when the Thing has a special status. For example "low battery" or "firmware update available".
 
Resolves #4063

When a Thing's `statusInfo.status === 'ONLINE'` and its `statusInfo.description is not empty` then this PR adds a "blue dot" information badge to the overview list as shown in the screenshot below.

image

image

#### Example code for setting and removing the badge
```
ThingStatusInfo status = thing.getStatusInfo();

// show the badge
updateStatus(status.getStatus(), status.getStatusDetail(), "I have a special status");

// hide the badge
updateStatus(status.getStatus(), status.getStatusDetail(), null);
```


dilyanpalauzov:
Why not just render the description directly, instead of putting the description in a tooltip?

andrewfg:
> Why not just render the description directly

The text may be quite long. So on the Things list page we show a badge. Whereas on the Thing detail page the text is shown in full.

Also the idea is that the text may be dynamically added or removed so the badge is only displayed when there is something special to say. For example it can display a badge when there is a software update available, and no badge when there is not.

In other words the badge only shows when something new has happened. Rather like the blue dot your inbox when there is mail. Or the blue dot on an app on your phone when it has an update. Or when a device has a low battery, etc.

---
EDIT: I added some sample code for setting and removing the badge in the top post above.

dilyanpalauzov:
> The text may be quite long. So on the Things list page we show a badge. Whereas on the Thing detail page the text is shown in full.

I agree this is a valid point of view.

andrewfg:
### Pending!!

Perhaps a [blue dot status badge](https://framework7.io/docs/badge) would look better ??

bundles/org.openhab.ui/web/src/pages/settings/things/things-list.vue
```diff
@@ -217,6 +223,25 @@
   margin-bottom calc(var(--f7-fab-size) + 2 * calc(var(--f7-fab-margin) + var(--f7-safe-area-bottom)))
 </style>
 
+<style scoped>
```
florian-h05: We use Stylus in all parts of the UI:

`<style scoped lang="stylus">`

You can then leave out the brackets, colons and semicolons.

andrewfg: Many thanks @florian-h05 for the tip. I have changed the code to use stylus. There were two style blocks but since both of them now use stylus I merged them into one block. NOTE: that the original block was not 'scoped' and now it is 'scoped'. I think that is Ok, but please confirm.

florian-h05: The issue with scoped is that if you need to manipulate the styling of a F7 component used in your component that might not work (at least in my experience, hence we usually don’t use scoped styles but instead have a „wrapper class“ at the top level of the component.

andrewfg: @florian-h05 sorry but I dont really understand the issue; I am happy if it is with or without scoped; so please tell me what you prefer.

andrewfg: I removed the 'scoped' ..

florian-h05 (approved):
LGTM, thanks!

I've changed the intendation of the new styling to ensure it is encapsulated inside .things-list.

PR #4095, opened 2026-04-03
Update setup wizard images

florian-h05:
Follow-up for #3845.


florian-h05:
@mherwege WDYT?

mherwege:
Much better.

Just the last one would be nice if colors of the arrows and titles would match. One of the equipment has almost the same color as the location. It looks a bit inconsistent.

Still better then what I provided, so a definite improvement.

bundles/org.openhab.ui/web/src/images/wizard-semantics.png
jsjames, line 1: I'm not sure I understand the color here?  Color is used on location, equipment and point on the labels below the image, but then the image uses different colors for Location, Equipment and Point.

florian-h05:
I have now updated the images once again after "refactoring" them in Affinity Photo, keeping the overall structure, but using rounder edges, more transparency, a consistent color set etc.

mherwege:
Did you try them on a dark theme? Some of the parts in the last image have a white background, which may look awkward.
Do the grey arrows work well?

florian-h05:
image
image
image

I think it looks good in dark mode.

mherwege:
LGTM

boehan:
It may be extremely nitpicking and I'm absolutely fine, if you disagree (given that the images are usually only shown once to the average user), but I wanted to share a few points I noticed with those images:
- the icon for the switch(?) item in the first image looks rather "uncommon" to me
- not sure which kind of equipment the right one in the 3rd image should be, but it looks a bit too abstract IMHO
- consistency
  - the light bulbs in image 1 and 3 are different designs
  - thermometers differ between the first two and the third image (and slightly within image 3 again)
  - different design concepts: 3d-icons (lightbulb/-switch, thermostat, windows) vs. 2d-icons (strategies, all in image 3, slider in 1)

As written, these are just some things I noticed when overlooking the PR and wanted to mention. Unfortunately, I can't provide a better suggestion, but maybe there is a suitable icon set somewhere (e.g. on iconify) that could fit most purposes.

florian-h05:
I haven't found any icon set that provides all icons we need and looks good, but I have improved the images further, removing most of the inconsistencies.

Issue #2463, opened 2024-03-08
Open things, items and rules in new tab when pressing middle mouse button


mueller-ma:
## The problem

Pressing the middle mouse button usually opens links in new tabs. This doesn't work in Main UI.
E.g. on `/settings/things/` the middle mouse button click on a thing opens `/settings/things/` in a new tab, but not the clicked thing.

PR #4106, opened 2026-04-10
Revert "list navigation: Fix middle-mouse button navigation (#4036)"

### LINKED ISSUE #4098: [MainUI] Multiple thing selection does not work correctly
## The problem

Not sure if it's on my configuration, or if it's global.
When I try to select multiple thing in the thing UI admin list using the Ctrl + click.
I've see the checkbox for selection, but just after, my browser navigate to the thing details page.
And so the multiple selection is not possible anymore.

Do someone else observe this ?

## Your environment

Windows + Chrome browser

### LINKED ISSUE #4104: Ctrl + click in lists is broken
## The problem

Ctrl + click in lists used to enable "select mode". Now instead, it opens the item, although it seems like select mode is briefly entered before leaving the page. It seems like the router action is no longer prevented.

## Expected behavior

I would expect "select mode" to be enabled while remaining on the list page.

## Steps to reproduce

Go to any list view, e.g "Rules". Ctrl + click a rule and observe what happens.

## Your environment

I'm running 47cd64672073ac4a5e3f75d536949ef3b3427728 in dev mode.

jsjames:
This reverts commit 3d964429a682124afa69c1e1aafd29f9f8b31926.

Fixes #4104
Fixes #4098

Reopens #2463

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/routes.js
+++ b/bundles/org.openhab.ui/web/src/js/routes.js
@@ -545,10 +545,7 @@ export default [
             path: ':uid',
             beforeEnter: [enforceAdminForRoute],
             beforeLeave: [checkDirtyBeforeLeave],
-            async: loadAsync(WidgetEditPage, ({ to }) => (to.params.uid === 'add' ? { createMode: true } : {})),
-            options: {
-              animate: false
-            }
+            async: loadAsync(WidgetEditPage, ({ to }) => (to.params.uid === 'add' ? { createMode: true } : {}))
           }
         ]
       },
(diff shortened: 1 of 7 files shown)


florian-h05 (approved):
Quite unfortunate that we have to revert this for now.

jsjames:
> Quite unfortunate that we have to revert this for now.

Agree, I looked for easy way to make sure f7 would not always follow link, but couldn’t find. I think implementing the middle mouse button move in another way vs relying on f7 is the better way to proceed.

PR #4109, opened 2026-04-12
list-navigation: Fix middle-mouse navigation

### LINKED ISSUE #2463: Open things, items and rules in new tab when pressing middle mouse button

jsjames:
Fixes #2463

Browser handles middle-mouse button based on the link href and framework7 does not handle. So, to support this, the link must be specified.

When a link is specified, one must include prevent-router to prevent framework7 from automatically routing based on a click. We then handle the route navigation appropriate in the click event.


florian-h05 (approved):
LGTM, thanks!

PR #4111, opened 2026-04-12
Support invoking multiple widget actions

svnsssd:
## Description
This PR adds support for multiple widget actions.

Changes included:
- Extend action definition handling for actions.
- Update widget action execution.

## Testing
- Verified action execution in widget action flow.
- Built UI bundle locally with Maven (`mvn clean install -DskipTests -pl :org.openhab.ui`).

needed by DahuaDoor Intercom PR: https://github.com/openhab/openhab-addons/pull/20570


florian-h05 (changes requested):
Thanks for your contribution @svnsssd!

While I see the benefit of such a feature, I think it is implemented the wrong way. If we start merging this, the next PR will be about combining two other actions in such an intransparent way. (Exaggerating a bit.)

I'd rather make `action` support multiple values, so one could select `http` and `variable`. Implementation-wise that would mean that the `action` param definition would need a `.m()` method call. The action composable would need to be adjusted to handle a `,` concatenated string of `action`s instead of a single one IIRC the implementation of multiple correctly.

WDYT?

svnsssd:
@florian-h05 
> I'd rather make `action` support multiple values, so one could select `http` and `variable`. Implementation-wise that would mean that the `action` param definition would need a `.m()` method call. The action composable would need to be adjusted to handle a `,` concatenated string of `action`s instead of a single one IIRC the implementation of multiple correctly.
> 
> WDYT?

Very good suggestion! 
I have changed the code accordingly and tested with following adapted section of the [dahuadoor intercom widget](https://community.openhab.org/t/dahua-door-intercom-widget/169092):
```yaml
              - component: oh-button
                  config:
                    iconF7: phone_fill_arrow_up_right
                    iconSize: "28"
                    iconColor: green
                    visible: =!!props.thingUid
                    action: [http, variable]
                    actionHttpMethod: POST
                    actionUrl: ='/dahuadoor/sip/answer?thing=' + props.thingUid
                    actionVariable: localMicEnabled
                    actionVariableValue: true
```
So it's now very explicit and one can use two actions at a time.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
```diff
@@ -213,13 +227,13 @@ export const actionParams = (paramPrefix?: string, groupName?: string) => {
     )
       .a()
       .v((_value, configuration) => {
-        return ['command', 'toggle', 'options', 'rule', 'http'].indexOf(configuration[paramPrefix + 'action'] as string) >= 0
+        return actionSelected(configuration, 'command', 'toggle', 'options', 'rule', 'http')
       }),
     pt(paramPrefix + 'actionVariable', 'Variable', 'The variable name to set').v((_value, configuration) => {
-      return ['variable'].indexOf(configuration[paramPrefix + 'action'] as string) >= 0
+      return actionSelected(configuration, 'variable', 'http')
```
florian-h05: Why `http` here?

bundles/org.openhab.ui/web/src/components/widgets/useWidgetAction.ts
```diff
@@ -159,268 +168,329 @@ export function useWidgetAction(context: WidgetContext, config: ComputedRef<Widg
...
+              break
+            case Action.http:
+              const actionHttpUrl = actionConfig[`${processedPrefix}actionUrl`]
+              if (!actionHttpUrl) {
+                actionResult = false
+                break
+              }
+              const actionHttpMethod = actionConfig[`${processedPrefix}actionHttpMethod`] || 'GET'
+              const actionHttpBody = actionConfig[`${processedPrefix}actionHttpBody`]
+              const actionHttpVariable = actionConfig[`${processedPrefix}actionVariable`]
```
florian-h05: What do we need the action http variable code for? I will not accept that code.
It should be enough to select both `http` and `variable` actions.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetAction.ts
```diff
@@ -53,6 +53,16 @@ function showActionFeedback(prefix: string, actionConfig: ActionConfig, text?: s
   }
 }
 
+const normalizeActionList = (value: unknown): Action[] => {
```
florian-h05: Please import the function from actions.ts and use it with a type assignment:

```js
    const actionList = normalizeActionList(actionValue) as Action[]
```

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
```diff
@@ -4,6 +4,21 @@ import type { ConfigDescriptionParameterGroup } from '@/api'
 import { po, pt, pi, pb } from './helpers.ts'
 import { aggregationTypeOptions } from './chart/options.ts'
 
+const normalizeActionList = (value: unknown): string[] => {
```
florian-h05: `unknown` is generally considered bad practise:

```suggestion
export const normalizeActionList = (value: undefined | string | string[]): string[] => {
```

When calling it below, assume a type there:

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
```diff
@@ -4,6 +4,21 @@ import type { ConfigDescriptionParameterGroup } from '@/api'
...
+  const values = Array.isArray(value) ? value : [value]
+  const actions = values.flatMap((entry) => {
+    if (typeof entry === 'string') return entry.split(',')
+    return [String(entry)]
+  })
+  return actions.map((action) => action.trim()).filter(Boolean)
+}
+
+const hasAction = (configuration: Record<string, unknown>, actionKey: string, ...actions: string[]): boolean => {
+  const selectedActions = normalizeActionList(configuration[actionKey])
```
florian-h05:
```suggestion
  const selectedActions = normalizeActionList(configuration[actionKey] as string | string[] | undefined)
```

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
```diff
@@ -4,6 +4,21 @@ import type { ConfigDescriptionParameterGroup } from '@/api'
 import { po, pt, pi, pb } from './helpers.ts'
 import { aggregationTypeOptions } from './chart/options.ts'
 
+const normalizeActionList = (value: unknown): string[] => {
+  if (value === undefined || value === null) return []
+  const values = Array.isArray(value) ? value : [value]
+  const actions = values.flatMap((entry) => {
+    if (typeof entry === 'string') return entry.split(',')
```
florian-h05: With the more strict type definition above, you can remove the if and always `return entry.split...`

florian-h05 (changes requested):
Thanks for the next iteration!

florian-h05:
Can you please modify `generate-widget-component-ts.config.js`, we need to fix the array type in the generated type definitions:

```js
export default {
    defaultExportAs: 'enum',
    '*': { // global modifier applied to all components
        modifier: (e) => e
          .replace('action?: Action', 'action?: Action | Action[]') // Fix array type
          .replace('taphold_action?: Taphold_action', 'taphold_action?: Taphold_action | Taphold_action[]') // Fix array type
...
```

Afterwards, run `npm run generate-widget-component-ts`

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
svnsssd: Fixed. Should not be there. Leftover from previous code

bundles/org.openhab.ui/web/src/components/widgets/useWidgetAction.ts
svnsssd: Yes. Removed.

bundles/org.openhab.ui/web/src/components/widgets/useWidgetAction.ts
svnsssd: Imported normalizeActionList from actions.ts and cast the result to Action[].

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
svnsssd: Changed normalizeActionList signature to string | string[] | undefined.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
svnsssd: Changed as suggested.

bundles/org.openhab.ui/web/src/assets/definitions/widgets/actions.ts
svnsssd: Simplified normalizeActionList to always split entries and trim.

svnsssd:
@florian-h05 
Thanks for your review! Implemented your suggestions!

florian-h05 (approved):
LGTM, thanks!

PR #4112, opened 2026-04-13
useStatesStore: Reactivity fixes

florian-h05:
* Removed Side-Effects from Proxy Getter: Previously, when an Item state was first requested, the Proxy's get handler would immediately add a placeholder state to the itemStates Map. In Vue 3, mutating reactive state (the Map) during the evaluation of a computed property can cause the dependency tracker to ignore that access. Modifies ensureItemTracking to return a placeholder without mutating the Map if the item is not yet tracked. Ensures Vue correctly records the dependency on that item.
* Reactive Proxy: Changes `trackedItems` from a ref of a Proxy to a reactive Proxy. Improves integration with Vue's reactivity system, making it more reliable for dependency tracking when accessed in expressions.

This fixes an issue where expressions in widget configuration were incorrectly evaluated due to missing Item states.


jsjames (approved):
LGTM

PR #4113, opened 2026-04-13
Charts: Remove default for `barBorderRadius`

florian-h05:
Fixes an issue where the default value changed the appearance of existing charts that set a border radius array.


jsjames (approved):
LGTM

PR #4115, opened 2026-04-17
Add readonly mode for file-based pages and widgets

jimtng:
UI implementation of https://github.com/openhab/openhab-core/pull/5493

Note: this PR changes the YAML syntax in the Code tab:
- remove special keys for the slots. Instead, the slots are displayed in the YAML as is. 
- always show the `component:` key.

This is so that the yaml code is compatible with the file-based syntax, making it copy-pasteable to/from file.

image
image


mherwege:
> Note: this PR changes the YAML syntax in the Code tab:
> 
> * remove special keys for the slots. Instead, the slots are displayed in the YAML as is.
> * always show the `component:` key.

Would that render some of the examples on the forum disfunctional? How big would the impact be?

jimtng:
> > Note: this PR changes the YAML syntax in the Code tab:
> > 
> > * remove special keys for the slots. Instead, the slots are displayed in the YAML as is.
> > * always show the `component:` key.
> 
> Would that render some of the examples on the forum disfunctional? How big would the impact be?

Good point. Maybe I'll also accept the old yaml structure (the special keys for the slots) as well as the new structure when parsing but use the new syntax when "generating" the yaml.

jimtng:
> > > Note: this PR changes the YAML syntax in the Code tab:
> > > 
> > > * remove special keys for the slots. Instead, the slots are displayed in the YAML as is.
> > > * always show the `component:` key.
> > 
> > 
> > Would that render some of the examples on the forum disfunctional? How big would the impact be?
> 
> Good point. Maybe I'll also accept the old yaml structure (the special keys for the slots) as well as the new structure when parsing but use the new syntax when "generating" the yaml.

Done.

jimtng:
This has now been rebased on top of #4144 so that needs to be merged first, then I'll rebase this PR against main.

jimtng:
I've made the yaml code to look like the file-format syntax, e.g.

```yaml
version: 1
pages:
  pageuid:
    the_rest_here
```

Unfortunately this adds two extra levels of indentation, but it's consistent with things / items, etc.
If this is super annoying to people, we can add a toggle to hide them - but it should be added across all file-format code (items, things, etc too)

It should also accept the old syntax without this so old yaml can still be copy-pasted into the UI.

mherwege:
> I've made the yaml code to look like the file-format syntax, e.g.
> 
> ```yaml
> version: 1
> pages:
>   pageuid:
>     the_rest_here
> ```
> 
> Unfortunately this adds two extra levels of indentation, but it's consistent with things / items, etc. If this is super annoying to people, we can add a toggle to hide them - but it should be added across all file-format code (items, things, etc too)
> 
> It should also accept the old syntax without this so old yaml can still be copy-pasted into the UI.

Does this mean you have also implemented the multi-select and copy function on the pages list? Because that would allow to put mutiple objects into a single file.

jimtng:
> Does this mean you have also implemented the multi-select and copy function on the pages list? Because that would allow to put mutiple objects into a single file.

Not yet. I can work on that next.

jimtng:
> > Does this mean you have also implemented the multi-select and copy function on the pages list? Because that would allow to put mutiple objects into a single file.
> 
> Not yet. I can work on that next.

Feature added.

mherwege:
The remove and copy buttons have moved to the right in your screenshots. They are centered in all other places I believe (and the remove button was centered before).

jimtng:
> The remove and copy buttons have moved to the right in your screenshots. They are centered in all other places I believe (and the remove button was centered before).

They _are_ centered there too. The screenshot was partial, didn't take it all the way to the full width of the list control / screen.

jimtng:
Updated screenshots - this time the whole width of the list items

image

---

image

jimtng:
@florian-h05 would you mind if I squash all the commits? It would make rebasing easier.

florian-h05:
I don't think I need to review by individual commits here, so feel free to squash 👍

jimtng:
> I don't think I need to review by individual commits here, so feel free to squash 👍

Great! It rebased nicely!

jimtng:
The "Awaiting other PR" label can be removed now.

bundles/org.openhab.ui/web/src/pages/yaml-file-format.ts
```diff
@@ -0,0 +1,110 @@
...
+        const { uid, editable, ...objWithoutUidEditable } = yamlObject
+        acc[uid] = objWithoutUidEditable
+        return acc
+      },
+      {} as Record<string, unknown>
+    )
+  }
+
+  return YAML.stringify(fileObject)
+}
```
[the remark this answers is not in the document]
jimtng: great idea. it was originally set I think

bundles/org.openhab.ui/web/src/pages/yaml-file-format.ts
```diff
@@ -0,0 +1,110 @@
...
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+function createYamlStructureError(yamlElement: string, uid: string | null): Error {
+  const singularElement = yamlElement.endsWith('s') ? yamlElement.slice(0, -1) : yamlElement
+  const message = uid
+    ? `The YAML must contain a single ${singularElement} with the uid "${uid}" under the "${yamlElement}" key.`
+    : `The YAML must contain a single ${singularElement} under the "${yamlElement}" key with a "uid" property.`
+  return new Error(message)
```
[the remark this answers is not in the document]
jimtng: At the moment only a single entry is indeed supported, even though toFileYamlSyntax supports outputting multiple objects.

When we add the feature for importing multiple objects, this will have to be refactored of course, but we're not there yet.

bundles/org.openhab.ui/web/src/pages/yaml-file-format.ts
```diff
@@ -0,0 +1,110 @@
...
+ * @returns The JavaScript object corresponding to the specified `yamlElement` and `uid`.
+ * @throws An error if the YAML structure does not match the expected format or if the specified `yamlElement` and `uid` are not found.
+ */
+
+export function fromFileYAMLSyntax(yamlElement: string, yamlString: string, uid: string | null): unknown {
+  const obj = YAML.parse(yamlString) as unknown
+
+  if (!isRecord(obj)) {
+    throw createYamlStructureError(yamlElement, uid)
+  }
```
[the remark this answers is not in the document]
jimtng: added prettyerrors

bundles/org.openhab.ui/web/src/components/util/not-editable-notice.vue
```diff
@@ -0,0 +1,19 @@
...
+  </f7-block>
+</template>
+
+<script>
+export default {
+  props: {
+    subject: {
+      type: String,
+      default: 'page'
+    }
```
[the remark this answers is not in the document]
jimtng: will leave this as is for now.

bundles/org.openhab.ui/web/src/components/util/not-editable-notice.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <f7-block>
+    <f7-block-footer class="no-margin">
+      <f7-icon f7="lock_fill" size="12" color="gray" />&nbsp;Note: this {{ subject }} is not editable because it has been provisioned from a
+      file.
```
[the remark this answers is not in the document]
jimtng: Changed to `This`. The line break was enforced by our formatter

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-grid-layout.vue
```diff
@@ -43,8 +43,8 @@
 
     <grid-layout
       ref="vueGridLayout"
-      :is-draggable="!!context.editmode"
-      :is-resizable="!!context.editmode"
+      :is-draggable="!!context.editmode && context.isEditable"
+      :is-resizable="!!context.editmode && context.isEditable"
       v-model:layout="layout"
```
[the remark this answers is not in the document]
jimtng: I believe they are as they should be. Even when isEditable == false, we still want to show the "design time" look vs "run mode".

bundles/org.openhab.ui/web/src/pages/yaml-file-format.ts
```diff
@@ -0,0 +1,110 @@
...
+  // Allow ingesting old syntax with uid at root level
+  if ('config' in obj || 'slots' in obj) {
+    return obj
+  }
+
+  const wrappedObj = obj[yamlElement]
+  if (uid === null) {
+    // create mode - extract uid from YAML
+    if (!isRecord(wrappedObj)) {
+      throw createYamlStructureError(yamlElement, uid)
```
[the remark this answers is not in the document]
jimtng: it is not critical do enforce single items here. This method is only used for single entity editors, so it's expected to just parse one uid / element.

jimtng:
It seems that copilot is running out of ideas. I don't have any more changes to push at this point.

bundles/org.openhab.ui/web/src/components/pagedesigner/page-settings.vue
```diff
@@ -8,14 +8,14 @@
...
-        :clear-button="createMode"
+        :clear-button="createMode && !readOnly"
         :info="createMode ? 'Required. Note: cannot be changed after the creation' : ''"
         input-id="input"
         required
         validate
         pattern="[A-Za-z0-9_]+"
         error-message="Required. A-Z,a-z,0-9,_ only"
-        :disabled="!createMode ? true : null">
+        :disabled="readOnly || (!createMode ? true : null)">
```
florian-h05:
```suggestion
        :disabled="readOnly || !createMode">
```

bundles/org.openhab.ui/web/src/components/pagedesigner/page-settings.vue
```diff
@@ -34,10 +34,16 @@
         @input="page.config.label = $event.target.value"
         required
         validate
-        clear-button />
+        :disabled="readOnly ? true : null"
```
florian-h05: Since this is a f7 component we don't need to unset props:

```suggestion
        :disabled="readOnly"
```

bundles/org.openhab.ui/web/src/components/pagedesigner/page-settings.vue
```diff
@@ -34,10 +34,16 @@
...
+        :clear-button="!readOnly" />
       <f7-list-item v-if="page.uid !== 'overview'" accordion-item title="Sidebar &amp; Visibility">
         <f7-accordion-content>
-          <f7-list-item ref="pageVisibility" title="Visible only to" smart-select :smart-select-params="{ openIn: 'popover' }">
+          <f7-list-item
+            ref="pageVisibility"
+            title="Visible only to"
+            smart-select
+            :smart-select-params="{ openIn: 'popover' }"
+            :disabled="readOnly ? true : null">
```
florian-h05:
```suggestion
            :disabled="readOnly">
```

bundles/org.openhab.ui/web/src/components/widgets/types.d.ts
```diff
@@ -38,6 +38,10 @@ export interface WidgetContext {
    */
   ctxVars?: ContextVarObj
   editmode?: EditMode
+  /**
+   * Whether the page/widget is editable. When false, edit controls are hidden.
+   */
+  isEditable?: boolean
```
florian-h05: Have you thought about only defining `context.editmode` if the page is actually editable??
This would allow us to not require and additional boolean prop, we'll likely miss adding it in some places.

jimtng: no I haven't considered it. It will take me some time to try to figure it out.

jimtng: I'm exploring this idea now.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -289,13 +281,6 @@
...
       <f7-tab v-if="ready && hasSource" id="source" :tab-active="currentTab === 'source'">
-        <f7-icon
-          f7="lock"
-          class="float-right margin"
-          style="opacity: 0.5; z-index: 4000; user-select: none"
-          size="50"
-          color="gray"
-          tooltip="Source code is not editable" />
         <editor v-if="currentTab === 'source'" class="rule-source-viewer" :mode="sourceType" :value="source" :readOnly="true" />
       </f7-tab>
```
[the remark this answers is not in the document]
jimtng: "Source code is not editable" doesn't explain the reason why it's not editable, so I think the default msg is better.

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-item.vue
```diff
@@ -30,37 +30,41 @@
...
-              text="Bring to Front" />
-            <f7-menu-dropdown-item @click="context.editmode.moveWidgetDown(context.component, context.parent)" href="#" text="Move Up" />
-            <f7-menu-dropdown-item @click="context.editmode.moveWidgetUp(context.component, context.parent)" href="#" text="Move Down" />
-            <f7-menu-dropdown-item
-              @click="context.editmode.sendWidgetToBack(context.component, context.parent)"
-              href="#"
-              text="Send to Back" />
-            <f7-menu-dropdown-item divider />
-            <f7-menu-dropdown-item @click="context.editmode.removeWidget(context.component, context.parent)" href="#" text="Remove Item" />
+              :text="context.editmode.isEditable ? 'Edit YAML' : 'View YAML'" />
```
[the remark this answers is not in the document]
jimtng: ah good spotting! I'll change them all to a single "xxxx Settings" so they remain consistent and predictable for editable/non editable version.

bundles/org.openhab.ui/web/src/components/widgets/types.d.ts
jimtng: I've moved isEditable into the editmode. We still need a separate isEditable flag to show read-only editmode controls in design mode.

bundles/org.openhab.ui/web/src/pages/settings/pages/unknown/unknown-edit.vue
```diff
@@ -0,0 +1,102 @@
...
+    return {
+      page: null,
+      suggestions: []
+    }
+  },
+  methods: {
+    onPageAfterIn() {
+      const uid = this.uid || this.f7route?.params?.uid
+      if (!uid) return
+      this.$oh.api
```
florian-h05: Pages are available from `useComponentsStore`. Please use them from there and don't load them here.

florian-h05 (approved):
Overall looks quite good, many thanks! I will shortly push a number of commits fixing various minor bugs/regressions.

jimtng:
Thanks for the help, @florian-h05!

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -1,13 +1,4 @@
 <template>
-  <f7-icon
-    v-if="readOnly"
-    f7="lock"
-    class="float-right margin"
-    style="opacity: 0.5; z-index: 4000; user-select: none"
-    size="50"
-    color="gray"
-    :tooltip="readOnlyMsg" />
-
```
Nadahar, line 10: Why was the read-only icon removed? It's certainly needed for rules...

jsjames: it was refactored to be included in the script-editor

bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
```diff
@@ -1,5 +1,13 @@
 <template>
-  <codemirror ref="cm" class="code-editor-fit" :model-value="value" :extensions="extensions" @ready="onCmReady" @change="onCmCodeChange" />
+  <div class="code-editor-fit">
+    <f7-icon
+      v-if="readOnly"
+      f7="lock"
+      size="50"
+      class="code-readonly-icon"
+      :tooltip="readOnlyMsg || 'Read-only: provisioned from a file'" />
```
Nadahar, line 8: I wish we would stop assuming that read-only = file. It's misleading. *Only* those objects that are provided by the one managed provider are writeable, everything else is read-only. That applies if the object is provided by e.g an add-on.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: > it was refactored to be included in the script-editor

How do you pass on `readOnlyMsg` now then? Script-editor is a child of code-editor, so you have no way to pass anything directly to script-editor when you used code-editor..?

bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
jimtng: Are there any real-world examples of such read-only objects that aren't from a file?
What's your suggestion instead of this message?

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
jsjames: I agree - that looks like it was removed in this PR and if needed should be added back in.

bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
Nadahar: I don't know if there are "real-world examples" or not. That's not the point, the point is that associating it with files is wrong.

As to what it should say instead, if you want to have a "generic" text, it should probably just say "Read-only: Cannot be edited in the UI" or something like that, but it's more useful if it can reference the actual object type, e.g. rule. Passing it on seems to not be an option anymore though.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: "Needed" is relative - but I've done a lot of work on rules, and suddenly things are reorganized and removed, which is frustrating. I'm pretty sure that my "shared context" icon will be broken now for example, and I will need to do that again.

But, as far as I know, the `readOnlyMsg` was used - so by removing it, that will no longer work. And, since the generic text that has replaced it is wrong on top of everything, I don't think that's a good move.

Nadahar: I've already found one place where it's botched now - on the rules "Source" tab, the tooltip used to be:
```
tooltip="Source code is not editable"
```

Now, it will show the generic text, which is completely misleading here, instead. The "source" is never editable, because it shows the "source code" that *created* the rule. It's not actually a part of the rule, it's the origin of the rule. As such, it doesn't matter if it was "provisioned from a file" or not.

jimtng: > The "source" is never editable, because it shows the "source code" that _created_ the rule. It's not actually a part of the rule, it's the origin of the rule. As such, it doesn't matter if it was "provisioned from a file" or not.

What's the scenario where the source is not the same as the rule / file?

jimtng: I see - you've done this for DSL by extrapolating / translating the rule into YAML syntax :(

jimtng: Although the representation is inaccurate, it's still provisioned from a file. In any case, hopefully #4118 restores the ability to set the readonlymsg for you

Nadahar:
This is how "shared context" looks now:

bilde

(The two icons used to be next to each other and without pushing the text down)

By refactoring it into `script-editor`, it has become less flexible for customization, and I guess I will have to make the "shared context" functionality also part of the `script-editor` now, even though it is only relevant for rules. I'm not sure doing this reorganization s an improvement, but I'll find a way to make it work.

jimtng:
> This is how "shared context" looks now:
> 
> bilde
> (The two icons used to be next to each other and without pushing the text down)

Can this be solved with css?



> I guess I will have to make the "shared context" functionality also part of the `script-editor` now, even though it is only relevant for rules.

That doesn't sound like a good idea.

Nadahar:
> Can this be solved with css?

I don't know, the way it were, they were both "floating right". I'm no "CSS expert" and can't say for sure what can and can't be done, but they used to be "on the same level" in the HTML, now they aren't. How can you make them "respond to each other" so that if only one of them are present, they are rightmost, and if both are there, they are beside each other, without having them as "equals"?

> > I guess I will have to make the "shared context" functionality also part of the `script-editor` now, even though it is only relevant for rules.
> 
> That doesn't sound like a good idea.

No, it doesn't, but I don't quite see any other way with this refactoring.

IMO, it was better when this was handled "locally" in the components, because you have the necessary flexibility without having to affect other pages.

Now I have an additional problem, because now the padlock no longer "floats", but is positioned absolutely, which won't work when they are both there. Was there any particular reason for changing it to a fixed position? It worked very nicely as it was when they were floating.

PR #4124, opened 2026-04-22
Refactor dirty-mixin to TS composable

jsjames:
- dirty-mixin is now a composable.
- Separated `switchTabs` to a `useTabs` composable, so `useTabs` composable functionality can be expanded in the future.


florian-h05 (approved):

Nadahar:
I have no idea what a mixin is or what a composable is. It doesn't matter so much what you call it, I understand code, but I must understand the logic. After rebasing my work branch, this PR has changed `rule-edit.vue` so that the logic I had made no longer works.

There are somewhat complex handling both for `switchTabs()` and "dirty", this logic used to reside in methods in `rule-edit.vue`, but that's not possible now. Where should this logic be put?

florian-h05:
Where can I find your branch? I’ll have a look.

Normally it should be enough to call useDirty() in the setup function and export the return values from useDirty() as done in the changes of this PR.

Nadahar:
> Where can I find your branch? I’ll have a look.
> 
> Normally it should be enough to call useDirty() in the setup function and export the return values from useDirty() as done in the changes of this PR.

The branch is in *very early stages*, full of TODOs and commented code/testing. But, it is here: https://github.com/Nadahar/openhab-webui/tree/yaml-rules

I think that the tab handling will pose the biggest challenge, more logic have to be added there. Tabs will have to be hidden/revealed dynamically and conversion to/from the file format (which it already kind of is, but with lots of conditional behavior).

florian-h05:
Wrt to useDirty: Just import it in the setup function (as done in this PR), the new composable allows to write to the dirty flag as before.

Wrt to useTabs: I'd recommend exporting it under a different name, then calling it in your new switchTab function. In your setup function:

`return { ...., switchTabInternal: switchTab, ... }`

Nadahar:
> Wrt to useTabs: I'd recommend exporting it under a different name, then calling it in your new switchTab function. In your setup function:
> 
> `return { ...., switchTabInternal: switchTab, ... }`

I understand that conflicts can be avoided by using different names, what I don't understand is the purpose of this.

From the little I've seen of mixins, it seems like they are "shared code" that come into effect as long as you don't overwrite them internally. So, they are confusing but flexible, you can still do more complex tasks, they just function as a sort of "default" that's there if you do nothing.

The composables seem very different in that they look like they are completely inflexible. As far as I can tell, you can't override parts of them, so you must either use the "common code" without modification, or do "everything" yourself. It feels like I must be missing something, because such rigidity can't possibly be considered an improvement? As of now, I've just commented out `useTabs` entirely, and it seems to work, so I'm wondering what it's for at all.

If I rename one of them, I will have to let the calls go to "my" method (because the composable isn't up to the complexity here). If so, when will it ever be called at all - what's it for?

jsjames:
Regarding useTabs, that is just a convenience composable and may not fit all cases (in fact, item-edit and thing-details don't use it at all and has its own switchTab method.  This was always the case, useTabs just encapsulates the switchTab method that was originally in the dirty-mixin and creates the ref currentTab so the component doesn't have to create.

> From the little I've seen of mixins, it seems like they are "shared code" that come into effect as long as you don't overwrite them internally. So, they are confusing but flexible, you can still do more complex tasks, they just function as a sort of "default" that's there if you do nothing.

Regarding the "file-definition-mixin", we shouldn't be writing new mixin's - those can't be used when we convert components to composition api. There are many benefits to using composables, but if you are just using javascript/typescript methods, then don't make a composable and just import separate .js/.ts methods.  The big question to ask, does it need to interact with vue (reactivity, component lifecycle, ...). In a brief look at the file-definition-mixin, I think you can easily just make that a typescript module and import the methods.  Also, typescript has an enum as well which can be used for your ObjectType.

florian-h05:
> The composables seem very different in that they look like they are completely inflexible. As far as I can tell, you can't override parts of them, so you must either use the "common code" without modification, or do "everything" yourself. It feels like I must be missing something, because such rigidity can't possibly be considered an improvement?

Composables provide strong encapsulation, which in 99% is a good thing IMO as it avoids having a ton of not-obvious as with mixins, where a mixin could depend on state or methods of its "user" and its user could completely overwrite parts of the mixin without noticing it. 

Together with TS composables improve the overall quality of the code.
But sure, there are rare cases where the "merging" of the mixin into the context of your component is beneficial as it allows for modifications.

Nadahar:
> The big question to ask, does it need to interact with vue (reactivity, component lifecycle, ...). In a brief look at the file-definition-mixin, I think you can easily just make that a typescript module and import the methods. Also, typescript has an enum as well which can be used for your ObjectType.

Just so that's said, *I* haven't made the file-definition-mixin. I don't know enough about Vue, TypeScript, JavaScript, or the latest trends to have any opinion about such "design decisions". My focus is the logic, I'm trying to provide the functionality I'm after, and I try to find ways within/that are similar to what I already find. That's all there is to it really.

But, I do quite a few things before I see that it's done other places, because I don't want to break things and I don't have the overview of the implications. So, all I really needed to know here, is that I *don't have to use* `useTabs`. Nothing will magically break if I don't. The kind of logic that is needed to handle the tabs here isn't easily generalizable anyway.

> Composables provide strong encapsulation, which in 99% is a good thing IMO as it avoids having a ton of not-obvious as with mixins, where a mixin could depend on state or methods of its "user" and its user could completely overwrite parts of the mixin without noticing it.
> 
> Together with TS composables improve the overall quality of the code. But sure, there are rare cases where the "merging" of the mixin into the context of your component is beneficial as it allows for modifications.

As I said, I know almost nothing about the different concepts or fashions, but I see that mixins are "out" and composables "in". I did read a little bit in some discussions, and my impression is that they shouldn't really be thought of as "either or", but as different concepts. You can't just "convert" mixins to composables, you must "redesign the logic" to fit what composables do. And yes, I see that you can achieve some flexibility by sending functions/callbacks, but things quickly become very complex when you do that, and the whole "reuse benefit" might be lost.

So, what I get out of this is: Composables are more rigid, they are thus "friendlier" with TypeScript. They can't really do all that mixins can, but "we're willing to pay the price" because it's considered "backwards" to use mixins, one day they might become "deprecated" and people obsessed with trends will complain about their use. Composables are also clearly more aligned with the "functional crowd", and they have a lot of impact on fashion these days. On top of that, mixins are a bit messy and hard to keep track of if used "irresponsibly".

To me, it is a symptom of stubbornly refusing to embrace proper OO by JavaScript. By using things like mixins, you get flexibility that is akin to multiple inheritance, but you don't get the headaches because you don't formalize the relationships. Which is why you get entirely different headaches; it becomes "unpredictable" and bug-prone. What you really need is a way to generalize certain aspects of a system, and use them where needed, while retaining the flexibility to tweak their behavior when the "standard" doesn't fit. Inheritance largely provides that, but is off the table because of past sins and inertia. So, the search goes on for all kind of "substitute concepts" that might provide something similarish.

In the meanwhile, enormous amounts of effort is wasted on refactoring/redesigning code following whatever idea has the most traction at a given time.

I think I'll try to focus on making the logic in a way I can figure out without having to learn about "web design trends", and then it can be refactored to fit into whatever paradigm is "in" later.

Nadahar:
The file-definition-mixin looks more like a "shared library" of functions than most other mixins I've seen, so it should probably just become a "library", "module", whatever it's called, what I would call a class in OO, that can then be imported and invoked where needed. I don't think it's a good idea that I try to do that, given how much I don't know about all these things, so I'll just work with what is there. It shouldn't be hard to modify what I've done to get access to these methods "in another way" when the mixin is converted to something else.

PR #4133, opened 2026-04-25
Fix rule module reordering bug

### LINKED ISSUE #4132: Recording rule steps get lost after saving and leaving the page
## The problem
When reordering rule steps they seem to get lose after saving, leaving the page and reopening the rule.

Image

## Expected behavior

That the rule steps keep the new order.

## Steps to reproduce

Reorder any multi step rule, press save, leave page, return to rule, confirm steps did not keep the new order.

## Your environment
OpenHAB 5.2.0

Nadahar:
Reordering rule modules doesn't work because the reordered modules aren't stored in the correct location. This fixes that.

Fixes #4132 

It stems from #3350, perhaps other reordering refactorings should be checked for the same issue.


jsjames (approved):
Thanks - LGTM

Nadahar:
Perhaps a backport to 5.1.x should be considered?

jsjames:
> Perhaps a backport to 5.1.x should be considered?

done

PR #4134, opened 2026-04-26
Widget edit: Don't cache expressions in edit mode & throttle rerender on input

jsjames:
Fixes #4092

Widget-edit context needs to have editmode=true
also debounced editor input


jsjames:
FYI - I'm going to see if throttling would be a better experience vs. debounce, so I converted to draft

jsjames:
I moved to use throttling vs. debounce as it gives more realtime updates.

florian-h05 (approved):

PR #4139, opened 2026-04-27
Update OpenAPI REST client

florian-h05:


florian-h05:
@Nadahar FYI

Nadahar:
@florian-h05 Maybe it would be a good idea to modify `generate-rest-client.mjs` so that it could accept an option URL argument. If the URL was specified, it would first download and format `openhab-api-spec.json`, and then do the generation from that file. If run without the URL argument, it would just process `openhab-api-spec.json` like it does now.

The reason I'm suggesting that is that the formatting is hard to get consistent. I've tried to format the JSON both with Notepad++ and VSCode now, and none of them are an exact match with whatever you have used. Some fields are sorted differently etc. It leads to unnecessary manual work to make the diff only what it should be. If the script did the formatting of the JSON, it should be able to do the exact same formatting each time, so that we only got "real changes" in the diff without the manual work I'm currently doing.

jsjames:
> @florian-h05 Maybe it would be a good idea to modify `generate-rest-client.mjs` so that it could accept an option URL argument. If the URL was specified, it would first download and format `openhab-api-spec.json`, and then do the generation from that file. If run without the URL argument, it would just process `openhab-api-spec.json` like it does now.
> 
> The reason I'm suggesting that is that the formatting is hard to get consistent. I've tried to format the JSON both with Notepad++ and VSCode now, and none of them are an exact match with whatever you have used. Some fields are sorted differently etc. It leads to unnecessary manual work to make the diff only what it should be. If the script did the formatting of the JSON, it should be able to do the exact same formatting each time, so that we only got "real changes" in the diff without the manual work I'm currently doing.

That was originally the intent (to download from the demo server) and that code is currently commented out in the script. However, there continue to issues with the openapi spec, so keeping a static copy tied to the current checkin of code keep things in sync and not tied to a moving openhab-core.

Nadahar:
> That was originally the intent (to download from the demo server) and that code is currently commented out in the script. However, there continue to issues with the openapi spec, so keeping a static copy tied to the current checkin of code keep things in sync and not tied to a moving openhab-core.

Yeah, I didn't mean to download it automatically anyway, because that would work for what I'm doing. I'm developing one branch in core and one in MainUI in parallel, so I need to generate the spec from my development branch that has created new REST endpoints.

What I mean was just that it was possible to get the script to download it from a specified, optional, URL, because it would allow the JSON formatter to be consistent. As it is now, the exact formatting of the JSON depends on the tool you use, which leads to a lot of "changes" that aren't actually changes.

Nadahar:
I will need to do this repeatedly as I make changes to the REST endpoints when I discover something that is missing or "unfortunate" when I try to implement the UI side. For every such "adjustment", I must regenerate, and that leads to quite a lot of manual "merging" of the document.

jsjames:
I haven't tested in a while, but if you look at the build-rest-client script, it allows you to specify a url and it will fetch the spec.  You could use that for development - but we wouldn't want to check that change in.

```

// const url = `https://api.swaggerhub.com/apis/wborn/openhab-rest-api/8/swagger.json`
// const url = `https://demo.openhab.org/rest/spec`
const url = null
const file = './build/openhab-api-spec.json'

const outpath = './src/api'

let openApi = null

if (url) {
  console.log('Fetching OpenAPI spec from URL:', url)
  try {
    openApi = await fetch(url).then((res) => res.json())
  } catch (error) {
    console.error('Error fetching OpenAPI spec:', error)
  }
}

**Comment by Nadahar** [comment]
> I haven't tested in a while, but if you look at the build-rest-client script, it allows you to specify a url and it will fetch the spec.

I have tried that, but it doesn't work that well. I need the file to see what has actually changed. The reason is that not all endpoints are available in my dev installation where I generate the spec, because some of them rely on Karaf (which I don't use). There are also some bundles I simply don't run because I don't need them. So, some things will be missing from what I generate. Doing that merge manually is fine, I see when there are whole parts missing and include those parts. But, when the sorting of the fields are slightly different as well, the whole procedure becomes much more work.

**Comment by florian-h05** [comment]
We need oxfmt (our formatter) to format the JSON so we have a consistent formatter.
The demo server cannot be used as a source because we need OpenAPI spec from a snapshot, not the last release.

**Comment by Nadahar** [comment]
> We need oxfmt (our formatter) to format the JSON so we have a consistent formatter.

I'm not sure if we're talking past each other. How do you run oxfmt is basically what I think I'm asking. What is the procedure you use to fetch the unformatted data from `/rest/spec` and format it into `openhab-api-spec.json`? I was hoping that this process could be automated so that it was consistent, but lacking that, it would help me a lot to know how to run this formatter after I paste the content into the file.

**Comment by Nadahar** [comment]
Just to test what I mean, I had Copilot modify the script to use oxfmt like I'm suggesting. However, the output still don't match. It doesn't sort identically to how it's sorted in the committed file, in addition, it doesn't encode unicode characters as hex, like the current committed file does. Why is it so hard to generate the same format?

```diff
 .../web/build/generate-rest-client.mjs             | 71 +++++++++++++++++-----
 1 file changed, 55 insertions(+), 16 deletions(-)

diff --git a/bundles/org.openhab.ui/web/build/generate-rest-client.mjs b/bundles/org.openhab.ui/web/build/generate-rest-client.mjs
index 5d12bcf0..40c4e01c 100755
--- a/bundles/org.openhab.ui/web/build/generate-rest-client.mjs
+++ b/bundles/org.openhab.ui/web/build/generate-rest-client.mjs
@@ -2,34 +2,71 @@
 
 import fetch from 'node-fetch'
 import fs from 'node:fs'
+import path from 'node:path'
+import { spawnSync } from 'node:child_process'
 import { createClient } from '@hey-api/openapi-ts'
 
-// const url = `https://api.swaggerhub.com/apis/wborn/openhab-rest-api/8/swagger.json`
-// const url = `https://demo.openhab.org/rest/spec`
-const url = null
-const file = './build/openhab-api-spec.json'
+// Usage: node build/generate-rest-client.mjs [SPEC_URL]
+// If SPEC_URL is provided, the script will download it, pretty-print
+// and save it to './build/openhab-api-spec.json', then generate from
+// that saved file. If no argument is provided, it will generate from
+// the existing './build/openhab-api-spec.json'.
 
+const argUrl = process.argv[2] ?? null
+const file = './build/openhab-api-spec.json'
 const outpath = './src/api'
 
 let openApi = null
 
-if (url) {
-  console.log('Fetching OpenAPI spec from URL:', url)
+if (argUrl) {
+  console.log('Downloading OpenAPI spec from URL:', argUrl)
   try {
-    openApi = await fetch(url).then((res) => res.json())
+    const res = await fetch(argUrl)
+    if (!res.ok) {
+      throw new Error(`Failed to fetch spec: ${res.status} ${res.statusText}`)
+    }
+
+    const text = await res.text()
+    let parsed
+    try {
+      parsed = JSON.parse(text)
+    } catch (err) {
+      throw new Error('Downloaded spec is not valid JSON')
+    }
+
+    // ensure build directory exists and write pretty JSON (initial write)
+    fs.mkdirSync(path.dirname(file), { recursive: true })
+    fs.writeFileSync(file, JSON.stringify(parsed, null, 2), 'utf8')
+    console.log('Saved OpenAPI spec to file:', file)
+
+    // Try to pretty-print the spec using oxfmt (local install preferred)
+    try {
+      const localOxfmt = path.join(process.cwd(), 'node_modules', '.bin', `oxfmt${process.platform === 'win32' ? '.cmd' : ''}`)
+      if (fs.existsSync(localOxfmt)) {
+        console.log('Formatting spec with local oxfmt...')
+        const r = spawnSync(localOxfmt, ['--write', file], { stdio: 'inherit' })
+        if (r.error || r.status !== 0) console.warn('oxfmt (local) failed with', r.error || `exit ${r.status}`)
+      } else {
+        console.log('Local oxfmt not found, trying npx oxfmt...')
+        const r = spawnSync('npx', ['oxfmt', '--write', file], { stdio: 'inherit' })
+        if (r.error || r.status !== 0) console.warn('npx oxfmt failed with', r.error || `exit ${r.status}`)
+      }
+    } catch (err) {
+      console.warn('Formatting with oxfmt failed:', err?.message || err)
+    }
   } catch (error) {
-    console.error('Error fetching OpenAPI spec:', error)
+    console.error('Error downloading OpenAPI spec:', error)
+    process.exit(1)
   }
 }
 
-if (file) {
-  console.log('Reading OpenAPI spec from file:', file)
-  try {
-    const data = fs.readFileSync(file, 'utf8')
-    openApi = JSON.parse(data)
-  } catch (error) {
-    console.error('Error reading OpenAPI spec from file:', error)
-  }
+console.log('Reading OpenAPI spec from file:', file)
+try {
+  const data = fs.readFileSync(file, 'utf8')
+  openApi = JSON.parse(data)
+} catch (error) {
+  console.error('Error reading OpenAPI spec from file:', error)
+  process.exit(1)
 }
 
 try {
@@ -50,6 +87,8 @@ try {
       }
     ]
   })
+  console.log('✅ Done! Your output is in', outpath)
 } catch (error) {
   console.error('Error creating client:', error)
+  process.exit(1)
 }
```

florian-h05:
> How do you run oxfmt is basically what I think I'm asking.

Best from the command line, but that's a bit tricky. I'll add it to the npm script that regenerates the client.

> What is the procedure you use to fetch the unformatted data from /rest/spec and format it into openhab-api-spec.json?

I accessed it through the browser ... but I agree a script would be much nices.

florian-h05:
There we go: https://github.com/openhab/openhab-webui/pull/4141

Nadahar:
> Best from the command line, but that's a bit tricky. 

"From the command line" probably implies that you run Linux and have a bunch of stuff configured and installed? Anyway, my problem isn't to *get it to run*, but to get it to format the JSON identically to what's there already, so that diff isn't full of noise.

> I'll add it to the npm script that regenerates the client.

Keep in mind that this must be an optional operation, and that you must be able to specify the URL. I'd hate if it overwrote the JSON every time you wanted to regenerate something else. And, to work with branches not merged into core, it's necessary to be able to point to a URL of your choosing, where you run the version whose spec you want to use.

florian-h05:
> but to get it to format the JSON identically to what's there already, so that diff isn't full of noise.

That's not possible since it wasn't formatted with Prettier before ...

> Keep in mind that this must be an optional operation, and that you must be able to specify the URL. I'd hate if it overwrote the JSON every time you wanted to regenerate something else.

I meant to add the formatting to the package script (i.e. npm run generate-rest-client), not the download to a script as in the sense of a script file. Sorry for the confusion.

Nadahar:
> That's not possible since it wasn't formatted with Prettier before ...

I didn't mean "before" as in how it's formatted right now. I meant that it should be formatted consistently with whatever the current `main` version is, so that you can see what has actually changed.

> I meant to add the formatting to the package script (i.e. npm run generate-rest-client), not the download to a script as in the sense of a script file. Sorry for the confusion.

Ok, so you'll format the JSON every time you generate? It will still "work" for me, I'll just have to run the generation twice. First I run it to format the JSON, then I manually go through the Git diff and corrects whatever is off, and then I run the actual generation.

florian-h05:
> Ok, so you'll format the JSON every time you generate?

Exactly, every time before generating.

PR #4162, opened 2026-05-01
useChart & useWidgetAction: Ensure composables are fully reactive to props

florian-h05:
Similar to https://github.com/openhab/openhab-webui/pull/4061.


jsjames:
Did you try to see if toRef worked vs. making it a computed?  Probably not a big deal, but (according to ai) would have slightly less overhead.

florian-h05:
toRef didn’t work when working on #4016. I didn’t try here again, I assume the behaviour is the same.

jsjames (approved):
LGTM

PR #4163, opened 2026-05-03
Charts: Fix inconsistencies with `week`/`weekday` fixed to start on Sunday vs. depending on locale

florian-h05:
Previously, in some places `week` start of a period was forced to Sunday in some places and `weekday` indexing was dynamic based on the locale.
This caused issues with `oh-aggregate-series` `week` charts, where the axis and the data did not match.
This PR fixes this by having `week` start of period always depend on the locale to align with the indexing of `weekday`.

Additionally, it removes the `startOnSunday` config parameter from `oh-aggregate-series` as it now starts the week based on the locale default for `ChartType.week` or starts with Monday with `ChartType.isoWeek`.

If you want a chart based on Monday = start of week: `isoWeek` always starts on Monday, `isoWeekday` uses other indexing and `weekday`.


jsjames (approved):
LGTM

PR #4164, opened 2026-05-03
Upgrade to Vite 8

florian-h05:
## Description

This upgrades from the Vite Rolldown preview to Vite 8, which finally ships Rolldown as default.

See https://vite.dev/blog/announcing-vite8.
For upgrading from Vite 7, see
https://vite.dev/guide/migration#migration-from-v7. Note that to us, only NRV-labeled sections apply.

The most important change for users is the change of the browser targets. The Chrome, Edge & Firefox versions now required were all released about three years ago, Safari 11.1 => 15.4 drops support for iPhone <= 6, iPad Air 1, iPad mini <= 3. I do think that dropping support for those relatively old devices is reasonable.

`vite-plugin-dynamic-import` doesn't support Vite 8 anymore, so all dynamic imports were migrated to `import.meta.glob` which is handled by Vite natively.

### DIFF
--- a/bundles/org.openhab.ui/web/package-lock.json
+++ b/bundles/org.openhab.ui/web/package-lock.json
@@ -147,7 +147,6 @@
         "typescript-eslint": "^8.58.2",
         "vite": "^8.0.10",
         "vite-plugin-compression2": "^2.5.3",
-        "vite-plugin-dynamic-import": "^1.6.0",
         "vite-plugin-pwa": "^1.2.0",
         "vite-plugin-top-level-await": "^1.6.0",
         "vite-plugin-vue-devtools": "^8.1.1",
@@ -8250,11 +8249,6 @@
         "node": ">= 0.4"
       }
     },
-    "node_modules/es-module-lexer": {
-      "version": "1.7.0",
-      "dev": true,
-      "license": "MIT"
-    },
     "node_modules/es-object-atoms": {
       "version": "1.1.1",
       "license": "MIT",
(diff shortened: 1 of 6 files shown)


florian-h05:
@jsjames Do you think we should wait with Vite 8 for openHAB 5.3.0? This way we'd keep support for iPad Air 1st Gen, iPhone <= 6 etc. for openHAB 5.2.0

jsjames:
> @jsjames Do you think we should wait with Vite 8 for openHAB 5.3.0? This way we'd keep support for iPad Air 1st Gen, iPhone <= 6 etc. for openHAB 5.2.0

I don’t see an issue waiting. I also don’t recall where the dynamic import plugin was used.

florian-h05:
> I also don’t recall where the dynamic import plugin was used.

Charts don't load without it, so I suspect there, but haven't checked myself.

jsjames:
> > I also don’t recall where the dynamic import plugin was used.
> 
> Charts don't load without it, so I suspect there, but haven't checked myself.

I investigated a bit - vite does not out of the box handle dynamic imports from node_modules - so it is used to import the locales for both echarts and dayjs.  Not sure if vite8 will support this natively?

florian-h05:
Vite 8 supports `import.meta.glob` natively, so I migrated our dynamic imports of i18n files, Dayjs locales and ECharts locales to this. 
Dev server looks good to me now, just waiting for the mvn build to verify that as well.

florian-h05:
Production build also looks good, this is ready for merging!

jsjames (approved):
LGTM

Nadahar:
This seems to have broken MainUI for me, I just get a blank page and this logged in the console:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'key')
    at App.vue:915:98
```

There needs to be taken into account that the locale might not exist, so that `dayjsLocale` is undefined, like it is for me with a Norweigan system.

Nadahar:
I don't know if it's the "correct" fix, but changing line 915 to this, makes it work for me again:

```js
            const dayjsLoader = dayjsLocalesGlob[`../node_modules/dayjs/esm/locale/${dayjsLocale?.key}.js`]
```

Nadahar:
By the way, the reason it's undefined, is that `locale` is `no-no`. That doesn't match any key, so it fails. The "translation" to `nb` for `no` happens too late, by then, it has already failed to match.

florian-h05:
Thanks for the report. I've created #4252.

PR #4165, opened 2026-05-03
Fix token auth on SSE reconnection

thestr4ng3r:
The headers containing the token for the SSE endpoint were initialized only once with some initial token. On future reconnections, for example if the connection breaks this old token was still used even though it was possibly expired.
In that case, the UI will show an error message and only update states again after manually pressing the reload button.

This became visible for me when having the iOS app in the background for over an hour, which does not kill the app, but close the SSE connection after a while. On re-entereing the app I always saw the "Communication failed" toast of the main ui.

The issue can be reproduced by opening the web ui on an instance with the implicit user role disabled in API security settings, waiting for the initially aquired token to expire and then breaking the /states SSE connection somehow (e.g. disconnecting network for a few seconds).
When the main ui tries to reconnect, it will retry indefinitely with the expired token:

Bildschirmfoto 2026-05-03 um 20 50 57

To not have to wait an entire hour, one can use this patch on rg.openhab.core.io.rest.auth:

```diff
diff --git a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
index d3e78dca7..c34b3d43f 100644
--- a/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
+++ b/bundles/org.openhab.core.io.rest.auth/src/main/java/org/openhab/core/io/rest/auth/internal/TokenResource.java
@@ -98,7 +98,7 @@ public class TokenResource implements RESTResource {
             + "=%s; Domain=%s; Path=/; Max-Age=2147483647; HttpOnly; SameSite=Strict";

     /** The default lifetime of tokens in minutes before they expire */
-    public static final int TOKEN_LIFETIME = 60;
+    public static final int TOKEN_LIFETIME = 1;

     private final UserRegistry userRegistry;
     private final JwtHelper jwtHelper;
```

This patch is for 5.1.x. I tried to reproduce this with the latest development state on main as well, but there the SSE connection did not even work on the first connection for me, so it will need some further investigation there.

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/openhab/sse.ts
+++ b/bundles/org.openhab.ui/web/src/js/openhab/sse.ts
@@ -29,24 +29,27 @@ function newSSEConnection (
 ): KeepaliveEventSource {
   let eventSource: KeepaliveEventSource
   let reconnectSeconds = 1
-  const headers: Record<string, string> = {}
-
-  // Setup headers for authentication
-  const accessToken = getAccessToken()
-  if (accessToken && getRequireToken()) {
-    if (getTokenInCustomHeader()) {
-      headers['X-OPENHAB-TOKEN'] = accessToken
-    } else {
-      headers['Authorization'] = 'Bearer ' + accessToken
-    }
-  }
-  const basicCreds = getBasicCredentials()
-  if (basicCreds) {
-    headers['Authorization'] = 'Basic ' + btoa(basicCreds.id + ':' + basicCreds.password)
-  }
 
   // Core initialization logic
   function initEventSource (): KeepaliveEventSource {
+    const headers: Record<string, string> = {}
+    // Setup headers for authentication.
+    // We make sure to always use the latest token here, otherwise it may
+    // have already expired when initEventSource() is called again after
+    // a connection failure.
+    const accessToken = getAccessToken()
+    if (accessToken && getRequireToken()) {
+      if (getTokenInCustomHeader()) {
+        headers['X-OPENHAB-TOKEN'] = accessToken
+      } else {
+        headers['Authorization'] = 'Bearer ' + accessToken
+      }
+    }
+    const basicCreds = getBasicCredentials()
+    if (basicCreds) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/js/openhab/sse.ts
```diff
@@ -31,22 +31,22 @@ function newSSEConnection (
   let reconnectSeconds = 1
   const headers: Record<string, string> = {}
```
florian-h05: I think you should also move `headers` into `initEventSource` then.

florian-h05 (commented):
Thank you for your contribution!
Your changes make sense and thanks for fixing the issue.
One minor comment:

And can you please add a short code comment explaining that the headers should be initialised inside initEventSource so they are re-initialised when the event source attemps to reconnect?

florian-h05 (approved):

florian-h05:
Just a heads up for future PRs: Please try to avoid force pushing, this resets the review progress, which is bad in large PRs.

florian-h05:
BTW, normally we only develop against main and backport patches.

thestr4ng3r:
Thank you!

> Just a heads up for future PRs: Please try to avoid force pushing, this resets the review progress, which is bad in large PRs.

Sure, will keep this in mind for further PRs.

> BTW, normally we only develop against main and backport patches.

This was my initial approach too, but I was not able to get the SSE connection to work there at all without digging deeper:

> This patch is for 5.1.x. I tried to reproduce this with the latest development state on main as well, but there the SSE connection did not even work on the first connection for me, so it will need some further investigation there.

florian-h05:
> This was my initial approach too, but I was not able to get the SSE connection to work there at all without digging deeper:

I cherry picked your changes to main, the look good and make sense, I don’t need to reproduce the issue and the fix. In any way, it’s an improvement.

PR #4166, opened 2026-05-04
Add support for pulling OpenAPI spec from URL & Regenerate REST client

florian-h05:


florian-h05:
@Nadahar With these changes you can update the REST client like this:

```shell
 npm run generate-rest-client -- --url http://localhost:8080/rest/spec  
```

Only a single command, formatting, file update, generation - all will be taken care of.

florian-h05:
@jimtng Please merge using rebase merging so the two commits stay separated.

jimtng:
> @jimtng Please merge using rebase merging so the two commits stay separated.

which PR? AFAIK I rebased the non-editable PR on top of the "split" branch. Once the "split" PR is merged I planned to look at it and rebase.

Nadahar:
> Only a single command, formatting, file update, generation - all will be taken care of.

Yeah - that's great, except why do you need so many `--`s? I'd never guess that in a million years...

But, as I said before, this won't work anyway when working against a core branch - unless you don't care about what it ruins in other areas. Only the REST bundles that are running will be part of the generated spec. So, what I need to do, is to download and format it, then go through it manually with a merge tool to look at the changes and pick parts from both versions to keep, and then finally, generate the client.

But, there might be situations where you don't care that it ruins other parts of the spec, and then that'd be handy.

The whole concept is something to think about though, if add-ons make additional REST endpoints available, they won't be a part of the generated client. But, maybe that's OK, maybe those endpoints will just have to be "mapped manually" if needed?

florian-h05:
@jimtng Sorry, wrong ping. I meant @jsjames

florian-h05:
> Yeah - that's great, except why do you need so many --s? I'd never guess that in a million years...

When running a script through npm run, the first -- denotes the begin of the arguments that npm should pass to that script.
This is in the CONTRIBUTING.md.

> But, as I said before, this won't work anyway when working against a core branch - unless you don't care about what it ruins in other areas.

Yes, that’s correct, but with a typed API client, I guess we’ll never make this work. In case you work on the UI part of a core change, I’d recommend just using the OpenAPI spec from that core build (but it has to be a full build).

After the core PR is merged we’ll regenerate the REST client on main and you should be able to rebase your branch.

> The whole concept is something to think about though, if add-ons make additional REST endpoints available, they won't be a part of the generated client.

The UI generally is not adjusting/implementing things for specific add-ons, and in any case add-ons can add servlets, but I think the REST API should be something only core handles.

Nadahar:
> This is in the CONTRIBUTING.md.

Yes, but it's not in my mind because there's a gazillion such rules for everything, and I can only remember so many.

> Yes, that’s correct, but with a typed API client, I guess we’ll never make this work. In case you work on the UI part of a core change, I’d recommend just using the OpenAPI spec from that core build (but it has to be a full build).
> 
> After the core PR is merged we’ll regenerate the REST client on main and you should be able to rebase your branch.

Yes, I understand how it works - as long as the core PR is merged first, there should be no problem. My "problem" is that I develop both in parallel, when I find that I need something in the REST API, I make that, then modify the UI. Between each such change, I must "regenerate" the client, do the manual merge etc.

> The UI generally is not adjusting/implementing things for specific add-ons, and in any case add-ons can add servlets, but I think the REST API should be something only core handles.

I'm not saying that bindings should modify the REST API, but add-ons can be anything. There could be an add-on that created new REST endpoints for some particular purpose, with a "UI add-on" (whatever form that would take) to utilize those endpoints, but there would be no way to get the generated client to know about those endpoints. But as I said, in such situations, I guess the good old "manual mapping" is the way to go.

That said, it would be nice if there was a way to add functionality to MainUI itself via add-ons, but I don't know how easy that is to make or how it would look. You could then create an add-on with a "corresponding UI add-on" that provides a much more tailored experience.

jsjames (approved):
LGTM

Nadahar:
Something really weird is going on with oxfmt. It will not generate a matching spec by itself. But, I've found that if I edit `.oxfmtrc.json` and set:

```json
  "printWidth": 20,
```

...and format the document, it will "split" everything a lot. After that, if I *then* run it with 

```json
  "printWidth": 140,
```

...it formats it the way it is formatted in `main`. But, if I just paste the content manually and run it with 140, it formats it completely differently. It's super-weird.

PR #4169, opened 2026-05-05
Charts: Add option for showing display instead of raw state

florian-h05:
UI part of https://github.com/openhab/openhab-core/issues/4697.

- Add a new parameter to chart series to request the display instead of the raw state.
- Use the new `unit` field in the REST response instead of the fixed `Item::unitSymbol` to display the unit in the tooltip.
- Refactor `$oh.utils` to TypeScript.
- Extract `simpleHash` from `oh-repeater` to `$oh.utils`.


jsjames (approved):
LGTM

PR #4170, opened 2026-05-05
WebSocket util: Unify close/cleanup logic

florian-h05:
Refactor WebSocket close handling to remove connection from openWSConnections array only in onclose event.
Fixes a bug where a not explicitly closed connection was not cleaned-up from `openWSConnections`.


jsjames (approved):
LGTM

PR #4171, opened 2026-05-06
Align rule UID validation with core

Nadahar:
This updates MainUI's rule UID validation to match https://github.com/openhab/openhab-core/pull/5467.

- renamed "rule ID" to "rule UID" since that's what it's called everywhere else
- adds disable option to oh-nav-content for Save link (disableSaveLink)
- minor code cleanup

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/rule/rule-general-settings.vue
+++ b/bundles/org.openhab.ui/web/src/components/rule/rule-general-settings.vue
@@ -4,8 +4,8 @@
       <f7-col>
         <f7-list inline-labels no-hairlines-md>
           <f7-list-input
-            ref="ruleId"
-            :label="`${type} ID`"
+            ref="ruleUid"
+            :label="`${type} UID`"
             type="text"
             :placeholder="`A unique identifier for the ${type.toLowerCase()}`"
             :value="rule.uid"
@@ -14,17 +14,17 @@
             :disabled="!createMode ? true : null"
             :info="createMode ? 'Required. Note: cannot be changed after the creation' : ''"
             input-id="input"
-            pattern="[A-Za-z0-9_\-]+"
-            error-message="Required. A-Z,a-z,0-9,_,- only"
+            :pattern="$oh.utils.ruleUidPattern()"
+            error-message="Invalid rule UID. It can't contain '/', '\' or have leading or trailing whitespace"
             @input="rule.uid = $event.target.value"
             :clear-button="createMode">
             <template #inner>
               <f7-link
-                v-if="createMode && $refs.ruleId?.state?.inputInvalid && rule.uid.trim()"
+                v-if="createMode && !uidValid"
                 icon-f7="hammer_fill"
                 style="margin-top: 4px; margin-left: 4px; margin-bottom: auto"
-                tooltip="Fix ID"
-                @click="$oh.utils.normalizeInput('#input')" />
+                tooltip="Fix UID"
+                @click="this.normalizeUid('#input')" />
             </template>
           </f7-list-input>
           <f7-list-input v-if="!createMode && templateName" label="Template" type="text" :value="templateName" disabled />
@@ -67,7 +67,7 @@
       <f7-col class="skeleton-text skeleton-effect-blink">
         <f7-list inline-labels no-hairlines-md>
           <f7-list-input
-            label="Rule ID"
+            label="Rule UID"
             type="text"
             placeholder="Required"
             value="_______"
@@ -134,6 +134,13 @@ export default {
       if (this.inScriptEditor) return 'Script'
       if (this.inSceneEditor) return 'Scene'
       return 'Rule'
+    },
+    uidRegExp() {
+      return new RegExp('^' + this.$oh.utils.ruleUidPattern() + '$')
+    },
+    uidValid() {
+      if (!this.rule || !this.rule.uid) return false
+      return this.rule.uid.match(this.uidRegExp)
     }
   },
   methods: {
@@ -144,6 +151,11 @@ export default {
     isSceneTag(tag) {
       if (this.inSceneEditor !== true) return false
       if (tag === 'Scene') return true
+    },
+    normalizeUid(id) {
+      const inputElement = document.querySelector(id)
+      inputElement.value = inputElement.value.trim().replace(/\/|\\/, '_')
+      inputElement.dispatchEvent(new Event('input'))
     }
   }
 }
--- a/bundles/org.openhab.ui/web/src/components/rule/rule-general-settings.vue
+++ b/bundles/org.openhab.ui/web/src/components/rule/rule-general-settings.vue
@@ -8,7 +8,7 @@
             :label="`${type} UID`"
             type="text"
             :placeholder="`A unique identifier for the ${type.toLowerCase()}`"
-            :value="rule.uid"
+            v-model:value="rule.uid"
             required
             :validate="editable"
             :disabled="!createMode ? true : null"
@@ -16,7 +16,6 @@
             input-id="input"
             :pattern="uidPattern"
             error-message="Invalid rule UID. It can't contain '/', '\' or have leading or trailing whitespace"
-            @input="rule.uid = $event.target.value"
             :clear-button="createMode">
             <template #inner>
               <f7-link
@@ -24,7 +23,7 @@
                 icon-f7="hammer_fill"
                 style="margin-top: 4px; margin-left: 4px; margin-bottom: auto"
                 tooltip="Fix UID"
-                @click="this.normalizeUid('#input')" />
+                @click="normalizeUid()" />
             </template>
           </f7-list-input>
           <f7-list-input v-if="!createMode && templateName" label="Template" type="text" :value="templateName" disabled />
@@ -33,18 +32,16 @@
             type="text"
             :placeholder="`${type} label for display purposes`"
             :info="createMode ? 'Required' : ''"
-            :value="rule.name"
+            v-model:value="rule.name"
             required
             validate
             :disabled="!editable ? true : null"
-            @input="rule.name = $event.target.value"
             :clear-button="editable" />
           <f7-list-input
             label="Description"
             type="text"
-            :value="rule.description"
+            v-model:value="rule.description"
             :disabled="!editable ? true : null"
-            @input="rule.description = $event.target.value"
             :clear-button="editable" />
         </f7-list>
         <f7-list inline-labels no-hairlines-md>
@@ -75,24 +72,9 @@
             :validate="editable"
             :disabled="true"
             :info="createMode ? 'Note: cannot be changed after the creation' : ''"
-            @input="rule.uid = $event.target.value"
             :clear-button="createMode" />
-          <f7-list-input
-            label="Name"
-            type="text"
-            placeholder="Required"
-            required
-            validate
-            :disabled="true"
-            @input="rule.name = $event.target.value"
-            :clear-button="editable" />
-          <f7-list-input
-            label="Description"
-            type="text"
-            value="__ _____ ___ __ ___"
-            :disabled="true"
-            @input="rule.description = $event.target.value"
-            :clear-button="editable" />
+          <f7-list-input label="Name" type="text" placeholder="Required" required validate :disabled="true" :clear-button="editable" />
+          <f7-list-input label="Description" type="text" value="__ _____ ___ __ ___" :disabled="true" :clear-button="editable" />
         </f7-list>
         <f7-list inline-labels no-hairlines-md>
           <f7-list-group>
@@ -114,8 +96,11 @@
 import TagInput from '@/components/tags/tag-input.vue'
 import { RULE_UID_PATTERN } from '@/js/openhab/uid.ts'
 
+const UID_REGEX = new RegExp('^' + RULE_UID_PATTERN + '$')
+
 export default {
   props: {
+    // This component intentionally edits the shared rule object in place.
     rule: Object,
     ready: Boolean,
     createMode: Boolean,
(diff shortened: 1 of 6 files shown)


Nadahar:
Regarding putting the regexp pattern in `utils.js`. It doesn't really fit there, but I didn't know where to put it. I could probably have been made in TS instead, but it's easy for you guys to organize/move it the way you want it.

florian-h05:
I‘ll look for a better place for the RegExp.

Nadahar:
As always, I found some other stuff that I addressed while doing this, some of which should be mentioned. One is that it turns out that you could actually save a rule with an invalid UID, the field validation didn't actually prevent saving. It does now, in `rule-edit`, but I'm pretty sure that the same problem still exists in `script-edit` and `scene-edit`.

I also resurrected the "fix UID" functionality that was there, but didn't work. If you have entered an invalid UID, a red hammer icon appears, and clicking it "cleans" the string so that it becomes valid.

bilde

Clicking the hammer changes it to this:

bilde

florian-h05:
Wrt the location: I’d add a new uid.ts file to the js/openhab/ directory, I guess we can move other UID validation stuff there as well. No need to create a method to return the RegEx, you can directly export the constant. Another place would be somewhere in assets, where other definitions live.

Nadahar:
> Wrt the location: I’d add a new uid.ts file to the js/openhab/ directory, I guess we can move other UID validation stuff there as well. No need to create a method to return the RegEx, you can directly export the constant. Another place would be somewhere in assets, where other definitions live.

Can I just create `uid.ts` and that will work/be importable?

The reason I create the RegEx in a computed method is that I don't know if it's "safe" to share a RegExp instance between different pages/vue files. Also, I need to have the string available for use in the `pattern` attribute. Will it work to use a TS file directly in an HTML/template attribute?

I thought that TS would have to be somehow "compiled into" JS before use, will the reference be handled correctly for an attribute?

florian-h05:
> The reason I create the RegEx in a computed method is that I don't know if it's "safe" to share a RegExp instance between different pages/vue files. Also, I need to have the string available for use in the pattern attribute. Will it work to use a TS file directly in an HTML/template attribute?

You can export the RegEx as string and as RegEx. As long as you export it as a constant you cannot accidentally modify it and it is safe to share it. At least I never heard it wouldn’t be safe to do so.

> I thought that TS would have to be somehow "compiled into" JS before use, will the reference be handled correctly for an attribute?

Yes, the browser needs JS instead of TS, but you can import TS into JS just as you would import JS into JS. Importing JS into TS however isn’t that simple because TS wants types for the import. The imports and the TS to JS compilation is handled by Vite and done inside the dev server or in the prod build.

Nadahar:
> At least I never heard it wouldn’t be safe to do so.

I'm not claiming that it isn't, I just know too little of the "rules" in this environment, so I often err on the side of caution. Is there any way to "export it as a constant" here? If a function returns an instance, can't it do pretty much what it wants with that instance? The is true in Java too, but the exception is immutable instances (like `String`s). When the instance itself can't be mutated, it's safe to share, but is there such a thing as "immutable" in JS?

This is what Google says:

> JavaScript regular expressions are objects, and while their source pattern and flags are generally immutable, they maintain mutable state through the lastIndex property.

> Side Effects: Reusing the same regex instance for multiple tests without resetting lastIndex to 0 can lead to unexpected results, as it may skip parts of the string or fail to find existing matches.

So, it seems like a good idea to create local RegExp instances to me. It's amazing how they managed to fuck that up though, things like `lastIndex` should live in the `Matcher` instance, not the `RegExp`. Given this behavior, it might not even be safe to share the same instance within the same page/file, as it might be used for matching in several, independent places in the code. At the same time, compiling regex'es is known to be an expensive operation, so you don't want to constantly recycle them. Confusing, I guess bringing my ideas of "safety" into an environment like this just doesn't work...

Nadahar:
Oh my... more "problems": https://www.githubstatus.com/incidents/f5pb5d5mr9yh

Nadahar:
@florian-h05 Look at my latest commit - doing it this way seems to work (I couldn't get the "attribute" to import it directly, but by making it a "data property", I could access it). Is this a better/acceptable design?

Nadahar:
@jsjames Thanks, I don't know enough about what I can and cannot do, so I rely heavily on what I find of existing solutions. The DOM manipulation stuff looked a bit dodgy to me, but it's the way it's done other places - and it seemed to work.

bundles/org.openhab.ui/web/src/components/rule/rule-general-settings.vue
```diff
@@ -132,14 +132,6 @@ export default {
...
   },
   methods: {
-    isScriptTag(tag) {
-      if (this.inScriptEditor !== true) return false
-      if (tag === 'Script') return true
-    },
-    isSceneTag(tag) {
-      if (this.inSceneEditor !== true) return false
-      if (tag === 'Scene') return true
-    },
```
Nadahar: Are you sure that these are unused? The component isn't only used by `rule-edit`, so I assumed that they were used elsewhere.

jsjames:
> @jsjames Thanks, I don't know enough about what I can and cannot do, so I rely heavily on what I find of existing solutions. The DOM manipulation stuff looked a bit dodgy to me, but it's the way it's done other places - and it seemed to work.

no problem - that's part of the problem with the codebase - things are done differently and it will be nice to eventually get things somewhat aligned.

If you could test and verify, i can merge.

@florian-h05, one other UI design question, currently with oh-nav-content we only have the option of displaying or not the Save/Create link. Would it be better to add a "disable" flag to disable the link (vs. it not showing at all) when the fields are not valid?

bundles/org.openhab.ui/web/src/components/rule/rule-general-settings.vue
jsjames: Yeah, it looks like the code was refactored at some point and these were embedded in the tag-input component. (That's another issue with the codebase, there are many areas that need some cleanup).

Nadahar:
> one other UI design question, currently with oh-nav-content we only have the option of displaying or not the Save/Create link. Would it be better to add a "disable" flag to disable the link (vs. it not showing at all) when the fields are not valid?

I completely agree with this, I looked for a way to disable it, but when I couldn't find it, I chose to hide it instead. It's not that good a solution anyway, because the keyboard shortcut still works, so I also had to check this in the save method itself.

If disabling was possible, the padlock that is shown for read-only rules could be in a disabled state permanently, there's no point in it being "clickable" only to immediately abort the save method. Also, it would make it possible to show a tooltip when "Save" was disabled because of unfulfilled requirements.

bundles/org.openhab.ui/web/src/components/rule/rule-general-settings.vue
Nadahar: It seems like this code has "moved" to `tag-input.vue` and that somebody forgot to delete them?

edit: GitHub didn't show me your reply until *after* I wrote mine 😒

Nadahar:
> If you could test and verify, i can merge.

As far as I can tell, it still works the same.

jsjames:
> > one other UI design question, currently with oh-nav-content we only have the option of displaying or not the Save/Create link. Would it be better to add a "disable" flag to disable the link (vs. it not showing at all) when the fields are not valid?
> 
> I completely agree with this, I looked for a way to disable it, but when I couldn't find it, I chose to hide it instead. It's not that good a solution anyway, because the keyboard shortcut still works, so I also had to check this in the save method itself.
> 
> If disabling was possible, the padlock that is shown for read-only rules could be in a disabled state permanently, there's no point in it being "clickable" only to immediately abort the save method. Also, it would make it possible to show a tooltip when "Save" was disabled because of unfulfilled requirements.

Let me know what you think of the added disableSaveLink option. I think its a better UI representation.

jsjames:
> > one other UI design question, currently with oh-nav-content we only have the option of displaying or not the Save/Create link. Would it be better to add a "disable" flag to disable the link (vs. it not showing at all) when the fields are not valid?
> 
> I completely agree with this, I looked for a way to disable it, but when I couldn't find it, I chose to hide it instead. It's not that good a solution anyway, because the keyboard shortcut still works, so I also had to check this in the save method itself.

Correct, the disabling in the UI is just a nod to the user that it can't be saved - it still should be validated (as you did) in the save method.

florian-h05:
I agree it makes sense to be able to disable the save link. Though the option to hide it should be kept.

Nadahar:
I think it looks better, but I would have preferred with the padlock shown for read-only stuff was "disabled" as well. But, when I add the "disabled" class to `f7-link`, the tooltip stops working, so it's probably better to leave it "enabled"





> Though the option to hide it should be kept.

I think that's still an option if you simply don't set `saveLink` - but perhaps not so logical: if `editable` is false, the padlock will still show. But, if neither `editable` nor `saveLink` is set, it should be hidden as far as I can tell.

Nadahar:
I would say that it would make more sense if: No `saveLink` means that nothing is shown, neither padlock nor "Save". If `saveLink` is set, but it's not editable, the padlock is shown - otherwise the text.

jsjames:
> But, when I add the "disabled" class to f7-link, the tooltip stops working, so it's probably better to leave it "enabled"

Just to confirm, you are referring to only the padlock here and not the Create/Save link, correct?  If so, I can merge at this point?

Nadahar:
> Just to confirm, you are referring to only the padlock here and not the Create/Save link, correct? If so, I can merge at this point?

Yes, there's no tooltip for the create/save text anyway. Only the padlock has one, and it stops working if the padlock is "disabled". Just go ahead and merge, those navbar details can be tweaked later independently of this PR if desired.

Nadahar:
I discovered a nice "bonus" - now "Save" is disabled also while the page is loading 👍

jsjames:
> > But, when I add the "disabled" class to f7-link, the tooltip stops working, so it's probably better to leave it "enabled"
> 
> Just to confirm, you are referring to only the padlock here and not the Create/Save link, correct? If so, I can merge at this point?

@florian-h05 - since I made changes here, what are the guidelines for merging - should I or you do the merge?

Nadahar:
Some tweaking must be allowed IMO, otherwise it becomes hard to do anything. If we want to be really pedantic, let's have @florian-h05 push a commit too, and it has become unmergeable 😎

jsjames:
> Some tweaking must be allowed IMO, otherwise it becomes hard to do anything. If we want to be really pedantic, let's have @florian-h05 push a commit too, and it has become unmergeable

:)  yeah, at some point "what is a tweak" vs. a large change that should get further reviewed.  I'm fine merging, just didn't know if there are guidelines on this that should be followed.

jsjames (approved):
LGTM

florian-h05 (commented):
FTR: LGTM

florian-h05:
> since I made changes here, what are the guidelines for merging - should I or you do the merge?

If it's only tweaking, you can merge yourself. If you do substantial changes, ask another maintainer to merge. It's up to you to decide whether its tweaking or substantial changes.

> Some tweaking must be allowed IMO, otherwise it becomes hard to do anything. If we want to be really pedantic, let's have @florian-h05 push a commit too, and it has become unmergeable 😎

Yeah, that illustrates the problem very exactly, you cannot be to pedantic, especially when there are not that many maintainers.

Nadahar:
@jsjames I've now rebased my larger "rules code tab refactoring" (introducing YAML/DSL view etc.) on the updated `main`, and I've got a problem.

There used to be a bug in `rule-general-settings` where if the UID, label/name or description became `undefined`, the controls wouldn't update and still show the old value (which might not be true anymore, if the code has been changed from the code tab, for example). I fixed it by doing this, and according to my testing, it works flawlessly this way:

https://github.com/Nadahar/openhab-webui/commit/4b8c2bfb87d13e7cd7436fd360881babc0c67eda

But, when you changed it from `:value` to `v-model:value` and removed the `@input` handler, I don't get the opportunity to "translate" the values anymore. The point is that `undefined` in the `rule` object must be "translated" to and from and empty string in the control for things to work properly. Is it possible to do this "translation" using v-model, or should I just revert to the previous way of doing it to retain the fix?

Nadahar:
When i read about v-model, I find this:

> v-model can be used on a component to implement a two-way binding.
> 
> First let's revisit how v-model is used on a native element:
> template
> 
> ```vue
> <input v-model="searchText" />
> ```
> 
> Under the hood, the template compiler expands v-model to the more verbose equivalent for us. So the above code does the same as the following:
> template
> 
> ```vue
> <input
>   :value="searchText"
>   @input="searchText = $event.target.value"
> />
> ```

This makes it look like "v-model" is just syntactic sugar to achieve what was already there with less code, but it removes the flexibility to "translate" the values. As a result, I think I should just revert it to the previous behavior where my fix can be used. Agreed?

jsjames:
> When i read about v-model, I find this:
> 
> > v-model can be used on a component to implement a two-way binding.
> > First let's revisit how v-model is used on a native element:
> > template
> > ```
> > <input v-model="searchText" />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Under the hood, the template compiler expands v-model to the more verbose equivalent for us. So the above code does the same as the following:
> > template
> > ```
> > <input
> >   :value="searchText"
> >   @input="searchText = $event.target.value"
> > />
> > ```
> 
> This makes it look like "v-model" is just syntactic sugar to achieve what was already there with less code, but it removes the flexibility to "translate" the values. As a result, I think I should just revert it to the previous behavior where my fix can be used. Agreed?

Correct, that is all v-model is - in the original implementation that was all it was doing. Implementing with a separate @input is fine.

Nadahar:
> Correct, that is all v-model is - in the original implementation that was all it was doing. Implementing with a separate @input is fine.

Yes - I know that it was all the code did before my bugfix - but not after 😉 All I really wanted to know is that I don't do anything wrong by reverting it to not use v-model 👍

PR #4173, opened 2026-05-06
Channel list: Scroll to the bottom after "Select All" is clicked

jimtng:
This is a small QoL improvement, I think. In "Add equipment to model" -> Channel list when clicking Select All, the list of channels would expand to display more information. This results in the "Select All" and the "Add to Model" buttons to disappear off the screen because they get pushed down very far.

## Before clicking:
image

# After clicking

After clicking "Select All", the current position is about 20% towards the top of the screen - at the start of the channel list which has now all expanded much longer. The user would have to scroll way down before they can see the "Select All" button that they clicked and to find "Add To Model" (the natural next step). This "jump" is quite jarring from a UX perspective.


florian-h05 (approved):
LGTM, thanks!

PR #4174, opened 2026-05-06
Sidebar Menu: Enable expanding/collapsing menu sections without requiring item selection

jimtng:
This lets you jump straight to any sub-menu without having to bunny-hop onto the top-level menu first.

The main top-level menu items are still clickable. The expand/collapse action is only done by clicking the chevron.

- Down chevron indicates that the menu item has sub-items, and it can be expanded to reveal them
- Up chevron indicates that the sub-items can be collapsed
- No chevron means there are no sub-items
- The currently selected item is highlighted as usual and has a right chevron. 

menu-expand-collapse

### DIFF
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -8,7 +8,7 @@
     <!-- Left Panel -->
     <f7-panel v-show="ready" left :cover="showSidebar ? true : null" class="sidebar" :visible-breakpoint="960">
       <f7-page>
-        <f7-link href="/overview/" class="openhab-logo no-ripple" panel-close>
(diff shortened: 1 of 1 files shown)


jimtng:
The only thing that's still a bit inconsistent is that the pages (above the menu) all have right chevrons whether or not they're currently selected. Since this PR is about the menu, the pages chevrons haven't been changed. Should we remove the right chevrons on the pages, and only show one when that page is actually selected, to be consistent?

jimtng:
rebased

florian-h05:
> Should we remove the right chevrons on the pages, and only show one when that page is actually selected, to be consistent?

Yes :+1:

jimtng:
> > Should we remove the right chevrons on the pages, and only show one when that page is actually selected, to be consistent?
> 
> Yes 👍

done

florian-h05 (approved):
LGTM, thanks!
This makes navigating around way faster 🚀

Issue #3490, opened 2025-11-20
[MainUI] Loading /settings/items/ shows only one item; `Uncaught (in promise) DOMException: The operation is insecure.`


dilyanpalauzov:
When I load over HTTPS directly `/settings/items/`, it shows one item, alphabetically the first one of all items.

When I click somewhere else, then go back to Settings → Items then all items are shown.

It might be unrelated, but the browser console contains `Uncaught (in promise) DOMException: The operation is insecure.`:
```
Uncaught (in promise) DOMException: The operation is insecure. 
[index-Bpy8UFX_.js:7720:54](https://192.168.0.11:8443/assets/index-Bpy8UFX_.js)
    __tla https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:7720
    (Async: promise callback)
    __tla https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:7719
    register https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:7714
    init https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:7769
    forEach self-hosted:157
    init https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:7768
    __tla https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:4626
    forEach self-hosted:157
    __tla https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:4625
    forEach self-hosted:157
    emit https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:4620
    n2 https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:4992
    init https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:4996
    __tla https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:10752
    __tla https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:1478
    Nn https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:1028
    Pn https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:1035
    __weh https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:1470
    Xn https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:1117
    Qn https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:1134
    (Async: promise callback)
    qn https://192.168.0.11:8443/assets/index-Bpy8UFX_.js:1092
[...]

PR #4175, opened 2026-05-07
Items list: Fix virtual-list rendering on hard refresh

jimtng:
Prevent mounting the Framework7 virtual-list while hidden and trigger a short delayed resize after initial population so the list remeasures its scroll height.

The bug: When refreshing the items list using the browser's refresh button or Cmd+R, the items list will end up showing one entry. 

image

One must navigate away then click Items to get the normal listing.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/items-list-vlist.vue
@@ -79,7 +79,7 @@
         </f7-list>
       </f7-col>
 
-      <f7-col v-show="ready && items.length > 0">
+      <f7-col v-if="ready && items.length > 0">
         <f7-block-title class="no-margin-top">
           <span>{{ listTitle }}</span>
           <template v-if="showCheckboxes && listedItems.length">
@@ -92,13 +92,7 @@
         <f7-list v-if="!listedItems.length">
           <f7-list-item title="Nothing found" />
         </f7-list>
-        <f7-list
-          v-show="listedItems.length > 0"
-          class="searchbar-found col"
-          ref="itemsList"
-          media-list
-          virtual-list
-          :virtual-list-params="vlParams">
+        <f7-list class="searchbar-found col" ref="itemsList" media-list virtual-list :virtual-list-params="vlParams">
           <ul>
             <f7-list-item
               v-for="(item, index) in vlData.items"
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
The problem description seems similar to https://github.com/openhab/openhab-webui/issues/3490.

jimtng:
> The problem description seems similar to #3490.

Indeed it's similar.

In my case, the https one works fine, but http exhibits this problem. As I understand it you have the problem under https? 

Can you test this PR to see if it solves your problem?

dilyanpalauzov:
As this happens only sometimes, and I cannot say when, it cannot be tested: if the problem does not appear during testing with the current change, it does not mean, that the problem is solved.  I tried it now without the current change, and the problem does not appear.

jimtng:
I have two openhab instances. I can consistently trigger the problem on one (test instance) about 80% of the time, but I cannot trigger it on the other (production) no matter what I tried.

florian-h05 (approved):
LGTM, thanks!
I haven't been able to reproduce the original issue, but have seen it from time to time. The changes make sense, let's see whether the issue ever appears again.

jsjames:
I also could not duplicate the issue, but I did ask co-pilot what it thought was the issue. Not sure if that makes sense or if it was tried - it might be a better fix than issuing the resize? Thoughts?

```
1. Component mounts fresh → listedItems = [] → v-show="listedItems.length > 0" is false → the f7-list element has display: none
2. API returns → replaceAllItems(this.items) is called while the container is hidden
3. Framework7's virtual list calculates the visible area based on clientHeight = 0, so it renders only 1 item (its minimum)
4. updateListedItems() runs in a nextTick, populating listedItems, making v-show true — the list is now visible
5. But F7 never recalculates, so only 1 item stays rendered

On SPA navigation, the component is typically not destroyed between visits, so listedItems already has items from the last visit. The container is visible when replaceAllItems is called, and F7 calculates correctly.

The fix is to call f7VirtualList.update() in the existing nextTick at the end of load(), by which point updateListedItems()'s own $nextTick has already run, the DOM has updated, and the container is visible:
```

jimtng:
I removed the setTimeout -> it broke again
I replaced setTimeout with `this.$refs.itemsList.$el.f7VirtualList.update()` -> Still broken
I replaced the resize inside the setTimeout with virtualList.update() -> works.

So we can replace the resize with update if preferred? It's probably more "obvious" / self explanatory rather than the resize.

dilyanpalauzov:
> > The problem description seems similar to #3490.
> 
> Indeed it's similar.
> 
> In my case, the https one works fine, but http exhibits this problem. As I understand it you have the problem under https?
> 
> Can you test this PR to see if it solves your problem?

It does not solve the problem with version 5.2.0 build 5389 Main UI Commit ba322023, at least not for insecure http.

PR #4188, opened 2026-05-11
Restore readOnlyMsg prop on code-editor

jimtng:
Regression from #4115

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
@@ -56,6 +56,7 @@
           :object="item"
           :object-id="item.name"
           :read-only="!editable"
+          read-only-msg="This Item is not editable because it has been provisioned from a file."
           @save="save()"
           @parsed="updateItem"
           @changed="onCodeChanged" />
--- a/bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
@@ -6,6 +6,7 @@
       :value="code"
       :hint-context="hintContext"
       :read-only="readOnly"
+      :read-only-msg="readOnlyMsg"
       @input="onEditorInput"
       @save="$emit('save')" />
   </div>
@@ -105,7 +106,8 @@ export default {
     objectType: String, // the type of the object, e.g. 'items', 'things' which corresponds to the yaml element name.
     objectId: String,
     hintContext: Object,
-    readOnly: Boolean
+    readOnly: Boolean,
+    readOnlyMsg: String
   },
   // @parsed  event is emitted when the code has been parsed back into an object
   //          as a result of calling the parseCode() method
--- a/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
@@ -1,11 +1,6 @@
 <template>
   <div class="code-editor-fit">
-    <f7-icon
-      v-if="readOnly"
-      f7="lock"
-      size="50"
-      class="code-readonly-icon"
-      :tooltip="readOnlyMsg || 'Read-only: provisioned from a file'" />
+    <f7-icon v-if="readOnly" f7="lock" size="50" class="code-readonly-icon" :tooltip="readOnlyMsg || 'Not editable'" />
     <codemirror ref="cm" :model-value="value" :extensions="extensions" @ready="onCmReady" @change="onCmCodeChange" />
   </div>
 </template>
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/thing-details.vue
@@ -313,6 +313,7 @@
           :object="thing"
           :object-id="thing.UID"
           :read-only="!editable"
+          read-only-msg="This Thing is not editable because it has been provisioned from a file."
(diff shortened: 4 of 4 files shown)


jimtng:
FYI @Nadahar

Nadahar:
I guess you used the old text, and that it already used "file". I guess I should pick my battles, but it just feels "wrong" every time I see it.

What about this in `script-edit` (line 175) that was changed from

```vue
       :tooltip="isBlockly ? 'Cannot edit the code generated by Blockly' : 'This code is not editable'" />
```

to

```vue
        :read-only-msg="isBlockly ? 'Cannot edit the code generated by Blockly' : null"
```

This applies, AFAIK, when you edit an action of a rule. Is there a particular reason why there should no longer be a tooltip there?

jimtng:
> I guess you used the old text, and that it already used "file". I guess I should pick my battles, but it just feels "wrong" every time I see it.

If and when we have other sources, we would then need to add _source name_ (file? filename? binding?) as part of the metadata so we can then display it properly.

Until such time that we do have uneditable sources other than files, I believe we should stick with calling it "file". Just saying "not editable" is confusing - you can look at it and just see not editable but why and where can I edit it?



> What about this in `script-edit` (line 175) that was changed from
> 
> ```
>        :tooltip="isBlockly ? 'Cannot edit the code generated by Blockly' : 'This code is not editable'" />
> ```
> 
> to
> 
> ```
>         :read-only-msg="isBlockly ? 'Cannot edit the code generated by Blockly' : null"
> ```
> 
> This applies, AFAIK, when you edit an action of a rule. Is there a particular reason why there should no longer be a tooltip there?

I'm trying to reduce the instances where this message is written (DRY). Essentially null means "use the default message".

See script-editor.vue line 8.

Nadahar:
> If and when we have other sources, we would then need to add _source name_ (file? filename? binding?) as part of the metadata so we can then display it properly.
> 
> Until such time that we do have uneditable sources other than files, I believe we should stick with calling it "file". Just saying "not editable" is confusing - you can look at it and just see not editable but why and where can I edit it?

I just want it to be correct - and "not editable in the UI" *is* as correct as it can be as far as I can tell - because the point isn't that this object isn't editable - on the contrary, you just can't edit it from the UI.

I think trying to show the "reason" it is read-only is too complex and not worth it. I think there already are examples of add-ons that create rules, but I only have a vague memory of it, so I can't point to it.

> I'm trying to reduce the instances where this message is written (DRY). Essentially null means "use the default message".
> 
> See script-editor.vue line 8.

Yes, but again you replace it with the misleading "default" that drags files into the mix instead of the, perhaps less informative, but at least correct and not misleading "This code is not editable".

Nadahar:
The bottom line is that, as I see it, the UI doesn't *know* why it isn't editable. It just knows that it isn't. So, it can't give any more information without making assumptions, unless a system to relay the reason why is implemented.

jimtng:
This is why I want to know in which instances / examples would it not be editable but it's not from a file?

Nadahar:
> This is why I want to know in which instances / examples would it not be editable but it's not from a file?

From *any* `Provider` that is not the `ManagedProvider`. So, it could be a rule that is generated by an add-on for some purpose, it could be definitions that are provided by a database, from a webpage, pick any source imaginable. They will all be read-only in the UI.

jimtng:
> > This is why I want to know in which instances / examples would it not be editable but it's not from a file?
> 
> From _any_ `Provider` that is not the `ManagedProvider`. So, it could be a rule that is generated by an add-on for some purpose, it could be definitions that are provided by a database, from a webpage, pick any source imaginable. They will all be read-only in the UI.

I understand that. But do such things exist now? I know you have said that's not the point, but please humour me.

jimtng:
PS: This whole "it has been provisioned from a file" wasn't my invention. It has been like this for many years.

So perhaps if you want to debate about it, open a new issue?

The bottom line for this PR is that the readOnlyMsg is restored.

Nadahar:
> I understand that. But do such things exist now? I know you have said that's not the point, but please humour me.

As I've already said, I think it does for rules, but I don't remember the details. It's not difficult to imagine how it could exist, so I don't really understand the relevance of whether I can come up with an example of it here and now. It is, wrong to claim that it comes from a file, and we know that. Why do we still want to claim that it is, even if that is the most common source?

But, as I said before, if "everybody else" think that making erroneous assumptions somehow benefits the user, go ahead. It's not important enough to invest any more energy in. I've explained why it's wrong, and I can't really do anything more.

Nadahar:
> PS: This whole "it has been provisioned from a file" wasn't my invention.

I know that it was like that for Things and Items, but it wasn't for rules. So, this "spread it further", when editing actions in particular.

jimtng:
I've made the default just "Not editable". The Things and Items message remain as is.

jimtng:
Related: openhab/openhab-jruby#535

bundles/org.openhab.ui/web/src/pages/settings/items/item-edit.vue
```diff
@@ -56,6 +56,7 @@
           :object="item"
           :object-id="item.name"
           :read-only="!editable"
+          read-only-msg="This Item is not editable because it has been provisioned from a file."
```
florian-h05, line 59: Just as a side note: JS Scripting supports provisioning Items, but that really only makes sense with file-based scripts, so technically, "from a file" is correct. Wrt Items, maybe we could extend the API to add information about the class name of the provider that provides an Item. Similar for Things below.

florian-h05 (approved):
LGTM, thanks!

I think the whole discussion about the text should be moved to an issue. I agree that the current texts are not optimal, and I think we need to enrich the REST API response with information that the UI can use to actually tell where something comes from.

PR #4193, opened 2026-05-12
[habot] Bump bouncycastle to 1.84

holgerfriedrich:
Supersedes #4116 (I cannot push to the dependabot branch, so I created it from scratch)


holgerfriedrich:
@florian-h05 can you help out here? OH is on 1.8.4.

florian-h05 (approved):
Thanks for the upgrade. I haven't tested it, but since its a minor release and everything still compiles, it should be fine.

PR #4195, opened 2026-05-13
Fix app drawer when dev sidebar is open

jimtng:
Steps to reproduce:
- Go to home page / overview
- Open dev sidebar click on the icon or shift+alt+d
- Click on the app drawer icon 

When the devsidebar is open, clicking on the appdrawer icon (from the home page) creates an extra blank vertical column. The appdrawer didn't show up.

Also made the appdrawer panel resizable, since its size is otherwise only governed by the right panel size which is only resizable through the devsidebar.

We possibly need to backport this to 5.1 because this bug exists there too. I haven't checked 5.0.

This bug didn't exist in 4.x, so it was probably introduced in the vue2 -> vue3 upgrade.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -347,7 +347,7 @@
     </f7-panel>
 
     <!-- Right Panel -->
-    <f7-panel v-if="ready" right reveal dark>
+    <f7-panel v-if="ready" right reveal dark resizable>
       <panel-right />
       <!-- <f7-view url="/panel-right/"></f7-view> -->
     </f7-panel>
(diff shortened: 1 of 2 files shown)


jimtng:
TBH although this "fix" seems almost identical, I don't quite understand why it works when the original didn't.

florian-h05:
Now the UI switches from the dev sidebar to the app drawer, and closing the app drawer does not restore the dev sidebar. I don't have a problem with that, just to confirm this is identended behaviour?

jimtng:
> Now the UI switches from the dev sidebar to the app drawer, and closing the app drawer does not restore the dev sidebar. I don't have a problem with that, just to confirm this is identended behaviour?

Yes it's fine I think, although if someone really wanted to, we can show the dev sidebar again.

Basically once the appdrawer is shown, clicking _anything_ outside the app drawer will close the app drawer. At first I tried "clicking" one of the other icons (e.g. the app drawer icon again to "toggle" it, or clicking "dev sidebar" to open the sidebar) but then I realised that it's not actually clicking any icons outside. It's all just a "backdrop" click anywhere outside the app drawer, to close it.

florian-h05 (approved):

PR #4199, opened 2026-05-15
Update MainUI styling, add searchbar to config sheet, add advanced parameter indicators

### LINKED ISSUE #4198: Allow Collapse/Expand in Widget Properties
In openHAB custon widgets there are times when there are many “Properties” needing to be configured and the widget editor allows the creation of Groups to simplify things for the user when entering.  

This still creates a challenge as the editing of this can be many pages and difficult to scroll and see where one is

## Your suggestion

Allow the Collapse/Expand of the Groups and even a default of if its expanded or collapsed similar to the Accordion functionality in Main UI for example below is the definition of a group and the relevant Accordian options in bold
/
+-----------------------+
| Enter Properties...         |
| (Click < to expand).       |
+-----------------------+
| < Group1                        |
| > Group2                        |
|   - Name2                       |
|   - Item2                       |
| < Group3                        |
+-----------------------+

e.g. 

props:
  parameters

 - context: string
       default: The Name
       description: The name for 1
       label: Name1
       name: name1
       required: true
       type: TEXT
       groupName: Key
   - context: item
       description: The Item 1
       label: The Item 1
       name: item1
       required: true
       type: TEXT
       groupName: Key 
parameterGroups:
     **_- config: 
        accordianEnabled: true_**
    - name: Group1
        context: Key
        label: Group1
        description: Item 1 Parameters
    - name: Group2
        context: Item2
        label: groupd2
        description: Item 2 Parameters
 

https://community.openhab.org/t/dual-inverter-gauge-widget/162396

## Your environment
openHAB 5.1.3
Openhabian (trixie)
Raspberry Pi 5 8Gb

```yaml

```

jimtng:
- Convert boxy information groups into rounded rectangle, outlined boxes
- Add a searchbar for the configuration sheet
- Mark advanced configuration parameters and advanced channels with a blue left border for easy identification
- Apply cosmetic improvements to text input fields, including distinct background fills and focused state outlines
- Sort the well-known Item Metadata namespaces according to the well-known namespace ordering
- Add an icon to list filter header

Resolves #4198 - with future enhancement to show category list, possibly in a future PR.

# A Search feature for config sheet:

- Advanced parameters are indicated with a blue left border
- Focused input fields are outlined for better visibility

image

---

# Settings Screen

image

## On a narrow screen:

The sides snap to the edges, the rounded corners disappear and left/right borders removed.

image

---

# Developer Tools page

image

---

# List Pages:

- Add an icon for "Filter"
- The item counter + select all / deselect in the list header. Select All / deselect all right-aligned

image

---

image

---

## The advanced channels are also indicated with blue left border

image

---

image

---

And more.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/config-parameter.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-parameter.vue
@@ -2,7 +2,7 @@
   <f7-list
     v-show="configDescription.visible ? configDescription.visible(value, configuration, configDescription, parameters) : true"
     ref="parameter"
-    class="config-parameter"
+    :class="['config-parameter', { 'advanced-param': configDescription.advanced }]"
     :no-hairlines-md="
       configDescription.type !== 'BOOLEAN' &&
       (!configDescription.options || !configDescription.options.length) &&
@@ -39,8 +39,6 @@
 </template>
 
 <script>
-import { f7 } from 'framework7-vue'
-
 // import ScriptEditorPopup from './config/script-editor-popup.vue'
 import ParameterThing from './controls/parameter-thing.vue'
 import ParameterBoolean from './controls/parameter-boolean.vue'
@@ -164,6 +162,7 @@ export default {
     padding-left 0
 .param-description
   padding-left 16px !important
+  padding-right 10px !important
   &.block-footer
     margin-top 2px
     margin-bottom 1rem
@@ -187,4 +186,6 @@ export default {
       white-space nowrap
       margin-top 0
       margin-bottom 0
+  &.advanced-param
+    border-left 2px solid var(--f7-color-blue)
 </style>
--- a/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/config-sheet.vue
@@ -1,51 +1,64 @@
 <template>
   <f7-block v-if="parameters" class="config-sheet no-margin" ref="sheet">
-    <div v-if="hasAdvanced" style="text-align: right" class="padding-right padding-bottom">
-      <label class="advanced-label">
-        <f7-checkbox v-model:checked="showAdvanced" />
-        Show advanced
-        <f7-badge
-          v-if="advancedNonDefaultCount"
-          style="margin-left: 2px"
-          color="blue"
-          class="count-badge"
-          tooltip="Non-default advanced parameter">
-          {{ advancedNonDefaultCount }}
-        </f7-badge>
-      </label>
+    <div v-if="showFilterControls" class="config-sheet-filters">
+      <f7-searchbar
+        v-if="showSearchbar"
+        ref="searchbar"
+        custom-search
+        :backdrop="false"
+        placeholder="Search configuration"
+        :disable-button-text="null"
+        @searchbar:search="onSearch"
+        @searchbar:clear="clearSearch" />
+      <f7-chip
+        v-if="hasAdvanced"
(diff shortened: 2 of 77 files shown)


jimtng:
This PR is no longer dependent on another PR.

mherwege:
Not sure how to do this better, but the blue border next to advanced parameters draws more attention to it, rather than less. Fine to leave it like that if no other ideas pop up.

jimtng:
> Not sure how to do this better, but the blue border next to advanced parameters draws more attention to it, rather than less. Fine to leave it like that if no other ideas pop up.

The blue is meant to correspond to the blue in the "Advanced" button itself, as you may be aware.

I considered adding a little icon / flourish after the param name, etc. instead, but it would still be hard to identify / scan visually, unlike the blue border.

As to whether your attention should be more on the advanced or not, I figured since you actually toggled to show Advanced params, you are probably more interested in them, therefore the highlighting works perfectly.

I'm open to better ideas or or objections etc. I just want them clearly identifiable, unlike the current situation.

jimtng:
@openhab/webui-maintainers please remove the "awaiting other PR" tag.

jsjames:
@jimtng - I'd like to use css styles for the truncation if possible in the item-picker - let me know if this provides the effect you wanted.

jimtng:
> @jimtng - I'd like to use css styles for the truncation if possible in the item-picker - let me know if this provides the effect you wanted.

Thank you for the fixes! Using CSS is much better.

jimtng:
The only thing I'm wondering about is whether this css for title vs after can be made global, but that might be for another PR since that might affect something else.

jimtng:
rebased. I noticed there are some things still set to "blue" although unrelated to this PR. Perhaps another pass should be done at some point.

jsjames (approved):
Good cleanup of the UI - Thanks.

florian-h05:
I overall like the new look, however I feel like its "wasting" too much space in config dialogs, what previously fitted on a FHD screen without scrolling now needs (a lot of) scrolling which is very annoying. We already lost "condenseness" when upgrading Framework7 last year, I don't think we should further reduce it.

florian-h05:
image

Even with a single configuration parameter, you can already see the "loss" of space. I unfortunately have no better example as the amount of Things on the demo server is quite limited with a even smaller subset of Things I also have at my openHAB.

This also very extreme with Thing action popups, where it feels like the required height has tripled ...

jimtng:
Some ideas:

We could remove some of the spacings:
- Remove the main title line "Thing Details", "Item Details", "Rule Details", etc. - this was added in this PR
- Reduce input field margins

These two changes would probably restore the heights back to the original.

But, I like the spacings especially the input fields and the group-box title - things aren't too cramped and easier to tell input fields vs static text. It also helps with deteriorating visions but not quite to the level of "vision impaired".

Maybe we could add an option for "Compact" vs "Standard"? In the compact mode, we could remove the spacings between one groupbox to the next and have a bolder lines to delineate the separations between one group box and the next?

Thing action popups - we could reduce the top/bottom margins of each list-item

The group-box has the ability to collapse things in accordions, but that means you can't see everything although we can add a way to collapse all / expand all, or a mode where you expand one, it collapses the others, but that still hides things.

Lastly, we could add a "section index" on the left column (not sure if it will fit horizontally) so people can jump between configuration parameter groups more easily (i.e. like Windows File Explorer / manager but only one level, not tree-like, or like OSX Preview TOC when viewing a PDF). This left index can be shown/hidden much like our left panel menu bar, for example.

What are your thoughts on this?

PR #4202, opened 2026-05-16
[MainUI] Rule and RuleTemplate file format conversion

Nadahar:
This is the "companion PR" of https://github.com/openhab/openhab-core/pull/5572. It implements file format conversion and "the updated Code tab" for rules, to make it work similarly to Things and Items.

As a result, this PR depends on https://github.com/openhab/openhab-core/pull/5572 and won't "work" if used against a core version without those changes. It will therefore be difficult (but not impossible) to test until https://github.com/openhab/openhab-core/pull/5572 has been merged.

After https://github.com/openhab/openhab-core/pull/5572 has been merged, the MainUI REST client should be regenerated and this PR rebased on top of that, to make sure that the REST client is as it should. I don't trust that the REST client I've generated is fully compliant with unrelated parts of the REST API.

As far the changes themselves, the commits provide a coarse description of what each step does.  This has been a bit complicated with lots of corner cases to cover, so I expect that there exists "corner cases" that I haven't thought of testing and that don't behave optimally.


Nadahar:
During [the development of the twin PRs](https://community.openhab.org/t/rules-and-rule-templates-yaml-integration/168568) I've created quite a few GIFs and at least one video that shows parts of what this does. I'll include some of them here, to give a better idea of what the changes are (not all features exist in all the GIFs, as these were created during development):

ShowAllDesktop

EditRuleSync

EditNewRule

EditIncompatibleDSL

EditYamlHints1

EditYamlHints2

CodeEditToolbarResize2

RuleSharedContext

A video showing "mass export" of rules: https://youtu.be/VcOpyv2Oy7E

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -152,16 +215,43 @@ export default {
...
+            onSuccessCallback()
+          }
+          return
+        }
+      }
+      const ruleShowAll = targetMediaType === 'application/yaml+rule' && this.isShowAll
+      targetMediaType = targetMediaType.split('+')[0] // remove the +thing, +item or +rule suffix, if present
+      const params = new URLSearchParams()
+      if (ruleShowAll) {
+        params.set('ruleSerializationOption', 'INCLUDE_ALL')
```
mherwege, line 246: I am not a big fan of doing this inside this code-editor component. This is rule specific. I would think this should be handled from the parent component, and you already have a mechanism in place with the extra template to handle extra buttons. `parseCode` and `generateCode` in this component are meant to be called from a parent component. Why not give these functions extra arguments for the parameters and pass these in from the parent? In my view this is too rule specific and the code editor is for many different code types.

Nadahar: I agree in principle, but had to try to find a balance between refactoring/generalizing and getting things done. I'm not very well versed in JavaScript, and I have to look up pretty much how to do everything. That means that "designing" is a slow process, because I don't know what possible constructs are - not what is considered "best practices".

I initially did more directly in the code-editor, and then later refactored things to be more general. Maybe I missed this one. Are you thinking that instead you could just send one or more "general parameters" that would be added as query parameters? What about path parameters or body parameters? It could quickly become complicated to make it fully generic, and again, I just don't know what constructs are considered "good" and "bad", so any suggestions about how a good design would look are appreciated.

mherwege: I was looking at this, and also had other parameters in the create endpoint that I could have used for `showAll` (I didn't in the end). But these parameters are different from the ones for rules, so there is no general mechanism. I would have had to further enhance that code-editor with again extra code just for semantics. That is not tenable. And you already create the template mechanism that fits the bill and could do exactly what you need in my view.

Nadahar: I think maybe you misunderstood what I wrote - or I misunderstood your answer, but I agree: It shouldn't be there. I'm just not sure what the "right design" is. Can this be done using the template mechanism, isn't that about the HTML, not about "logic"? I think I'm missing something...

Nadahar: I've looked at it a bit now, and starts to remember a bit about the considerations I made.

First, code-editor itself isn't very "generalized", there is very little flexibility there without making code changes that lead down a road where it's harder and harder to maintain. It's not at all how I would have written something in a language I know. But, this is a language where I'm ill-suited to make "structural design decisions", I simply don't know what can be done. There is no OO/inheritance to work with, which would have made things easier. Instead of trying to "redesign" things, I just tried to make the functionality I needed with what was already there. I think it makes much more sense if somebody more familiar with the language refactors the code into a better structure if it's deemed necessary, than that I try to do something I don't have the skills to do.

As far as I can tell, this can't be done with templates alone, `generateCode()` is called within the component without involving the parent component, and the request must be modified based on the selected language and this option. Changing this would mean a larger redesign using callbacks or something similar, so that the parent component could then modify the request. I'm not sure how that would look or how elegant it would be.

There's also the layout part. The "Select All" option is "a part of" the group with the language buttons. It scales and moved differently than it would have done if it were merely supplied as a template. I think the current behavior is preferable to it just "floating around" with the other buttons.

I made it so that "Select All" could easily be used for other "media types" as well. That's why I made it like this:

```js
      const ruleShowAll = targetMediaType === 'application/yaml+rule' && this.isShowAll
```

It will only activate if the "media type" is a rule. Other types could use similar logic which could then control how they would need to modify the request.

But, if you need anything more than "Show All", then this won't do. It will require a larger redesign. I have more options available for rules, I could have had use for a more detailed set of choices, but there's the space restraint to consider as well. For this to work on very small screens, things have to be dumbed-down. So, I decided that "Show All" would have to be good enough, that it would cover the most important choice.

All in all, I think this is the best I can manage in this situation. It does work, it's not super-flexible, but it's probably very easy to redesign later by somebody more capable in Vue/JS.

bundles/org.openhab.ui/web/src/api/core/serverSentEvents.gen.ts
```diff
@@ -79,7 +79,7 @@ export type ServerSentEventsResult<TData = unknown, TReturn = void, TNext = unkn
   >;
 };
 
-export function createSseClient<TData = unknown>({
+export const createSseClient = <TData = unknown>({
```
jsjames: When I run gernerate-rest-client, this generates the original code here vs. the change to export const - just curious if there is a version difference or something? any ideas?

bundles/org.openhab.ui/web/build/openhab-api-spec.json
```diff
@@ -957,7 +957,7 @@
                 "schema": {
                   "type": "array",
                   "items": {
-                    "$ref": "#/components/schemas/Template"
+                    "$ref": "#/components/schemas/RuleTemplateDTO"
```
jsjames, line 960: I presume the "dependent on another PR" is related to the changes here?  When they get merged into core and we generate a new openhab-api-spec, we should update the openhab-api-spec and regenerate as a separate PR so we can easily track and document the changes.

bundles/org.openhab.ui/web/src/api/core/serverSentEvents.gen.ts
Nadahar: There are quite a few changes in core from https://github.com/openhab/openhab-core/pull/5572 that needs to be in for the generated client to match. That said, I wouldn't expect this to be affected. The current generated client in this PR is now from an older version of core - maybe something that affects this method has changed in the meanwhile?

Anyway, I never planned to use the generation in this PR, because what I generate doesn't properly match what you guys generate. I don't know if it's a Windows <-> Linux issue or something else. The formatting doesn't match either.

So, I always planned that once https://github.com/openhab/openhab-core/pull/5572 has been merged, some of you guys can generate a new client, and then I just remove the generation commit from this PR. It feels "safer". But, I need to have it in here in the meanwhile, or the methods I use won't exist.

bundles/org.openhab.ui/web/build/openhab-api-spec.json
Nadahar: Yes, it depends on https://github.com/openhab/openhab-core/pull/5572 which makes quite a few changes to the REST API, both new and modified endpoints. Without it, this PR simply doesn't work.

Nadahar: The specific change here comes from this commit, but there are many others: https://github.com/openhab/openhab-core/pull/5572/changes/f64dea15f391ede0a8c04892e22d03b383e25142

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
jsjames: I agree with the discussion that it would be good to make the code-editor more general. My thought here is the following:

- move parseCode and generateCode to a separate file (outside of the code-editor). Those can then be imported and called from the parent component to render the code appropriately where there is more "expertise".
- create a separate code-editor toolbar component - this can then be instantiated and customized in the parent component as well. Actions such as re-rendering the code when DSL/YAML is clicked would then all be managed by the parent component.

If I can find some time tomorrow, I can try to work on this - unless someone else wants to take a crack at it.

Nadahar: The thing is that this PR already does quite a few changes to it, and it currently works. I'm not very eager to do all that over again, so I would prefer if such refactoring took place *after* this PR was merged. Then it would be easy for whoever did the refactoring to take into account the needs of this PR as well.

mherwege: @jsjames Wasn't code-editor supposed to be the more generalized component on top of script-editor? We seem to be building layer on top of layer here. I do see opportunity for a major refactoring but would indeed ask you to refrain from doing it until https://github.com/openhab/openhab-webui/pull/4234 and this PR get finalized and merged. It becomes very difficult otherwise. And you are certainly better positioned to do this refactoring then myself or @Nadahar. We both get around in js and vue but are far from experts.


> As far as I can tell, this can't be done with templates alone, `generateCode()` is called within the component without involving the parent component, and the request must be modified based on the selected language and this option. Changing this would mean a larger redesign using callbacks or something similar, so that the parent component could then modify the request. I'm not sure how that would look or how elegant it would be.

Sure, but have a look at the code in my PR. The controls you built into the template can trigger functions that set properties that get passed down to the component again. And these property changes down in code-editor can impact how things are done there. So even if you call `generateCode` from inside the code-editor, it can still look at code-editor properties to influence behaviour. That way you can make this more neutral inside the component.


> There's also the layout part. The "Select All" option is "a part of" the group with the language buttons. It scales and moved differently than it would have done if it were merely supplied as a template. I think the current behavior is preferable to it just "floating around" with the other buttons.

I struggled with that part myself. I needed a 3 button segmented control inside the code-editor to switch what is shown, and as I use your template approach, it doesn't align with the left media type selection button. This is especially noticable if switching buttons switches the component from editable to read-only, as the Revert button will disappear and the buttons are aligned badly again.
At the same time, there may be better solutions to this visualization problem. I didn't spend much time on it.

Nadahar:
@mherwege I can't figure out how to merge 4a0de997bd1f4e55d0bbe90496ae07253406d826 with this PR, since I have no idea why you're doing the things you do.

You seem to have changed your mind and gone back on the whole "send the parsedObject" to `generateCode()` idea - which is fine with me, but then you have added a `nextTick()` in there. I don't know why, I assume that it has something to do with letting reactivity do its thing. But, my changes to the same code are also substantial, and they clash, so I think we'll have to figure out together how to solve it.

I'm also not sure how the whole "array instead of object" thing will work with my code. I have changed the `parseCode()` to return a `Promise` that I wait for, which achieves the same I guess - by the time I'm done waiting for the `Promise`, reactivity will have done its thing.

mherwege:
> You seem to have changed your mind and gone back on the whole "send the parsedObject" to `generateCode()` idea - which is fine with me, but then you have added a `nextTick()` in there. I don't know why, I assume that it has something to do with letting reactivity do its thing. But, my changes to the same code are also substantial, and they clash, so I think we'll have to figure out together how to solve it.

Directly using the object without going through the calling component did not do the necessary pre-processing on the object to be parsed again. It also messed up the dirty state in the parent. And that logic sits in the parent for much of the object pre-processing. Therefore, I always use the object the editor gets from the parent through the property, but when switching it means the property needs to wait for nextTick to get updated.



> I'm also not sure how the whole "array instead of object" thing will work with my code. I have changed the `parseCode()` to return a `Promise` that I wait for, which achieves the same I guess - by the time I'm done waiting for the `Promise`, reactivity will have done its thing.

Before, the parser could only parse a single object. But the semantic tags are an array of objects. I explicitly needed an array as input, but made it so it did not change anything for all other uses of code editor. It wouldn’t change anything for what was already there, only allow my use case. There is no further change anywhere else in the code.

I am afraid if you return a Promise and use that inside the code, you risk creating issues with the usage in other places in the code.

Nadahar:
> Directly using the object without going through the calling component did not do the necessary pre-processing on the object to be parsed again. It also messed up the dirty state in the parent. And that logic sits in the parent for much of the object pre-processing. Therefore, I always use the object the editor gets from the parent through the property, but when switching it means the property needs to wait for nextTick to get updated.

I never used the "direct" functionality, but I had resolved this PR to work despite it. So, now I have to revert that, and I'm just trying to understand what I can and cannot do without clobbering something else.

The situation for rules is much more complicated than I dare say anything else. Which formats are available can change while the user is editing. There is a separate API endpoint that checks if a rule can be serialized to which format, which has to be called every time anything has changed.

All this used to work when I wrote it, before tags and sitemaps were created. The "magic trick" here is to try to make it work again, without breaking any of what has been added since.

I've had no need to use nexttick, and I assume that is because I use promises to wait for the various API calls along the way. The question now is how to reconcile all this. If you look at `switchCodeType()` in this PR, you'll see that it's quite a lot more complex. I basically have to ask the parent component which formats are still valid every time.



> There is no further change anywhere else in the code.

No, but this PR makes quite a few changes, and I have to make sure that it still works with those integrated.

> I am afraid if you return a Promise and use that inside the code, you risk creating issues with the usage in other places in the code.

The method didn't use to return anything, so I don't know why that would cause a conflict. It's not emitting a Promise, it's returning it. That allows waiting for it to complete.

Nadahar:
@mherwege I need to understand exactly what problem nexttick solves, so that I can verify that I can find a way to not break that.

Nadahar:
I think perhaps I found a way to combine the nexttick system with my code - it still needs to be tested, so a way to reproduce the situation it solved would be helpful.

mherwege:
> @mherwege I need to understand exactly what problem nexttick solves, so that I can verify that I can find a way to not break that.

Any property that is passed down and is based on something being watched or computed needs to wait for the next cycle. My object wasn't refreshes evzn when the recalculation had finished.

Nadahar:
> Any property that is passed down and is based on something being watched or computed needs to wait for the next cycle. My object wasn't refreshes evzn when the recalculation had finished.

Yes, I understand the principle, but what do I do to test that it "still works"? What was the situation where it failed?

Nadahar:
This looks a bit strange, is this something my changes have caused?

image

Nadahar:
There's something else very strange going on in the tag editor - and I can't understand how that can be related to any of my changes. If I try to create a custom tag, and give it a name, I can only type one letter at the time before it shows "Nothing selected", and there's a lot of red in the browser console:

EditTagName

edit: I think the errors in the browser console might be unrelated, I visited the block editor just before entering the tags, and I suspect that the errors came from the blocks view.

Nadahar:
However, trying to save my new tag fails:

image

Nadahar:
I think I've done what I can to reconcile this PR with other changes that have happened in the meanwhile. As far as I know, it's ready for testing and review.

@mherwege It would be very nice if you could check this out locally and give the things you've made a spin to check that nothing has been broken. I've tried to do it, but I'm not sure that I know what to look for.

mherwege:
> This looks a bit strange, is this something my changes have caused?
> 
> image

This looks correct. The segmented buttons and icons are intentional (and visible/not visible depending on screen width).

mherwege:
> There's something else very strange going on in the tag editor - and I can't understand how that can be related to any of my changes. If I try to create a custom tag, and give it a name, I can only type one letter at the time before it shows "Nothing selected", and there's a lot of red in the browser console:
> 
> EditTagName
> edit: I think the errors in the browser console might be unrelated, I visited the block editor just before entering the tags, and I suspect that the errors came from the blocks view.

That is indeed strange and not intended. I am also not sure it is because of your changes. All of the focus and testing has been on the code view in my PRs, not on editing in the treeview, apart from changes showing up  there properly. I don’t know when and where that broke if it did.

EDIT: Did you see it has to start with a capital?

Nadahar:
> That is indeed strange and not intended. I am also not sure it is because of your changes. All of the focus and testing has been on the code view in my PRs, not on editing in the treeview, apart from changes showing up there properly. I don’t know when and where that broke if it did.

It would be very strange if it's related to my changes, I haven't changed anything on that page. What seems to happen is that some kind of "filtering" being applied when I type, that filters away the tag I'm editing. But, it might be a side effect of the validation failing, perhaps the exact action taken when the validation fails should be inspected closer.

> EDIT: Did you see it has to start with a capital?

Yes, I realize that once I finally read the error message, however I didn't at first, since I've seen an almost identical message before, and took it for an error that was part of whatever else was going on.

That said, I found it peculiar that a capital letter is required there, and that's because of the general ID/UID/Name/Label confusion that is found throughout OH. Since they are used "without much consistency", I kind of assume that the first is the ID and the second is the name, which is why I wrote the first in lower-case. I finally realized that here, it's a new twist: `Name` is the *English* name *and* ID, while `Label` is the translation. Elsewhere, "Name" and "Label" are used somewhat interchangeably, so I think of them as synonyms in the OH sphere.

I guess this all has its history...

mherwege:
I now opened a test with your branch.

First, I can confirm editing in the treeview is broken and is not because of your PR. It looks like it broke with my PR (the commit before still works). I will look into that.

Second I tried to replicate why I needed `nextTick`. I couldn't with the final state of the PR as it was merged. I was pretty sure it had to do with editing something in the code editor and switching to the treeview or another code visualization (edit/custom/all), and losing the dirty state when doing that. But I couldn't replicate it. I am using another old laptop for this test, running linux, while I had done the development and test on a Windows laptop. But I didn't expect that would make a difference, except in timings maybe, which could explain why it worked on my development machine and not this laptop. That's a long shot though.

I can say that the sitemap editor still works as expected with your PR. I am at a loss why there are so many problems with the semantics editor that I must have overlooked when working on it.

Nadahar:
> Second I tried to replicate why I needed `nextTick`. I couldn't with the final state of the PR as it was merged. I was pretty sure it had to do with editing something in the code editor and switching to the treeview or another code visualization (edit/custom/all), and losing the dirty state when doing that. But I couldn't replicate it. I am using another old laptop for this test, running linux, while I had done the development and test on a Windows laptop. But I didn't expect that would make a difference, except in timings maybe, which could explain why it worked on my development machine and not this laptop. That's a long shot though.

I have kept the nexttick logic though: "My" code works with promises, so it doesn't have that timing problem, as it waits for promises to complete, which probably means that reactivity has the time to do what it needs. What I did was make it so that *if* there's a promise to wait for, it will wait for the promise. If not, it will instead wait for nexttick. I think it should hopefully work for "everybody".

Nadahar:
@florian-h05 @jsjames I hope you guys have this one in mind, now that the core PR has been merged.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -96,22 +106,25 @@
...
     margin-left 4px
 
-@media (max-width: 475px)
-  .code-editor-toolbar
-    --f7-toolbar-height var(--f7-tabbar-labels-height)
-    font-size var(--f7-tabbar-label-font-size)
-    .button-label
-      display none
-    .toolbar-inner
+.code-editor-toolbar.toolbar-narrow
```
jsjames, line 109: did the @media not work here?

Nadahar: It works, but it's very difficult to get everything "in sync" when you have to keep editing pixel values all over the place. Using `toolbar-narrow` instead allows several things to react synchronously. There already was one such problem which I "cheated my way out of" by changing line 393 in `semantic-tags-edit.vue`. The hiding of the icons and the hiding of the text wasn't "in sync", which lead to buggy behavior at certain widths.

These things can be organized many different ways, but I believe using classes is much better than hard-coding pixel widths. In any case, I spent more than two days just getting the toolbar to resize correctly under different circumstances in the "rule code view", so I'm not very eager to start messing with it again. Thus, when I had to resolve the conflicts from the things that had been changed since I made the code, I tried to do so in a way that would break as little as possible. I therefore chose to stick with the class-based approach, which is the one I used when doing all my testing.

If you guys want to arrange it differently later, that's great, but I'd really prefer if I didn't have to do any more layout work than I've already done on this PR.

mherwege: > There already was one such problem which I "cheated my way out of" by changing line 393 in `semantic-tags-edit.vue`. The hiding of the icons and the hiding of the text wasn't "in sync", which lead to buggy behavior at certain widths.

The different widths for showing/not showing text/icons were on purpose there. The text was relatively long. Very small widths - only icons, intermediate - only text, large - text and icons. If you standardized that, you may have broken that.

Nadahar: No, I didn't break that - that wasn't how it worked when I tested, because I think the "narrow width" had been modified after you made it, which is my point. So, what it did was: icon only -> icon and text -> text only -> icon and text. It was the "first" icon and text that bugged, the buttons were drawn over each other. It's very easy to replicate, just undo the change I did on line 393 and see for yourself.

When pixel widths are set in various places, any adjustment must take all these into account, which can be hard to do.

mherwege: The first icon and text should not exidt at all if the widths I defined are correct and have the same boundaries. So that's where it would go wrong.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -11,18 +11,28 @@
...
+          outline
+          small
+          :key="type"
+          :active="uiOptionsStore.codeEditorType === type"
+          :tooltip="`Show code as ${type}`"
+          @click="switchCodeType(type)">
+          {{ type }}
+        </f7-button>
+      </f7-segmented>
+      <f7-checkbox
```
mherwege, line 28: Why have this here and not in the extra controls in the calling component (additional-panel-controls)? This is very specific and would be more logical in my view.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -188,15 +224,41 @@ export default {
...
       const sourceMediaType = MediaType.JSON
       let targetMediaType = this.mediaTypes[codeType]
-      targetMediaType = targetMediaType.split('+')[0] // remove the +tag, +thing or +item suffix, if present
+      if (this.isObjectEmpty && this.emptyMediaTypeTemplates) {
+        const emptyTemplate = this.emptyMediaTypeTemplates[targetMediaType]
+        if (emptyTemplate) {
+          let emptyCode = typeof emptyTemplate === 'function' ? emptyTemplate() : emptyTemplate
+          this.code = emptyCode
+          this.originalCode = emptyCode
+          this.dirty = false
```
mherwege, line 233: This cannot be right. This resets the dirty flag on the calling component, making it lose the reference for changes that were done somewhere else than in the code editor.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
mherwege, line 267: It probably comes from here, and this is not right either. The dirty flag should only be reset in the parent component, not in here, as you may not be generating code for the original object, but an already modified object.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: Yes, I found out, that's why I had to adjust it. My point is that if `:not(.toolbar-narrow)` had been used instead of a pixel value, they would never have been able to "lose sync".

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: This goes back to the same discussion as before, with the shared used of `code-editor` and the lack of generalization. I did my best to generalize this, I've made it an option that can be used by simply enabling it. It doesn't mean the same as the "show all" in semantic tags, it's not about filtering the elements, it's about showing the hidden YAML nodes. To me, this was something I imagined would be useful for other use cases as well, which is why I did all the extra to generalize it. The text, "Show all", is really too general, so it can be interpreted in many ways, but I couldn't find anything else that was short and would work.

This is closely tied to the chosen format, it is a "sub option" of the YAML option. It thus belongs/should be attached to the YAML select button, so I had to find a way to make them behave "together" layout wise. It would have little meaning of it just "floated" among the other additional panel controls with no affiliation to the target format button. The "alternative" as I see it, would have been to make two different YAML buttons, one "normal" and one "include all", but it would just have ended up requiring more space.

In my mind, this could be used e.g. for Things for "show default values" so that without it, default values were hidden, and with "show all" they were included. I also use it to show for example empty collections.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: I must look at it closer to remember the details here, but I'm not so sure that it does what you think. I think `this.dirty` points to the code editors dirty flag, not that of the parent component. In `rule-edit` we have essentially two dirty flags, which are watched and handled like this:

```vue
    ruleDirty: function () {
      this.dirty = this.ruleDirty || this.codeDirty
    },
    codeDirty: function () {
      this.dirty = this.ruleDirty || this.codeDirty
    }
```

Whenever one of them changes, it affects the "total dirty status", but they are independent and can each toggle their dirty status as required.

I think `this.dirty` here only means the "dirty status" of the code editor, and that it probably only really affects the "Revert button".

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: The same applies here, I don't think it has anything to do with the flag in the parent component.

mherwege (commented):
> This cannot be right. This resets the dirty flag on the calling component, making it lose the reference for changes that were done somewhere else than in the code editor. 

Indeed, I am wrong here. I have seen some strange behaviour with the semantic editor losing its dirty state in some cases, so that's where I got off track.

mherwege (commented):
> This goes back to the same discussion as before, with the shared used of `code-editor` and the lack of generalization. I did my best to generalize this, I've made it an option that can be used by simply enabling it. It doesn't mean the same as the "show all" in semantic tags, it's not about filtering the elements, it's about showing the hidden YAML nodes. To me, this was something I imagined would be useful for other use cases as well, which is why I did all the extra to generalize it. The text, "Show all", is really too general, so it can be interpreted in many ways, but I couldn't find anything else that was short and would work.

> 
> This is closely tied to the chosen format, it is a "sub option" of the YAML option. It thus belongs/should be attached to the YAML select button, so I had to find a way to make them behave "together" layout wise. It would have little meaning of it just "floated" among the other additional panel controls with no affiliation to the target format button. The "alternative" as I see it, would have been to make two different YAML buttons, one "normal" and one "include all", but it would just have ended up requiring more space.
> 
> In my mind, this could be used e.g. for Things for "show default values" so that without it, default values were hidden, and with "show all" they were included. I also use it to show for example empty collections. 

I understand that. But the 3 button choice in the semantics editor is exactly the same type of situation (switching between what code to generate or not, switching read only vs read/write at the same time), only I need 3 options, and not a checkbox on/off. So this solution is not general at all. And what I did with the semantics editor shows it can be done differently.

I agree that the styling may be less optimal as the controls don't stick to the left language choice button. I was just hoping there is another way to solve that. I don't see it as a priority though.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
mherwege: > This cannot be right. This resets the dirty flag on the calling component, making it lose the reference for changes that were done somewhere else than in the code editor.

Indeed, I am wrong here. I have seen some strange behaviour with the semantic editor losing its dirty state in some cases, so that's where I got off track.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
mherwege: I understand that. But the 3 button choice in the semantics editor is exactly the same type of situation (switching between what code to generate or not, switching read only vs read/write at the same time), only I need 3 options, and not a checkbox on/off. So this solution is not general at all. And what I did with the semantics editor shows it can be done differently.

I agree that the styling may be less optimal as the controls don't stick to the left language choice button. I was just hoping there is another way to solve that. I don't see it as a priority though.

Nadahar: > And what I did with the semantics editor shows it can be done differently.

Of course it can, but for rules, there are also a lot of others controls to show as well:

CodeEditToolbarResize2

As said before, ideally I'd need up to 4 different options for YAML, but I couldn't find a way to solve it "layout-wise", so I just dropped the "less important variants" here. They are available when you use the "Copy" button, so I figured that would have to do.

I'm not saying that any of this is ideal, I'm saying "it's working as it is now", let's use this for now and then refactor the whole thing to make it more general, if it can at all be done. If not, consider not sharing the code-editor, and use customized versions for "complicated cases".

mherwege:
I did a little test with a rule that has a single trigger and DSL script action defined in the UI. Switching from the UI to code worked as expected. I then did a small change in the code from the YAML code view (changes a var declaration to val, but the details don't matter I think). I switched from YAML to DSL and the change was preserved and the dirty flag remained.
I switched back to the design view and the dirty flag remained. Very good. I did not save the rule.
I then clicked on the `Execute an inline script` link for the DSL script. Before opening it, the modified rule was saved and the dirty flag was gone. I did not expect that. I would have expected the dirty flag to remain without a save, or a refusal to open the script as the rule was dirty. But not an unattended save.

Nadahar:
> I then clicked on the `Execute an inline script` link for the DSL script. Before opening it, the modified rule was saved and the dirty flag was gone. I did not expect that. I would have expected the dirty flag to remain without a save, or a refusal to open the script as the rule was dirty. But not an unattended save.

This is nothing new and nothing I've made, it's been like this all along. I guess the underlying reason is that rules are executed by the rule engine, and it works with registered rules only. So, there's no way to run a "temporary rule". The decision to just save and run without informing the user is another thing, there might be room for improvement there. But, I don't think it's viable to actually "run" an unsaved rule.

edit: I guess it would be easy to just show a message saying that the rule must be saved before it can be run, instead of just saving it... would that be better or worse than the current behavior?

edit2: I was too quick to read. You're talking about the automatic saving when opening the script action. It's not new either, and I agree that it's not ideal. If you try to open the script action before you've given the rule a name, you'll even get an error. The underlying reason here is, I guess, that you're really opening a new page, that will retrieve the data using the API. There is no mechanism to transfer the "script content" back and forth between "rule edit" and "script edit".

That would probably be possible to do, but it's not something I've attempted to address. It's been like this for as long as I can remember.

mherwege:
> > I then clicked on the `Execute an inline script` link for the DSL script. Before opening it, the modified rule was saved and the dirty flag was gone. I did not expect that. I would have expected the dirty flag to remain without a save, or a refusal to open the script as the rule was dirty. But not an unattended save.
> 
> This is nothing new and nothing I've made, it's been like this all along. I guess the underlying reason is that rules are executed by the rule engine, and it works with registered rules only. So, there's no way to run a "temporary rule". The decision to just save and run without informing the user is another thing, there might be room for improvement there. But, I don't think it's viable to actually "run" an unsaved rule.

The `Execute an inline script` is not running the script! It is taking you to the script editor. Nothing gets executed at this point. It just saves without any warning or message modifications you had done in the code view.
This might have been OK before as you would not edit the script in the code view, but only the the script editor. It is not now, as you may have edited the script and it should not be saved without doing a save.

I would also think you should not be able to run the rule from the UI if you are modifying it and you have not saved it yet. I didn't check if that's the case. Again, it should not automatically save (and I am not sure it does, except for the case I identified). I did not check running the rule, I only checked going to all the screen of the rule editor without saving a modified rule.

Note that, if you modify the script in the script editor and you back out of it without saving, you get a warning and you lose the changes when not saving. There should be some consistency the other way around as well. It may not be new, but you could not modify the script from outside of that script editor before.

Nadahar:
> The `Execute an inline script` is not running the script! It is taking you to the script editor. Nothing gets executed at this point. It just saves without any warning or message modifications you had done in the code view. This might have been OK before as you would not edit the script in the code view, but only the the script editor. It is not now, as you may have edited the script and it should not be saved without doing a save.
> 
> I would also think you should not be able to run the rule from the UI if you are modifying it and you have not saved it yet. I didn't check if that's the case. Again, it should not automatically save (and I am not sure it does, except for the case I identified). I did not check running the rule, I only checked going to all the screen of the rule editor without saving a modified rule.

I edited my above post. This "issue" applies to both running rules and editing their actions in the script editor. I absolutely agree that it's far from ideal, but it's not of my making. I haven't attempted to fix it, I don't know how complicated it would be.

> t may not be new, but you could not modify the script from outside of that script editor before.

That's not true. You could modify the script in the "old" Code view as well. The situation is essentially identical to how it was, but I agree that it's not ideal. I have however tried to get this ready for 5.2.0, not to follow up on any potential for improvement.

mherwege:
> > The `Execute an inline script` is not running the script! It is taking you to the script editor. Nothing gets executed at this point. It just saves without any warning or message modifications you had done in the code view. This might have been OK before as you would not edit the script in the code view, but only the the script editor. It is not now, as you may have edited the script and it should not be saved without doing a save.
> > I would also think you should not be able to run the rule from the UI if you are modifying it and you have not saved it yet. I didn't check if that's the case. Again, it should not automatically save (and I am not sure it does, except for the case I identified). I did not check running the rule, I only checked going to all the screen of the rule editor without saving a modified rule.
> 
> I edited my above post. This "issue" applies to both running rules and editing their actions in the script editor. I absolutely agree that it's far from ideal, but it's not of my making. I haven't attempted to fix it, I don't know how complicated it would be.

If you try editing the trigger of a dirty rule, it will not save it. So at least, it is even inconsistent there. Editing the trigger in the YAML or DSL view will actually show the updated trigger and allow you to edit it in the design view without saving.

If you say that behaviour has not changed, it is a bug in my view, maybe not caused by this PR.

EDIT: tested on 5.1.4 and I can confirm the bug exists there as well.

Nadahar:
> If you try editing the trigger of a dirty rule, it will not save it. So at least, it is even inconsistent there. Editing the trigger in the YAML or DSL view will actually show the updated trigger and allow you to edit it in the design view without saving.
> 
> If you say that behaviour has not changed, it is a bug in my view, maybe not caused by this PR.

I'm not sure I follow here now. You can edit anything that can be changed in `rule-edit.vue` and `code-editor.vue` without it affecting saving/dirty state, but this isn't the case for `script-edit.vue`.

While `code-editor.vue` is loaded as a "child component" of `rule-edit.vue`, `script-edit.vue` is not, and that is the real cause for the difference in behavior. `script-edit.vue` is loaded as a separate page, and the router is involved in returning to the rule you came from. So, you're essentially switching between pages, and `script-edit.vue` has its own API calls to retrieve the data - which is why it must be saved first.

> tested on 5.1.4 and I can confirm the bug exists there as well.

This has been like this for as long as I've been using MainUI. It's annoying an inconsistent, yes, but I'm not sure I would call it a bug, but rather a "design flaw".

Nadahar:
I'm sorry for misreading your initial report, the thing is that I'm aware of this "issue" for both editing scripts and running rules, I've not introduced/made any of them, but I know that they are there. Therefore, when I saw "execute", my mind jumped to the wrong case.

I don't think any of these preexisting "challenges" should block this PR, and I'm tired of this PR and just want to get it merged. I'm not sure I'll undertake anything this large in MainUI again in the future, the constantly changing landscape, my lack of knowledge and all the existing "challenges" makes the experience a bit too frustrating for me.

edit: I just tested editing a rule and then running it, and as I thought, it does the exact same thing - saves it without informing the user, and then runs the rule.

mherwege:
> I'm sorry for misreading your initial report, the thing is that I'm aware of this "issue" for both editing scripts and running rules, I've not introduced/made any of them, but I know that they are there. Therefore, when I saw "execute", my mind jumped to the wrong case.
> 
> I don't think any of these preexisting "challenges" should block this PR, and I'm tired of this PR and just want to get it merged. I'm not sure I'll undertake anything this large in MainUI again in the future, the constantly changing landscape, my lack of knowledge and all the existing "challenges" makes the experience a bit too frustrating for me.
> 
> edit: I just tested editing a rule and then running it, and as I thought, it does the exact same thing - saves it without informing the user, and then runs the rule.

I am far from good at coding in the UI as well, and it indeed is sometimes frustrating. I wanted to have a look at this in more detail, but I have a lack of time, meaning I look at, and try, parts at different points in time, in between, which is not the most efficient way of looking at this.
Don't feel like any feedback I am giving is meant to be negative at all. I am not a maintainer, and won't make a final approval decision. I want to get this accross as much as you do.

Nadahar:
> I am far from good at coding in the UI as well, and it indeed is sometimes frustrating. I wanted to have a look at this in more detail, but I have a lack of time, meaning I look at, and try, parts at different points in time, in between, which is not the most efficient way of looking at this.
> Don't feel like any feedback I am giving is meant to be negative at all. I am not a maintainer, and won't make a final approval decision. I want to get this accross as much as you do.

As I've said, I completely agree with you that the automatic save behavior (and the fact that you can't edit the "script" dynamically/without saving is "less than ideal". My point is just that these things aren't really related to or affected by this PR. It was "bad" before, and it continues to be "bad".

I think it should be addressed in the future, but I don't want to start messing with that in this PR.

I think perhaps refusing to run the rule if it's dirty might be a better solution for "run". Regarding the editing of the script, I think `script-edit` would have to be loaded as a "child component" instead of as a separate, routable page, but I have no idea how big a project that would be. It might be relatively simple, or it might require a major rewrite. I guess a way could be found to not call the APIs if certain props are provided, but I'm very unsure of how to handle the rendering - would it have to be a "popover" or "sheet" that is placed on top of `rule-edit`? I don't know.

Nadahar:
@florian-h05 @jsjames It would be very useful to get this merged before the next milestone build, given that the core PR has been merged. We need this "last chance" for it to be tested before release.

Nadahar:
@florian-h05 It's a real shame this won't make it into the milestone build. I would consider merging it and let us deal with any problems swiftly, instead of having the 5.2.0 release be the "first release" of this UI change.

florian-h05:
We can merge it right now, I’ll be back in action on Wednesday, which gives us a short time interval to fix any regressions or newly introduced bugs. 
Just to note I’m unable to review it and also can’t test it at the moment.

Nadahar:
> We can merge it right now, I’ll be back in action on Wednesday, which gives us a short time interval to fix any regressions or newly introduced bugs. Just to note I’m unable to review it and also can’t test it at the moment.

I have tested it quite a lot, and @mherwege has also tested it some. I've in fact been running this version locally for quite a long time. I can't guarantee that there won't be any issues, but I doubt it's anything major.

kaikreuzer:
I've just requested a copilot review - if nothing critical comes up, I will merge and smaller things can still be fixed as a follow-up.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -243,22 +306,24 @@ export default {
...
                 onSuccessCallback(object)
               }
             } else {
-              this.$emit('parsed', object[0])
+              this.$emit('parsed', object[0], params)
               if (onSuccessCallback) {
                 onSuccessCallback(object[0])
               }
             }
+            return Promise.resolve()
```
kaikreuzer, line 320: This resolves `parseCode()` as successful after emitting `parsed`, but the parent handler can reject the parsed object (`rule-edit.updateRule()` returns `false` for invalid edits like UID changes or `source`/`sourceType`). Vue event handler return values are not propagated here, so save/tab-switch callers can continue, clear `codeDirty`, and save the old or partially updated rule while the editor still contains rejected code. Please make parent validation failure propagate back into `parseCode()` before treating the parse as successful.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -856,56 +1261,86 @@ export default {
...
+    },
+    copyRuleDefinitionToClipboard(type, serializationOption) {
+      if (!this.rule) {
+        return
+      }
+      const mediaType = type === 'DSL' ? 'application/vnd.openhab.dsl.rule' : 'application/yaml'
+      const progressDialog = f7.dialog.progress(`Loading '${this.rule.name}' ${type || 'YAML'} definition...`)
+      create(
+        {
+          serializationOption: serializationOption || undefined,
```
kaikreuzer: The generated `create()` client does not map `serializationOption` for `/file-format/create`; the query parameter is named `ruleSerializationOption`. As written, all single-rule YAML export choices here (`Include all`, `Stub only`, `Strip template`) are silently dropped and the backend uses `Normal`. Please pass `ruleSerializationOption` or use the route-specific client so the chosen option reaches the server.

kaikreuzer (changes requested):
I found two issues that should block merging this as-is.

kaikreuzer:
I'm tempted not to merge this yet - with 4 high copilot findings and 2 critical ones by the help of Codex, I think this should first be addressed. If we all agree that this should get into 5.2, I'd be fine to do an exception from the feature freeze for this as well after the milestone build.

Nadahar:
I'll get on these right now.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -188,15 +224,41 @@ export default {
...
       const payload = {}
       if (Array.isArray(this.object)) {
         payload[this.objectType] = this.object
       } else {
         payload[this.objectType] = [this.object]
       }
       this.$oh.api
-        .postPlain('/rest/file-format/create', JSON.stringify(payload), null, sourceMediaType, { accept: targetMediaType })
+        .postPlain(
+          `/rest/file-format/create${params.size ? '?' : ''}${params.toString()}`,
```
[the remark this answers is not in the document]
Nadahar: How come it works then? There's no doubt that the parameter is sent and received by the other side. This is what my debugger shows:

image

image

The effect of the parameters are obvious, so it "has to" work. Whether `params.size` is `undefined` or `0` makes no difference.

Also, MDN disagrees: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/size

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -856,56 +1261,86 @@ export default {
...
+          fileFormat: {
+            rules: [this.rule]
+          }
+        },
```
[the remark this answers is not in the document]
Nadahar: This must be some odd consequence of the autogenerated REST client, I know that it used to work. Maybe the name was changed in the core PR at one stage? Anyway, fixed and verified to be working.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -946,21 +1381,19 @@ export default {
...
-    opaqueModulesType() {
-      const modules = this.opaqueModules
-      if (!modules || !modules.length) return undefined
-      // "Opaque modules" implies that the rule is created through JSR223.
-      // The assumption is therefore that all opaque module types are of the same type/scripting language.
-      return modules.find((m) => m.configuration?.type)?.configuration?.type
-    },
     opaqueModules() {
       if (!this.rule) return []
       return [...(this.rule.actions || []), this.rule.triggers || [], this.rule.conditions || []].filter((m) => this.isOpaqueModule(m))
```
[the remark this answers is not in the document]
Nadahar: Opaque triggers and conditions doesn't actually exist, which is why this has gone unnoticed. It should however in principle work if they are created in the future. Fixed.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -946,21 +1381,19 @@ export default {
...
     },
+    hasSharedContextModule() {
+      return this.sharedContextModules.length > 0
+    },
+    sharedContextModules() {
+      if (!this.rule) return []
+      return [...(this.rule.actions || []), this.rule.triggers || [], this.rule.conditions || []].filter((m) =>
```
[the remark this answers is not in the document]
Nadahar: Shared context triggers and conditions doesn't actually exist, which is why this has gone unnoticed. It should however in principle work if they are created in the future. Fixed.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -856,56 +1261,86 @@ export default {
...
+            },
+            onError: () => {
+              f7.dialog.alert(`Error copying rule ${type || 'YAML'} definition to the clipboard`, 'Error')
+            }
+          })
+          this.copyPopupOpened = false
+        })
+        .catch((error) => {
+          progressDialog.close()
+          console.error('Failed to generate rule definiton', error)
```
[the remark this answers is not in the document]
Nadahar: Fixed

bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
```diff
@@ -701,6 +888,173 @@ export default {
...
+        }
+      } else {
+        this.dslCopyOk = []
+        this.dslCopyErrors = []
+        console.warn('Failed to check DSL serialization support:', yamlResult.reason)
+      }
+
+      this.copyPopupOpened = true
+      console.debug("Can't serialize to YAML:", this.yamlCopyErrors)
+      console.debug("Can't serialize to DSL:", this.dslCopyFailures)
```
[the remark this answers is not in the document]
Nadahar: I blame VS Code for not finding all references when renaming. Fixed.

bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
```diff
@@ -701,6 +888,173 @@ export default {
...
+                `Error copying rule ${type || 'YAML'} definition${this.selectedItems.length === 1 ? '' : 's'} to the clipboard`,
+                'Error'
+              )
+            }
+          })
+          this.copyPopupOpened = false
+        })
+        .catch((error) => {
+          progressDialog.close()
+          console.error(`Failed to generate rule definiton${this.selectedItems.length === 1 ? '' : 's'}`, error)
```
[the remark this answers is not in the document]
Nadahar: Fixed

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -264,31 +261,208 @@
...
+                color="blue"
+                tooltip="Copy YAML, where only the template and the configured template parameters are included, to clipboard"
+                @click="copyRuleDefinitionToClipboard('YAML', serializationOptions.STUB)">
+                Rule Stub Only
+              </f7-button>
+              <f7-button
+                v-if="rule.templateUID && rule.templateState === 'instantiated'"
+                fill
+                color="blue"
+                tooltip="Copy YAML, where the template and the configured parameters are removed, resulting in an indentical but fully independent rule, to clipboard"
```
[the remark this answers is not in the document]
Nadahar: Fixed

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: How did you find that? It should be said that the whole flow of the `parseCode()` and `generateCode()` is "too simplistic" to handle all that it needs to handle, and the whole thing is being pushed beyond its limits here, making intricate bugs like this likely. We have already discussed that this whole thing should be refactored once all these PRs are in (this is the last of them), to make for more sane handling of the complexities.

But, unless I'm mistaken, this bug has existed before I introduced the `Promise`, and I'm pretty sure that if tested, you'll find the same issue in the current code. I only use the `Promise` to wait for the operation to be complete, the whole solution with emitting the result and in parallel returning a `Promise` is questionable. But, since `rule-edit` needs to contact the API to check the results of a change, I needed to find some way to wait for the API call to complete before proceeding.

I'm not opposed to make the parent validation propagate back, I'm just not sure if I see how to do it without breaking something else in this "brittle construct".

Any suggestions are appreciated.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
Nadahar: This has already been addressed in one of the copilot comments. For some reason it has been renamed at one stage, it's a bit hard to catch with no type-checking and an autogenerated API where things change without notice. I did use to work though, because I've tested this before.

Nadahar:
@kaikreuzer I've addressed all but one of the Codex findings. The threshold for "critical" is quite low, it was mostly typos, a logging failure, and the one parameter that has changed name so that the generated output wouldn't respect the extra options.

The one remaining finding I must chew some more on, it's not apparent to me how to solve it.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
Nadahar: I'm not sure that I can wrap my head around the details here. The comment doesn't actually make sense to me. I can't figure out how the scenario described above can happen, it seems to be mixing different scenarios. Switchtab has its own handling for failure, it doesn't save anything, but shows the error and aborts the switch, so that the user must fix the error or revert the change. This logic is "inherited" from Things, Items etc. and isn't special for rules.

The `save()` method doesn't have any callbacks, it looks like this:

```js
          await editor.parseCode(undefined, undefined, { editorType: this.uiOptionsStore.codeEditorType, showAll: editor.isShowAll })
```

As can be seen, there are no callback functions in this case, so they cannot "fail". I think maybe what the comment above points to is that any failure resulting from processing the `emit` won't be reflected in the `Promise`, but how can it? These are independent "code paths". If some of the changes are rejected, then they won't be part of the saved object, but the user will be presented with an error dialog informing that the change was rejected.

In case of the tab switching, I don't think it will matter if the code that is left behind is "wrong", because it will be generated again from the actually "accepted" rule if the user later switches back to the "Code" tab.

As stated before, this whole thing is a bit messy because it has to handle situations more complex than it was designed to. It should be refactored, but that's both beyond the scope of this PR and beyond my JavaScript/Vue abilities. I can't find anything to actually rectify as this is now.

Nadahar:
As written above, I don't know what to do about the one Codex comment, I'm not even sure if there's a failure situation that can actually occur there.

bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
```diff
@@ -220,6 +237,104 @@
...
+                color="blue"
+                tooltip="Copy YAML, where only templates and configured template parameters are included, to clipboard"
+                @click="copyRuleDefinitionsToClipboard('YAML', serializationOptions.STUB)">
+                Rule Stub Only
+              </f7-button>
+              <f7-button
+                v-if="yamlCopyCanStrip"
+                fill
+                color="blue"
+                tooltip="Copy YAML, where templates and configured parameters are removed, resulting in indentical but fully independent rules, to clipboard"
```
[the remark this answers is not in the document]
Nadahar: Fixed.

bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
```diff
@@ -701,6 +888,173 @@ export default {
...
+          this.dslCopyErrors = reasons
+        } else {
+          this.dslCopyOk = []
+          this.dslCopyErrors = []
+          console.warn('Failed to check DSL serialization support, received an empty result')
+        }
+      } else {
+        this.dslCopyOk = []
+        this.dslCopyErrors = []
+        console.warn('Failed to check DSL serialization support:', yamlResult.reason)
```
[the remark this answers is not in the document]
Nadahar: Fixed.

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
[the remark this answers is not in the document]
Nadahar: I still don't know why the parameter has changed name in the generated rest client, this is the definition in the endpoint:

```java
            @DefaultValue("Normal") @QueryParam("serializationOption") @Parameter(description = "Decides what to include in serialized rules") RuleSerializationOption option,
```

So, the name is `serializationOption` and the type is `RuleSerializationOption`. How that ends up with a parameter named `ruleSerializationOption` is a mystery to me.

Nadahar:
I've tried to reproduce the scenarios described in the Codex finding. When switching tab, I don't think there is a problem, because, as I said, the code will be regenerated when you switch back:

UpdateFailureSwitch

When you save, there is a "problem", because you're left with an editor state that is "out of sync". But, you're still informed about the problem, and the invalid rule isn't saved. The only "real problem" is that the dirty flag is cleared. Because, if you switch to another tab, things are corrected, but if you stay in the code editor, it isn't. But, if you make further edits and tries to save again, you will still receive the same error message until the problem has been fixed.

The bug is that the UI "indicates", with the dirty indicator, that it has been saved when it has not. If it just stayed "dirty", all would be OK as far as I can tell. I don't know how to achieve that though, because these two code paths are completely independent of each other.

UpdateFailureSave

Nadahar:
I've "solved" the save situation in the same way that it is done for Things: I simply regenerate the code after every save. That makes it behave identical to how the Thing editor behaves - it still somewhat misleading shows a message about the thing/rule being "updated", but it overwrites the content of the editor with what the "actual rule" looks like:

UpdateFailureSave2

This is how the same operation looks for a Thing:

UpdateThingFailureSave

kaikreuzer (approved):
Thanks for your swift reaction, @Nadahar - I guess it should be good to be merged now.
If anything weird comes up, I am sure that one of our Main UI maintainers can help on it in future!

Nadahar:
@florian-h05 The situation with the API parameter that had changed name in the generated client shows that it can be a bit fragile. Is there any way to make TypeScript "detect" and make an error in such situations?

lolodomo:
I just made few tests and looks good.
Only point I can mention: on an Android phone, a way to scroll is missing when too many errors are displayed in the "Copy Rule File Definition" window. On my PC, it is ok, a scrollbar is present.

florian-h05:
> The situation with the API parameter that had changed name in the generated client shows that it can be a bit fragile. Is there any way to make TypeScript "detect" and make an error in such situations?

It should do so for TypeScript code. However, most of the existing UI code is plain JavaScript, where TypeScript doesn’t check for such errors. James and me have refactored a lot of code from JS to TS, but there is still much to do. You can move from JS to TS by adding lang=„ts“ to the <script> tag, ideally you then also refactor from options to composition api as this improves the overall typescript experience.

Nadahar:
> Only point I can mention: on an Android phone, a way to scroll is missing when too many errors are displayed in the "Copy Rule File Definition" window. On my PC, it is ok, a scrollbar is present.

I'll look and see what I can figure out.

Nadahar:
@lolodomo I think I fixed it: https://github.com/openhab/openhab-webui/pull/4292

lolodomo:
I also created issue #4293 (which is not critical at all).

PR #4210, opened 2026-05-20
Sitemap edit: Various fixes

mherwege:
This solves a few issues in the sitemap editor identified here: https://github.com/openhab/openhab-core/issues/5007#issuecomment-4485207602

- Don't automatically sort the Buttons in a Buttongrid when loaded or saved. The proposed solution is to not do that, but provide a sort button on the Button widget next to the up/down arrows. Instead of moving up/down, it will sort all sibling Buttons by row/column/visibility rule.
- Allow empty format for the label. This will make sure the format from the item itself is not applied and no state is shown. This was always removed before, and therefore the UI was always showing the formatted state if there was a format defined on the item. This is now solved by providing an extra switch 'No state'.

There was a bug when opening the add widget selection action sheet on narrow screens. This is fixed.

There was also a bug in the drag/drop logic. When dropping a widget in a not-allowed parent, it visually showed it could not be dropped, but it was still dropped and not reverted. This has also been fixed.

### DIFF
--- a/bundles/org.openhab.ui.basic/.classpath
+++ b/bundles/org.openhab.ui.basic/.classpath
@@ -30,16 +30,5 @@
 			<attribute name="annotationpath" value="target/dependency"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="src" path="target/generated-sources/annotations">
-		<attributes>
-			<attribute name="optional" value="true"/>
-		</attributes>
-	</classpathentry>
-	<classpathentry kind="src" output="target/test-classes" path="target/generated-test-sources/test-annotations">
-		<attributes>
-			<attribute name="optional" value="true"/>
-			<attribute name="test" value="true"/>
-		</attributes>
-	</classpathentry>
 	<classpathentry kind="output" path="target/classes"/>
 </classpath>
--- a/bundles/org.openhab.ui.basic/.classpath
+++ b/bundles/org.openhab.ui.basic/.classpath
@@ -30,5 +30,16 @@
 			<attribute name="annotationpath" value="target/dependency"/>
 		</attributes>
 	</classpathentry>
+	<classpathentry kind="src" path="target/generated-sources/annotations">
+		<attributes>
+			<attribute name="optional" value="true"/>
+		</attributes>
+	</classpathentry>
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.ui.basic/.classpath
florian-h05, line 1: This is your second PR containing changes to Basic UI's classpath. I suspect your IDE is modifying it automatically ...

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/widget-details.vue
```diff
@@ -382,6 +397,22 @@ export default {
         delete this.widget[parameter]
       }
     },
+    updateItemFormat($event) {
```
florian-h05:
```suggestion
    updateItemFormat(event) {
```

The method body also needs adjustment.

bundles/org.openhab.ui/web/src/components/pagedesigner/sitemap/widget-details.vue
```diff
@@ -382,6 +397,22 @@ export default {
...
+      if ($event === null || $event === undefined || $event.length === 0) {
+        delete this.widget.format
+      } else {
+        this.widget.format = $event
+      }
+    },
+    toggleItemFormatOverride($event) {
```
florian-h05:
```suggestion
    toggleItemFormatOverride(event) {
```

The method body also needs adjustment.

florian-h05 (changes requested):
Overall LGTM, just my three comments.

bundles/org.openhab.ui.basic/.classpath
mherwege: Yes, indeed. I am doing the editing in VSCode, but also have Eclipse open for the core part (and run OH from there). BasicUI is open in Eclipse, and annoyingly keeps on modifying these `.classpath` files.

florian-h05: Probably that modification of the .classpath file is correct, and should be a dedicated PR. Just a though, I don't know.

florian-h05 (approved):
LGTM, thanks!

PR #4229, opened 2026-05-31
Log viewer: Fix dock y-overflow when window height is reduced

florian-h05:
Decrease the height of the log dock when reducing the height of the browser window and the log dock starts overflowing.


bundles/org.openhab.ui/web/src/App.vue
```diff
@@ -1343,6 +1357,7 @@ export default {
   beforeUnmount() {
     if (window) {
       window.removeEventListener('keydown', this.keyDown)
+      window.addEventListener('resize', this.avoidLogDockOverflow)
```
jimtng: shouldn't this be removeEventListener?

florian-h05: Corrected

jsjames (approved):
LGTM

PR #4231, opened 2026-05-31
Add a widget catalog for tools like agents

digitaldan:
This is for a new MCP tool for allowing agents to have access to our UI for editing , its actually now my most used MCP feature, claude is very very good at creating openhab pages/floorplans/widgets once it has a widget catalog and some specific instructions along with specialized endpoints that minimize data sizes (context windows).   I'll open a PR later today in our addons for the MCP tool.


florian-h05:
Do we want to have that JSON file tracked in this repo, or should we always generate it as part of the build process and keep it out of Git?

digitaldan:
Agreed, i actually did not mean to check that in!

florian-h05:
I think we need to configure the mvn build nom script to build the catalog, and gitignore it.
I can probably have a look at it.

florian-h05 (approved):
LGTM, thanks!

PR #4234, opened 2026-06-01
Semantic tag editor use conversion REST API

mherwege:
This enhances the semantic tag editor to use the new YAML conversion API introduced in: https://github.com/openhab/openhab-core/pull/5573

The logic in the treeview has been enhanced to only allow adding child tags inside a default tag or a managed tag to avoid issues of parent tags not being available when loading in core.

Not done in this PR, but a potential future enhancement: if tags are edited in the code editor, check the the parent-child relationships are still correct before accepting the change.

Here are some screenshots:

Show and edit editable (=managed) tags:

image

Show all custom tags (managed and unmanaged). If there are no unmanaged tags (nothing defined in files), this will be editable, if not this will be read-only:

image

Show all tags, including the default tags. This will always be read-only:

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
@@ -134,7 +134,7 @@ export default {
     Editor
   },
   props: {
-    object: Object,
+    object: [Object, Array],
     objectType: String, // the type of the object, e.g. 'items', 'things' which corresponds to the yaml element name.
     objectId: String,
     hintContext: Object,
--- a/bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
@@ -11,7 +11,7 @@
       @save="$emit('save')" />
   </div>
 
-  <f7-toolbar bottom class="code-editor-toolbar">
+  <f7-toolbar bottom :class="{ 'code-editor-toolbar': true, 'toolbar-narrow': $f7dim.width < 450 }">
     <f7-segmented>
       <f7-button
         v-for="type in Object.keys(mediaTypes)"
@@ -23,13 +23,15 @@
         {{ type }}
       </f7-button>
     </f7-segmented>
+    <slot name="additional-panel-controls" />
     <f7-button
       @click="copy"
       icon-ios="f7:square_on_square"
       icon-aurora="f7:square_on_square"
       color="blue"
+      tooltip="Copy code to clipboard"
       class="copy display-flex flex-direction-row">
-      &nbsp;Copy
+      &nbsp;{{ $f7dim.width < 390 ? '' : 'Copy' }}
     </f7-button>
     <f7-button
       v-if="!readOnly"
@@ -38,8 +40,9 @@
       icon-ios="f7:arrow_2_circlepath"
       icon-aurora="f7:arrow_2_circlepath"
       color="red"
+      tooltip="Revert local changes to the code"
       class="reset display-flex flex-direction-row">
-      &nbsp;Revert
+      &nbsp;{{ $f7dim.width < 390 ? '' : 'Revert' }}
     </f7-button>
   </f7-toolbar>
 
@@ -75,9 +78,36 @@
   position absolute
   .toolbar-inner
     padding-left 8px
+    .toolbar-options
+      align-items center
+      gap 8px
+      .opt-show-all
+        flex-wrap nowrap
+        align-items center
+        flex-direction row
+        span
+          white-space nowrap
+          padding-left 4px
   .segmented
     .button
       width 5em
+
+.code-editor-toolbar.toolbar-narrow
+  --f7-toolbar-height var(--f7-tabbar-labels-height);
+  font-size var(--f7-tabbar-label-font-size)
+  .toolbar-inner
+    padding-left 5px
+    .toolbar-options
+      gap 6px
+      .opt-show-all
+        flex-wrap nowrap
+        align-items center
+        flex-direction column
+        span
+          padding-left 0
+  .segmented
+    .button
+      width auto
 
 .code-editor-errors
   .item-title
@@ -120,7 +150,6 @@ export default {
     return {
       code: null,
       originalCode: null,
-      displayCodeSwitcher: false,
       dirty: false,
       errors: null
     }
@@ -147,18 +176,21 @@ export default {
      *
      * @param {string} codeType - Optional. The type of code to generate (e.g. YAML, DSL)
      * @param {function} onSuccessCallback - Optional. A callback function to call when the code has been generated
-     * @param {object} sourceObject - Optional. The object to generate code from. Defaults to the current prop value.
      */
-    generateCode(codeType, onSuccessCallback, sourceObject = this.object) {
+    generateCode(codeType, onSuccessCallback) {
       codeType ||= this.uiOptionsStore.codeEditorType
       if (!this.mediaTypes[codeType]) {
         codeType = Object.keys(this.mediaTypes)[0]
       }
       const sourceMediaType = MediaType.JSON
       let targetMediaType = this.mediaTypes[codeType]
-      targetMediaType = targetMediaType.split('+')[0] // remove the +thing or +item suffix, if present
+      targetMediaType = targetMediaType.split('+')[0] // remove the +tag, +thing or +item suffix, if present
       const payload = {}
-      payload[this.objectType] = [sourceObject]
+      if (Array.isArray(this.object)) {
+        payload[this.objectType] = this.object
+      } else {
+        payload[this.objectType] = [this.object]
+      }
       this.$oh.api
         .postPlain('/rest/file-format/create', JSON.stringify(payload), null, sourceMediaType, { accept: targetMediaType })
         .then((code) => {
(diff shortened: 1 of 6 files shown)


mherwege:
@Nadahar I have used your idea for adding extra buttons in the code editor from https://github.com/openhab/openhab-webui/pull/4202 in this PR. I hope that makes rebasing easier if one of these gets merged first.

Nadahar:
> @Nadahar I have used your idea for adding extra buttons in the code editor from #4202 in this PR. I hope that makes rebasing easier if one of these gets merged first.

Great, thanks. Mine is likely to be merged late since the core PR hasn't really moved, but as long as the same "principle" has been used, I think I'll manage the rebase. I just saw a potential nightmare where a completely different solution had been used.

jsjames:
@mherwege - are the changes in the openhab-api-spec from the latest 5.2.0M5 or were they from a different core build? I'm wondering if we should create a separate PR to update the openhab-api-spec and heyapi library so we can track this independently from the feature PRs?  Thoughts?

mherwege:
> @mherwege - are the changes in the openhab-api-spec from the latest 5.2.0M5 or were they from a different core build? I'm wondering if we should create a separate PR to update the openhab-api-spec and heyapi library so we can track this independently from the feature PRs? Thoughts?

No, they are all in the core milestone, but I was working on this before the final version got merged.. I removed them again. I hope it works.

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
```diff
@@ -38,7 +38,7 @@
       </div>
     </f7-toolbar>
     <f7-tabs class="semantics-editor-tabs">
-      <f7-tab class="design" id="tree" :tab-active="currentTab === 'tree'">
+      <f7-tab class="design" id="tree" :tab-active="currentTab === 'tree' ? true : null">
```
jsjames: Is the ternary operation needed here? For f7 components (even under vue3), it shouldn't be necessary to set to null to eliminate the tag.

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
```diff
@@ -168,14 +168,51 @@
...
+          :read-only-msg="
+            showCodeTags === 'custom'
+              ? 'Showing non-editable tags, switch to editable to edit'
+              : showCodeTags === 'all'
+                ? 'Showing default tags, switch to editable to edit'
+                : ''
+          "
+          :object="codeEditorTags"
+          @parsed="update"
+          @changed="onCodeChanged">
```
jsjames: you should keep the @save event here - this is so that :w in VIM emulation or future codemirror enhancements will save the doc.

bundles/org.openhab.ui/web/src/pages/useDirty.ts
```diff
@@ -99,13 +99,14 @@ export function useDirty(pageRefOrName: string | Ref<PageRef> | null) {
    * @param value - The ref to watch for changes (typically your data model)
    * @param dirtyRef - Optional, currently unused (kept for API compatibility)
    */
-  function setupDirtyWatch(value: Ref<unknown>, dirtyRef?: Ref<boolean>) {
-    let pristineValue = cloneDeep(value.value)
+  function setupDirtyWatch(value: WatchSource<unknown>, dirtyRef?: Ref<boolean>) {
```
jsjames, line 102: great enhancement here!!

lolodomo:
Maybe few screenshots to show what you implemented?

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
mherwege: Must have slipped in when trying to improve the logic to avoid tab switches when the code is not valid. I will remove.

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
mherwege: It looks like I missed updating the `save` function completely. I will put this back and adjust.

mherwege:
> Maybe few screenshots to show what you implemented?

I updated the first post.

lolodomo:
> > Maybe few screenshots to show what you implemented?
> 
> I updated the first post.

Great.
You were fast to implement that. Good point to have this in 5.2.

Nadahar:
@jsjames I have spent a lot of time making the toolbar layout work in #4202, I'm not going to do it all over again. Please don't refactor so that #4202 won't fit - I will just give up on it then.

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
```diff
@@ -10,8 +10,8 @@
         :f7router />
     </f7-navbar>
     <f7-toolbar tabbar position="top">
-      <f7-link @click="switchTab('tree')" :tab-link-active="currentTab === 'tree'" tab-link="#tree"> Design </f7-link>
-      <f7-link @click="switchTab('code')" :tab-link-active="currentTab === 'code'" tab-link="#code"> Code </f7-link>
+      <f7-link @click="switchTab('tree', switchTabTree)" :tab-link-active="currentTab === 'tree'"> Design </f7-link>
```
jsjames, line 13: Just a note - as you probably found out removing the tab-link does eliminate the red bar on the active tab that is added by framework7, however, when tab-link is present framework7 from automatically switching tabs.  I've looked into this quite a bit and there's not a good way around this. So what you have is good.

jsjames:
> @jsjames I have spent a lot of time making the toolbar layout work in #4202, I'm not going to do it all over again. Please don't refactor so that #4202 won't fit - I will just give up on it then.

The changes made here are just using css styles to drop the labels on revert/copy when the width is below a certain number of pixels - so its the same exact logic.  I did drop the labels at a slightly wider width as when I was testing on my system, they were still overlapping.  This should work fine for your PR.

bundles/org.openhab.ui/web/src/components/config/controls/code-editor.vue
```diff
@@ -184,10 +218,11 @@ export default {
...
      * @param {function} onSuccessCallback - Optional. A callback function to call when the code has been parsed. Receives the parsed object.
      * @param {function} onFailureCallback - Optional. A callback function to call when parsing fails or no object is found
+     * @param {boolean} parseAll - Optional. If true, all parsed objects will be returned as an array. By default, only the first object is returned for backward compatibility with existing code that expects a single object.
      */
-    parseCode(onSuccessCallback, onFailureCallback) {
+    parseCode(onSuccessCallback, onFailureCallback, parseAll = false) {
       let sourceMediaType = this.mediaTypes[this.uiOptionsStore.codeEditorType]
-      sourceMediaType = sourceMediaType.split('+')[0] // remove the +thing or +item suffix, if present
+      sourceMediaType = sourceMediaType.split('+')[0] // remove the +tag, +thing or +item suffix, if present
       const targetMediaType = MediaType.JSON
```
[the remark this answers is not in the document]
mherwege: When only YAML is available, the codeEditorType is correctly set generateCode, so not a problem.

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
```diff
@@ -397,38 +438,67 @@ export default {
...
+      }
     }
   },
   methods: {
     onPageAfterIn() {
       if (window) {
         window.addEventListener('keydown', this.keyDown)
       }
+      this.setupDirtyWatch(() => this.editableTags)
     },
```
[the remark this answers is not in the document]
mherwege: Correctly identified a problem, but the solution does not work. The second parameter in setupDirtyWatch does not do anything. Fixed in another more appropriate way.

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
```diff
@@ -668,40 +764,25 @@ export default {
...
+      })
+      tags.push(...updatedTags)
       this.semanticTags = tags
-      return true
+      this.codeDirty = false
+    },
+    onCodeChanged(codeDirty) {
+      this.dirty = this.dirty || codeDirty
+      this.codeDirty = codeDirty
     }
```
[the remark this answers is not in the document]
mherwege: Fixed in a different way, the proposed fix does not work.

bundles/org.openhab.ui/web/src/pages/developer/semantics/semantic-tags-edit.vue
mherwege: I actually missed noticing it. But I now provided a fix using CSS.

jsjames: Great solution here in using CSS.  BTW, you don't have to create a new class here or assign the class in the f7-link there is already a "tab-link-active" class which gets applied when you switchTabs, so the below should be sufficient.

```
.semantics-editor-tabbar
  .link
    border-bottom 2px solid transparent
  .link.tab-link-active
    border-bottom-color var(--f7-tabbar-link-active-border-color)
```

There are quite a few other pages in the UI which could use this framework now as there wasn't a good way to prevent a tab from switching if/when the parsing fails.  The only "limitation" I see to using css styling here, is that the swoosh animation of the red border does not happen when the tab changes, but I think that is a very minor tradeoff to get this better functionality.  Thanks!

jsjames (approved):
LGTM, thanks for the enhancements!

PR #4252, opened 2026-06-02
Fix Day.js locale loading regression from #4164

florian-h05:
Regression from #4164.

#4164 broke Day.js locale loading and thereby Main UI for locales that aren't included in Day.js.
This also fixes the translation of locale identifiers, which wasn't working anymore.


jsjames (approved):
LGTM

Issue #3939, opened 2026-03-01
Variables seem to no longer work on OVERVIEW Page (Worked prior to OH 5.2.0.M1


Mark-VG:
## The problem

Prior to OH 5.2.0.M1 I used variables to determin teh visibility of componets on my widgets.

After installing the Milestone this functionality stopped working and the ```visibility``` seemed to be always ```false```

The functionality still seems to work on other pages as well as the Overview in "run" mode.

After posting in the community it was suggested to create a limited widget to demonstrate the issue.

https://community.openhab.org/t/openhab-5-2-milestone-discussion/168410/2
https://community.openhab.org/t/openhab-5-2-milestone-discussion/168410/23

Similar issue was resolved in the past: https://github.com/openhab/openhab-webui/issues/2046

The following widget code demonstrates the issue when used on the Overview Page:
```
component: oh-cell
config:
  title: ="Test Visibility " + vars.visibleMonthlyPop
slots:
  background:
    - component: f7-block
      config:
        bgColor: none
        style:
          bottom: 5px
          left: 0px
          margin: 0
          padding: 0
          position: absolute
          width: 100%
      slots:
        default:
          - component: oh-button
            config:
              action: variable
              actionVariable: visibleMonthlyPop
              actionVariableValue: true
              popupOpen: .month-pop-test
              style:
                height: 100%
                width: 100%
              text: OPEN
    - component: f7-popup
      config:
        class: month-pop-test
        style:
          --f7-popup-tablet-height: 720px
          --f7-popup-tablet-width: 360px
          background-color: rgb(220,220,220)
          border-radius: 30px
          height: 720px
          text-overflow: ellipsis
          width: 360px
      slots:
        default:
          - component: f7-block
            config:
              style:
                --f7-popup-tablet-height: 720px
                --f7-popup-tablet-width: 360px
                background: rgb(220,220,220)
                border-radius: 20px
                height: 720px
                margin: 0
                padding: 0
                position: absolute
                top: 0px
                width: 360px
            slots:
              default:
                - component: oh-button
                  config:
                    action: variable
                    actionVariable: visibleMonthlyPop
                    actionVariableValue: false
                    clearVariable:
                      - visibleMonthlyPop
                    large: true
                    popupClose: .modal-in
                    style:
                      background: rgb(220,220,220)
                      color: white
                      display: flex
                      height: 30px
                      left: 5px
                      position: absolute
                      top: 5px
                      width: 30px
                      z-index: 999
                  slots:
[...]

PR #4257, opened 2026-06-07
Pages: Code improvement for applying defineVars

florian-h05:
Improves the fix for #3939 from PR #4038 by using watchers instead of having side-effects in computed.


jsjames (approved):
lgtm

PR #4258, opened 2026-06-07
Item picker: Fix grayed out like disabled if empty

florian-h05:
Regression from #4117.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/item-picker.vue
@@ -81,9 +81,6 @@
     padding 0
   .item-inner:after
     display none
-  .item-picker-empty
-    .item-title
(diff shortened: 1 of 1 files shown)


florian-h05:
@mherwege What are your thoughts on this?

mherwege:
Yes, it looks like that was a mistake. The title is not a placeholder. It was the placeholder I wanted to be greyed out.

florian-h05:
Okay, then we have to check that with this PR the placeholder still is grayed out.

mherwege:
My main concern was when using the item picker in the sitemap editor. There actually is no placeholder there. The thing about the item picker is that it shows an icon before the title, and that's what got me confused assuming it was a placeholder. So it should be fine.

jsjames (approved):
LGTM

jsjames (approved):

PR #4260, opened 2026-06-07
Fixes for navbar on mobile

florian-h05:
- Hide navbar save link text on mobile.
- Developer dock icons: Only add right margin if not last element.
- persistence-edit: Hide `(Ctrl-S)` hint from save link if not on desktop.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/navigation/oh-nav-content.vue
+++ b/bundles/org.openhab.ui/web/src/components/navigation/oh-nav-content.vue
@@ -2,7 +2,7 @@
   <f7-nav-left class="oh-nav-content">
     <f7-link v-if="menuIcon" class="menu-icon" icon-ios="f7:menu" icon-aurora="f7:menu" icon-md="material:menu" panel-open="left" />
     <f7-link v-if="!theme.md && backLink" icon-f7="chevron_left" :href="backLinkUrl" @click="back">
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.ui/web/src/components/navigation/oh-nav-content.vue
```diff
@@ -2,7 +2,7 @@
   <f7-nav-left class="oh-nav-content">
     <f7-link v-if="menuIcon" class="menu-icon" icon-ios="f7:menu" icon-aurora="f7:menu" icon-md="material:menu" panel-open="left" />
     <f7-link v-if="!theme.md && backLink" icon-f7="chevron_left" :href="backLinkUrl" @click="back">
-      {{ backLink }}
+      {{ $f7dim.width > 500 ? backLink : null }}
```
jsjames, line 5: What are your thoughts to moving to use css styling to hide / display elements? I think the html is cleaner, but know that most of the ui uses v-if.

florian-h05: I think using CSS here would make things unnecessarily complicated. 
v-if as well IMO, as we would have to introduce a third <f7-link> without the text.

jsjames (approved):
LGTM

PR #4261, opened 2026-06-07
Support interactive SVGs in floor plans

digitaldan:
This extends our interactive SVG support to Floor plans as well as normal pages.  It moves  `oh-embedded-svg-mixin.js` to a shared package, that both plans and pages use.  Couple other enhancements:

- State styling now works for all item types (previously only on/off worked)
- Colors can target stroke for outline (fill:none) shapes
- Any valid CSS color is now accepted instead of falling back to red.
- Couple state related fixes

I'll also update the doc project as well with new features for floor plans


ghys:
Oh wow, this is what I wanted for so long...

digitaldan:
👀

digitaldan:
@jsjames @florian-h05  any chance to get a review on this for our upcoming release?   I'm now using this extensively in my own home.....so would like to switch off of using my one off UI bundle and use the distro version.

jsjames:
> @jsjames @florian-h05 any chance to get a review on this for our upcoming release? I'm now using this extensively in my own home.....so would like to switch off of using my one off UI bundle and use the distro version.

I'm looking at this one now and implementing one of my SVGs as interactive to test. However, when I assign an item to one of the [openhab] paths, it applies that item to all the paths.  Here's my image, and I have only assigned an item (and off/on color) to one of the paths.  However, all the paths are now set to that item and that color.

image

digitaldan:
@jsjames can you post your SVG and YAML?  i can't tell what going on without that. Also i'm not clear if this was an existing SVG you have that was working and now its not, or is this your first time trying one?

My guess right now is your SVG elements do not have IDs ?  or they all have the same ID (at least the ones that are highlighted) ?

jsjames:
> @jsjames can you post your SVG and YAML? i can't tell what going on without that. Also i'm not clear if this was an existing SVG you have that was working and now its not, or is this your first time trying one?
> 
> My guess right now is your SVG elements do not have IDs ? or they all have the same ID (at least the ones that are highlighted) ?

That is the issue - thanks for the heads up.

jsjames (approved):
LGTM - Great add!

PR #4264, opened 2026-06-07
oh-aggregate-series: Load look-back timeframe for diff_last/diff_first aggregation functions

florian-h05:
`diff_last` and `diff_first` aggregation functions need data from before the start time to calculate their first value.
Load this data from persistence to calculate the first aggregated value.
Also return `NaN` from `diff_last` instead of 0 when unable to calculate aggregated value.

Attempt to fix https://community.openhab.org/t/echart-problem-with-oh-aggregate-series/168922/23.


florian-h05:
Should be backported to 5.2.x

jsjames (approved):
LGTM

PR #4267, opened 2026-06-10
[MainUI] Enable server pinia storage

Nadahar:
This doesn't do anything useful at the moment, it's more of a "sketch" of what I'm thinking of.

I got the idea when looking at #4196, I think it can be very useful, but also think that the fact that you must repeat in on every browser on every device kind of undermines the usefulness.

The idea here is that it will be stored per logged-in user instead. Most people probably only have one OH user, but it is possible to have more if that is desirable. Not all pinia stores are suitable for server/user storage, so my idea is to make a general way to easily facilitate either local browser storage or server storage per pinia store.

This is what I've managed to come up with, with heavy assistance by "AI". I don't know if this is the most reasonable way to organize it, but I think it serves to present the idea: That a pinia plugin handles the actual storage, and that each store only has to specify `local` or `server` when they are created, and the rest is handled transparently.

In addition, the stores have gotten a `ready` property, that can be used when determining if the page is loaded, so that the placeholder skeleton will stay in place until the store has been populated as well.

Endpoints for storage and retrieval would have to be created in core, and it would also have to be figured out exactly how to handle the per-user storage. When it comes to the store itself, it's just a JSON object, so I was imagining to treat it like a simple string in core - so that core don't actually interfere with the store data at all (if it's not parsed/understood, there can be no manipulation/changes)

I think I can make the core part relatively easily, but I won't start that unless you guys think it's something worth pursuing. So, @florian-h05 and @jsjames, what do you think of the idea, not so much the implementation?

### DIFF
--- a/bundles/org.openhab.ui/web/src/js/stores/useUIOptionsStore.ts
+++ b/bundles/org.openhab.ui/web/src/js/stores/useUIOptionsStore.ts
@@ -12,345 +12,351 @@ export interface LogHighlightFilter {
   active: boolean
 }
 
-export const useUIOptionsStore = defineStore('uiOptions', () => {
-  // States
-  // shared with Basic UI
-  const _storedDarkMode = localStorage.getItem('openhab.ui:theme.dark') || 'auto'
-  const storedDarkMode = ref<StoredDarkModeType>(
-    ['auto', 'dark', 'light'].includes(_storedDarkMode) ? (_storedDarkMode as StoredDarkModeType) : 'auto'
-  )
-  const darkModeChange = ref<number>(0) // Used to trigger recomputation of darkMode
+export const useUIOptionsStore = defineStore(
+  'uiOptions',
+  () => {
+    // States
+    // shared with Basic UI
+    const _storedDarkMode = localStorage.getItem('openhab.ui:theme.dark') || 'auto'
+    const storedDarkMode = ref<StoredDarkModeType>(
+      ['auto', 'dark', 'light'].includes(_storedDarkMode) ? (_storedDarkMode as StoredDarkModeType) : 'auto'
+    )
+    const darkModeChange = ref<number>(0) // Used to trigger recomputation of darkMode
 
-  const _storedBars = localStorage.getItem('openhab.ui:theme.bars') || 'light'
-  const bars = ref<'light' | 'filled'>(['light', 'filled'].includes(_storedBars) ? (_storedBars as 'light' | 'filled') : 'light')
+    const _storedBars = localStorage.getItem('openhab.ui:theme.bars') || 'light'
+    const bars = ref<'light' | 'filled'>(['light', 'filled'].includes(_storedBars) ? (_storedBars as 'light' | 'filled') : 'light')
 
-  const _storedNavBar = localStorage.getItem('openhab.ui:theme.home.navbar') || 'default'
-  const homeNavBar = ref<'default' | 'simple' | 'large'>(
-    ['default', 'simple', 'large'].includes(_storedNavBar) ? (_storedNavBar as 'default' | 'simple' | 'large') : 'default'
-  )
+    const _storedNavBar = localStorage.getItem('openhab.ui:theme.home.navbar') || 'default'
+    const homeNavBar = ref<'default' | 'simple' | 'large'>(
+      ['default', 'simple', 'large'].includes(_storedNavBar) ? (_storedNavBar as 'default' | 'simple' | 'large') : 'default'
+    )
 
-  const _storedHomeBackground = localStorage.getItem('openhab.ui:theme.home.background') || 'default'
-  const homeBackground = ref<'default' | 'standard' | 'white'>(
-    ['default', 'standard', 'white'].includes(_storedHomeBackground)
-      ? (_storedHomeBackground as 'default' | 'standard' | 'white')
-      : 'default'
-  )
+    const _storedHomeBackground = localStorage.getItem('openhab.ui:theme.home.background') || 'default'
+    const homeBackground = ref<'default' | 'standard' | 'white'>(
+      ['default', 'standard', 'white'].includes(_storedHomeBackground)
+        ? (_storedHomeBackground as 'default' | 'standard' | 'white')
+        : 'default'
+    )
 
-  const _storedExpandableCardAnimation = localStorage.getItem('openhab.ui:theme.home.cardanimation') || 'default'
-  const disableExpandableCardAnimation = ref<boolean>(_storedExpandableCardAnimation === 'disabled')
+    const _storedExpandableCardAnimation = localStorage.getItem('openhab.ui:theme.home.cardanimation') || 'default'
+    const disableExpandableCardAnimation = ref<boolean>(_storedExpandableCardAnimation === 'disabled')
 
-  const blocklyRenderer = ref<string | null>(localStorage.getItem('openhab.ui:blockly.renderer'))
-  const disablePageTransitionAnimation = ref<boolean>(localStorage.getItem('openhab.ui:theme.disablepagetransition') === 'true')
+    const blocklyRenderer = ref<string | null>(localStorage.getItem('openhab.ui:blockly.renderer'))
+    const disablePageTransitionAnimation = ref<boolean>(localStorage.getItem('openhab.ui:theme.disablepagetransition') === 'true')
 
-  const hideChatInput = ref<boolean>(localStorage.getItem('openhab.ui:theme.home.hidechatinput') === 'true')
+    const hideChatInput = ref<boolean>(localStorage.getItem('openhab.ui:theme.home.hidechatinput') === 'true')
 
-  // shared with Basic UI
-  const webAudio = ref<boolean>(localStorage.getItem('openhab.ui:webaudio.enable') === 'true')
+    // shared with Basic UI
+    const webAudio = ref<boolean>(localStorage.getItem('openhab.ui:webaudio.enable') === 'true')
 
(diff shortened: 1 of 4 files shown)


Nadahar:
By the way, I have no idea why it reformatted the whole `useUIOptionsStore.ts`. All I actually changed there was to add the `persistTo: 'local'` as a third parameter, but when I ran the formatter, it completely rearranged the file.

jsjames:
One way to implement this (without having to change the openhab-core) is to store the settings in a ui: component.  That would require the entire ui:component to be written each time a setting is updated - so there is a tradeoff. 

If you want to work on a more fine-grained setting of individual settings, you could start the work on openhab-core.  I think I'd still prefer to breaking up the useUIOptions store though.  We could adapt the vueuse/useStorage to support writing to the settings to the server - but keep more generic vs. having a lot of the logic for things in the store.

Nadahar:
The stores can be broken up however you like, I'm just saying that if this were to become a reality, things should be kept using stores, not reading/writing directly to/from LocalStorage.

The whole UIComponent storage in core is overly generic already in my view, there's "no type safety" there, the system can return a page as a widgets etc., and core would have no idea that it delivers the wrong element type. So, I'd rather see that made "more detailed" in the future, than jam more stuff in as generic UI components.

My idea was also to make this a per-user thing, so that what is returned by the API depends on the authenticated user. To achieve that, I must make a new way to store them, UIComponents aren't "per user".

I haven't looked at the details of how to do this in core, but I don't think it should be problematic. I just didn't want to proceed until you guys agreed on the concept itself.

Am I to interpret it as that you think the concept itself is sound, making it so that stores can either be per user or per browser? We could even perhaps make that in itself configurable for some stores, I haven't thought that through. But, the idea is that the MainUI code should be able to interact with the stores in the same way, regardless of the nature of the underlying storage.

jsjames:
> The stores can be broken up however you like, I'm just saying that if this were to become a reality, things should be kept using stores, not reading/writing directly to/from LocalStorage.
> 
> The whole UIComponent storage in core is overly generic already in my view, there's "no type safety" there, the system can return a page as a widgets etc., and core would have no idea that it delivers the wrong element type. So, I'd rather see that made "more detailed" in the future, than jam more stuff in as generic UI components.
> 
> My idea was also to make this a per-user thing, so that what is returned by the API depends on the authenticated user. To achieve that, I must make a new way to store them, UIComponents aren't "per user".
> 
> I haven't looked at the details of how to do this in core, but I don't think it should be problematic. I just didn't want to proceed until you guys agreed on the concept itself.
> 
> Am I to interpret it as that you think the concept itself is sound, making it so that stores can either be per user or per browser? We could even perhaps make that in itself configurable for some stores, I haven't thought that through. But, the idea is that the MainUI code should be able to interact with the stores in the same way, regardless of the nature of the underlying storage.

Yes, the concept to me makes sense.

My thinking for the client implementation is to still use the vueuse/useStorage composable.  It takes a "store" as an argument - today it supports either the local-store or session-store. We should be able to write our own store interface as long as it conforms to the below structure and have it interface with any new settings core api.  That way, we can use the common useStorage composable for any storage - local, session or user (via core apis).

```
export interface StorageLikeAsync {
  getItem: (key: string) => Awaitable<string | null>
  setItem: (key: string, value: string) => Awaitable<void>
  removeItem: (key: string) => Awaitable<void>
}
```

So, if the proposed core api could have a getItem, setItem and removeItem that should cover our bases (sorry for the baseball expression :) ).

Nadahar:
> We should be able to write our own store interface as long as it conforms to the below structure and have it interface with any new settings core api. That way, we can use the common useStorage composable for any storage - local, session or user (via core apis).

Yes, that sounds great, the more "universal" the better. Could you give me an idea of where that should be implemented (feel free to push examples/suggestions here, think of this PR as a "scratchpad").

There is one thing we must handle somehow, and that is that the user/server store will have to be async. Is there any way in the interface to "integrate" a promise somehow, so that all store handing would take "potential async behavior" into account? Local and session storage would then just resolve immediately.

Nadahar:
I see now that the interface seems to take async into account, which is great. I don't fully understand how I must go about to implement it, but this sounds like a great way. I can certainly make 3 REST endpoints that can serve these 3 operations.

jsjames:
> nt it, but this sounds like a great way. I can certainly make 3 REST endpoints that can serve these 3 operations.

I'll can work on the client code to implement this.  I'll check something in which use useStorageAsync to show how it would work. Then, we just need to write the new StorageLikeAsync method to call the new server APIs.  I presume the 3 endpoints would actually be one endpoint withe a GET, POST, DELETE?

Some other things we need to think about:
- For any one configuration, would we support either local or server?  Would we have a setting which allows local or server storage?
- Moving to server could cause many new server queries on startup - This could slow down mobile use-cases.  While more complicated, one option is to use local as a cache so it is immediately available, then after getting the result from the server, we would use that value?
- We need to think about what the upgrade path would look like.

Nadahar:
> I'll can work on the client code to implement this. I'll check something in which use useStorageAsync to show how it would work. Then, we just need to write the new StorageLikeAsync method to call the new server APIs. I presume the 3 endpoints would actually be one endpoint withe a GET, POST, DELETE?

That'd be great. I don't need anything "finished"/polished, but when making the server endpoints, I need to mock up something that I can use for testing.

> * For any one configuration, would we support either local or server?  Would we have a setting which allows local or server storage?

I haven't made that many thoughts about that. I see that it could be useful at least for some stores, to be able to choose between user and local. But, it would be hard to know what should be the default then. Because, if you make LocalStorage store the choice, and the default is "local", it won't be possible to configure OH so that this is "preconfigured correctly" for new clients. I think that's something some users would wish to be able to, make some kind of (textual) configuration server-side that made sure that the defaults were set in a specific way.

I haven't looked at all the stores, but I assume that for some, server/user storage isn't a relevant option?

> * Moving to server could cause many new server queries on startup - This could slow down mobile use-cases.  While more complicated, one option is to use local as a cache so it is immediately available, then after getting the result from the server, we would use that value?

I'm not sure if that would be very heavy compared to many other things that go on? We're probably talking about very small amounts of data here? Page loading will typically ask for quite a lot of other information, so would a few store options really make a difference?

But, we should probably think of caching the results somehow, so that the same info isn't requested over and over again. There's also the question of how to handle if one session changes a user setting that is also used by others clients. I guess that we don't have a "notification mechanism" here..? Ideally, if the server could notify clients about changes, clients could just query each store once, and then work with its cache. But, I don't know if that's possible to do, it would probably at least be complicated.

If we cache a value in LocalStorage and use that value first, and when the result from the server is in, it turns out to be different, how could the setting then be "updated"? Again, we need some kind of "callback"/event mechanism if we want the store to be able to "push" changes.

> We need to think about what the upgrade path would look like.

That depends a bit on the store I guess, but these settings are very "volatile" today. You'll probably end up having to set the same options over and over again over time already, so I'm not sure that it "resetting to default" one more time would be that much of an issue, as a one-time occurrence when the store moved to user storage.

mherwege:
Overall, I like the idea of moving this to the server. One warning though. This effectively makes settings user dependent instead of user and device dependent. And that may not be right in all cases.
E.g. imagine you have a tablet on a wall running MainUI. The settings for that may be different from when you open MainUI on your computer or phone. If you make it all user specific, you get the same on all devices unless you create separate users. So I am not convinced moving from a per device configuration model to a per user configuration model makes sense for everything. For many configurations, the right level may be per user and per device. But than the question becomes what is gained by moving it to the server at all.
I am not against moving settings to the server, but this has to be tought through carefully for each individual setting. Some are per device, some may be per user.

ErikDB87:
Wouldn't a tablet on the wall require its own user (e.g. `walltablet`)?

Nadahar:
> This effectively makes settings user dependent instead of user and device dependent. And that may not be right in all cases.

It was never my intention to make everything user dependent. I've always said that it would probably only make sense for *some* stores. We've also discussed above if we could make that itself configurable in some cases, so that the user could choose between user and local storage.

When it comes to the tablet on the wall, that could be a situation where it would make sense to create a separate user to get completely "separate behavior" anyway.

jsjames:
> verything user dependent. I've always said that it would probably only make sense for _some_ stores. We've also discussed above if we could make that itself configurable in some cases, so that the user could choose between user and local storage.

We could offer a vscode/chrome like sync-settings option where the user on any device can select whether their settings are synced.  I think that would be pretty straight forward.

Nadahar:
> We could offer a vscode/chrome like sync-settings option where the user on any device can select whether their settings are synced. I think that would be pretty straight forward.

I'm not sure how that "sync setting option" looks, but I'm not really worried about how to make the option available in the UI, but how to implement it underneath. How do we tell the store whether to do this or that, if we only implemented the above-mentioned interface? But, maybe that's not necessary, maybe we can communicate that through a "side-channel". I can't quite imagine what can and cannot be done until I see how it would be implemented (the interface).

What I think is important, is that where the user can select local or user, the default should be user. Because, that's the only way to enable "pre configuring" OH so that the UI behaves a certain way.

jsjames:
> I'm not sure how that "sync setting option" looks, but I'm not really worried about how to make the option available in the UI, but how to implement it underneath.

The underlying implementation we are proposing would work the same. We would just have a setting in localStorage called "sync" - if that is enabled, it would pull/store options from the server and if disabled, it would pull/store locally. That way someone with a tablet on a wall one could have this disabled and set device specific settings. Then, for other devices, you would have it enabled and it would pull the settings from the server.

Nadahar:
> We would just have a setting in localStorage called "sync" - if that is enabled,

Yeah, but it would miss my point about the default. Honestly, I think "sync" is a bad fit here, but the name is details. It's much better to have a choice between "user" and "local" storage, than have people believe it's about somehow synchronizing the local to a server.

PR #4278, opened 2026-06-14
Add a chat interface to Main UI

### LINKED ISSUE #2995: Add a intelligent smart home chat bot to the UI
I am thinking of having a smart home chatbot for openHAB 5, a bit like HABot but more intelligent, integrated into Main UI and not only limited to smart home related stuff.

This would require the following bits:

- [x] A powerful, LLM-based human language interpreter available: Something like https://github.com/openhab/openhab-addons/pull/17320.
- [ ]  New core APIs that not only allow answering with text to a prompt (like current HLI interface), but provide a payload for the UI that can e.g. make it render a card widget to show an Item state.
- [ ]  Probably: Shared core logic to be used by LLM based HLIs for Item control, so we don't have to duplicate code for OpenAI API, Google Vertex API etc.
- [ ] The chatbot UI. Framework7 provides something we can use for that: https://framework7.io/docs/messages

Unfortunately anything LLM-based is unlikely to work locally on embedded hardware like Pis or NAS, however is OpenAI's GPT 4o Mini relatively cheap and I don't have large privacy concerns as it is only used to interpret a text prompt - it does not have direct access to openHAB.

florian-h05:
Resolves #2995.

This PR introduces a new feature called HABAssist—a modern, interactive chat interface built into Main UI. HABAssist allows users to control and query their smart home using natural language by routing conversation messages through the openHAB human language interpreter and LLMs.

## Key Features

1. A HABAssist entry has been added to the sidebar:
  

3. Welcome Screen & Suggestion Cards:
    A welcome screen display 4 suggestion that are sent once they are clicked.
  

5. Real-Time Interactive Messaging Feed:
    Built on top of Framework7's messaging component.
7. Interactive In-line Item Control (Special Tool Rendering):
    When the LLM decided to interact with items through the `get-item-state` or `send-item-command` LLM tools, the UI renders the default card widget for that Item.

https://github.com/user-attachments/assets/4aed2013-9246-4c40-9170-c033a4cb3880

9. Tool Customization & Tool Call Display Settings:
    Allows to select the tools that are provided to the interpreter, and allows to enable/disable the displaying of all tool calls. When disabled, only the item related tool calls are rendered, when enabled, other tool calls are rendered in a generic component.
  

## Technical Details

- The conversation is persisted on the server, each instance of Main UI generated is own random conversation id.
- Selected LLM tools and the tool call display setting are persisted in the browser's local storage.
- Messages are sent to the server over the REST API, all incoming messages are received through the event bus over SSE after the initial loading of the conversation history from the REST API.

## Requirements

- A very recent snapshot build.
- A human language interpreter that is capable enough to handle real chatting. At the moment, such an implementation is only available for Gemini models: https://github.com/openhab/openhab-addons/pull/20922
- The `gemini` HLI should be set as default HLI in the Voice settings.


florian-h05:
@jsjames Can you please review and merge this before the feature freeze on Friday? This is one of our major features for openHAB 5.2.
I will leave for vacation tomorrow, so I am not available to address review feedback. However, if there are any issues, we can fix them after the feature freeze. Also, since this is a new feature and nearly 100% new code, the regression risk is very low. 

/cc @digitaldan @kaikreuzer @holgerfriedrich You may want to take a look.

rkoshak:
Maybe not for this PR, but this should deprecate the HABot chat interface built into MainUI, right? It would be confusing to continue with both for the long term. 

image

florian-h05:
Yes. I’ve already had a look at it, but given my current lack of time decided to postpone this.

florian-h05:
What do you think about the naming? HABAssist?
Any other ideas?

jsjames:
This looks good to me. I didn't personally test it since I haven't upgraded my production system.

jsjames (approved):
LGTM

rkoshak:
> What do you think about the naming? HABAssist?

I dont' partticularly like it because it makes me think of Home Assistant. It's kind of funny though becuause it doesn't make me think of Google Assistant in the same way and even if it did that wouldn't bother me as much. 

I'm not sure I have any better ideas though. Naming things is not my forte and this mental assosiation with HA may just be me. 

- HABChat seems too cute and the letters don't look good next to each other, maybe HABchat?
- Talos: giant autamoton in Greek myth and if we ever go to voice it makes an OK wakeup word and I'm not aware of anyone using the name for anything in home automation (it's used in robotics though which may be too close)
-  HABctl might trip up some users but HABcontrol seems too long but you'd say it the same so maybe that doesn't matter; but it also doesn't imply the LLM aspect of it

I can think on it some more and if something comes up I'll come back.

florian-h05:
In the first place I named it „Chat“ and took care of implementing it. I then decided that Chat is basically saying nothing and thought about Assist or Assistant. I didn’t choose Assistant because of Google. HA has a comparable feature called Assist, so I went with HABAssist. 
If we were Apple we would name it HABot AI 😂

Ultimately the naming is just I18n strings, so it is super easy to change.

kaikreuzer:
> In the first place I named it „Chat“ 

I just noticed the HABAssist menu entry after upgrading - tbh, I would very much prefer "Chat", which is pretty neutral and describes what the menu entry is about without trying to introduce any new kind of product name.

florian-h05:
What about plain „Assist“?

kaikreuzer:
Assist does not really point towards speech. And it still reminds one of HA. 😉

florian-h05:
Then let’s change it to chat.

kaikreuzer:
Like that? https://github.com/openhab/openhab-webui/pull/4298

ghys:
Fine piece of work right there.

You could call it HABot 2 if you wish, it isn't as if I trademarked the name or anything. In fact I would be honored if it were the new HABot.

Reminds me of 2019, except in 2019, you didn't have ChatGPT, it was 4 years away, damn kids!

https://memepediadankmemes.fandom.com/wiki/Old_Man_Yells_at_Cloud

florian-h05:
And now ChatGPT is nearly 4 years old — time flies!

HABot 2 would probably be a good pick wrt recognition among users. @kaikreuzer WDYT?

kaikreuzer:
> wrt recognition among

... old users. 😉 

I appreciate HABot being far ahead of its time, but we are here imho talking about a chat feature of the Main UI - so I would continue to stick to "Chat" for the menu entry here. It is not a UI on its own, which the name HABot 2 would suggest.

PR #4279, opened 2026-06-15
Properly support read-only block libraries

Nadahar:
This PR depends on https://github.com/openhab/openhab-core/pull/5590.

Support for providing widgets and pages through YAML have been implemented previously, but block libraries were left out. Since these are almost identical from core's point of view, I decided to add block libraries as well in https://github.com/openhab/openhab-core/pull/5590 when implementing support for hosting these in "the new YAML format" on the marketplace.

This would leave the UI with one "inconsistency", because the UI doesn't know that a block library might not be editable, and it won't know to "wrap" the YAML in `version:` etc.

This PR aims to address these shortcomings, but there's a problem: I don't know Blockly, I have no idea what a block library is or how you use or view it. Since Blockly is used to render these blocks in the UI, and I have no idea about any of it, there were some challenges in making the UI work for "read-only entries". I've tried my best, I haven't found a way to edit the read-only entries, but ultimately, I can't "verify" that the result makes sense. Somebody that knows/understands Blockly should verify the functionality for read-only libraries. The behavior for editable libraries shouldn't have changed.

Here are a couple of screenshots of a read-only library:

image

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/blockly-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/blockly-editor.vue
@@ -1181,11 +1181,7 @@ Blockly.ContextMenuItems.registerCommentOptions()
 export default {
   props: {
     blocks: String,
-    libraryDefinitions: Array,
-    readOnly: {
-      type: Boolean,
-      default: false
-    }
+    libraryDefinitions: Array
   },
   emits: ['mounted', 'ready', 'change'],
   data() {
@@ -1285,7 +1281,7 @@ export default {
       this.addLibraryToToolbox(libraryDefinitions || [])
 
       const options = {
-        toolbox: this.readOnly ? null : this.$refs.toolbox,
+        toolbox: this.$refs.toolbox,
         horizontalLayout: !this.$device.desktop,
         theme: useUIOptionsStore().darkMode === 'dark' ? DarkTheme : undefined,
         zoom: {
@@ -1303,7 +1299,6 @@ export default {
         },
         trashcan: false,
         showLabels: false,
-        readOnly: this.readOnly,
 
         // Multi-select-options
         multiselectCopyPaste: {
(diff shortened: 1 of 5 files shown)


Nadahar:
@rkoshak perhaps you have something to add?

florian-h05 (commented):
Code changes LGTM.
I haven't tested this though. In case the core PR is merged before the feature freeze, feel free to merge this after having a short look at it @kaikreuzer.
I'll be unavailable from tomorrow until Wednesday next week.

mherwege:
The preview does not make sense to me. The preview is meant to be able to test the blocks in the Blockly rule editor and verify the generated js code, as well as if the blocks properly connect. You removed the ability to test, which I think is still useful, just leaving a preview window.
I think you should either keep the preview as is, also for file based libraries, or completely disable it for file based blockly libraries.

And I indeed do think there is an existing bug, when creating a test Blockly rule, the blockly library itself should not go dirty.

Nadahar:
@mherwege The changes I did to the preview window are very easy to revert or disable. I just couldn't figure it out. I spent a long time trying to "discuss" this with both Copilot and Gemini, without becoming much smarter. There's a read-only property that can be set for the preview window, and when you do, it becomes just a blank canvas. I assume this is something that Google/Blockly does, it creates a new, blank workspace when you enable read-only from what I understand. 

But, since you have no use for a blank canvas where you can't do anything, I had to find a way to populate it with the blocks from the library. If the "live"/editable preview window can be allowed to exist and used without affecting the block library, that would absolutely be the best, but why does it become dirty?

If you want to test/look at this, I would very much appreciate it - as I don't know quite what to do without having to familiarize myself with how the whole Blockly system works - something that isn't very tempting since I don't think I'll even be remotely tempted to use it.

mherwege:
In the quick tests I did, the dirty flag on the preview incorrectly propagates to the library editor anyway, but that is not specific to read only libraries, that is a general error. You don't have to fix this bug as it is not introduced with your code. Let's look at that separatly and keep the preview for all block libraries as is.

Nadahar:
There's a specific reason why it becomes dirty, `blockly-editor` has `@change="dirty = true"` set. It is easy enough to remove, I just wonder why it was set in the first place. This seems to have been the case since block libraries was introduced: 5bf193dd73698d801920155cd5005aaf9a47a353

Nadahar:
@mherwege I think it's just wrong. I put a watch on both `blocksDefinition` and `blocks`, and neither change no matter what I do in the preview window (not that I know fully what can be done there). We can probably just remove the "change means dirty directive".

mherwege:
> There's a specific reason why it becomes dirty, `blockly-editor` has `@change="dirty = true"` set. It is easy enough to remove, I just wonder why it was set in the first place. This seems to have been the case since block libraries was introduced: [5bf193d](https://github.com/openhab/openhab-webui/commit/5bf193dd73698d801920155cd5005aaf9a47a353)

I didn't look at it carefully, but I expect this component to be used when actually creating Blockly rules as well. It won't be in a popup, but use the full screen. And then it definitely needs to be setting dirty. It is only when it is used called from the Blockly library preview that it shouldn't happen. The @change emit should be ignored by the library editor.

Confirmed, it is called from `script-edit.vue` and `transformation-edit.vue`, and that is where the real uses are to create rule scripts.

Nadahar:
> The @change emit should be ignored by the library editor.

Yes, but the "attribute" I suggested removing *is* in the library editor, so it will have no consequences elsewhere, as far as I can understand.

I have done it now, and is testing while still having the watches enabled. As far as I can tell, everything behaves.

There's another problem though, and that's a conflict between Blockly and Vue's reactivity, leading to lots of console errors and failure for the blocks to attach. I think I remember that there was something similar after the upgrade to Vue3 for the "rules" blockly editor, which was fixed. The block library editor was probably forgotten, so that it is still bugged. I'm trying to see if I can figure it out.

Apparently, Vue wraps the workspace in some proxy objects that makes the blockly code fail, because it uses a lot of instance comparison etc. It probably worked with Vue2 because reactivity wasn't "everywhere", but in Vue3 it is. So, the solution is probably to find a way to make the workspace not reactive.

Nadahar:
I changed my mind about the reactivity conflict. This must be solved within `blockly-editor.vue` from what I can see, and that's just too scary for me, since it's a shared component used in other places as well. I'll just leave that bug be and let somebody more capable solve that some other time.

rkoshak:
> . I'll just leave that bug be and let somebody more capable solve that some other time.

It's probably a good idea to create an issue with all the details you know about it right now so the info doesn't get lost top time.

Nadahar:
I changed my mind again, I discovered that the reactivity problem wasn't in `blockly-editor.vue`, it was in `block-preview.vue`, which is only used by the block library editor. So, I disabled reactivity for the workspace, and the errors are gone. I think all the "quirks" are solved now, the preview dialog is back in full, changes don't propagate to the dirty flag in the library editor and the reactivity problems are gone.

> It's probably a good idea to create an issue with all the details you know about it right now so the info doesn't get lost top time.

Sure, except that I think I solved it. That said, there wasn't much details needed - only to open the browser console and view block libraries - the console would be flooded with errors.

Nadahar:
This is the preview dialog for a read-only block library now:

image

Nadahar:
@florian-h05 I guess my way through combining the CSS changes from #4259, and the layout doesn't seem broken to me, so I think maybe it worked out.

Nadahar:
The PR this was awaiting has been merged, and there's very little "feature" in this PR, so I'd more or less consider it a "bugfix". What it does is make MainUI aware of read-only block libraries so that they aren't editable, and to make the "code" that is shown consistent with the "new YAML" format used elsewhere. It also fixes a bug with the Blockly renderer and reactivity in the preview - this bug has been around since the transition to Vue3. It also fixes the fact that changes in the preview dialog used to make the block library "dirty" despite not actually changing anything.

@florian-h05 Any thoughts about whether this could/should be included in 5.2.0?

florian-h05:
I think it is reasonable to merge this for 5.2 :+1:

florian-h05 (approved):
LGTM, thanks! I fixed some overflow issues.

PR #4290, opened 2026-06-18
Unhide "scripts"/triggerless rules from "Rules" view

Nadahar:
As was discussed here: https://github.com/openhab/openhab-core/pull/5462#issuecomment-4187804231, there seem to be some agreement that we should move away from the "script" concept, even though the exact road to get there isn't clear.

There is no technical difference between a triggerless rule and a "script", except that a "script" has the `Script` tag applied to it. This tag triggers specific functionality in MainUI that hides the rule under "Rules" in `rules-list.vue` and makes it show up under "Scripts" instead. However, "Scripts" doesn't allow the user to access the rule itself, it opens the script editor for the first `Action` directly.

 DSL rules that have the `.script` extension automatically gets the `Script` tag applied. For other scripting languages, you must specifically create it under "Scripts" to have the tag applied - but it also prevents you from seeing the YAML view, modifying various details, and adding triggers or conditions. If you create it under "Rules" instead, you don't have all these restrictions, and the only thing you can't do with such a rule is to add the `Script` tag, because that tag is "banned" and is ignored if you try.

This PR simply removes the hiding of rules with the `Script` tag from the "Rules" view and unbans the `Script` tag so that it's possible to add manually. Apart from that, nothing else is changed, "Scripts" behave as before.

I see this as a "soft" first step towards removing the artificial difference in treatment between rules with and without the `Script` tag.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/tags/tag-input.vue
+++ b/bundles/org.openhab.ui/web/src/components/tags/tag-input.vue
@@ -72,7 +72,7 @@ export default {
       this.pendingTag = ''
       // Block adding of Scene or Script tags in the wrong editor
       // Adding them would otherwise lead to a situation where the rule/scene/script is not visible in the UI
-      if ((!this.inScriptEditor && newTag === 'Script') || (!this.inSceneEditor && newTag === 'Scene')) {
+      if ((this.inSceneEditor && newTag === 'Script') || (!this.inSceneEditor && newTag === 'Scene')) {
         return
       }
       if (newTag && this.item.tags.indexOf(newTag) === -1) {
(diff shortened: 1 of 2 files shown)


florian-h05:
I would like to discuss this properly in an issue before changing it, also with an outline what to change and how.

Nadahar:
Sure, I had just had this in the back of my mind for a while that it would be a nice "first step" so that "Scripts" weren't suddenly removed before users had "gotten used to" them showing in "Rules", and then suddenly the 5.2.0 release was getting very close, so I just did this.

The thing is that the whole thing is quite unnatural as it is, because you can create rules without triggers already. The only difference between that and a script is the `Script` tag, and the fact that once you have created something "as a script", you're not allowed to use conditions or add a trigger later.

The exception is DSL rules, where `.script` files are automatically tagged with `Script` (so that they end up under "Scripts"), so those you "can't free" from the limitations of the "Scripts view". But, since triggerless `.rules` files are now allowed, you can just "convert" it to a `.rules` file instead, which is the exact same thing, only that it shows up under "Rules" instead.

To me, changing this just means getting rid of an inconsistency/obstacle.

PR #4292, opened 2026-06-20
Fix scrolling of "Copy File Definition" popup on narrow screens & Fix DSL "but only if" syntax highlighting

### LINKED ISSUE #4293: "but only if" not all in red in Main UI code tab
For a DSL rule, I would expect to see the text "but only if" with the same (red) color..

Currently, "if" is in red but "but only" is not.

Image

Nadahar:
The title says it all. It turns out this is a known F7 issue, with popups that needs scrolling when the popup becomes "full screen", and the fix seems to be pretty simple.

Wide screen:

image

Narrow screen:

image

Fixes #4293


Nadahar:
I added one more tiny fix to this PR, the syntax coloring of the "but only if" phrase. It now looks like this:

image

Nadahar:
Thanks for #4301 @jimtng, I've rebased this PR and re-compiled the grammar, but it seems that my "manual attempts" had been mostly successful, the only difference was a tiny formatting change.

Nadahar:
Lol, oxfmt didn't like the formatting change, so after formatting, it is identical to what it was. I therefore just removed https://github.com/openhab/openhab-webui/commit/a7455c1511939446d705184d4260ff9e90534935 from the PR again. It looks like it's fine the way it is.

florian-h05 (approved):
LGTM, thanks for the fixes. I will merge using rebase to retain the individual commits - I don't think we should squash those.

PR #2623, opened 2024-06-24
Rules list: Improve performance of rule status tracking

florian-h05:
Fixes #2621.
Fixes #2593.

This improves performance of the SSE event source event processing for rule status changes by using a separate object for the rule statuses (allows O(1) modification) instead of finding the rule in the rules array (O(n) for linear search) and ignoring RUNNING status.
The RUNNING status can be ignored IMO because rules tend to run such a short time that you don't see the state shown as RUNNING, so we can use this "trick" to reduce the amount of processed events.
The rule-status-mixin was updated with short circuit evaluations for the IDLE status, which is very likely the most processed status, and therefore reducing the amount of comparisons done.

PR #4296, opened 2026-06-20
Rules list: Resurrect RUNNING status

Nadahar:
`RUNNING` status for rules in `rules-list.vue` was disabled in #2623 for performance reasons. I question the idea, the sending and receiving of the events are still there, the only thing that is "saved" is the actual rendering. I certainly can't notice any difference in performance when I re-enable the status badge. There's a limit to how many rules can be shown on the screen at the same time, and I assume that Vue or the browser is smart enough to not render changes to elements that aren't visible. There's no animation or other "heavy processing", it's just a simple update of what's shown on the screen.

It's true that it's often so quick that it's not noticeable, but that depends entirely on what the rule does. And, perhaps especially, in those cases where the rules run for longer, it might be important that the user can see that. It might or might not be "by design". The fact that a rule says `IDLE` when it's running is also somewhat of a "lie". I don't think that's a good thing, when information is presented, it should be correct.

So, I'm hereby suggesting to re-enable `RUNNING` status for rules in `rules-list.vue`.

RulesListRunning

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rules-list.vue
@@ -689,8 +689,6 @@ export default {
           case 'state':
             const uid = topicParts[2]
             const newStatus = JSON.parse(event.payload)
-            // skip status updates for RUNNING for performance reasons (can be easily skipped as it was never really shown due to the short execution time of rules)
-            if (newStatus.status === 'RUNNING') return
 
             if (!this.ruleStatuses[uid]) this.ruleStatuses[uid] = {}
             this.ruleStatuses[uid].status = newStatus.status


JustinGeorgi:
> The fact that a rule says `IDLE` when it's running is also somewhat of a "lie". I don't think that's a good thing, when information is presented, it should be correct.

I agree that if `RUNNING` is not an option in the for the badge, then `IDLE` is misleading and should probably just be `ENABLED`. On the other hand, I also agree that the information is just not that useful on the rules list page. It's absolutely critical on the rule detail page when you are editing a rule, but here in the whole list such transient information is of very limited use.

It would be a much larger change (presumably with some core updates as well), but what about some tracking of the rule's last run and duration which is then displayed for each rule in the list? That way accidental long-running rules can still be identified, but the information isn't transient or even dependent on the page being open at the right time. That also helps identify rules that are not triggering when the user is expecting them to (or triggering when they are not expected to).

Nadahar:
I'm sure there are other things that could be done and would be useful, but I just don't see a reason to hide this information. It used to be there, and as far as I can tell, there's no reason why I can't still be. Additional information for diagnostics could be made in addition, but it would be a somewhat larger project because you not only have to gather and keep the information, but figure out how to communicate it.

I actually think it's very useful if you can see in the overview page that a rule is stuck in `RUNNING`.

Also, the badge doesn't only show enabled/disabled. It shows the state of the rule, uninitialized, initializing, idle, running, template error and perhaps others I don't remember. So, reducing it to just show enabled/disabled would be removing a lot more information.

HolgerHees:
just a few words from my side. I reported this bug/feature.

In the past I saw from time to time the state Running when I was scrolling through the rule list. In this moment of scrolling, not useful, just that I saw, a lot of things happens.

Most confusing thing was for me the state "Idle", because I was still expecting a state "Running" sometimes.

Another aspect was, that in this list you can identify endless running rules. 

The situation would improve a lot if we just have the state Enabled/Disabled to avoid confusion. 

The cherry on the cake would be the "running" state, which also looks pretty cool, as you can really see how much is going on. And you can identify endless rules.

florian-h05:
> There's a limit to how many rules can be shown on the screen at the same time, and I assume that Vue or the browser is smart enough to not render changes to elements that aren't visible. 

That’s not correct. Normal lists render all entries. If only the visible entries should be rendered, we need to use virtual lists, which is done only for the item list.

> There's no animation or other "heavy processing", it's just a simple update of what's shown on the screen.

Before #2623 O(n) operations were needed to update the state. After #2623 only O(1) operations should be needed, so this should be fine to re-enable RUNNING. 

Out of interest: How many rules do you have?

HolgerHees:
I'm have 141 active rules.

Nadahar:
> That’s not correct. Normal lists render all entries. If only the visible entries should be rendered, we need to use virtual lists, which is done only for the item list.

As I said, it was an assumption, but I don't understand what it means to render something that is outside the visible screen. Does it mean that the list is actually redrawn, without changes, if the status of a rule not shown changes?

If this is solved by using virtual lists, what is the downside? Why not use virtual lists "everywhere" then?

Nadahar:
One more thing to note: as far as I can understand, the current solution only removes the update *to* `RUNNING`. The update back to `IDLE` is still performed, even though no visible change happens. If you have 1000 rules and this causes re-rendering for out-of-view rules, it could still be a significant number if renderings - depending on how often the rules are triggered of course.

florian-h05:
> If this is solved by using virtual lists, what is the downside? Why not use virtual lists "everywhere" then?

Virtual lists have the downside that you have to assume or compute a height for every single item for the list to render correctly. As the item height is depending on description, tags etc. that’s not really straightforward. So you use virtual lists only where you expect really many entries (e.g. 1000 Items), in other cases you trade in the optimal performance against better developer experience and more stable rendering (you don’t need to maintain a height calculation).

> The update back to IDLE is still performed, even though no visible change happens. If you have 1000 rules and this causes re-rendering for out-of-view rules, it could still be a significant number if renderings - depending on how often the rules are triggered of course.

Since IDLE -> IDLE doesn’t really change anything I suspect there is no re-rendering happening. I think the algorithmic complexity optimisation in #2623 made the biggest performance impact, removing RUNNING doesn’t change that much in comparison I would think.

florian-h05 (approved):

PR #4297, opened 2026-06-20
Developer sidebar: Fix duplicate keys in events list v-for

### LINKED ISSUE #4295: Developer Sidebar fills browser console with warnings
When I open the developer sidebar and starts streaming events, my browser console quickly fills with warnings like this:

```
[Vue warn]: Duplicate keys found during update: 1781981636963 Make sure keys are unique. 
  at <F7List media-list="" > 
  at <F7Block class="no-margin no-padding" > 
  at <F7Block class="developer-sidebar" > 
  at <DeveloperSidebar key=2 activeToolTab="events" searchFor=null > 
  at <F7PageContent ptr=false ptrDistance=undefined ptrPreloader=true  ... > 
  at <F7Page class="developer-dock" > 
  at <DeveloperDock dock="tools" helpTab="current" toolTab="events"  ... > 
  at <AsyncComponentWrapper dock="tools" helpTab="current" toolTab="events"  ... > 
  at <F7Panel key=1 right="" visible-breakpoint=1280  ... > 
  at <App>
<html>
<body>

Promise.then
  (anonymous) @ developer-sidebar.vue:1187
  (anonymous) @ sse.ts:119
```

It almost seems like there's one warning per streamed event. The events (or at least some) still appear, but this doesn't look right. Also, it completely drowns out anything else that might happen in the browser console, making it useless.

Nadahar:
As described in #4295, the developer toolbar fills my browser console with warnings. I also think it means that many events aren't shown. When looking at it, the reason seems pretty obvious: The "current time" in with milliseconds precision is used as an index. Since a millisecond is a long time for a computer, it can and will happen that multiple events are created within the same millisecond, and when that happens, Vue's "index" fails because of duplicated. For me, it happened at an alarming rate, the browser console grew "like crazy".

I simply introduced a counter instead. Simple, but I think, effective. It will have to run for many thousands of years to reach the maximum value, so I think we're safe there.

Fixes #4295.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
+++ b/bundles/org.openhab.ui/web/src/components/developer/developer-sidebar.vue
@@ -502,7 +502,7 @@
           <f7-list media-list>
             <f7-list-item
               v-for="event in sseEvents"
-              :key="event.time.getTime()"
+              :key="event.sequenceId"
               :title="event.topic"
               :subtitle="event.type"
               :footer="event.payload" />
@@ -1184,6 +1184,7 @@ export default {
         '',
         (event) => {
           event.time = new Date()
+          event.sequenceId = this.sseEvents.length > 0 ? this.sseEvents[0].sequenceId + 1 : 1
           this.sseEvents.unshift(...[event])
           this.sseEvents.splice(20)
         }
--- a/bundles/org.openhab.ui/web/src/pages/developer/developer-tools.vue
+++ b/bundles/org.openhab.ui/web/src/pages/developer/developer-tools.vue
@@ -142,7 +142,7 @@
                 <f7-list media-list>
                   <f7-list-item
                     v-for="event in sseEvents"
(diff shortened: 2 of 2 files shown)


florian-h05:
I remember that this problem also occurred in the SSE/WS Connection Tester on the developer tools page. Do you mind checking how I solved it there? Probably replace my solution with a simple counter as well to align the implementations.

Nadahar:
> I remember that this problem also occurred in the SSE/WS Connection Tester on the developer tools page. Do you mind checking how I solved it there? Probably replace my solution with a simple counter as well to align the implementations.

You did this:

```vue
:key="event.time.getTime() + '#' + f7.utils.id()"
```

A simple counter is probably better, I'll modify that as well.

Nadahar:
I've applied the same to the SSE and WS tester events, and tested both without any entries appearing in the browser console.

florian-h05 (approved):

PR #4298, opened 2026-06-20
[MainUI] Rename HABAssist to Chat

kaikreuzer:
## Summary
- update the Main UI sidebar entry title from HABAssist to Chat
- rename assist page i18n title from HABAssist to Chat
- rename assist settings title from HABAssist Settings to Chat Settings


bundles/org.openhab.ui/web/src/App.vue
```diff
@@ -42,9 +42,9 @@
           </f7-list-item>
         </f7-list>
 
-        <!-- Assistant -->
+        <!-- Chat -->
         <f7-list class="admin-links">
-          <f7-list-item link="/assist" title="HABAssist" no-chevron panel-close :class="{ currentsection: currentPath.assist }">
+          <f7-list-item link="/assist" title="Chat" no-chevron panel-close :class="{ currentsection: currentPath.assist }">
```
florian-h05: I think ideally we should use the assist.title I18n message here.

florian-h05 (commented):
In addition to changing the translation strings, I would rename the path and the components accordingly.

bundles/org.openhab.ui/web/src/App.vue
kaikreuzer: That makes me think that we should probably also update the path and the i18n keys to `chat`...

florian-h05 (approved):

PR #4299, opened 2026-06-21
Fix tabbed page navigation issues

digitaldan:
For the past ~2 months i have noticed that on tabbed pages, selecting a tab and then navigating into an element and back (quickly) would sometimes land on /page/myPage/NaN instead of the correct tab index (like 0 or 1) , rendering a blank page, which is super annoying. This now guards against 'NaN' values and uses the previous tab state for tab index selection when returning. Have been running this for a couple days on my production system.


florian-h05 (approved):
That has also annoyed me a lot, thanks for tracing it down and fixing it!

PR #4300, opened 2026-06-21
Rule edit: Update rule code when the source changes

Nadahar:
This is a small fix for something that was forgotten. Under normal circumstances, there are no reasons to update the code in `load()` because the "Code tab" is never the "start tab". But, with a file-based rule, if you already have the "Code tab" open and the source file is updated, there is a need to regenerate the code to make sure that it reflects the changes made to the source file.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
@@ -636,6 +636,16 @@ export default {
           this.loading = false
           if (!this.createMode && !this.stubMode && this.hasOpaqueModule && this.hasSource) {
             this.switchTab('source')
+          } else if (this.currentTab === 'code') {
+            this.resolveEditorTypes().then(() => {
+              if (!this.hasCode) {
+                showToast('This rule cannot be shown in code form because it contains elements that cannot be serialized to YAML or DSL.')
+                this.currentTab = 'design'
+                f7.tab.show('#design')
+              } else {
+                this.$refs.codeEditor.generateCode()
+              }
+            })
           }
         })
       }
--- a/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
@@ -641,7 +641,6 @@ export default {
               if (!this.hasCode) {
                 showToast('This rule cannot be shown in code form because it contains elements that cannot be serialized to YAML or DSL.')
                 this.currentTab = 'design'
-                f7.tab.show('#design')
               } else {
                 this.$refs.codeEditor.generateCode()
               }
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
```diff
@@ -636,6 +636,16 @@ export default {
           this.loading = false
           if (!this.createMode && !this.stubMode && this.hasOpaqueModule && this.hasSource) {
             this.switchTab('source')
+          } else if (this.currentTab === 'code') {
+            this.resolveEditorTypes().then(() => {
+              if (!this.hasCode) {
+                showToast('This rule cannot be shown in code form because it contains elements that cannot be serialized to YAML or DSL.')
+                this.currentTab = 'design'
+                f7.tab.show('#design')
```
florian-h05: Why do we need f7.tab.show? currentTab was enough in the past to control the visible tab ...

Nadahar: I don't know - I've just done the same as is done in `switchTab()`. I don't know why setting `currentTab` would be enough unless that variable is watched, because I thought that was just a variable with no "magical properties". Perhaps there was a mixin or something that did the actual change?

Calling `switchTab()` here isn't desirable, because that would mean that the `resolveEditorTypes()` (an API call) would be executed again, and it has just been called.

florian-h05: `currentTab` decided whether the tab-active class is applied in the Vue template:

```html
      <f7-tab id="design" :tab-active="currentTab === 'design'">
```

Nadahar: You're right - I've removed all the `f7.tab.show` calls, and things still seem to work. But, that means that it should be removed everywhere, not just here?

I didn't "invent it", but I think I started out with the `switchTab()` from `thing-details.vue`, which explains the origin. I'll remove all of those in an additional commit then..

florian-h05: > But, that means that it should be removed everywhere, not just here?

Yes, everywhere where we also have it in the Vue template.

florian-h05 (approved):

bundles/org.openhab.ui/web/src/pages/settings/rules/rule-edit.vue
Nadahar: I meant everywhere in `rule-edit.vue`, I didn't touch other files.

florian-h05: I noticed you only touched rule-edit.vue, but we could to the same change in other files. Something for 5.3….

PR #4304, opened 2026-06-22
Fix widget action regression from #3959

Nadahar:
During refactoring in #3959, the order of two method parameters were swapped, but callers weren't updated. The result is not surprising that things don't work as expected.

(One of) the bug(s) is discussed here: https://community.openhab.org/t/action-context-parameter-groups-are-not-properly-displayed-on-configuration-dialogs/169608


florian-h05 (approved):
Thanks for the fix!

PR #4308, opened 2026-06-24
Block library editor: Fix blocks not refreshed on code change

florian-h05:
This makes the blocks automatically refresh on code changes.
Throttling is used to avoid too much rerendering.


jsjames (approved):
LGTM

PR #4313, opened 2026-06-25
Setup wizard: Fix persistence strategy selection showing invalid translations

florian-h05:


florian-h05:
Self-merging this small fix so it gets into the RC build.

PR #4315, opened 2026-06-25
[blockly] Add blocks to get and set Thing enabled status

### LINKED ISSUE #4312: [blockly] Add block to Enable/Disable Things
## Your Environment

- openHAB version used (check Main UI -> About): 5.2 Milestone 5
- openhab-js version used (run `console.log('openhab-js ' + utils.OPENHAB_JS_VERSION)` from a JS script): Current default version included in 5.2 M5
- Operating System and version (e.g. desktop or mobile, Windows 10, Raspbian Buster, ...): Debian GNU/Linux 12 (bookworm) / Kernel 6.12.75 (64-Bit Raspberry Pi OS) running openHAB via Docker

## Reference Scenario
Often things are hides behind a OnOff Plug. In case the Device is switched Off the Binding of the devices might trigger over and over again the device and create lots of Traffic.

## Is your feature request related to a problem? Please describe.
Currently things require a code block to enable disable things. 
This caused in the past issues an even today I faced an issues with the update on OH 2.5 M5 which created an issue with Multitasking as i put the whole functionality into a script and executed the script from the Main Script.

## What is needed
A Blocky block with 
input: Thing UUID, Should enable (true/false)

In case its true the Thing is enabled.
in case its false the Thing gets disabled.

Thanks for supporting as last time and sorry for the unneeded text blocks here.
@Nadahar: I really want to excuse for the initial version of this Report I

Nadahar:
Resolves #4312.

This is my attempt to solve #4312, but as I know next to nothing about Blockly, somebody will have to test/verify it.

It seems to deliver the desired functionality to me though.

Image

Image

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-things.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-things.js
@@ -62,6 +62,58 @@ export default function defineOHBlocks(f7) {
     return [code, 0]
   }
 
+  Blockly.Blocks['oh_getthing_isenabled'] = {
+    init: function () {
+      this.appendValueInput('thingUid').appendField('get thing enabled status').setCheck(['String', 'oh_thing', 'oh_thingtype'])
+      this.setInputsInline(false)
+      this.setOutput(true, 'Boolean')
+      this.setColour(0)
+      this.setTooltip('Checks if the given thing is enabled')
+      this.setHelpUrl('https://www.openhab.org/docs/configuration/blockly/rules-blockly-items-things.html#get-thing-enabled-status')
+    }
+  }
+
+  javascriptGenerator.forBlock['oh_getthing_isenabled'] = function (block) {
+    const thingUid = valueToCode(block, 'thingUid', javascriptGenerator.ORDER_ATOMIC)
+    const inputType = blockGetCheckedInputType(block, 'thingUid')
+    const code = inputType === 'oh_thingtype' ? `${thingUid}.isEnabled` : `things.getThing(${thingUid}).isEnabled`
+    return [code, 0]
+  }
+
+  Blockly.Blocks['oh_setthing_enabled'] = {
+    init: function () {
+      this.appendValueInput('thingUid').appendField('set thing enabled status').setCheck(['String', 'oh_thing', 'oh_thingtype'])
+      this.appendValueInput('enabledState').appendField('to').setCheck('Boolean')
+      this.setInputsInline(true)
+      this.setPreviousStatement(true, null)
+      this.setNextStatement(true, null)
+      this.setColour(0)
+      this.setTooltip('Enables or disables the given thing')
+      this.setHelpUrl('https://www.openhab.org/docs/configuration/blockly/rules-blockly-items-things.html#set-thing-enabled-status')
+    },
+
+    onchange: function (event) {
+      if (!this.workspace || this.workspace.isDragging()) {
+        return
+      }
+
+      if (!this.getInputTargetBlock('enabledState')) {
+        this.setWarningText('A True/False value is required here')
+      } else {
+        this.setWarningText(null)
+      }
+    }
+  }
+
+  javascriptGenerator.forBlock['oh_setthing_enabled'] = function (block) {
+    const thingUid = valueToCode(block, 'thingUid', javascriptGenerator.ORDER_ATOMIC)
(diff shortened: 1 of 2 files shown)


Nadahar:
By the way, it's an interesting confusion in the block definitions where `oh_thing` seems to mean `ThingUID` and `oh_thingtype` seems to mean `Thing`. Don't ask me what a `ThingType` would be - `oh_thingtypetype`?

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-things.js
```diff
@@ -62,6 +62,58 @@ export default function defineOHBlocks(f7) {
...
+      this.setColour(0)
+      this.setTooltip('Checks if the given thing is enabled')
+      this.setHelpUrl('https://www.openhab.org/docs/configuration/blockly/rules-blockly-items-things.html#get-thing-enabled-status')
+    }
+  }
+
+  javascriptGenerator.forBlock['oh_getthing_isenabled'] = function (block) {
+    const thingUid = valueToCode(block, 'thingUid', javascriptGenerator.ORDER_ATOMIC)
+    const inputType = blockGetCheckedInputType(block, 'thingUid')
+    const code = inputType === 'oh_thingtype' ? `${thingUid}.isEnabled` : `things.getThing(${thingUid}).isEnabled`
```
florian-h05, line 79: Not sure if we want to make the code more resistant to non existing Things:

```suggestion
    const code = inputType === 'oh_thingtype' ? `${thingUid}.isEnabled` : `things.getThing(${thingUid})?.isEnabled`
```

This will avoid that the execution fails, but will evaluate to falsy, possibly causing unexpected behaviour for users. Might be better to have the execution fail ... WDYT?

florian-h05 (approved):

bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-things.js
Nadahar: Yeah, I think that an error is probably better than a "wrong result".

florian-h05:
> By the way, it's an interesting confusion in the block definitions where oh_thing seems to mean ThingUID and oh_thingtype seems to mean Thing. Don't ask me what a ThingType would be - oh_thingtypetype?

The main issue with block IDs is that once you have them defined you cannot change them. This would break existing blocklies as Blockly wouldn't be able to understand its XML. So someone chose a kind of unfortunate naming years ago and we have to stick with it ...

Nadahar:
While not technically a bugfix, it provides access to a function that's already in the helper library. Should these blocks wait for 5.3.0, or be backported?

florian-h05:
I've rejected backporting another enhancement in the last days, so either I have to also backport both or none.

Nadahar:
> I've rejected backporting another enhancement in the last days, so either I have to also backport both or none.

I've never understood the OH backporting policy and what it's supposed to achieve, but since it's none of my business, all I can do is ask.

As I see it, releases are too far apart to wait for "fixes". This PR is a response to a concrete problem found on the forum, and for it to have any value for those having this problem, it must be available before Yule. I'm completely on-board with the idea that entirely new "features" must wait for the next release, but I think the rules should be much more "relaxed" right after a release than when you start getting closer to a new one.

There is so much stress before a release, that many details don't make it or are forgotten, or nobody has thought about it. You typically discover many of these things right after release, and it makes sense to be able to mend and release that, even though they aren't "critical bug fixes". Whether you call it an "enhancement" or not is subjective. In the strictest/OH sense, it seems that an enhancement is anything that isn't strictly a bugfix. I see a third category, that is "filling in holes" - providing things that "should have been there", and that should be allowed 1–2 months after a new release, at a minimum, as long as it doesn't risk breaking anything.

From what I've seen in response from maintainers, the consensus among maintainers are very different. So, all I can do is ask, but when I think something shouldn't have to wait, I will do so. I think it might also be in its place to question what the goal of the "strategy" is. If it's only to make sure that there are "enough enhancements" to "qualify" for a new release, so that they must be "saved", I think that's the wrong focus. What's important to the users is that they have the tools they need for the system to work well, not that a new release is a big "wow factor".

florian-h05:
It’s definitely not to „save“ features for the big release. When I became maintainer, I was told that we only backport bug fixes. Seeing other maintainers also backport some enhancements, I think it is justified to backport this PR here.

PR #4317, opened 2026-06-26
Chat page: Show a splash screen if no LLM-based HLI is available

florian-h05:
Note: Do not explicitly require gemini HLI as this would prevent us from shipping other LLM-based HLI implementations through the marketplace.

image


florian-h05:
@jsjames Whilst this is a feature and we are in feature freeze, I would merge this before the RC. For new features, we usually allow enhancements and tweaks during feature freeze.

jsjames (approved):
LGTM

PR #4326, opened 2026-06-30
Fix actionParams prefix given reorder of parameters

jsjames:
Regression from #4304


jsjames:
@florian-h05, this should get in to 5.2R2 since the taphold config settings (plus others) won't be an option in the UI.

florian-h05 (approved):
LGTM, thanks.

PR #4340, opened 2026-07-01
Fix OpenAPI spec missing optional slot field for `UiComponent`

jsjames:
- Updates the openapi spec to the latest from RC1
- applies fixes to the spec to address missing optional slots in the UIComponent (https://github.com/openhab/openhab-core/issues/5686)
- Also makes the slots optional in the RootUIComponent (since it is inherited from UIComponent)
- fixed several typescript errors due to the change

@florian-h05, I'd like to get swagger to generate the slots correctly, but per the above issue, it is confused due to the recursive nature. Hopefully we can address that and then we can remove the tweaking of the openapi spec.

### DIFF
--- a/bundles/org.openhab.ui/web/build/generate-rest-client.mjs
+++ b/bundles/org.openhab.ui/web/build/generate-rest-client.mjs
@@ -36,7 +36,7 @@ if (url) {
   process.stdout.write('DONE\n')
 
   try {
-    // temporary fix for UIComponent schema in OpenAPI spec - see https://github.com/openhab/openhab-core/issues/5686
+    // TODO: temporary fix for UIComponent schema in OpenAPI spec - see https://github.com/openhab/openhab-core/issues/5686
     process.stdout.write(`Updating UIComponent in ${file} ... `)
     execSync(`jq '.components.schemas.UIComponent = input' ${file} ./build/UIComponent.json > tmp.json &&  mv tmp.json ${file}`, {
       stdio: 'ignore'
@@ -47,7 +47,7 @@ if (url) {
   }
 
   try {
-    // temporary fix for RootUIComponent schema in OpenAPI spec
+    // TODO: temporary fix for RootUIComponent schema in OpenAPI spec
     process.stdout.write(`Updating RootUIComponent in ${file} ... `)
     execSync(`jq '.components.schemas.RootUIComponent = input' ${file} ./build/RootUIComponent.json > tmp.json &&  mv tmp.json ${file}`, {
       stdio: 'ignore'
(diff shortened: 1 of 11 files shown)


jsjames:
@florian-h05 - i'm not sure why the Build failed - here?  Also, I tried adding the rebuild label to see if this was just a fluke, but that doesn't seem to have any effect.

jsjames:
> @florian-h05 - i'm not sure why the Build failed - here? Also, I tried adding the rebuild label to see if this was just a fluke, but that doesn't seem to have any effect.

You can ignore this, this was just a fluke in the build system so forced a new version and it built correctly.

However, I don't think the rebuild label is working correctly - not sure where this is looked at in the build system?

jsjames:
Converting to draft, will wait for the proper fix with https://github.com/swagger-api/swagger-core/pull/5232

jsjames:
@florian-h05 - given that they are taking longer to accept the PR at swagger-core, let's go ahead and merge this PR.  When they update swagger-core and we get that integrated into core, we can then remove the unnecessary patches to the open api spec.  Since I have 2 additional PRs waiting on this one, it will allow moving forward.

bundles/org.openhab.ui/web/build/generate-rest-client.mjs
```diff
@@ -35,6 +35,28 @@ if (url) {
   fs.writeFileSync(file, JSON.stringify(openApi, null, 2))
   process.stdout.write('DONE\n')
 
+  try {
+    // temporary fix for UIComponent schema in OpenAPI spec - see https://github.com/openhab/openhab-core/issues/5686
```
florian-h05: Probably add a TODO in front of the comment to it's more "visible":


```suggestion
    // TODO: temporary fix for UIComponent schema in OpenAPI spec - see https://github.com/openhab/openhab-core/issues/5686
```

bundles/org.openhab.ui/web/src/components/widgets/useWidgetAction.ts
```diff
@@ -368,7 +368,7 @@ export function useWidgetAction(context: Ref<WidgetContext>, config: Ref<WidgetA
                       return el
                     }
                   }
-                  throw new Error('Invalid actionPhotos parameter format in ' + ctx.component.component)
+                  throw new Error('Invalid actionPhotos parameter format in ' + (ctx.component as api.RootUiComponent).component)
```
florian-h05: Why do we need that? ctx.component is either RootUiComponent or UiComponent, both have a component property (string).

florian-h05 (commented):
Apart from two minor comments LGTM.

florian-h05 (approved):

PR #4346, opened 2026-07-04
oh-image - Allow embedding of SVG to oh-image and oh-image-card

### LINKED ISSUE #4324: mutli-select drag/drop does not work on oh-canvas-layout
Somewhere in the migration to vue3, vue-draggable-resizeable, re-factor of oh-canvas-item to ts the multi-select drag-drop function no longer works.

jsjames:
Fixes #4324

This adds support to embed SVGs to any oh-image and oh-image-card (in addition to oh-canvas-layout and oh-plan-page).

In addition:
- converts the svg-embedded mixin to a composable which is reused across these components
- converts oh-canvas-layout and oh-plan-page to composition api and typescript
- adds appropriate svg-embedding fields in definitions (and adds some which were not listed, but are used in the ui)
- uses the stronger typescript typing for the Config parameters
- Adds DOMPurify to eliminate any XSS attacks with embedded SVGs
- Adds attribute data-state to the openHab SVG Elements to simplify CSS styling with [data-state="ON"] format

Addresses build issue caused by: https://github.com/Menci/vite-plugin-top-level-await/issues/79

```
<style>
#HouseOutline[data-state='READY'] {
  fill: RGBA(0, 150, 0, .5) !important
}
#MotionZones [data-state='CLOSED'] {
  fill: none !important
}
[data-state='CLOSED'] {
  stroke: green !important
}
[data-state='OPEN'] {
  stroke: red !important
}
[data-state='UNDEF'] {
  stroke: yellow !important
}
</style>
```

Note:
- For oh-image and oh-image-card, the embeddedSvgActions can only be edited in the code editor and not directly in the UI. To support editing in ui will be more involved as it will require more extensive changes throughout the widget editor and other components.


jsjames:
@digitaldan - I just added an additional data-state attribute which has the actual state of the item. This allows simplifying setting up the css styles and allows more than just an ON and OFF state (for instance, I wanted a separate color for an UNDEF state - see my stylesheet below).

This would eliminate the need for any class style config parameters for an ON and OFF state. Since you would need to define those in the stylesheet anyways, I think using the data-state would be simpler and we could deprecate those settings in the config params?

```
      stylesheet: |
        [data-state="CLOSED"] {
          stroke: green !important
        }
        [data-state="OPEN"] {
          stroke: red !important
        }
        [data-state="UNDEF"] {
          stroke: yellow !important 
        }
```

jsjames:
@digitaldan - before you make any PR, let me work through these co-pilot issues so we don't conflict each others changes.

jsjames:
> @digitaldan - before you make any PR, let me work through these co-pilot issues so we don't conflict each others changes.

FYI, I got the edits for the co-pilot review all in.

jsjames:
@florian-h05 @digitaldan, I think using the data-state attribute in CSS styles is much simpler than specifying classes for styling in the UI (parameters stateOnAsStyleClass and stateOffAsStyleClass). My assumption is that folks that are comfortable adding classes to use for these parameters would be comfortable for writing CSS styles using the format below.  I'd recommend deprecating those two parameters and documenting the below to simplify the code/etc.  Thoughts?

digitaldan:
I think data-state is better, so sounds good to me.  I opened a small PR against this one to support numeric levels using css, for instance tying a dimmer level to opacity or "glow" on elements (i use this to show brightness levels on my floorplan).

jsjames:
> I think data-state is better, so sounds good to me. I opened a small PR against this one to support numeric levels using css, for instance tying a dimmer level to opacity or "glow" on elements (i use this to show brightness levels on my floorplan).

Sounds good.  I'll add in a "deprecated" field in the parameter definitions so we can flag this for 5.3 and then remove down the road.  Unless, @florian-h05 thinks we should just remove and mention in the upgrade notes?

digitaldan:
I think "deprecated" is fine, i would not remove, i assume that would break existing user's SVGs, so i think if its a small amount of overhead, keep supporting it.    FYi, this is working great for me, its powering a AppleTV widget i plan on bundling with the new binding (and thats another discussion, allowing bindings to provide custom widgets.....)

image

digitaldan:
FYI, i'm good with this if @jsjames and @florian-h05 you are too

florian-h05:
@jsjames Can you please rebase?

jsjames:
> @jsjames Can you please rebase?

done

digitaldan:
Any reason to not merge this?  I'm having to keep a custom build of the WebUI just for this feature ;-)

jsjames:
> Any reason to not merge this? I'm having to keep a custom build of the WebUI just for this feature ;-)

@florian-h05 - let me know if there are any changes required on this one.

florian-h05:
It’s a rather large PR and I haven’t found the time yet to review it. I will have a look after next week when all exams are finally finished.

PR #4347, opened 2026-07-05
Thing Add: Show binding logo in binding selection list

jimtng:
Improve the "Choose Binding" screen (Settings -> Things -> Add) by adding the binding icons. This should make it easier to "spot" the binding more quickly.

Downside: makes the list longer so we may need to scroll more.

Before:

image

After:

image

Hovering on the (i) icon shows the binding uid (binding-astro shown in the above screenshot).


florian-h05 (approved):
LGTM!

PR #4348, opened 2026-07-05
Thing Add: speed up the loading of installed binding list

### LINKED ISSUE #2337: Loading the "Add Thing" screen is slow
Clicking on Things -> + leads to the "Choose Binding" page. This page can be quite slow to load, randomly. Usually the first load is slow, then subsequent loads are faster albeit still with a small delay. Then randomly, it would become slow again. This makes for a frustrating experience when trying to add multiple Things.

- Speed up the first load
- Make the subsequent load instantaneous

Could it cache the list of bindings, and show the cache upon first load, and then refresh the list of binding after the page is loaded.
I'm also thinking perhaps once user is on the "Things List" page, start loading the cache of list of binding, so when user clicks Add, it's already loaded / loading.

jimtng:
Fixes #2337

Depends on: openhab/openhab-core#5692

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-binding.vue
+++ b/bundles/org.openhab.ui/web/src/pages/settings/things/add/choose-binding.vue
@@ -97,7 +97,7 @@ export default {
   methods: {
     onPageAfterIn() {
       this.loading = true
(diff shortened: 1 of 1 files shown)


jimtng:
https://github.com/openhab/openhab-core/pull/5692 has been merged!

florian-h05 (approved):

PR #4352, opened 2026-07-09
Add-on Details: Fix back link on non-official addons, set target to type subpage

jimtng:
Non-official addons ids don't have the addon type prefix e.g. `marketplace:xxxx` or `jar:xxxx` instead of `binding-hue`, or `automation-groovy` so the code that extracts this prefix failed, leading to an incorrect path.

This PR extracts the addon type directly from the addon object instead of relying on the addon id pattern.

It also returns the correct back link title depending on the addon type.


florian-h05 (approved):
LGTM, thanks!

PR #4357, opened 2026-07-10
Support URL parameter `hideLeftPanel=true` to hide Main UI menu for kiosk displays

### LINKED ISSUE #4356: Add a supported way to hide the Main UI sidebar for kiosk / wall display setups
### Problem

I am using openHAB Main UI on a dedicated kiosk / wall display system, running Chromium in kiosk mode on a Raspberry Pi. The device is intended to operate without a mouse and keyboard, and in some cases without convenient touch interaction.

With openHAB 5.2, the left Main UI sidebar can remain visible depending on the effective viewport width / breakpoint behavior. For a kiosk or wall display, this is undesirable because the sidebar consumes valuable screen space and there is currently no obvious supported way to permanently hide it via configuration, URL parameter, or page setting.

The current workarounds are not ideal for unattended kiosk systems:

- manually unpinning / hiding the sidebar once with mouse, keyboard, touch, or VNC
- manipulating browser LocalStorage manually
- changing Chromium scaling to force a different responsive breakpoint

I found that the following LocalStorage value can hide the sidebar:

```js
localStorage.setItem('openhab.ui:panel.visibleBreakpointDisabled', 'true')
```

However, this is an internal browser-side setting and not a documented or supported kiosk configuration option.

### Use case

A typical setup is:

- Raspberry Pi or similar device
- Chromium / Chrome in kiosk mode
- openHAB Main UI page displayed full-screen
- no keyboard or mouse attached
- fixed wall-mounted display, for example 1024×600, or similar
- display should show only the selected openHAB page content, without the persistent Main UI sidebar

For such installations, it would be very useful to have a supported way to hide the sidebar automatically.

### Requested feature

Please add an official and documented way to hide or disable the Main UI sidebar for kiosk / wall display scenarios.

Possible implementation options could be one or more of the following:

1. **URL parameter**

```text
/page/my-page?hideSidebar=true
```

or:

```text
/page/my-page?kiosk=true
```

2. **Page configuration option**

A setting in the page configuration, for example:

```yaml
hideSidebar: true
```

or:

```yaml
kioskMode: true
```

3. **Main UI / user preference**

A persistent user preference in Main UI settings, exposed through the UI, which controls whether the sidebar should be hidden or whether the visible breakpoint should be disabled.

4. **Dedicated kiosk mode**

A documented kiosk mode that hides the sidebar and optionally other navigation elements, so a selected page can be displayed cleanly on wall-mounted displays.

### Expected behavior

When the option is enabled:

- the left Main UI sidebar should not be displayed persistently
- the selected page should use the full available screen width
- the behavior should be stable after browser restarts
- it should not require manual interaction with the UI after boot
- it should work in Chromium kiosk mode
- it should be documented for wall display / kiosk installations

### Current workaround

At the moment, I can only achieve the desired behavior by setting the following LocalStorage value for the openHAB origin:

```js
localStorage.setItem('openhab.ui:panel.visibleBreakpointDisabled', 'true')
```

This works, but it is not ideal because:

- it depends on an internal implementation detail
- it may break in future openHAB versions
- it requires scripting through browser DevTools Protocol or manual setup
- it is not discoverable for users building kiosk dashboards

### Why this would be useful

Kiosk and wall display setups are common in smart home installations. In these setups, the available screen space is limited and the UI should be predictable after every reboot. Having an official way to hide the sidebar would make openHAB Main UI much easier to use on dedicated displays.

### Environment

- openHAB version: 5.2
- UI: Main UI
- Browser: Chromium in kiosk mode
- Device: Raspberry Pi
- Display examples: 800×480 / 1024×600
- Input devices: no mouse / keyboard during normal operation

jimtng:
Resolves #4356

### DIFF
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -1275,6 +1275,9 @@ export default {
           if (page.route.url.includes('/log-viewer/')) {
             this.logViewerPageActive = true
           }
+          if (page.route.query.hideLeftPanel === 'true') {
+            f7.panel.get('left').disableVisibleBreakpoint()
+          }
         }
       })
 


florian-h05 (approved):

florian-h05:
Please add this to the Main UI docs.

PR #4358, opened 2026-07-12
Add-ons Store: Add a refresh button for explicit add-on list reload

jimtng:
Allow users to manually trigger an immediate reload of the add-on list, so that core will re-fetch the list of remote add-ons (e.g. marketplace add-ons) and return a fresh result.

The reload can be done through either of these two mechanisms:
- A refresh FAB in the bottom right corner
- Browser refresh (typically F5, Cmd+R, or the refresh icon on the browser)

Previously the add-on list was reloaded in core based on a 15-minute expiring cache. This resulted in random stutterings when loading the add-ons store page. Additionally, the user must wait for this cache to expire before they see the updated list of remote addons, and they don't know exactly when this happens.

Depends on https://github.com/openhab/openhab-core/pull/5710 

The core PR removes the expiring cache mechanism and provides a way for the UI to request a refresh on demand.

This improves responsiveness and gives users predictable control over when the data is refreshed. This would be very handy if you're posting or updating a marketplace addon and want to test it - you can instantly refresh the listing without waiting for the "15 minute" cache to expire or resorting to restarting the addon.marketplace bundle to force a refresh.

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
+++ b/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
@@ -298,6 +298,8 @@ import { AddonIcons, AddonTitles, AddonSuggestionLabels, AddonConnectionTypes, A
 
 import { useRuntimeStore } from '@/js/stores/useRuntimeStore'
 
+let handledReloadInCurrentDocument = false
+
 export default {
   mixins: [AddonStoreMixin],
   props: {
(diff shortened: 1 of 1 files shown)


jimtng:
The prerequisite core PR has been merged, so this one is ready to proceed.

jimtng:
I'm investigating also supporting browser refresh into this pr.

jimtng:
> I'm investigating also supporting browser refresh into this pr.

Browser refresh is now also supported. I've added this information to the original post.

florian-h05 (approved):

Issue #4135, opened 2026-04-27
make tab row behavior configurable


mstormi:
I’m using tabbed pages like many do. Ever felt annoyed by the tabs row disappearing?
It always goes away when I scroll down, and to make it return, I have to scroll up first.
I guess anf heard this annoys quite some more people.

It probably can be easily changed via some f7 config option.
Please make it a configurable OH UI option if the tab row is always there or if the current dynamic behavior applies.
Please also provide text config option for that so one can preconfigure an OH instance to have a static row right from the start.

florian-h05:
> Please make it a configurable OH UI option if the tab row is always there or if the current dynamic behavior applies.

Addressed in #4361.

PR #4361, opened 2026-07-14
Add local setting to disable hiding tabs when scrolling

Nadahar:
I didn't go very in-depth with this, I just mimicked what I found was already there. The name of the option, the description string, the placement, these things should all be evaluated - since I just "picked something that seemed reasonable".

Except for that, it does what the title says, it allows disabling the hiding of "bars" (I suspect that's more than just the tab bar) when scrolling, for those that find that annoying/impractical. Unfortunately, it's only a local storage setting, so one must do it over and over again on every browser/device, but that problem is shared with all the other "UI options".

Partially addressed #4135.


florian-h05 (approved):
LGTM, thanks!

PR #4363, opened 2026-07-15
Charts: Automatically create missing axes for oh-aggregate-series

florian-h05:
... and set the `chartType` to `day` (dynamic is not supported).


jsjames (approved):
LGTM

florian-h05:
@jsjames Should we consider backporting this?

PR #4365, opened 2026-07-16
Various add-on related fixes

Nadahar:
There are some small tweaks I've made while testing against https://github.com/openhab/openhab-core/pull/5720. If merged, it can lead to minor changes like that e.g. `countries` aren't in the JSON if none are specified. MainUI isn't ready for that, so a few tweaks are needed.

But, these changes don't rely on https://github.com/openhab/openhab-core/pull/5720, they should work equally well without that being merged, so there's no reason for this PR to wait for https://github.com/openhab/openhab-core/pull/5720. The other way is true though, this should be merged if https://github.com/openhab/openhab-core/pull/5720 is.

### DIFF
--- a/bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
+++ b/bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
@@ -1,7 +1,7 @@
 <template>
   <f7-page @page:beforein="onPageBeforeIn" @page:beforeout="onPageBeforeOut" ref="addondetails" class="page-addon-details">
     <f7-navbar :transparent="true" class="addon-details-nav">
-      <oh-nav-content :title="addon?.label" :back-link="backLinkTitle" :back-link-url="'/addons/' + addon?.type" :f7router />
+      <oh-nav-content :title="addon?.label ?? ''" :back-link="backLinkTitle" :back-link-url="backLinkUrl" :f7router />
     </f7-navbar>
     <f7-block v-if="ready && addon" class="block-narrow addon-details">
       <f7-row>
@@ -231,8 +231,11 @@ export default {
     backLinkTitle() {
       return AddonTitles[this.addon?.type] || 'Add-on Store'
     },
+    backLinkUrl() {
+      return this.addon ? '/addons/' + this.addon.type : '/addons/'
+    },
     showConfig() {
-      return this.addon && this.addon.installed && (this.addon.configDescriptionURI || this.addon.loggerPackages.length > 0)
+      return this.addon && this.addon.installed && (this.addon.configDescriptionURI || this.addon.loggerPackages?.length > 0)
     },
     realAddonId() {
       if (!this.addon) return null
--- a/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
+++ b/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue
@@ -133,11 +133,11 @@ const emit = defineEmits<{
 
 // computed
 const autoHeadline = computed<string>(() => {
-  const likeCount = props.addon.properties.like_count as unknown as number | undefined
+  const likeCount = props.addon?.properties?.like_count as unknown as number | undefined
   if (likeCount && likeCount >= 20) return 'Top'
-  const views = props.addon.properties.views as unknown as number | undefined
+  const views = props.addon?.properties?.views as unknown as number | undefined
   if (views && views >= 1000) return 'Popular'
-  const postsCount = props.addon.properties.posts_count as unknown as number | undefined
+  const postsCount = props.addon?.properties?.posts_count as unknown as number | undefined
   if (postsCount && postsCount >= 15) return 'Hot'
   return ''
 })
--- a/bundles/org.openhab.ui/web/src/js/openhab/sse.ts
+++ b/bundles/org.openhab.ui/web/src/js/openhab/sse.ts
@@ -130,10 +130,7 @@ function newSSEConnection(
         errorCallback()
       }
 
-      // Handle reconnection logic
-      // Note: readyState === 2 is defined as CLOSED in EventSource spec
-      if (es.readyState === 2) {
-        console.log('%c=!= Event source connection broken...', 'background-color: red; color: white')
+      const scheduleReconnect = () => {
         console.debug(`Attempting SSE reconnection in ${reconnectSeconds} seconds...`)
 
         setTimeout(() => {
@@ -148,6 +145,28 @@ function newSSEConnection(
             eventSource = initEventSource() // Reassign the outer scope's eventSource
           }
         }, reconnectSeconds * 1000)
+      }
+
+      // Handle reconnection logic
+      // Note: readyState === 2 is defined as CLOSED in EventSource spec
+      if (es.readyState === 2) {
+        console.log('%c=!= Event source connection broken...', 'background-color: red; color: white')
+
+        // Since SSE hides the response status, do a regular fetch to figure out if our credentials are rejected or not
+        fetch(path, { method: 'HEAD', headers: headers })
+          .then((response) => {
+            if (response.status === 401) {
+              console.debug(`SSE: The server responded with "${response.status}", aborting reconnect.`)
+              es.close()
+              es.clearKeepalive()
+            } else {
+              scheduleReconnect()
+            }
+          })
+          .catch((err) => {
+            console.debug(`SSE: Coudn't connect to the server, scheduling reconnect: ${err}`)
+            scheduleReconnect()
+          })
       }
     }
 
--- a/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
+++ b/bundles/org.openhab.ui/web/src/pages/addons/addons-store.vue
@@ -532,9 +532,10 @@ export default {
       if (this.regionReady) {
         if (this.regionType === 'exclude_other') {
           isInRegionFilter =
-            addon.countries.length > 0 ? addon.countries.map((c) => c.toUpperCase()).includes(this.region.toUpperCase()) : true
+            addon.countries?.length > 0 ? addon.countries.map((c) => c.toUpperCase()).includes(this.region.toUpperCase()) : true
         } else if (this.regionType === 'only_region') {
-          isInRegionFilter = addon.countries.map((c) => c.toUpperCase()).includes(this.region.toUpperCase())
+          isInRegionFilter =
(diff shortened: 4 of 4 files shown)


bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
```diff
@@ -231,8 +231,11 @@ export default {
     backLinkTitle() {
       return AddonTitles[this.addon?.type] || 'Add-on Store'
     },
+    backLinkUrl() {
+      return this.addon?.type ? '/addons/' + this.addon.type : '/addons'
```
jimtng: Is there a case where addon.type is undefined, when addon itself is defined?

Nadahar: I guess there shouldn't be, but yes, I got errors in the browser console complaining that it was undefined. I think perhaps it happens during initialization/loading, where maybe an empty `addon` object is created before it is populated. I didn't investigate, I just did this, and got rid of the error.

Nadahar: I've checked though the code, and can't find that it's initialized to `{}` anywhere, so I can't really explain how it can happen. Perhaps the REST API can return an empty object at times?

jimtng: I suppose this can't hurt. I'm just curious because I added that recently :( not anticipating that it could be undefined, and not knowing that JS string + undefined results in "stringundefined" 🤦

Nadahar: I checked in core, and there is an explicit check in the constructor that `type` isn't empty, so it shouldn't be possible.

Nadahar: I removed the `type` check and browsed around without getting any errors in the console, so I don't know what caused the errors I saw earlier. But, `addon` itself can be undefined, so that check needs to be there.

bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
```diff
@@ -232,7 +232,7 @@ export default {
       return AddonTitles[this.addon?.type] || 'Add-on Store'
     },
     backLinkUrl() {
-      return this.addon?.type ? '/addons/' + this.addon.type : '/addons'
+      return this.addon ? '/addons/' + this.addon.type : '/addons'
```
jimtng: I'm curious again. What's the difference between this and the original code i.e. 

this.addon?.type 
vs
this.addon ? ... : ....

Furthermore, I tried going to http://192.168.1.10:8081/addons and I got a 404, so the trailing `/` is needed -> therefore the original code was fine as is.

jimtng: the original code:

```
 <oh-nav-content :title="addon?.label" :back-link="backLinkTitle" :back-link-url="'/addons/' + addon?.type" :f7router />
```

Nadahar: You're right. The `/` needs to be there, so I don't know anything anymore. I changed it because it caused errors, but now the "fix" seems worse than the original. So, perhaps there has been some weird cache issues causing those original errors?

Nadahar: I still couldn't get those errors after reverting the whole change to the backlink, so I just removed that whole change.

Nadahar:
Now the formatter got mad that the line got too long 😖

bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <f7-page @page:beforein="onPageBeforeIn" @page:beforeout="onPageBeforeOut" ref="addondetails" class="page-addon-details">
     <f7-navbar :transparent="true" class="addon-details-nav">
-      <oh-nav-content :title="addon?.label" :back-link="backLinkTitle" :back-link-url="'/addons/' + addon?.type" :f7router />
+      <oh-nav-content :title="addon?.label ? addon?.label : ''" :back-link="backLinkTitle" :back-link-url="'/addons/' + addon?.type" :f7router />
```
jimtng: This is also unnecessary, because `:title="undefined"` or `:title="null"` would simply not render the title, so it's "effectively" the same as `''`

Nadahar: Reverting that, results in this in the browser console:

```
routes.js:102
 [Vue warn]: Invalid prop: type check failed for prop "title". Expected String with value "undefined", got Undefined  
  at <OhNavContent title=undefined back-link="Add-on Store" back-link-url="/addons/undefined"  ... > 
  at <F7Navbar transparent=true class="addon-details-nav" > 
  at <F7Page onPage:beforein=fn<bound onPageBeforeIn> onPage:beforeout=fn<bound onPageBeforeOut> ref="addondetails"  ... > 
  at <AddonDetails key="1784173452820_2" addonId="eclipse:binding-androidtv" f7router= 
Router {eventsParents: Array(1), eventsListeners: {…}, params: {…}, isAppRouter: false, app: Framework7, …}
 > 
  at <F7View url="/" main=true class="safe-areas"  ... > 
  at <App>
```

Nadahar: I didn't "make any of these up", each change was driven by console errors. I can't explain why I got those errors for the backlink, but generally: I wasn't looking for things I thought "might be causing problems", I only changed things that I found to cause problems.

Nadahar: Check the backlink URL in the above message, and you'll see that it doesn't really work out as intended with the backlink either.

Nadahar: JS has a very strange handling, the whole typeless thing is a mess IMO, but because of the whole falsey evaluation thing, I think we imagine that we're dealing with booleans a lot of times when we're actually dealing with strings. It explains why it comes up with those strange strings with "undefined" in them.

jimtng: Yes I see the same warning. Sorry for my questions. I didn't check / notice the warnings earlier.

Nadahar:
@florian-h05 This PR now also fixes MainUI when the implicit user role is disabled. We've had numerous reports on the forum of MainUI just being a "blank page", and I believe this could be the reason. Currently, with the implicit user role is disabled, you don't even get a login screen - there's no way to actually log in.

Also, if you were already logged in, SSE would fail to use authentication, which would lead it to fail constantly.

This whole PR is just bugfixes, some of them "aren't yet" bugs because the core PR isn't merged, but I can't see any harm of making MainUI handle that these fields can be missing even if they aren't today.

The PR now does more than its title indicates, but I'm not sure what to rename it to so that it's "covering" these bugfixes that are a bit all over the place.

I'd say that this should be a strong candidate for a backport.

Nadahar:
@openhab/webui-maintainers I've now rebased this on latest main and removed the fixes that were "common" with #4371, which should make this very easy to review.

bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <f7-page @page:beforein="onPageBeforeIn" @page:beforeout="onPageBeforeOut" ref="addondetails" class="page-addon-details">
     <f7-navbar :transparent="true" class="addon-details-nav">
-      <oh-nav-content :title="addon?.label" :back-link="backLinkTitle" :back-link-url="'/addons/' + addon?.type" :f7router />
+      <oh-nav-content :title="addon?.label ? addon?.label : ''" :back-link="backLinkTitle" :back-link-url="backLinkUrl" :f7router />
```
florian-h05: Slightly more elegant:

```suggestion
      <oh-nav-content :title="addon?.label ?? ''" :back-link="backLinkTitle" :back-link-url="backLinkUrl" :f7router />
```

Nadahar: Sure, I'm just not comfortable that I know well enough what `??` does to be confident that it will do what I want. But, I guess you are 😉

Nadahar:
I added the code that prevents SSE retries if authentication fails. The idea is that if it fails, it won't magically start working, so to keep retrying is just futile and generates a lot of log noise on both sides, and some traffic. But, since #4371, chances that SSE connections without a valid token will be attempted established has drastically reduced. So, one could question whether there is any point in doing this.

bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
florian-h05: So `addon?.label` either evaluates to the label property of `addon` if `addon` is defined or to `undefined` if `addon` is not defined. If the `addon?.label` expression evaluates to undefined or null (either through addon or addon.label), the  `??` evaluates to its right hand side.

florian-h05:
I think the SSE code doesn't hurt, however it doesn't really belong into this PR here.

bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue
Nadahar: It's just that I don't use these enough to remember exactly what they do, and when I'm not 100% sure, I like to avoid it.

Nadahar:
> I think the SSE code doesn't hurt, however it doesn't really belong into this PR here.

I had all the other SSE stuff here too, the stuff we both fixed. So, it's just that it's the only thing that remains after rebasing. It's a bit tiresome to make a separate PR just for that, if it's about the changelog, can't you just make a two-line commit message when you squash? Or won't that "be the same"?

florian-h05:
If you squash the add-on commits so we end up with 2 commits in your PR, one for add-ons, one for SSE, and add the PR number in braces with a hashtag I can rebase merge and retain those two individual commits.

Nadahar:
> If you squash the add-on commits so we end up with 2 commits in your PR, one for add-ons, one for SSE, and add the PR number in braces with a hashtag I can rebase merge and retain those two individual commits.

Like this?

florian-h05 (approved):

florian-h05:
Yes, exactly like this.

PR #4366, opened 2026-07-16
Thing Details: Improve binding installation link logic

jimtng:
When the Thing status is HANDLER_MISSING_ERROR
- Offer to install the binding only if the binding is truly not installed.
- If it's installed, suggest to check the logs instead.

Please consider backporting to 5.2. Currently the "Install binding" can appear even when the binding is already installed but failed to start. It causes confusion when user clicks install binding but the binding is already installed.


jsjames (approved):
LGTM

florian-h05:
@jsjames WDYT about backporting?

PR #4364, opened 2026-07-15
Home page: Add `hideNavbar` config to control navigation bar visibility

jimtng:
Resolves #4254

image

image

PR #4370, opened 2026-07-18
Support swipe gesture to show the left menu

jimtng:
As discussed in #4364 this PR provides a way to access the left menu when the navigation bar is hidden and the menu is not pinned.

Changes:
- Adds an edge-swipe gesture to reveal the hidden left menu.
- Adds a configuration option to disable the swipe gesture, allowing users to restore the old behavior.


florian-h05 (approved):
LGTM, thanks!

PR #4371, opened 2026-07-19
Various auth related fixes

### LINKED ISSUE #4354: Main UI unusable when implicit user role is disabled
## The problem

Since 5.2.0, when disabling the implicit user role, when the user is not logged in, the UI does not fall back to the login screen anymore but somehow crashes after encountering a 401 on the /rest/tags endpoint.

## Expected behavior

When the implicit user role is disabled and no login session is present, the UI should show the login form.

## Steps to reproduce

Start for example with a fresh openhab in a container:
```
podman run -e "CRYPTO_POLICY=unlimited" -p 127.0.0.1:8080:8080 docker.io/openhab/openhab:5.2.0
```

Then:
* Open the web ui and finish the wizard
* Go to Settings > API security and disable the implicit user role
* Sign out of the admin account

After that, one is locked out of the UI:

Image

## Your environment

```
podman run -e "CRYPTO_POLICY=unlimited" -p 127.0.0.1:8080:8080 docker.io/openhab/openhab:5.2.0
```

## Browser console

see screenshot above

## Browser network traffic

see screenshot above

## Additional information

I briefly mentioned I could not get to the point of SSE in https://github.com/openhab/openhab-webui/pull/4165, being under the impression that this issue was only temporary due to ongoing development:

> This patch is for 5.1.x. I tried to reproduce this with the latest development state on main as well, but there the SSE connection did not even work on the first connection for me, so it will need some further investigation there.

Unfortunately it seems that it made it into the release.

florian-h05:
Fixes #4354: Implicit user role disabled breaks Main UI.

Fixes a regression from #3575 with reverse proxy-based AA. The more advanced error handling in `App.vue::fetchData`` is designed for XHR, not `fetch`.

Also simplifies the `auth.ts` & `auth-mixin.js` code by always sending an access token with SSE and removing the `getRequireToken()` method.

### DIFF
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -907,7 +907,7 @@ export default {
             return Promise.reject('openHAB REST API connection failed with error: ' + err.message || err.status)
           }
         })
-        .then(res => res.data)
+        .then((res) => res.data)
         .then((rootResponse) => {
           // store the REST API services present on the system
           useRuntimeStore().setRootResource(rootResponse)
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -617,6 +617,7 @@ import { useSemanticsStore } from '@/js/stores/useSemanticsStore'
 import { useModelStore } from '@/js/stores/useModelStore'
 
 import { getRoot } from '@/api'
+import { isUnauthorized } from '@/js/hey-api'
 
 const dayjsLocalesGlob = import.meta.glob('../node_modules/dayjs/esm/locale/*.js', { import: 'default' })
 
@@ -915,7 +916,7 @@ export default {
           return useSemanticsStore().loadSemantics(i18n)
         })
         .catch((err) => {
-          if (err === 'Unauthorized' || err === 401) {
+          if (isUnauthorized(err)) {
             console.info('openHAB REST API implicit user role is disabled. Authorizing ...')
             this.authorize(false) // will redirect to auth page, redirecting back to Main UI triggers new load
           } else {
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -618,6 +618,7 @@ import { useModelStore } from '@/js/stores/useModelStore'
 
 import { getRoot } from '@/api'
 import { isUnauthorized } from '@/js/hey-api'
+import { request } from 'framework7'
 
 const dayjsLocalesGlob = import.meta.glob('../node_modules/dayjs/esm/locale/*.js', { import: 'default' })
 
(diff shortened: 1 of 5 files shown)


florian-h05:
@jsjames This should be backported to 5.2.x.

Nadahar:
@florian-h05 Did you even look at what I did in #4365? I bet it's pretty much the same things.

Nadahar:
Ok, I see that you have changed things much more drastically. I actually got the existing logic to work by fixing the fact that it didn't pick up on 401s where it should.

Nadahar:
If there's no need to ever do SSE without a token, your simplification makes sense - I just don't know if that's the case. The existing logic allows not keep sending the token when the implicit user role is enabled. I don't know if the bandwidth use is of any relevance.

florian-h05:
> Did you even look at what I did in https://github.com/openhab/openhab-webui/pull/4365? I bet it's pretty much the same things.

No, I didn't look at it yet because its add-on related fixes acc. to its title and I decided to look at it later. 

> Ok, I see that you have changed things much more drastically.

While looking at issue #4354 I noticed that not only the 401 detection is broken, but very likely the whole detection of reverse proxy authentication in front of openHAB, so I reverted the changes I considered problematic as well.

> I don't know if the bandwidth use is of any relevance.

I don't think it is of any relevance.
I guess when implicit user role is enabled, a good part of Main UI users isn't logged in, so there is simply no access token to send with the SSE request.
If a user is logged in, we send the token with every normal HTTP request -- SSE requests ultimately are HTTP requests that expect an event stream as response type, so it doesn't really differ. Also, an SSE connection is not established that often (the tracking list is updated through a separate HTTP request that carries the access token).

Nadahar:
> No, I didn't look at it yet because its add-on related fixes acc. to its title and I decided to look at it later.

It's not always easy to name a PR. What the PR is, is me browsing around in MainUI and finding warnings and errors in the browser console, which I then address. It started out with doing that for add-ons against my core PR for add-ons, but those changes are just a few `?`s here and there. But, when I found issues in the console that wasn't related to add-ons, I dealt with them as well.

So, it's more of a "universal warnings and errors fixes" PR. It should be very easy to review, the only thing in it that isn't immediately obvious is that I made it stop trying to reconnect SSE if it fails because of invalid credentials. There is just no point in it doing that over and over again, causing "error logging" on both sides. To fix it, the user must log in, after which new SSE's will be initialized, so that particular connection will never connect.

Nadahar:
> I don't think it is of any relevance.
> I guess when implicit user role is enabled, a good part of Main UI users isn't logged in, so there is simply no access token to send with the SSE request.
> If a user is logged in, we send the token with every normal HTTP request -- SSE requests ultimately are HTTP requests that expect an event stream as response type, so it doesn't really differ. Also, an SSE connection is not established that often (the tracking list is updated through a separate HTTP request that carries the access token).

I'm just wondering why it was made that way in the first place. I agree that it shouldn't matter to send the token for those few requests, but when I see something like this, I always assume that there's a reason I don't know about, so I try to make it work without changing what I don't understand. But, I agree with the assessment, it's very hard to see why it would serve any purpose, and the code can be simplified without it.

jsjames (approved):
LGTM

Nadahar:
I hope you guys are planning on backporting this, we can't wait for 5.3.0 to make MainUI work with the implicit user role is disabled.

florian-h05:
Backported 👍

PR #4372, opened 2026-07-20
[mainui] Interactive SVG support in widgets

digitaldan:
I have a new found love for using SVGs in openHAB  ;-) .  

This extends our enhanced SVG support to image widgets.  Our 5.2 release actually did the majority of the work by moving SVG support to a shared mixin, this PR uses that to support interactive SVGs in wdigets.

I originally created a new SVG component and card type, but quickly realized this duplicated much of the existing  image versions, so i instead decided to extend our  `oh-image`  and `oh-image-card` to support the SVG interactive features.

example SVG

```xml
  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 120">
    <rect width="200" height="120" rx="10" fill="#e5e7eb"/>
    <!-- clickable, state-colored lamp: the openhab attribute marks it interactive -->
    <circle id="lamp" openhab="true" cx="60" cy="60" r="30" fill="#9ca3af" stroke="#4b5563" stroke-width="3"/>
    <!-- text label fed with the Item's (display) state -->
    <text x="138" y="66" font-size="16" font-family="sans-serif" text-anchor="middle" fill="#111827">
      <tspan id="lampState" openhab="true">--</tspan>
    </text>
  </svg>
```

Use in a image card

```yaml
component: oh-image-card
config:
  title: Kitchen
  url: /static/demo-lamp.svg
  embedSvg: true
  embeddedSvgActions:
    lamp:
      action: toggle
      actionItem: KitchenLight
      stateItems: [KitchenLight]
      stateOnColor: "#ffb700"
```

Custom widget

```yaml
 uid: svg-lamp-demo
props:
  parameters:
    - name: item
      label: Lamp Item
      type: TEXT
      context: item
      description: Switch Item to control
component: oh-image
config:
  url: /static/demo-lamp.svg
  embedSvg: true
  embeddedSvgActions:
    lamp:                       # id of the clickable element
      action: toggle
      actionItem: =props.item
      actionCommand: ON
      actionCommandAlt: OFF
      stateItems:
        - =props.item
      stateOnColor: "#ffb700"   # fill while ON
    lampState:                  # id of the <tspan>, tracks the Item state
      stateItems:
        - =props.item
      useDisplayState: true        
```

I would definitely want to add some docs about this as creating a GUI to configure this is a future issue.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/standard/cards.ts
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/standard/cards.ts
@@ -167,10 +167,12 @@ export const OhToggleCardDefinition = () =>
 
 // OhImageCard
 import ImageParameters from '../system/image.ts'
+import SvgEmbeddingParameters from '../system/svg.ts'
 export const OhImageCardDefinition = () =>
   new WidgetDefinition('oh-image-card', 'Image Card', 'Display an image (URL or Image item ) in a card')
     .paramGroup(CardParameterGroup(), CardParameters())
     .paramGroup(pg('image', 'Image'), ImageParameters())
+    .paramGroup(pg('svgEmbedding', 'SVG Embedding'), SvgEmbeddingParameters())
     .paramGroup(actionGroup(undefined, undefined, 'Action to perform when the image is clicked'), actionParams())
 
 // OhVideoCard
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/index.ts
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/index.ts
@@ -1,4 +1,4 @@
-import { WidgetDefinition, pt, pb } from '../helpers.ts'
+import { WidgetDefinition, pg, pt, pb } from '../helpers.ts'
 import { actionGroup, actionParams } from '../actions.ts'
 
 const VariableParameter = pt('variable', 'Variable', 'Name of the variable to set on input change')
@@ -53,9 +53,11 @@ export const OhIconDefinition = () =>
   new WidgetDefinition('oh-icon', 'Icon', 'Display an openHAB icon').paramGroup(actionGroup(), actionParams()).params(IconParameters())
 
 import ImageParameters from './image.ts'
+import SvgEmbeddingParameters from './svg.ts'
 export const OhImageDefinition = () =>
-  new WidgetDefinition('oh-image', 'Image', 'Displays an image from a URL or an item')
+  new WidgetDefinition('oh-image', 'Image', 'Displays an image from a URL or an item, optionally as an embedded interactive SVG')
     .paramGroup(actionGroup(), actionParams())
+    .paramGroup(pg('svgEmbedding', 'SVG Embedding'), SvgEmbeddingParameters())
     .params(ImageParameters())
 
 import VideoParameters from './video.ts'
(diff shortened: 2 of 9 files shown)


jsjames:
@digitaldan - did you see #4346.  I haven't fully looked at your PR yet, but 4346 adds interactive SVG support to OhImage and OhImageCard?  It also updates it to a composable, typescript and composition api (for oh-canvas-layout and oh-plan-page.

digitaldan:
DOH 🤦  I did not see that !   Let me give it a spin, but yeah i think that makes this PR pretty redundant ;-)

jsjames:
I have one additional update I want to check in which takes advantage of some of the additional typescript improvements and fixes the multi-select when editing the canvas-layout.

digitaldan:
So very similar, there are a couple of features from this pr that would be nice to have if thats possible, i have been playing with my PR all weekend, so have some real world tweaks from testing it.  I think these are still valid after reviewing your pr, but i would need to probably go a bit deeper to be sure. 

  - Derive a viewBox from the SVG's fixed width and height, then remove those attributes so the image scales to fit the widget
  - Look up style class targets inside the widget's own SVG rather than the whole document, so two copies of the same SVG don't affect each other
  - Unsubscribe from Item states when the page is no longer visible
  - Stop the image refresh timer when SVG embedding is turned on
  - Move the flashing option under the advanced settings (i had it non advanced too at first, but it kinda confused me when i first saw it, and since its not conditional on SVG enabled, maybe is confusing to most users just using images) 

I would be happy to open PR with these to your PR if that helps.

jsjames:
> So very similar, there are a couple of features from this pr that would be nice to have if thats possible, i have been playing with my PR all weekend, so have some real world tweaks from testing it. I think these are still valid after reviewing your pr, but i would need to probably go a bit deeper to be sure.
> 
> * Derive a viewBox from the SVG's fixed width and height, then remove those attributes so the image scales to fit the widget
> * Look up style class targets inside the widget's own SVG rather than the whole document, so two copies of the same SVG don't affect each other
> * Unsubscribe from Item states when the page is no longer visible
> * Stop the image refresh timer when SVG embedding is turned on
> * Move the flashing option under the advanced settings (i had it non advanced too at first, but it kinda confused me when i first saw it, and since its not conditional on SVG enabled, maybe is confusing to most users just using images)
> 
> I would be happy to open PR with these to your PR if that helps.

For sure - that would be great!

jsjames:
> > So very similar, there are a couple of features from this pr that would be nice to have if thats possible, i have been playing with my PR all weekend, so have some real world tweaks from testing it. I think these are still valid after reviewing your pr, but i would need to probably go a bit deeper to be sure.
> > 
> > * Derive a viewBox from the SVG's fixed width and height, then remove those attributes so the image scales to fit the widget
> > * Look up style class targets inside the widget's own SVG rather than the whole document, so two copies of the same SVG don't affect each other
> > * Unsubscribe from Item states when the page is no longer visible
> > * Stop the image refresh timer when SVG embedding is turned on
> > * Move the flashing option under the advanced settings (i had it non advanced too at first, but it kinda confused me when i first saw it, and since its not conditional on SVG enabled, maybe is confusing to most users just using images)
> > 
> > I would be happy to open PR with these to your PR if that helps.
> 
> For sure - that would be great!

On your #2, I believe that won't be necessary if using the data-state property - we won't have to do any "class" insertion at the javascript level.  For instance, I have a map of my house with different zones (motion and windows/doors) and use this embedded stylesheet in the svg - so these only apply to the SVG and don't rely on javascript to insert any special classes.

```
<style>
#HouseOutline[data-state='READY'] {
  fill: RGBA(0, 150, 0, .5) !important
}
#MotionZones [data-state='CLOSED'] {
  fill: none !important
}
[data-state='CLOSED'] {
  stroke: green !important
}
[data-state='OPEN'] {
  stroke: red !important
}
[data-state='UNDEF'] {
  stroke: yellow !important
}
</style>
```

PR #2579, opened 2024-05-18
[BasicUI] Switch with buttons: value displayed if pattern on widget

lolodomo:
When a Switch widget leads to the rendering with buttons, the value should not be by default also displayed at the left of the buttons (redundant information), except if it is requested through a state pattern defined on the widget.

Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #4374, opened 2026-07-20
Fix unlogical behavior

### LINKED ISSUE #3517: BasicUI: Switch with mapping ignores label definition

- [x] Basic UI
- [ ] HABPanel
- [ ] HABot
- [ ] CometVisu

## The problem

BasicUI does not show the value of the label when using a switch with a mapping.
The Android App does show the value as expected. 
Behavior should be aligned otherwise it's very confusing.
I suggest BasicUI in Browser should also show the label.

## Steps to reproduce

item definition
```
String  MyStringItem  "MyLabel [%s]" 
```

sitemap definition
```
Switch  item=MyStringItem       icon="none" mappings=[ABC="abc", ASDF="asfd"]
```

Observe that the current value is not shown.
Open Sitemap in app and observe value is there.

## Your environment

```
openHAB 5.0.2
```

spacemanspiff2007:
Closes #3517 by reverting #2579

### DIFF
--- a/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SwitchRenderer.java
+++ b/bundles/org.openhab.ui.basic/src/main/java/org/openhab/ui/basic/internal/render/SwitchRenderer.java
@@ -119,23 +119,15 @@ public class SwitchRenderer extends AbstractWidgetRenderer {
         String snippet = getSnippet(snippetName);
         State state = itemUIRegistry.getState(w);
 
+        snippet = preprocessSnippet(snippet, w);
+        
         if (nbButtons == 0) {
-            snippet = preprocessSnippet(snippet, w);
-
             if (OnOffType.ON.equals(state)) {
                 snippet = snippet.replaceAll("%checked%", "checked=true");
             } else {
                 snippet = snippet.replaceAll("%checked%", "");
             }
         } else {
-            // Show the value at left of all the buttons only if a state pattern is set on the sitemap widget
-            if (!hasValue(w.getLabel())) {
-                snippet = snippet.replace("%value%", "");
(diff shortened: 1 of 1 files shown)


lolodomo:
I am against this change.

lolodomo:
... but let see what other maintainers think.

jimtng:
I haven't tested it and have only read through #3517 briefly.

My understanding of this PR
- When the Item's label includes state, button widget with mapping will always show the value
- The current behaviour does not show it, unless you add `label=[%s]` to the sitemap to show it.

My question is, after this PR, what if I don't want to display the value even though my item's label has it?

spacemanspiff2007:
For BasicUI there is a clear priority (from lowest to highest)
- Default
- Configuration on the item (if available)
- Configuration of the sitemap (if available)

So I want to display the item format in a certain way I can do this by configuring the item, e.g. ``label=[%.1f]``.
If I have one special place in the UI where I want the value to be different I can configure the label in the sitemap e.g. ``label=[%.0f]``

This logic works (or worked) consistently for all UI elements.
There was a breaking change where the label of the item configuration would no longer be respected for a switch (I'm not sure if it's only with a mapping.)

So this item definition with a value "MODE 1"
```
String  MyStringItem  "MyLabel [%s]" 
```

would render ``MyLabel MODE 1`` as expected for the sitemap config

```
Text item=MyStringItem
```

but for the sitemap config

```
Switch item=MyStringItem   mappings=[MODE 2="MODE 2"]
```

it would render nothing.
There was a PR that changed the precedence and the goal of this PR is to revert that and align the behavior again.



> My question is, after this PR, what if I don't want to display the value even though my item's label has it?

It has been the same since OH1: Just configure the label in the sitemap
```
Text item=MyItem label="Some Text Without Value"
```

PR #4385, opened 2026-07-22
[blockly] Add blocks to javaify and jsify data

Nadahar:
This PR does two things: It adds two new blocks for `javaify()` and `jsify()`. It also replaces the "run rule" block with one that can call either `runRule()` or `runAsync()`, and that returns a result if `runRule()` is used. The old block is unchanged to avoid breaking existing scripts, but it has been removed from the toolbar so that it can't be added to new scripts. 

This was done to avoid confusion by having two almost identical blocks. If you don't care about the result/output from the other rule, it can just fire of the other rule asynchronously and continue immediately. If you do want the result/output, you will have to wait for the execution of the other rule to finish. The one situation this doesn't "cover" is if you don't care about the result, but want to wait for the other rule to finish before progressing in the current rule. This can be achieved by "dumping" the result in some dummy block that you don't use.

There are a number of things I'm unsure of here, like the texts/name to show and where to put the new blocks in the toolbar. I've put them directly below `runRule` for now, but it doesn't really fit there. It doesn't really fit in any of the other categories either, so it might be better to create a new category..?

Also, I haven't been able to link to any documentation for `javaify()` and `jsify()`, because I'm pretty sure that doesn't exist. The question is if we should make that documentation first, and then link to it from the blocks, or if that isn't important enough to hold this back.

@florian-h05 @rkoshak I don't know to what degree any you have the answer to any of these questions, or if there is somebody else that should be asked - but you two are the only I can think of at the moment.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
@@ -68,6 +68,66 @@ export default function defineOHBlocks_Scripts(f7, transformationServices) {
   }
 
   /*
+   * Runs the rule with the specified UID, synchronously or asynchronously.
+   * Parameters can be provided with the special parameter block oh_scriptparam
+   * Blockly part
+   */
+  Blockly.Blocks['oh_runrule_v2'] = {
+    init: function () {
+      const dropdown = new Blockly.FieldDropdown(
+        [
+          ['asynchronously', 'ASYNC'],
+          ['synchronously', 'SYNC']
+        ],
+        this.validate.bind(this)
+      )
+
+      this.appendValueInput('ruleUID').setCheck('String').appendField('run rule or script').appendField(dropdown, 'MODE')
+      this.appendValueInput('parameters').appendField('with context').setCheck('Dictionary')
+      this.setInputsInline(false)
+
+      // Set the default
+      this.setPreviousStatement(true, null)
+      this.setNextStatement(true, null)
+
+      this.setColour(0)
+      this.setTooltip('Run a rule or script with a certain UID, and optional parameters. To retrive the result, use synchronous mode.')
+      this.setHelpUrl(
+        'https://www.openhab.org/docs/configuration/blockly/rules-blockly-run-and-process.html#run-rule-or-script-created-in-ui'
+      )
+    },
+
+    validate: function (newValue) {
+      if (newValue === 'SYNC') {
+        this.setPreviousStatement(false)
+        this.setNextStatement(false)
+        this.setOutput(true, 'Dictionary')
+      } else {
+        this.setOutput(false)
+        this.setPreviousStatement(true, null)
+        this.setNextStatement(true, null)
+      }
+    }
+  }
+
+  /*
+   * Runs the rule with the specified UID, synchronously or asynchronously.
+   * Parameters can be provided with the special parameter block oh_scriptparam
+   * Code part
+   */
+  javascriptGenerator.forBlock['oh_runrule_v2'] = function (block) {
+    const ruleUID = javascriptGenerator.valueToCode(block, 'ruleUID', javascriptGenerator.ORDER_ATOMIC)
+    const scriptParameters = javascriptGenerator.valueToCode(block, 'parameters', javascriptGenerator.ORDER_ATOMIC) || 'null'
+
+    if (block.getFieldValue('MODE') === 'SYNC') {
+      const code = `rules.runRule(${ruleUID}, ${scriptParameters})`
+      return [code, javascriptGenerator.ORDER_NONE]
+    } else {
+      return `rules.runAsync(${ruleUID}, ${scriptParameters});\n`
+    }
+  }
+
+  /*
    * Recursively convert any value, array or object to use Java types. Functions are not supported.
    * Blockly part
    */
--- a/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/blockly/blocks-scripts.js
@@ -68,6 +68,52 @@ export default function defineOHBlocks_Scripts(f7, transformationServices) {
   }
 
   /*
+   * Recursively convert any value, array or object to use Java types. Functions are not supported.
+   * Blockly part
+   */
+  Blockly.Blocks['oh_javaify'] = {
+    init: function () {
+      this.appendValueInput('INPUT_VALUE').appendField('javaify')
+      this.setOutput(true, null)
+      this.setColour(280)
+      this.setTooltip('Recursively convert any value, array or object to use Java types. Functions are not supported.')
+    }
+  }
+
+  /*
+   * Recursively convert any value, array or object to use Java types. Functions are not supported.
+   * Code part
+   */
+  javascriptGenerator.forBlock['oh_javaify'] = function (block) {
+    const inputValue = javascriptGenerator.valueToCode(block, 'INPUT_VALUE', javascriptGenerator.ORDER_NONE) || 'null'
+    const code = `javaify(${inputValue})`
+    return [code, javascriptGenerator.ORDER_NONE]
+  }
+
+  /*
+   * Recursively convert Java Lists, Sets, and Maps and their entries/values to their JS counterparts.
+   * Blockly part
+   */
+  Blockly.Blocks['oh_jsify'] = {
+    init: function () {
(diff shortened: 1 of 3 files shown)


Nadahar:
image

florian-h05:
> Also, I haven't been able to link to any documentation for javaify() and jsify(), because I'm pretty sure that doesn't exist. The question is if we should make that documentation first, and then link to it from the blocks, or if that isn't important enough to hold this back.

Probably we should write docs and link to them.
If you want to give it a go, open a PR to openhab-js updating its README.

Nadahar:
> Probably we should write docs and link to them.
> If you want to give it a go, open a PR to openhab-js updating its README.

Sure, I, like anybody else, just *love* writing documentation, especially in an area where I'm not really sure about anything 😛 

But, I can give it a try I guess.

Did you look at what I did with `runRule()`? Is that an acceptable way to "deprecate" the old version, simply removing it from the toolbox? Is the new version OK, or should it have a third option "synchronous without result" which would allow it to wait but not be connected to an "input block"? That would be the 100% match with the existing function, I'm just not sure if it's very useful in most situations, which is why I didn't "include" it. I suspect that most uses will benefit from just calling it async, but there *can* be cases where waiting for the other rule is essential, even if you don't care about the result. Is it then acceptable that they must attach it to a "dummy block", or should we provide for that option explicitly?

How is the dropdown menu that transforms the block seen? Is that a good/bad solution? Would it be preferable to just have separate blocks? I like the dropdown because it reduces "clutter" in the toolbox, and it is in a way something the user should decide when running a rule. But, others might disagree..?

Perhaps the old block should remain in the toolbox in addition to the new?

I have many questions here, and perhaps @rkoshak is more likely to have an opinion about some of these things?

By the way, who is "the authority" on Blockly in OH? I feel that there's a person "missing" in the discussion, the one that has the "full overview" over blocks, the organization and how things should be implemented...

Nadahar:
To demonstrate clearly, the "run rule" block has these two forms depending on what you select in the dropdown. The async form is the default:

image

The "old"/existing block runs synchronously, but doesn't return the result - so in a way, it waits for execution, but gets nothing in return.

Nadahar:
@florian-h05 The blocks aren't documented in the README, but in the [OH docs themselves](https://www.openhab.org/docs/configuration/blockly/rules-blockly-run-and-process.html#run-rule-or-script-created-in-ui). And, when looking in the README, there are lots of stuff not covered at all, `runRule()` for example simply isn't there.

If there was ever a good time to set an AI to work, I think it would be to compare the code with the README and find all that isn't documented - and perhaps even suggest documentation for what is missing. I have nothing that is suitable to a task that "large", but I guess you do.

rkoshak:
I'm not sure I like that the type of the block changes with the drop down. I know of no other block that can change like that. And what I mean is changing from a plain doc to a block with a plug. There is no precident for that AFAIK and I'm not sure what confusion that might cause. For example, let's say you started with a synchronous call and then decided to change it to asynchronous. Now what ever it used to be plugged into becomes disconnected and the block itself becomes unteathered to the rest of the blocks instead of being slotted in the same spot with the rest of the blocks. 

I can see that being very confusing and annoying. 

Also, I'm concerned about the upgrade path. The only seamless way to upgrade would be to provide a synchronous block that does not have the plug and that replaces the existing block. Then the usual upgrade path of "open your blockly rules and save" is viable. They need make no other changes to the blocks, just regenerate the JS code. 

So I would recommend two blocks: one that returns a value and one that doesn. The one that doesn't return a value has an asynchronous and synchronous option (could be a checkbox to make the block a little slimmer). The default is synchronous. This block replaces the existing runRule block with synchronous enabled. This will allow a one-for-one replacement of the runRule blocks already in use so the end users don't need to change anything, just save the rule. 

The second block with the plug is always synchronous so it doesn't need an option. And it always has the plug. 

I agree that the docs should probably be in place first. However, if there are other blocks docs already missing, maybe all the missing docs could be handled by a separate PR. I don't have a strong opinion on that. For blockly users, the shapes and options on the block often are sufficient docs to figure out how tio use it by themselves.

Nadahar:
> For example, let's say you started with a synchronous call and then decided to change it to asynchronous. Now what ever it used to be plugged into becomes disconnected and the block itself becomes unteathered to the rest of the blocks instead of being slotted in the same spot with the rest of the blocks.

The dropdown is "locked" if the block is connected, so it can't "break". You must select it before connecting it.

> Also, I'm concerned about the upgrade path. The only seamless way to upgrade would be to provide a synchronous block that does not have the plug and that replaces the existing block. Then the usual upgrade path of "open your blockly rules and save" is viable. They need make no other changes to the blocks, just regenerate the JS code.

I haven't removed the old block, I've just removed it from the toolbar so that you can't insert it into new code. Existing code using that block should be unaffected.

The "new" block is a separate block with a separate "block code", so there will never be any "mix-up" between the two.

Does this help your "fears" about upgrades etc.? I don't know why you'd want the default to be synchronous. It has been the only option in the past, so it's obviously widespread, but since there has been no return value in the past either, I'm betting that 90%+ would benefit from using the asynchronous version. The only exception is when you need to "orchestrate timing" between rules, so that you rely on waiting despite not wanting the result.

Nadahar:
One option would be to not hide the "old" block. I just think that it's slightly confusing, but then there would be no "confusion" at all - just a new block with different options.

Here is a screenshot of how that would look:

image

rkoshak:
> Does this help your "fears" about upgrades etc.?

Not really. Having a block in a script that appears to no longer exist in the menus posses a separate problems and sources for confusion. If the old block is still going to be used, it needs to be retained in the menu.

> I don't know why you'd want the default to be synchronous. It has been the only option in the past, so it's obviously widespread, but since there has been no return value in the past either, I'm betting that 90%+ would benefit from using the asynchronous version.

And it breaks for the remaining 10%. Where as if the default is synchronous it works the same as it always has for 100% of users. But the reasoning for making this the default is if the old runRule block is replaces with the new runRule block, which is how this sort of change has been handled in Blockly for all changes like this in the past. If the underlying code generated by a block needs to change, or the block has new options the old block is modified and the users apply the changes by opening and saving their rules. 

If you want to invent a new approach for this change :man_shrugging: I have no major comments. I'm not going to argue about it because I already know arguing is futile if you've already made up your mind.

Nadahar:
> If you want to invent a new approach for this change 🤷‍♂️ I have no major comments. I'm not going to argue about it because I already know arguing is futile if you've already made up your mind.

I haven't made up my mind at all - that's why I'm asking. I was just trying to explain why I made the choices I made. Given your views, I think the best solution is to just not hide the old block. A new block can behave however it wants to, without "disturbing" anything.

> And it breaks for the remaining 10%. Where as if the default is synchronous it works the same as it always has for 100% of users.

Nothing breaks. Nothing will be changed if they resave their scripts. It will keep working as before, but when they make new code, they would use the new block.

Nadahar:
I've "unhidden" the old block. What do you think of this solution?

Nadahar:
I've made another change: Modifying the "retrieve from cache" block. Since the underlying function has gotten the option whether to automatically apply `jsify()` or not when retrieving a value, the block needs to reflect that to. I've done it in a way that I think is completely backwards compatible, with the default option applied, it generates the exact same code:

These are the variants of the block:

image

...which produces this code:

```js
cache.private.get('a');

cache.shared.get('b');

cache.shared.get('c', null, false);
```

I've also made a dedicated tooltip for the new option:

image

florian-h05:
Pulling in @stefan-hoehn as our Blockly expert. WDYT about the block design?

Nadahar:
@stefan-hoehn Could you at least acknowledge that you have seen the ping?

florian-h05:
@Nadahar I‘ve just messaged Stefan, he’ll have a look the next days.

Nadahar:
I'm updating this to support the changes in https://github.com/openhab/openhab-js/pull/560.

This:

image

..now produces this code:

```js
cache.private.put('a', 'a');
cache.shared.put('b', 'b');
cache.shared.put('c', 'c', false);
```

Nadahar:
https://github.com/openhab/openhab-js/pull/560 incorporated.

I think the time to evaluate the block structure is getting very long; it can't be that much work.

I also discovered that the dynamic changing of output isn't that unprecedented; I happened to look at "contextual info" and it basically does the same thing. It doesn't visually change, because it always has an output, but it changes output type dynamically between string, number, and quantity, which would "break" attached blocks in the same way as "my design" will. I can't quite see why it is a problem, Blockly supports it and is meant to work that way, and it actually prevents you from making a choice that invalidates an existing connection. If you configure the block before connecting it, there will never be a problem, and if you try to change it after connecting it, you won't be able to.

PR #4399, opened 2026-07-27
Preliminary proposal for server storage of settings

jsjames:
@Nadahar - Here is a preliminary version of what i was thinking to support local settings.

Note, I am still working on the startup timing and so this doesn't fully work yet, but I wanted to get something uploaded so you can see the proposed API (apiStorage.ts), etc.

### DIFF
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -634,6 +634,8 @@ export default {
     const { locale, mergeLocaleMessage: globalMergeLocaleMessage } = useI18n({ useScope: 'global' })
     const { t, mergeLocaleMessage: localMergeLocaleMessage } = useI18n({ useScope: 'local' })
     const { startAudioWebSocket, triggerDialog } = useDialog()
+    const uiOptionsStore = {}
+
     // required for notReachable error screen:
     loadLocaleMessages('common', globalMergeLocaleMessage)
     loadLocaleMessages('about', localMergeLocaleMessage)
@@ -646,10 +648,12 @@ export default {
       globalMergeLocaleMessage,
       locale,
       startAudioWebSocket,
-      triggerDialog
+      triggerDialog,
+      uiOptionsStore
     }
   },
   data() {
+    console.log("App.vue data() called")
     let theme = localStorage.getItem('openhab.ui:theme')
 
     if ((!theme || theme === 'auto') && typeof window.OHApp?.preferTheme === 'function') {
@@ -674,7 +678,7 @@ export default {
         // theme: (document.documentURI && document.documentURI.indexOf('?theme=ios') > 0) ? 'ios'
         //   : (document.documentURI && document.documentURI.indexOf('?theme=md') > 0) ? 'md'
         //     : 'auto', // Automatic theme detection
-        autoDarkMode: useUIOptionsStore().isAutoDarkMode(),
+        autoDarkMode: this.uiOptionsStore.autoDarkMode,
         // App routes
         routes,
         // Enable panel left visibility breakpoint
@@ -832,6 +836,7 @@ export default {
     },
     loadData(useCredentials) {
       performance.mark('loadDataStart')
+      console.log('loadData')
       const useCredentialsPromise = useCredentials ? this.setBasicCredentials() : Promise.resolve()
       return useCredentialsPromise
         .then(() => {
@@ -945,7 +950,7 @@ export default {
                 })
               : Promise.resolve(null)
           }
-          return Promise.all([useComponentsStore().loadPagesAndWidgets(), dayjsLocalePromise])
+          return Promise.all([useComponentsStore().loadPagesAndWidgets(), dayjsLocalePromise, useUIOptionsStore()])
         })
         .then((data) => {
           this.pages = useComponentsStore()
@@ -962,6 +967,10 @@ export default {
             dayjs.locale(data[1], null, false)
             console.log('Day.js locale set to', dayjs.locale())
           }
+
+          console.log('App: UI Options loaded')
+          this.uiOptionsStore = data[2]
+
           // load & build the semantic model
           useModelStore().loadSemanticModel()
         })
@@ -988,7 +997,7 @@ export default {
     },
     getPageIcon,
     updateThemeOptions() {
-      if (useUIOptionsStore().visibleBreakpointDisabled) {
+      if (this.uiOptionsStore.visibleBreakpointDisabled) {
         nextTick(() => {
           f7.panel.get('left').disableVisibleBreakpoint()
         })
@@ -1019,7 +1028,7 @@ export default {
     },
     toggleVisibleBreakpoint() {
       f7.panel.get('left').toggleVisibleBreakpoint()
-      useUIOptionsStore().visibleBreakpointDisabled = f7.panel.get('left').visibleBreakpointDisabled
+      this.uiOptionsStore.visibleBreakpointDisabled = f7.panel.get('left').visibleBreakpointDisabled
     },
     setLogDockVisible(visible) {
       if (visible && !useUserStore().isAdmin()) return
@@ -1198,7 +1207,7 @@ export default {
       if (ev.target.closest('.list-item a, a[href]')) {
         // avoid flicker (collapse + reopen)
         setTimeout(this.collapseAllSections, 30)
-        if (useUIOptionsStore().visibleBreakpointDisabled) {
+        if (this.uiOptionsStore.visibleBreakpointDisabled) {
           f7.panel.get('left')?.close()
         }
       }
@@ -1239,13 +1248,14 @@ export default {
     }
   },
   mounted() {
+    console.log('App.vue mounted')
     f7ready(async (f7) => {
       this.$f7dim.width = f7.width
       this.$f7dim.height = f7.height
 
       this.updateThemeOptions()
 
-      this.tryExchangeAuthorizationCode()
+      await this.tryExchangeAuthorizationCode()
         .then((user) => {
           this.loggedIn = true
           this.loadData()
--- a/bundles/org.openhab.ui/web/src/App.vue
+++ b/bundles/org.openhab.ui/web/src/App.vue
@@ -968,8 +968,8 @@ export default {
             console.log('Day.js locale set to', dayjs.locale())
           }
 
-          console.log('App: UI Options loaded')
           this.uiOptionsStore = data[2]
+          console.log('App: UI Options loaded', this.ready)
 
           // load & build the semantic model
           useModelStore().loadSemanticModel()
@@ -1214,6 +1214,7 @@ export default {
     }
   },
   created() {
+    console.log('App.vue created')
     this.AddonIcons = AddonIcons
     this.AddonTitles = AddonTitles
 
--- a/bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
+++ b/bundles/org.openhab.ui/web/src/assets/i18n/theme-switcher/en.json
@@ -7,6 +7,7 @@
   "about.darkMode.dark": "Dark",
   "about.navigationBarsStyle": "Navigation bars style",
   "about.miscellaneous": "Miscellaneous",
+  "about.miscellaneous.sync.settings": "Sync settings across devices",
   "about.miscellaneous.home.navbar": "Home page navigation bar",
   "about.miscellaneous.home.navbar.default": "Default",
   "about.miscellaneous.home.navbar.simple": "Simple",
--- a/bundles/org.openhab.ui/web/src/components/theme-switcher.vue
+++ b/bundles/org.openhab.ui/web/src/components/theme-switcher.vue
@@ -53,6 +53,11 @@
         <f7-block-title>{{ t('about.miscellaneous') }}</f7-block-title>
         <f7-list>
           <f7-list-item>
+            <span>{{ t('about.miscellaneous.sync.settings') }}</span>
+            <f7-toggle v-model:checked="syncSettings" />
+
+          </f7-list-item>
+          <f7-list-item>
             <span>{{ t('about.miscellaneous.home.navbar') }}</span>
             <f7-segmented class="home-navbar-selection">
               <f7-button
@@ -221,6 +226,7 @@ export default {
       'homeNavBar',
       'homeBackground',
       'hideChatInput',
+      'syncSettings',
       'disableExpandableCardAnimation',
       'webAudio',
       'dialogEnabled',
--- a/bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
+++ b/bundles/org.openhab.ui/web/src/components/widgets/useWidgetExpression.ts
@@ -88,6 +88,7 @@ export type EvaluateExpressionFn = <T = unknown>(
 export function useWidgetExpression(properties: { context?: WidgetContext; props?: api.ConfigDescription } = {}) {
   // imports
   const userStore = useUserStore()
+  console.log('useWidgetExpression: useUIOptionsStore() called')
   const uiOptionsStore = useUIOptionsStore()
 
   const instance = getCurrentInstance()
--- a/bundles/org.openhab.ui/web/src/js/composables/useDialog.ts
+++ b/bundles/org.openhab.ui/web/src/js/composables/useDialog.ts
@@ -55,7 +55,7 @@ export function useDialog() {
       const ohURL = `${location.protocol}//${location.hostname}${port}`
 
       updatePageIcon(false, false, false)
-      let forceTrigger = dialogTriggerOnConnect
+    let forceTrigger = dialogTriggerOnConnect
       audioMain.value = new AudioMain(ohURL, getAccessToken, {
         onMessage: (message, level) => {
           if (level === 'error') {
--- a/bundles/org.openhab.ui/web/src/js/openhab/apiServerStorage.ts
+++ b/bundles/org.openhab.ui/web/src/js/openhab/apiServerStorage.ts
@@ -1,0 +1,84 @@
+// **** Temporary types and functions until the openapi spec supports
+import { client } from '@/api/client.gen'
+import * as api from '@/api'
+import type { RequestResult } from '@/api/client/types.gen'
+import { buildClientParams } from '@/api/client'
+
+type GetServerConfigValueResponses = {
+    200: string | null
+}
+
+type GetServerConfigValueErrors = {
+    404: unknown
+}
+
+export const getServerConfigValue = <ThrowOnError extends boolean = false>(parameters: {
+    key: string
+}, options?: api.Options<never, ThrowOnError>): RequestResult<GetServerConfigValueResponses, GetServerConfigValueErrors, ThrowOnError, 'data'> => {
+    const params = buildClientParams([parameters], [{ args: [
+            { in: 'query', key: 'key' }
+    ] }])
+    return (options?.client ?? client).get<GetServerConfigValueResponses, GetServerConfigValueErrors, ThrowOnError, 'data'>({
+        responseStyle: 'data',
+        url: '/client-store/{key}',
+        ...options,
+        ...params
+    })
+}
+
+type PutServerConfigValueResponses = {
+    200: unknown
+}
+
+type PutServerConfigValueErrors = {
+    400: unknown
+    404: unknown
+}
+
+export const putServerConfigValue = <ThrowOnError extends boolean = false>(parameters: {
+    key: string
+    body?: {
+        value?: string
+    }
+}, options?: api.Options<never, ThrowOnError>): RequestResult<PutServerConfigValueResponses, PutServerConfigValueErrors, ThrowOnError, 'data'> => {
+    const params = buildClientParams([parameters], [{ args: [
+                { in: 'path', key: 'key' },
+                { key: 'body', map: 'body' }
+            ] }])
+    return (options?.client ?? client).put<PutServerConfigValueResponses, PutServerConfigValueErrors, ThrowOnError, 'data'>({
+        responseStyle: 'data',
+        security: [{ scheme: 'bearer', type: 'http' }],
+        url: '/client-store/{key}',
+        ...options,
+        ...params,
+        headers: {
+            'Content-Type': 'application/json',
+            ...options?.headers,
+            ...params.headers
+        }
+    })
+}
+
+type DeleteServerConfigValueResponses = {
+    200: unknown
+}
+
+type DeleteServerConfigValueErrors = {
+    400: unknown
+    404: unknown
+}
+
+export const deleteServerConfigValue = <ThrowOnError extends boolean = false>(parameters: {
+    key: string
+}, options?: api.Options<never, ThrowOnError>): RequestResult<DeleteServerConfigValueResponses, DeleteServerConfigValueErrors, ThrowOnError, 'data'> => {
+    const params = buildClientParams([parameters], [{ args: [
+            { in: 'path', key: 'key' }
+        ] }])
+    return (options?.client ?? client).delete<DeleteServerConfigValueResponses, DeleteServerConfigValueErrors, ThrowOnError, 'data'>({
+        responseStyle: 'data',
+        security: [{ scheme: 'bearer', type: 'http' }],
+        url: '/client-store/{key}',
+        ...options,
+        ...params
+    })
+}
--- a/bundles/org.openhab.ui/web/src/js/openhab/ohStorage.ts
+++ b/bundles/org.openhab.ui/web/src/js/openhab/ohStorage.ts
@@ -1,0 +1,81 @@
+import type { Ref } from 'vue'
+import * as api from './apiServerStorage'
+import type { StorageLikeAsync } from '@vueuse/core'
+
+class OHStorage implements StorageLikeAsync {
+    constructor(syncSettings: Ref<boolean>, localPrefix: string = '', serverPrefix: string = '') {
+        this.syncWithServer = syncSettings
+        this.localPrefix = localPrefix
+        this.serverPrefix = serverPrefix
+    }
+
+    public localPrefix: string
+    public serverPrefix: string
+    public syncWithServer: Ref<boolean>
+
+    async getItem(key: string) : Promise<string | null> {
+        if (this.syncWithServer.value) {
+            let serverValue: string | null | undefined = null
+            try {
+                await new Promise(resolve => {          // temporary to model server fetch delay
+                    setTimeout(resolve, 300);
+                    console.log('Simulating server fetch delay for key:', key)
+                    return localStorage.getItem(`${this.localPrefix}${key}`)
+                })
+                /*
+                serverValue = await api.getServerConfigValue({ key: `${this.serverPrefix}${key}` })
+                if (serverValue) {
+                    localStorage.setItem(`${this.localPrefix}${key}`, serverValue)
+                }
+                return null
+                */
+            } catch (error) {
+                console.error('Failed to get item from server:', error)
+                // TODO: should we return the local value if server fetch fails? For now, we return null to indicate failure.
+                return null
+            }
+        } else {
+            return localStorage.getItem(`${this.localPrefix}${key}`)
+        }
+    }
+
+    async setItem(key: string, value: string): Promise<void> {
+        if (this.syncWithServer.value) {
+            await api.putServerConfigValue({ key: `${this.serverPrefix}${key}`, body: { value } })
+        }
+        localStorage.setItem(`${this.localPrefix}${key}`, value)
+    }
+
+    async removeItem(key: string) : Promise<void> {
+        if (this.syncWithServer) {
+            await api.deleteServerConfigValue({ key: `${this.serverPrefix}${key}` })
+        }
+        localStorage.removeItem(`${this.localPrefix}${key}`)
+    }
+
+    // @nadahar - consider adding a "key and clear" rest call to model similar calls to localStorage
+    async clear(): Promise<void> {
+        // Only clear items with the specific prefix
+        for (let i = localStorage.length - 1; i >= 0; i--) {
+            const key = localStorage.key(i)
+            if (key && key.startsWith(this.localPrefix)) {
+                localStorage.removeItem(key)
+            }
+        }
+    }
+
+    key(index: number) {
+        const key = localStorage.key(index)
+        if (key && key.startsWith(this.localPrefix)) {
+            return key.substring(this.localPrefix.length)
+        }
+        return null
+    }
+
+    setSyncWithServer(sync: boolean) {
+        this.syncWithServer = sync
+    }
+}
+
+
+export { OHStorage }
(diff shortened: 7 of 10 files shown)


Nadahar:
> Note, I am still working on the startup timing and so this doesn't fully work yet, but I wanted to get something uploaded so you can see the proposed API (apiStorage.ts), etc.

Great, thanks. I'll have a look.

Nadahar:
You have gone with the "sync" concept, which I'm not sure I understand quite what will accomplish. My idea was that we had 3 "level" of storage for a store, `session`, `local` and `user`. Only `user` would be stored on the server, and only stores where that "made sense" would use or offer `user` as an option.

When you say sync, it sounds like you still use `local` as the "main" storage, and just sync your local settings to a server. I assume that if you have "sync" enabled, that also means that the setting on the server will override your local setting, but it presents a couple of challenges:

* If you use the local setting first, and the server setting turns out to be different, you must re-render everything when the server answer is available. Is two renders very close together "better" than one that starts slightly later than the first one, but is probably finished sooner than the second one (because the rendering is likely to be slower than retrieving the results)? I guess you could say that there's a "benefit" when local and server agrees, because you don't have to re-render.
* Such a "sync setting" would typically have to be enabled on each client before coming effective. That would prevent `user` settings from actually being user settings in practice, and it would prevent preconfiguring MainUI to behave a certain way. I "know" what is coming if we make this, people will want to set the defaults for the user settings, probably in some kind of configuration file, so that when it's deployed, things already behave as desired. That can be handled in core if you make it possible to define defaults for user settings somewhere, so that if nothing is stored for a particular property for a user, the default is returned. But, if every client first has to enable "sync", this won't achieve this.
* Doing synchronization properly is actually quite complicated. You need to keep timestamps for every change, and then decide which setting that should apply based on the order of the timestamps. This also includes keeping timestamps for deletion/clearing of settings. Not having to handle synchronization but just working directly with the server as the storage for `user` sounds much less complicated to me.

jsjames:
> You have gone with the "sync" concept, which I'm not sure I understand quite what will accomplish. My idea was that we had 3 "level" of storage for a store, session, local and user. Only user would be stored on the server, and only stores where that "made sense" would use or offer user as an option.

Yeah, I think we should reuse a standard industry term/concept here vs creating something new.  This delivers the capability you want, correct?

- The use model - server still maintains the golden copy of the settings. The "sync" comes into play is that when you change a setting on device, it will "sync" with the server.
- When a setting is requested through the api, we wait for the api before proceeding. When a value is returned, we do write that value to local. This way, someone can sync settings, then then turn off for any device specific tweaks they want to make.
- We shouldn't complicated with a data stamp/etc, the server is the golden and that is where the settings come from. We need to decide what to do if the api response fails (for instance in the interim while this capability does not exist in core. Should we return localSetting or the default?
- Given comments in other threads, we shouldn't make 'sync' the default, but folks would enable this on any device where they want settings synced (again similar to industry practice for sync settings.

ghys:
You guys are amazing, it's a wonder I can still read your code - I recognize my old structure, that's probably why.

If I may offer some old timer advice:

During my time I have always carefully considered what the initial rendering of the UI would cost in terms of bandwidth and number of requests. It's not only because I'm old and have known 28kbps modems and so on: imagine you're hiking in the mountains, and you have only 1 bar of 4G, or god forbid, 3G or Edge. Every single request would then take seconds.

So there's why the initial rendering time is important, and every new HTTP request hinders that. I've been big on PWAs at the time because of that, since they could at least cache the static stuff, but then there's all the REST API calls that are necessary before the UI can be rendered. I've also tried some things to cache most of those (with Expires/ETag/Last-Modified headers and so on); gzipping the responses also help.

So I guess my message is: every new thing you add that delays the initial rendering has a cost, and you may not feel it on your home's gigabit connection, but others will.

jsjames:
> You guys are amazing, it's a wonder I can still read your code - I recognize my old structure, that's probably why.
> 
> If I may offer some old timer advice:
> 
> During my time I have always carefully considered what the initial rendering of the UI would cost in terms of bandwidth and number of requests. It's not only because I'm old and have known 28kbps modems and so on: imagine you're hiking in the mountains, and you have only 1 bar of 4G, or god forbid, 3G or Edge. Every single request would then take seconds.
> 
> So there's why the initial rendering time is important, and every new HTTP request hinders that. I've been big on PWAs at the time because of that, since they could at least cache the static stuff, but then there's all the REST API calls that are necessary before the UI can be rendered. I've also tried some things to cache most of those (with Expires/ETag/Last-Modified headers and so on); gzipping the responses also help.
> 
> So I guess my message is: every new thing you add that delays the initial rendering has a cost, and you may not feel it on your home's gigabit connection, but others will.

Good points.  Even on some of the 5G networks, latency can be an issue.

We will keep the "sync" feature as an opt-in feature, so if speed is an issue on your device, you don't have to enable. Further, the proposal would allow you to sync one time so your settings would be copied to localStorage and then you can disable syncing to improve the performance.

Nadahar:
> We will keep the "sync" feature as an opt-in feature, so if speed is an issue on your device, you don't have to enable. Further, the proposal would allow you to sync one time so your settings would be copied to localStorage and then you can disable syncing to improve the performance.

This is directly opposed to what I wanted to achieve. As I've said before, much of the point is to be able to preconfigure MainUI to behave consistently. Today, I have to open the "Help and About" on every single browser, on every single installation, after every singe "cache clearing" to change the UI to look what I consider "right". I doubt I'm alone in this, even if others don't agree on what the default should be. The point is that the default becomes way too powerful the way it is today, it's like MainUI is trying to get you to "give in" and accept the default of pure exhaustion. If you like the defaults, you'll never notice this.

The whole point of having "user" settings is that you can set it once, and then it will follow you around, without having to configure things over and over and over again. Having this default to use local storage, means that it won't make any difference - you will just have to flip a different setting over and over again.

I'd say that if you have a very bad connection, you might as well enable local storage explicitly to remedy that problem.

More broadly, I'd say that I'm not sure I that share the worry. I was planning to make the "get" endpoint as flexible as possible, so that you can retrieve a single property in a specific store, one whole specific store, or an array of specified stored. That way, there wouldn't be necessary to do a lot of requests, you could request everything you need in one request.

I haven't studied how much information is stored in all the stores, my idea was never to move everything to `user`, but to do so for the things you'd typically want to be the same regardless of what browser/device you use. This "sync" idea sounds more like you'd copy everything from local storage to server storage. That might make it more data, I don't know, but the size of the data I envisioned would be minimal. Just a few settings about layout, menu structure, theme etc. When you compare that with the often quite sizable chunks of data that are retrieved when you open many pages, it really amounts to nothing. I don't think what I had in mind would lead to any "measurable" difference in speed: One request with a few bytes of data. It would almost always be ready before the other resources were in anyway.

jsjames:
> > We will keep the "sync" feature as an opt-in feature, so if speed is an issue on your device, you don't have to enable. Further, the proposal would allow you to sync one time so your settings would be copied to localStorage and then you can disable syncing to improve the performance.
> 
> This is directly opposed to what I wanted to achieve. 

I'm not sure I understand here. To get a device to adopt your settings is a single switch - you turn that on and your setting syncs - pretty simple.  Not sure how that can be "directly opposed"??

Nadahar:
> I'm not sure I understand here. To get a device to adopt your settings is a single switch - you turn that on and your setting syncs - pretty simple. Not sure how that can be "directly opposed"??

It's not just "a single switch", it means first navigating to where that switch is, changing the setting, and then navigating back to where you were. I do a lot of testing, run a lot of test installations, use different browsers and computers, so I probably experience this much more than the average user, but yes: I very much long to get rid of that. I find it so frustrating that I sometimes abort doing what I was about to, to not have to go through that again. It's not that it's such a "huge task", it's just so annoying to be met with the default you don't like "an infinite number of times". It feels like those "nag dialogs" that some software use to get you to finally give in and accept that they steal your data or whatever the thing is. I'd like to be able to make the decision once, and then have that decision "respected".

jsjames:
> > I'm not sure I understand here. To get a device to adopt your settings is a single switch - you turn that on and your setting syncs - pretty simple. Not sure how that can be "directly opposed"??
> 
> It's not just "a single switch", it means first navigating to where that switch is, changing the setting, and then navigating back to where you were. I do a lot of testing, run a lot of test installations, use different browsers and computers, so I probably experience this much more than the average user, but yes: I very much long to get rid of that. I find it so frustrating that I sometimes abort doing what I was about to, to not have to go through that again. It's not that it's such a "huge task", it's just so annoying to be met with the default you don't like "an infinite number of times". It feels like those "nag dialogs" that some software use to get you to finally give in and accept that they steal your data or whatever the thing is. I'd like to be able to make the decision once, and then have that decision "respected".

If the issue is what we decide as the "default" - let's get other people's opinion.  We can prototype it up and measure it on different networks to provide that datapoint.

Nadahar:
> If the issue is what we decide as the "default" - let's get other people's opinion. We can prototype it up and measure it on different networks to provide that datapoint.

My point is that we will never agree on what the default should be, and that's fine. But, why can't we allow people to define their own defaults, making "as many as possible" happy, instead of trying to pick whatever suits the majority. Even a default that the majority supports can be disliked by a sizable number of people. We can avoid that situation completely by letting people define it themselves, and as I've said before, I had also imagined that I would find some way to define an "installation default" that could be defined in some file, and applied to any users that hadn't set a default. That way, users could have "their defaults" set the first time they start the installation, they could apply it across installations (for those that have more than one).

Why isn't that just better than subjecting everybody to *one default*?

jsjames:
> > If the issue is what we decide as the "default" - let's get other people's opinion. We can prototype it up and measure it on different networks to provide that datapoint.
> 
> My point is that we will never agree on what the default should be, and that's fine. But, why can't we allow people to define their own defaults, making "as many as possible" happy, instead of trying to pick whatever suits the majority. Even a default that the majority supports can be disliked by a sizable number of people. We can avoid that situation completely by letting people define it themselves, and as I've said before, I had also imagined that I would find some way to define an "installation default" that could be defined in some file, and applied to any users that hadn't set a default. That way, users could have "their defaults" set the first time they start the installation, they could apply it across installations (for those that have more than one).
> 
> Why isn't that just better than subjecting everybody to _one default_?

My point is let's get the data and then an educated decision can be made. Right now we are arguing about a situation where we don't have data.

mherwege:
In my view, it could be something like this (and I believe it is consistent with how sync works in browsers as well):
- Have a user setting 'sync UI' stored in the server JSONDB by user.
- First login for the user from a browser that has not seen OH for that user before, retrieve that setting and sync when the 'sync UI' setting is set to local storage. If not set, don't sync.
- Switching off that user setting in the UI will not update that server setting, but only a local storage setting. Switching it on will update the server setting.
- On subsequent logins, the local storage setting is checked first. If it is available, it takes precedence. If not, the server setting is used.

With this mechanism:
- Settings can be stored centrally and once sync is on, will be pushed to the clients.
- Individual clients can disable sync.
- I would opt for 'sync UI' off by default. This makes it consistent with current behaviour. But switching it on on a single client for a user will effectively enable it on all clients for that user from that point onwards, unless overridden by client.

I could also imagine an extra question in the setup wizard for first time installations: 'Sync UI settings accross browser clients?', that could set the value in the JSONDB for the default and admin users. And you could make that on by default.
Also, when creating a new user, you may want an option to set this 'sync UI' setting.

I think all settings that are currently in local storage (not session storage) should be treated the same if possible. It gets very confusing if some are in local storage and some are in server storage. It will create a lot of questions from users on what is where, why certain things are consistent accross browsers and others aren't. The concept should be simple. I dislike the idea of having this configurable by setting, as it creates a lot of complexity for the user. We need to strike a balance between detailed configurability and complexity. The idea is, what is currently stored locally in the browser can now be pushed to the server for the user.

jsjames:
> Switching off that user setting in the UI will not update that server setting, but only a local storage setting. Switching it on will update the server setting.

The issue I see here is that once the sync setting as been set on the server, it is permanently set and there is no way to turn it off, correct?

[EDIT] We could just prompt the user when they turn off this setting if they want this for their local machine or turn off globally.

Nadahar:
> My point is let's get the data and then an educated decision can be made. Right now we are arguing about a situation where we don't have data.

What data? Whether people are happy with the current default, or whether people would prefer to be able to set their own default? How exactly do you intend to acquire these data?

There are numerous examples from the forum where people ask for how to change various aspects of the defaults, so we already know that not everybody is happy with it. Especially the "Chat" sidebar entry and the lack of a Sitemap entry has gotten focus lately, but there are other examples. So, we already know that everybody isn't happy with the defaults, does it really matter if that group is in the majority or not? I think it's perfectly reasonable, and to be expected, that everybody won't be happy with the default, no matter what you pick.

@mherwege I appreciate your attempt to not let this "escalate" into something bigger, and I largely agree with what you suggest. The most important is that there's a way to preconfigure the default in my view, and that you don't have to keep doing these "manual customizations". Exactly how you get there is of less importance I guess. That said, I don't think turning the "sync by default" "on" on a client should change the server setting. I think perhaps that should be a completely independent setting, configurable somewhere in the "admin settings" section of the UI, and/or in configuration files. You might want to enable "sync" on some client(s) without that affecting every client (like e.g. a wall mounted tablet).

But, I want to explain what I originally suggested in some more detail, since all this originated in my suggestion:

I never considered that the current behavior was considered a "feature", something we would want to preserve because it's the better solution. I always considered it a technical limitation, something that was like it was because there was no better way to handle it. That's what I set out to change, and I am frankly somewhat surprised by the need to "preserve" the current behavior.

As I see it, there are two possible scenarios:

* You are happy with the current defaults, and this never bothered you. You would still be happy if some settings where user settings instead of local settings, you would in fact never notice the difference.
* You are unhappy with the current defaults. To what degree this bothers you depends on how many installations and clients you have to configure manually, but it's extremely unlikely that you actually *appreciate* having to make the same configuration over and over again.

I don't see the third scenario where the current behavior is the best solution, which is why I don't understand the desire to preserve it.

Next, the stores themselves. As I've said all along, I've wanted to make it possible to define stores as `session`, `local` or `user` backed. It never occurred to me that dumping everything in the local store to the server would be a good idea. When looking at the current stores, it still doesn't seem like a good idea to me:

image

I don't know how to tell the `session` stores from the `local` stores, so I haven't differentiated between the two. But, having looked through what these store, it appears to me that only `UIOptionsStore` would make any sense as `user` backed. *Maybe* also the `DeveloperStore`, but I'm not sure. *If* at all a point, I would think that having a separate setting for whether the "developer setting" should be `user` or `local` would make sense, independent of whether you'd like `UIOptionsStore` to be `local` or `user` (I suspect that 99% would want this to be `user` if they have an opinion about it at all).

This is what's currently stored in `DeveloperStore`:

```js
export const useDeveloperStore = defineStore('developer', () => {
  const pinnedObjects = reactive<PinnedObjects>({
    items: [],
    things: [],
    rules: [],
    scenes: [],
    scripts: [],
    pages: [],
    widgets: [],
    transformations: [],
    persistenceConfigs: []
  })
```

So, it seems to only be your pinned objects. I've never used this, so I don't quite know if you'd want it shared between devices. If you use it like a sort of "quick menu", I guess maybe the answer is yes, but isn't the dev toolbar only available on large screens, so that you simply can't use it on phones etc.? It would at least make having it being fetched from the server a waste for all those devices where you can't use it anyway.

So, `UIOptionsStore` seem to be what all this is about. And, it's a bit messy, because some settings are shared with BasicUI so they kind of must reside in local storage. To me, it would make sense to split this store into settings you'd want to be `user` and settings where it isn't desirable or possible.

The store with the `user` settings could then have an option to make it "local only", for those circumstances where you'd want a certain device to *not* follow the `user` settings.

Another thing that goes into this, is that some settings don't exist today because they don't make much sense when you have to set them on every device. Take #4361 for example. What is desirable there, largely depends on how you design your pages. If the pages are designed around tabs, it's very annoying that use lose the "navigation menu" if you're not at the top of the page. If you don't use the logic that way in the UI, and use a device with a tiny screen, it's likely that you prefer the "declutter" that give you a bit of extra "screen real estate" to use for actual content.

I'd say that the same is true for the new menu customization option. It's hard to see how it can be desirable to repeat this for every device.

I have other "ideas" like whether to use the OH defined language or that configured in the browser for translations and formatting. Also, whether you'd want to use the OH configured timezone or the local browser setting, and whether you'd want timezones converted into "the selected timezone" or see times in their "origin timezone". And, I'm sure there are quite a few more that I don't remember at the moment.

I haven't suggested making any of them because I think it's largely useless when you have to keep configuring it on every client, after ever "cache clearing" or software update that messes up your storage etc.

To me, this suggestion was about making things possible that don't really make much sense with the current browser based storage, in additions to getting rid of the annoying "defaults situation".

When it comes to using the "sync" term, I realize that I'm probably not in the majority here, but the reasons I dislike it is:

* It's somewhat of a lie unless you make proper sync logic with conflict resolution and timestamps.
* It has a negative association because of the way it's being used "in the industry". When I see a "sync" option, I *always* make sure it's off, because I don't see it as a "sync" setting. I see it as a disguise for "steal my data". Because using it required creating an account, which requires you to accept a privacy policy that you need a lawyer to understand, and which they can update whenever they like, without you having basically any say. So, I see it as mostly an "excuse" to make people accept the theft of data and linking their identity. What you get in return is a little bit of convenience, but the price is huge and in no way make up for the tiny convenience. Why would we want to mimic something that is so tainted?
* It would be unclear exactly what would be "synced", but I guess that could be helped with a bit more description that just the standard "sync settings" switch.

Nadahar:
> The issue I see here is that once the sync setting as been set on the server, it is permanently set and there is no way to turn it off, correct?

The way I see it, this would be a setting somewhere in the "Settings" menu in MainUI as a global setting. It would then be an inherited setting on the user level, that would be possible to manage from the console where you manage users, and if user management is ever implemented in MainUI, also from there. I don't think changing the setting on the device itself should ever change the "server setting".

mherwege:
> The way I see it, this would be a setting somewhere in the "Settings" menu in MainUI as a global setting. It would then be an inherited setting on the user level, that would be possible to manage from the console where you manage users, and if user management is ever implemented in MainUI, also from there. I don't think changing the setting on the device itself should ever change the "server setting".

In my view, but open for any opinion, if there is a local setting it is leading:
- Locally switched to off: will never load from the server, but will keep the local copy of settings that where there before turning off.
- Locally switched to on, server setting was off: implies the server does need to store settings, so if you switch on, and server sync was not yet on, it should be turned on both locally and on the server. At that point, local settings should be copied to the server.
- Locally switched to on, server setting was on: get the settings from the server and store them locally (potentially overwriting local settings that where already done).

When I say sync (and you can give it another name, but I actually think it conveys quite well what we want to achieve), I see it in a very simple way:
- When you open or refresh the browser and login as a specific user, check the local flag. If it is on, get the settings from the server. If it is off, don't do anything, keep working with what you have locally. If there is no local flag set, check the server flag. If the server flag is on, get the settings and set the local flag. If the server flag is off, don't do anything (don't set the local flag), and keep working with local storage.

It is true, once you switch on the server flag, it cannot be switched off anymore. But I don't see that as a real problem. You can still switch it off locally for a specific user on a specific device. But any new device you add, will be synced automatically as a starting point.

For me, I do like the idea of automatically getting and applying the settings in most cases. But I want the flexibility to switch that off on a user by user basis, and that only has to be a single switch if you logged in as a user on a device. I want to keep it simple for the user while getting the benefit of remembering how that user usually configures its UI.

To be further analyzed, but I agree that we are mainly talking about the UIOptionsStore and DeveloperStore. These stores are pinia concepts though and don't say much about where it is actually stored. They may just be backed by the API. To know what is stored locally in the browser, search for `localStorage` and `sessionStorage`. `sessionStorage` seems to only be used for user authentication. `localStorage` is persistent in the browser and is used in other places then the stores (browser logging, log viewer, main app component, theme switcher, sip client).

> Another thing that goes into this, is that some settings don't exist today because they don't make much sense when you have to set them on every device. Take #4361 for example. What is desirable there, largely depends on how you design your pages. If the pages are designed around tabs, it's very annoying that use lose the "navigation menu" if you're not at the top of the page. If you don't use the logic that way in the UI, and use a device with a tiny screen, it's likely that you prefer the "declutter" that give you a bit of extra "screen real estate" to use for actual content.
> 
> I'd say that the same is true for the new menu customization option. It's hard to see how it can be desirable to repeat this for every device.

I totally agree on that. Nothing I propose goes against that. I just want to make it as simple as possible from a user perspective, so as little options as possible that would achieve the stated goal. Once sync is on (default or not), they expect to see it come up as configured somewhere else except when explicitly switching it off. Why not make on the default straight away? That's just to avoid the initial impact of not having anything on the server after an upgrade and the UI settings being lost.

> I never considered that the current behavior was considered a "feature", something we would want to preserve because it's the better solution. I always considered it a technical limitation, something that was like it was because there was no better way to handle it. That's what I set out to change, and I am frankly somewhat surprised by the need to "preserve" the current behavior.

I actually do think it is somewhat of a feature. We could have implemented storing the UI settings in the JSONDB from the start and create a REST endpoint for that, but we didn't as the idea was that UI configuration was by device. You want a different configuration on a wall tablet, a phone or a desktop. As the number of options were limited, this was mostly fine. As we add more configuration options, it is worth revisiting.
I don't take the argument you should have separate users for the different devices to be able to have a different UI configuration for different devices. User management is not available in the UI as it stands, so that's not something most users will do. They go with admin and the default user. So let's keep to that for now and don't impose it. It even doesn't have to be imposed when we have better user management. So that's why you should be able to switch off the automatic UI settings retrieval and set it locally.


> I have other "ideas" like whether to use the OH defined language or that configured in the browser for translations and formatting. Also, whether you'd want to use the OH configured timezone or the local browser setting, and whether you'd want timezones converted into "the selected timezone" or see times in their "origin timezone".

That's a tall task you intend to take on, especially because much is server generated and already translated in the REST responses, not in the UI logic. You would need a way to get all responses from the server in a language requested by the browser.

> When it comes to using the "sync" term, I realize that I'm probably not in the majority here, but the reasons I dislike it is:

I understand your objections, but I am not with you on this. I think it is a simple and well understood term. And I don't see how someone would think that suddenly OH is tracking them because of using that terminology.



> The way I see it, this would be a setting somewhere in the "Settings" menu in MainUI as a global setting. It would then be an inherited setting on the user level, that would be possible to manage from the console where you manage users, and if user management is ever implemented in MainUI, also from there. I don't think changing the setting on the device itself should ever change the "server setting".

I disagree. If you have to manage this through the console, we miss the point. If we need to distinguish between a global setting and a user setting, we miss the point and make it more complicated then it should be. If you flip the switch in the UI, you are connected as a user (or the default user). So it should apply for that user, not others. It doesn't need more than that. I see the server setting also as a setting by user, just stored on the server. Why would you need any extra UI to switch that over?
The only case the server setting should be switched from the device, is if you switch it on on the device and it was off before. You do need the global flag to go on then or it cannot work. If you switch sync on somewhere, it applies it will be on everywhere for that user if it was not explicitly disabled on the device. That only makes sense to me.

jsjames:
> It is true, once you switch on the server flag, it cannot be switched off anymore. But I don't see that as a real problem. You can still switch it off locally for a specific user on a specific device. But any new device you add, will be synced automatically as a starting point.

Not sure if you saw my edit, but I think if the user is an administrator, we can prompt if they want to disable this setting for the specific device OR disable globally as the default.

Nadahar:
There are lots of "talking past each other here", but let me say one thing that I think can solve some of them: I always include the "null setting" when dealing with such concepts, so even a switch/boolean has 3 states: on, off, null. This affects everything, because it means that "user settings" will only override local/locally default settings if they have been set on the "user" level. So, there's no danger of "enabling sync" and then suddenly everything you have done locally is wiped out, if you haven't already configured how you want it to be.

The only "hole" in this model is that there isn't really a way to set the user settings back to `null`.

But, if we look at my suggested "use user storage" setting in this light, we would have:

* A global setting that is set to whatever default that is chosen by "the maintainers" or who it is that ultimately decides that. That setting remains in effect unless it has been changed, either through the "MainUI settings" on the OH settings page on the server, or through some yet to be invented file configuration setting.
* A user setting that defaults to `null` and can be managed wherever you manage users. If not set, it will just inherit the global setting.
* A browser local setting that defaults to `null`. If set, either to on or off, it becomes explicit on that device, if not, it will "follow" changes to the user and global setting. The "hole" is present here as well, in that you can't really "unset" the local setting and make it follow "parent defaults" again, but it can be implemented in the UI if really wanted, the data structure/logic supports it, there's just no precedent for how such a "clear" control would look for a switch (you could have the "X" next to it, but how do you "show" if it is in fact using a local or inherited setting?).

This is how I imagine everything, except that not everything might need/be suitable for a global setting.

mherwege:
I think that can work. But do we need the 3 levels? I think that can only create confusion. The question will become: which sync flag do I need to switch on to make sync work? And you need all the above text to explain. That's not what I call intuitive. Its logical, but not intuitive. I had envisioned only 2 levels:
- a server stored flag by user
- a local flag by user (and as it is local, by definition also by device)

All settings should be stored on the server and locally with a user key (that is not the case right now).

These 2 flags work together. Why do you need another global setting accross users? And as they work together, as I explained, they can logically be controlled with a single switch (which will act by user).

By device is actually also a bit of a misnomer. If you use 2 different browsers on the same machine, it will be by browser. But if sync is on, it should sync between these.

Nadahar:
> Why do you need another global setting accross users?

One of my many plans are to implement user management in MainUI. I imagine a future where users are actually used, where different family members actually use different users with different permissions and customization.

The need for the "global" level comes from being able to preconfigure an installation, basically decide what settings should apply to ever new user created.

One other thing I forgot to mention. I have started looking at the user system in core to figure out how to solve this, and I must make clear that the "implicit user role" is *not* a user. You aren't tied to an identity that can be used to store settings when not logged in with "implicit user role" enabled. So, in those situations, only global and local are possible "storage levels".

We could of course make some workaround for that, but I'd say that making user administration from the UI possible combined with enabling options for per-user configuration, would be a strong enough incentive that I would guess that many would opt to actually log in instead of using the implicit user role setting. Today, the setting is very prevalent, but I believe that's mostly because creating users requires that you use the console - or hack JSONDB.

Most installations therefor don't even have a non-admin user account that we could attach the "implicit role" to even if we wanted to.

> By device is actually also a bit of a misnomer. If you use 2 different browsers on the same machine, it will be by browser. But if sync is on, it should sync between these.

If you're referring to my terminology use, I'm sorry if I said device. I prefer to use "local" for that reason, of course it is per browser. But, if you're using a phone and the OH app, it might appear to you as if the setting is "for that device".

Nadahar:
I must also note, that I see some possible complications with the model I have in mind: There would have to be some kind of "type limitation" to what you could store in the user store. First, because they'd have to be serialized to JSON, but also because if it should be "within reason" to actually make UI configurable server side settings (for user/global) level, they would have to map to e.g. configuration description parameters. If we could make it work within those restrictions, making the UI for managing the global settings would be manageable. If not, it could become a huge task.

While you can use the actual settings you make in the client to set the user level settings (if storage is `user`, not `local`), I don't think it would ever be appropriate to let client settings "write" to the global settings.

This could pose challenges with e.g. storing the menu structure, I haven't looked at exactly how those data are organized today, but they might not be easy to fit inside the "config description parameter" limitations. Of course, it's possible to make custom controls for some things, but the amount of work grows quickly then, and might be the difference between a "pipe dream" and "reality".

mherwege:
> We could of course make some workaround for that, but I'd say that making user administration from the UI possible combined with enabling options for per-user configuration, would be a strong enough incentive that I would guess that many would opt to actually log in instead of using the implicit user role setting. Today, the setting is very prevalent, but I believe that's mostly because creating users requires that you use the console - or hack JSONDB.

It is true that, if user management was properly implemented, it may entice more people to do it properly. At the same time, why would one? I use the implicit user role at my house. My children are all adults. There has not been a need to differentiate. Why make it more complex than it is. Don't get me wrong. I strongly believe we do need proper user management. But I don't want to impose it. And the sync functionality should work without it.
To me, the implicit user role is a user. It may technically be different, but from the user perspective, it is the default user. And the logic I outline above can work for that. If there is no real user logged in, it is the implicit user role and it is stored without the user ID. It will then sync these values without client ID.
You could argue that is the global setting that should be used as the global master, but I think that is wrong. That implicit user should be treated as another user (even it requires a workaround), or people will get lost with that implicit user being taken away.

> If you're referring to my terminology use, I'm sorry if I said device. I prefer to use "local" for that reason, of course it is per browser. But, if you're using a phone and the OH app, it might appear to you as if the setting is "for that device".

I was the one using by device, so just wanted to make it clear what it meant. If there is any appology, it should be mine. But I think that terminology is not a problem though and I think we all understood what it meant.

> The need for the "global" level comes from being able to preconfigure an installation, basically decide what settings should apply to ever new user created.

I understand what you want to achieve, but it then requires a management UI for all the UI options, separate from the function where some of these currently reside. And if you take these out of context, it is much more difficult to understand what they do and how they should be set. As the number of the options grows this becomes a nightmare. Or you would need a master user, anyway login as a master user, do all the configuration, it gets saved to the server, and then create a new user from that master user.
Would it then not be smarter to have the ability to copy the UI config from one user to another user as an advanced admin function ones user management is implemented? You would configure the UI as a logged-in user. That is, in my view, a more natural flow. You see what you do. You can then copy over the config to another user. I still don't see why you need to global level for that.

Nadahar:
> But I don't want to impose it. And the sync functionality should work without it.
> To me, the implicit user role is a user. It may technically be different, but form the user perspective, it is the default user. And the logic I outline above can work for that. If there is no real user logged in, it is the implicit user role and it is stored without the user ID. It will then sync these values without client ID.

I never meant to impose anything, but I was thinking that when using the implicit user, user level settings might not be an option. That would mean that either the global defaults applied, or you used local settings, like today.

This might reduce the usefulness of the whole endeavor significantly though, so you're probably right that the "anonymous/implicit user" should be treated as a "special user" when it comes to store storage. It would make the benefits more immediate for people. I still hope people will see the benefit of customizing the UI to most easily access what "they use" once user management becomes more manageable though.

> I understand what you want to achieve, but it then requires a management UI for all the UI options, separate from the function where some of these currently reside. And if you take these out of context, it is much more difficult to understand what they do and how they should be set.

My approach is (almost always): Design the structure to be as flexible and robust as possible (within some reason), then build simpler solutions on top of that. Then you can expand it bit by bit, adding more value.

The "global" settings for example, could just be a hardcoded set of defaults from the start. They would essentially replace the hardcoded defaults inside MainUI if "sync" is enabled. The user settings will then be written by the client when enabled, without requiring any changes to the existing layout except the extra setting for "sync/local/user". But, with the logical structure in place, it would be possible to keep working on UI user management, global settings management, file-based settings management etc., and every "piece" added would add some value, and would fit into the structure and start "working" without breaking anything or requiring redesign of the structure itself.

This is what I envision, laying the foundation for something while making it usable before everything is built.

mherwege:
> My approach is (almost always): Design the structure to be as flexible and robust as possible (within some reason), then build simpler solutions on top of that. Then you can expand it bit by bit, adding more value.

I don't have anything against that. I don't think that more choice and flexibility always means more value though. Overall, OH is already on the more flexible side of the equation, not the user friendliness side. More options and more flexibility can also mean more effort to configure it, options nobody understands without reading the manual, questions about why something happens because very specific options are et in one way or the other on another device and and long forgotten about. It is about striking a balance.

Nadahar:
> More options and more flexibility can also mean more effort to configure it, options nobody understands without reading the manual, questions about why something happens because very specific options are et in one way or the other on another device and and long forgotten about. It is about striking a balance.

I largely agree with that, which is why I think it's important to create "well thought through" mechanisms and defaults, so that they can be there without adding work or confusion, or without people even being aware of it, until the day they need it. Then the functionality is there and can be used. Managing to create that is usually my "ultimate goal".

PR #4413, opened 2026-07-29
habot - Prefer `Stream.toList()` over `Stream.collect(Collectors.toList())`

dilyanpalauzov:


florian-h05 (approved):
LGTM, thanks!

Applying small patch exception for DCO.

PR #4414, opened 2026-07-29
Fix documentation link on State Description Metadata editor

jimtng:


rkoshak:
I literally hit this just now. I was coming to file an issue.

florian-h05 (commented):
IMO it is kind of unfortunate to link into the developer docs.

jimtng:
@florian-h05 so what should we do here? Should we rearrange the docs first and then point to the new docs? or fix it now and when the docs is rearranged, update it? I'd say merge this now and re-point later if needed, unless someone wants to work on the docs now.

PR #4415, opened 2026-07-30
 Re-evaluate `oh-context` values after item-state hydration

### LINKED ISSUE #2842: Item states not tracked at context variable and constant evaluation
oh-context variables and constant expressions are evaluated before the component is mounted. If the UI is newly opened (or recently reset) any item states are not available via the store when those expression are calculated for the very first time.

Replicate the issue

Here's a basic widget:
```yaml
uid: demo-context-error
tags: []
props:
  parameters: []
  parameterGroups: []
timestamp: Oct 27, 2024, 11:04:06 AM
component: oh-context
config:
  constants:
    testItemConst: =`testNumber state is ${items.testNumber.state}`
  variables:
    testItemVar: =`testNumber state is ${items.testNumber.state}`
slots:
  default:
    - component: f7-card
      config:
        title: =`testNumber state is ${items.testNumber.state}`
        content: =const.testItemConst
        footer: =vars.testItemVar
```
All three text outputs should be the same:
image

However, after a [ctrl + shft + f5] the result is different:
image

At this point, however, the item is now in the store because of the title expression so any further interaction (refresh the widget, leave the page and come back, etc.) will result in the proper calculations.

Even if you do activities first, for example open some other widgets or pages in the UI, the first time this widget is opened the item states are not properly available.

...except! Here's a second widget:
```yaml
uid: demo-context-error-2
tags: []
props:
  parameters: []
  parameterGroups: []
timestamp: Oct 27, 2024, 11:16:00 AM
component: f7-card
config:
  title: Add testNumber to store
  content: =items.testNumber.numericState
```
Opening this widget after a refresh, but before the first you open the first widget means that the first widget will be properly displayed the first time it opens. The testNumber item is now being tracked as a result of the second widgets expression being called prior to the first widget's use.

I assume there is a very simple solution to this, but after poking around for a few minutes, I admit, I don't see it.

mannemiethe:
## Fixes

Fixes #2842.

## Problem

`oh-context` constants and variable defaults are evaluated before the child widgets are displayed.

If the UI has just been opened or hard-refreshed, item states might not yet be available in the item state store when these expressions are evaluated. In that case, `const.*` and `vars.*` values can be initialized with missing, stale, or incomplete values.

Example from #2842:

```yaml
component: oh-context
config:
  constants:
    testItemConst: =`testNumber state is ${items.testNumber.state}`
  variables:
    testItemVar: =`testNumber state is ${items.testNumber.state}`
slots:
  default:
    - component: f7-card
      config:
        title: =`testNumber state is ${items.testNumber.state}`
        content: =const.testItemConst
        footer: =vars.testItemVar
```

Expected result:

```text
title   -> current testNumber state
content -> current testNumber state
footer  -> current testNumber state
```

Actual result after a hard refresh:

```text
title   -> correct item state
content -> missing or stale value from const.testItemConst
footer  -> missing or stale value from vars.testItemVar
```

The direct child expression:

```yaml
title: =`testNumber state is ${items.testNumber.state}`
```

can update once the item state becomes available, because it is evaluated inside the normal widget expression flow.

However, the `oh-context` values:

```yaml
constants:
  testItemConst: =`testNumber state is ${items.testNumber.state}`

variables:
  testItemVar: =`testNumber state is ${items.testNumber.state}`
```

were already evaluated once before the item state was available. They then kept that initial incorrect value.

## Why this happens

Before this change, `oh-context` evaluated constants and variable defaults once during initialization:

```javascript
this.const[key] = this.evaluateExpression(key, sourceConst[key])
this.localCtxVars[key] = this.evaluateExpression(key, sourceCtxVars[key])
```

At that point, the expression can access `items.testNumber.state`, but if `testNumber` is not yet present in the item state store, the expression result is based on the missing state.

Later, when the item state arrives, the child widget can update its own direct `items.*` expressions, but the already initialized `const.*` and `vars.*` values are not recalculated.

## Solution

This PR tracks which item states are accessed while evaluating `oh-context` constants and variable defaults.

During the initial evaluation, `oh-context` uses a small tracking wrapper around the item state store. This allows it to detect expressions such as `items.testNumber.state`.

If one of the accessed item states is not available yet, `oh-context` waits until all missing item states are present, then evaluates the constants and variable defaults once again.

Simplified behavior:

```text
initial evaluation
-> track accessed item states
-> detect missing item states
-> wait until missing states arrive
-> evaluate constants and variable defaults once again
-> pass corrected values to child widgets
```

## Reactivity note

This PR does **not** make `oh-context` constants or variable defaults fully reactive. That is intentional.

`oh-context` constants are meant to be stable values inside the child context. Variable defaults are also initial values; after initialization, variables can be changed by widget variable actions.

This PR only fixes the initial missing-state hydration problem. The new behavior is:

```text
initial evaluation
-> item state missing
-> wait until the missing state arrives
-> evaluate once again
-> keep that value
```

It does not mean:

```text
item state changes every time
-> const / variable default recalculates every time
```

Making these values fully reactive would change the existing semantics of `oh-context`. It could also break widgets that rely on constants being stable or overwrite variables that were changed later by widget actions.

## Result

After this change, `oh-context` constants and variable defaults that depend on item states are correctly initialized after a hard refresh or first page load, even when the item state was not available during the first evaluation.

Direct child expressions and `const.*` / `vars.*` values now start from the same item state once the initial item state has arrived, while preserving the existing non-reactive semantics of constants and variable defaults.

### DIFF
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/context.ts
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/context.ts
@@ -9,11 +9,11 @@ export default () => [
   pt(
     'constants',
     'Widget Constants',
-    'Object with key:constant pairs. Constants are available to expressions in all child components via the <code>const</code> object.<br />Constants are evaluated once before the widget is displayed the first time and cannot be changed.'
+    'Object with key:constant pairs. Constants are available to expressions in all child components via the <code>const</code> object.<br />Constants are evaluated before the widget is displayed and cannot be changed. If an initial item state dependency is not available yet, constants are re-evaluated once when the missing item state arrives.'
   ),
   pt(
     'variables',
     'Widget Variables',
-    'Object with key:variable default value pairs. Variables are available to expressions in all child components via the <code>vars</code> object and take precedence over variables with the same name from higher contexts.<br />Variables are evaluated once before the widget is displayed the first time . Their values can only be changed by other component variable actions (e.g. <a class="external text-color-blue" target="_blank" href="https://www.openhab.org/docs/ui/components/oh-button.html#action-variable">oh-button</a>)'
+    'Object with key:variable default value pairs. Variables are available to expressions in all child components via the <code>vars</code> object and take precedence over variables with the same name from higher contexts.<br />Variables are evaluated before the widget is displayed. If an initial item state dependency is not available yet, variable defaults are re-evaluated once when the missing item state arrives. Their values can later be changed by component variable actions (e.g. <a class="external text-color-blue" target="_blank" href="https://www.openhab.org/docs/ui/components/oh-button.html#action-variable">oh-button</a>).'
   )
 ]
--- a/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/context.ts
+++ b/bundles/org.openhab.ui/web/src/assets/definitions/widgets/system/context.ts
@@ -9,11 +9,11 @@ export default () => [
   pt(
     'constants',
     'Widget Constants',
-    'Object with key:constant pairs. Constants are available to expressions in all child components via the <code>const</code> object.<br />Constants are evaluated once before the widget is displayed the first time and cannot be changed.'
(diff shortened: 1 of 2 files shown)


florian-h05:
@mannemiethe Can you please rebase and resolve conflicts?

mannemiethe:
done

PR #4432, opened 2026-08-01
Fix missing Day.js locale definitions for `en`/`en-us` locale

florian-h05:


florian-h05:
Should be backported to 5.2.x.

jsjames:
Is this the best fix - I would think folks in the US would still want date formatted as M/D/YYYY (I know that is not the logical way to do it :) ).

From what I've read, basic en formats should be built in, however, more advance (LT, LTS, L, LL) need to be pulled in from the localeData plugin.

```
import dayjs from 'dayjs';
import localizedFormat from 'dayjs/plugin/localizedFormat';
import localeData from 'dayjs/plugin/localeData';

// Extend the core library functionality
dayjs.extend(localizedFormat);
dayjs.extend(localeData);

// Now localized formats will resolve correctly for 'en'
console.log(dayjs().format('L')); 
```

florian-h05:
So we need both localizedFormat and localeDate?

jsjames:
> So we need both localizedFormat and localeDate?

I think so - I believe they work together - but I didn't test it.

florian-h05:
Will try to have a look soon ...

jsjames:
@florian-h05 - I recommend this fix given the lower case formats are not "built-in".

florian-h05:
Your changes LGTM!
I haven’t tested them, I’m on the road right now. 

Since I’ve opened this PR it’s up to you to merge I think.

jsjames (approved):
LGTM

florian-h05:
Can we please backport this?

PR #4434, opened 2026-08-03
Code Editor: Show checkbox focus outline in codemirror search panel

### LINKED ISSUE #4433: Selection in search bar (code editor) not always visible
## The problem

(Copied from https://community.openhab.org/t/selection-in-search-bar-code-editor-not-always-visible/169992)

The new (although, new…) code editor has a nice search bar:

As everywhere, you can “hop around” the UI buttons with the TAB key. The relevant fields then get a bold edge, as seen with “all” in the above example, or you can see a cursor in the fields “Find” or “Replace”.

But between “all” and “Replace”, there are three check boxes, which get no visual ‘confirmation’ of being selected. So when I want to go from “Find” to “Replace”, I tap Tab until I see “all” being selected, and then I have to make sure that I tap Tab four times, because I’m not sure which one of the checkboxes is my “step in between”.

A very minor issue, of course, but I assume the fix is also pretty minor.

Should I create a GitHub issue for it?

## Expected behavior

I would expect the edges to be bold, when the checkbox is selected. 

## Steps to reproduce

Seems too obvious. 

## Your environment

OpenHAB 5.2.1; I can't imagine the rest is important.

## Browser console

I don't think this matters.

## Browser network traffic

I don't think this matters.

## Additional information

Not applicable.

jimtng:
Resolves #4433

Also made the checkbox and label to align vertically

codemirror-search-focus

This should be backported

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
+++ b/bundles/org.openhab.ui/web/src/components/config/controls/script-editor.vue
@@ -90,10 +90,11 @@
     .cm-panel.cm-search
       label:has(input[type="checkbox"])
         position relative
-        top 2px
+        vertical-align middle
         display inline-flex
-        input[type=checkbox]:focus-visible
-          outline 2px solid black
+        input[type="checkbox"]:focus-visible
+          outline 2px solid -webkit-focus-ring-color
+          outline-offset 2px
 </style>
 
 <script setup lang="ts">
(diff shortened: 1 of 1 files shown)


ErikDB87:
You mean horizontally? :)

jimtng:
Thanks for the improvements, @jsjames!

jsjames (approved):
approve with some minor tweaks

florian-h05:
@jsjames Please add labels and milestone.

Issue #4440, opened 2026-08-05
Main UI: unbounded JS heap growth on page navigation - never returns to baseline (client-side memory leak)


matushromada:
## The problem

Navigating between Main UI pages leaks JavaScript heap memory: it grows on every navigation and **never returns to baseline**. GC reclaims some of it (sawtooth), but each navigation permanently retains more.

This is **not a page cache**. Visiting only **two** built-in pages back and forth, the live DOM holds a **constant 3 page elements**, yet the heap climbs **linearly with no plateau** - a bounded cache cannot explain unbounded growth at constant page count. Measured on **completely stock openHAB**, using the **native sidebar navigation** between **built-in pages** (no custom widgets/pages/CSS/JS):

```txt
| Round-trips | JS heap | Pages in DOM |
| ----------: | ------: | -----------: |
|    0 (base) |   26 MB |            3 |
|           1 |   40 MB |            3 |
|           3 |   67 MB |            3 |
|           6 |  106 MB |            3 |
|           9 |  144 MB |            3 |
|          12 |  188 MB |            3 |
|          15 |  242 MB |            3 |
|          18 |  285 MB |            3 |
```

**Page count stays at 3 the entire time; heap grows +259 MB over 18 round-trips.**

~14 MB retained per round-trip, no plateau; a longer run eventually crashes the renderer tab (OOM). Growth is **proportional to the number of widgets on the page** (a light page ~2 MB/trip, a ~40-widget page ~8-14 MB/trip), pointing at per-page component/listener retention on unmount. Reproduced on **5.1.3, 5.2.0 (my infra) and 5.2.1 (official online demo)**

## Expected behavior

Navigating away from a page and back should return the JS heap to approximately its pre-navigation baseline. Repeatedly navigating between a fixed set of pages should reach a steady state, not grow without bound.

## Steps to reproduce

1. Stock openHAB, Main UI. Two pages using only built-in widgets, no custom CSS/JS.
2. Navigate back and forth between them repeatedly - via the **sidebar menu** (native clicks); a scripted `router.navigate()` reproduces it identically. Console sampler (replace `YOUR_PAGE_UID` with the UID of one of your pages):

```js
const r = document.querySelector('.view-main').f7View.router;
const mb = () => Math.round(performance.memory.usedJSHeapSize/1048576);
r.navigate('/overview/',{animate:false}); await new Promise(x=>setTimeout(x,600));
const base = mb(), s = [];
for (let i=1;i<=18;i++){
[...]

PR #4443, opened 2026-08-08
Fix several memory leak causes

florian-h05:
Related to #4440.

When squash merging this PR, please use this commit message (copy here and paste below):

```md
Addresses several memory leak causes, component lifecycle issues, event listener cleanup problems, and unintended side-effects across the codebase:

* `oh-chart` & `oh-chart-page`: Fix `this` references in `<template>`.
* `oh-list-item`: Clean up leftover event listener from PR #3862.
* `oh-canvas-layout`: Clean-up event listeners referencing non-existing method.
* `$oh.see`: Decouple exposed logical connection handle from underlying EventSource instance. Fixes stale EventSource references in consumers and `$oh.sse`.
* `developer-sidebar` & `developer-tools`: Ensure SSE and WebSocket connections are stopped before unmounting.
* Global event listeners: Always ensure global event listeners are properly removed.
* `model-dragdrop-mixin`: Only add or remove global event listeners if `window` is defined.
* AGENTS.md: Add documentation regarding unclosed SSE/WebSockets and global event listener cleanup.
* Memory Leaks and Async Lifecycle Safety in Framework7:
  * Router: Clear `page.pageFrom` references after `afterIn` callbacks to release retained DOM elements.
  * Tooltip: Extend the `pageBeforeRemove` hook to destroy all `f7Tooltip` instances within `$el` and `$navbarEl`.
  * Menu: Fix `menuClosed` listener never detached.
  * Framework7-Vue: Add `isMounted` checks to async `f7ready` callbacks and ensure explicit `.destroy()` calls on `f7Searchbar` and `f7Tooltip` during `onBeforeUnmount`.
* `overview-tab`: Fix `defineVars` assignment causing side-effects in computed properties (improving the fix from PR #4038).
* `home`: Prevent unnecessary component recreation on overview tab transitions by removing `this.overviewPageKey = f7.utils.id()` from `onPageBeforeIn()`,
   allowing unmount hooks (like window resize listeners in masonry layout) to clean up gracefully.

-----

Signed-off-by: Florian Hotze <dev@florianhotze.com>
```

### DIFF
--- a/bundles/org.openhab.ui/web/patches/framework7+7.1.5.patch
+++ b/bundles/org.openhab.ui/web/patches/framework7+7.1.5.patch
@@ -1,44 +1,39 @@
-diff --git a/node_modules/framework7/package.json b/node_modules/framework7/package.json
-index f607800..24e6369 100644
---- a/node_modules/framework7/package.json
-+++ b/node_modules/framework7/package.json
-@@ -4,12 +4,30 @@
-   "description": "Full featured mobile HTML framework for building iOS & Android apps",
-   "type": "module",
-   "exports": {
--    ".": "./framework7.esm.js",
--    "./core": "./framework7.esm.js",
--    "./bundle": "./framework7-bundle.esm.js",
--    "./lite": "./framework7-lite.esm.js",
--    "./lite/bundle": "./framework7-lite-bundle.esm.js",
--    "./lite-bundle": "./framework7-lite-bundle.esm.js",
-+    ".": {
-+      "default": "./framework7.esm.js",
-+      "types": "./framework7-types.d.ts"
-+    },
-+    "./core":  {
-+      "default": "./framework7.esm.js",
-+      "types": "./framework7-types.d.ts"
-+    },
-+    "./bundle": {
-+      "default": "./framework7-bundle.esm.js",
-+      "types": "./framework7-types.d.ts"
-+    },
-+    "./lite": {
-+      "default": "./framework7-lite.esm.js",
-+      "types": "./framework7-types.d.ts"
-+    },
-+    "./lite/bundle": {
-+      "default": "./framework7-lite-bundle.esm.js",
-+      "types": "./framework7-types.d.ts"
-+    },
-+    "./lite-bundle": {
-+      "default": "./framework7-lite-bundle.esm.js",
-+      "types": "./framework7-types.d.ts"
-+    },
-     "./less": "./framework7.less",
-     "./less/bundle": "./framework7-bundle.less",
-     "./css": "./framework7.css",
+diff --git a/node_modules/framework7/components/tooltip/tooltip.js b/node_modules/framework7/components/tooltip/tooltip.js
+index c4ffd27..533b1cc 100644
+--- a/node_modules/framework7/components/tooltip/tooltip.js
++++ b/node_modules/framework7/components/tooltip/tooltip.js
+@@ -99,13 +99,12 @@ export default {
+     },
+ 
+     pageBeforeRemove(page) {
+-      const app = this;
+-      page.$el.find('.tooltip-init').each(el => {
++      page.$el.find('*').each(el => {
+         if (el.f7Tooltip) el.f7Tooltip.destroy();
+       });
+ 
+-      if (app.theme === 'ios' && page.view && page.view.router.dynamicNavbar && page.$navbarEl && page.$navbarEl.length > 0) {
+-        page.$navbarEl.find('.tooltip-init').each(el => {
++      if (page.$navbarEl && page.$navbarEl.length > 0) {
++        page.$navbarEl.find('*').each(el => {
+           if (el.f7Tooltip) el.f7Tooltip.destroy();
+         });
+       }
+diff --git a/node_modules/framework7/modules/router/router-class.js b/node_modules/framework7/modules/router/router-class.js
+index 01963aa..c35bb06 100644
+--- a/node_modules/framework7/modules/router/router-class.js
++++ b/node_modules/framework7/modules/router/router-class.js
+@@ -1072,6 +1072,10 @@ class Router extends Framework7Class {
+     $pageEl.trigger(colonName, page);
+     router.emit(camelName, page);
+ 
++    if (callback === 'afterIn' && page.pageFrom) {
++      page.pageFrom = null;
++    }
++
+     if (callback === 'beforeRemove' || callback === 'beforeUnmount') {
+       detachEvents();
+ 
 diff --git a/node_modules/framework7/shared/utils.d.ts b/node_modules/framework7/shared/utils.d.ts
 index c88ac4c..e29088b 100644
 --- a/node_modules/framework7/shared/utils.d.ts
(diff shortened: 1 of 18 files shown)


florian-h05:
This was heavily assisted by AI (harness: Antigravity 2.x, Claude Sonnet 4.6 for analysis, Gemini Flash 3.6 Low for fixing findings). I did three rounds of memory debugging consisting of:

1. Doing a baseline heap snapshot at the overview page.
2. Performing 10 navigation roundtrips between overview page and an oh-layout-page (EMS overview on demo.openhab.org).
3. Doing a target heap snapshot at the overview page.
4. Triggering GC, waiting 15 seconds and doing a final heap snapshot.

Feeding that into Sonnet 4.6 lead to these analysis results for each round:

[memory_leak_analysis_R1.md](https://github.com/user-attachments/files/30855427/memory_leak_analysis_R1.md)
[memory_leak_analysis_R2.md](https://github.com/user-attachments/files/30855428/memory_leak_analysis_R2.md)
[memory_leak_analysis_R3.md](https://github.com/user-attachments/files/30855430/memory_leak_analysis_R3.md)

florian-h05:
Here is an AI generated summary of the measured (in the heap snapshots) improvements:

***

### 🚀 Memory Leak Analysis & Improvements Summary

Through heap snapshot comparison using `memlab`, we identified and successfully resolved major memory leaks that caused heap usage to spike from **65 MB to 120+ MB** over repeated round-trips:

#### 1. **Resolved: Previous Page DOM Retention (`f7Page.pageFrom.el`)** ✅
* **Issue:** Framework7's router retained the entire DOM tree of the previously-visited page (`page-previous`) to support animations. Because this page tree contained hundreds of Vue widget nodes, it kept thousands of reactive Vue objects (`ComputedRefImpl`, `Dep`, `ReactiveEffect`) pinned in memory.
* **Fix:** Excluded the heavy reactive subtrees from the retention path (e.g. by nulling out `pageFrom.el` after page transitions), resulting in an **~86% reduction in leaked heap mass** (leaked mass dropped from **~165 MB to ~23 MB** in development, with production being even cleaner).

#### 2. **Resolved: Orphaned Window Listeners via Key-Cycling (`masonry-grid`)** ✅
* **Issue:** Key-cycling the `overviewPageKey` on every navigation back to the home page forced a complete destruction and recreation of the widget subtree. Due to timing overlaps with F7's transitions, the `onBeforeUnmount` hook in `masonry-grid.vue` was skipped for some instances, leaving orphaned `resize` event listeners on the `window` object that leaked the component closure.
* **Fix:** Corrected the timing/lifecycle of the key-cycling (or removed it in favor of reactive store updates) to ensure Vue unmount lifecycles propagate cleanly and event listeners are properly unregistered.

#### 3. **Remaining Minor Item: Bounded F7 Tooltip Leak** ℹ️
* **Observation:** A minor, bounded leak persists where F7 tooltip elements (`f7Tooltip.el`) are detached but not fully destroyed when navigating away from navbar items. This is bounded (~150-180 KB total) and does not grow infinitely. *(Optional: A subsequent cleanup in `onPageBeforeOut` using `f7.tooltip.destroy` will address this).*

bundles/org.openhab.ui/web/src/js/openhab/sse.ts
```diff
@@ -141,8 +141,15 @@ function newSSEConnection(
             // Close the current broken connection
             es.close()
             es.clearKeepalive()
-            // Reinitialize the connection
+            // Reinitialize the connection and update openSSEClients so that
+            // SSEService.close() can reach the new live EventSource.
+            const oldIndex = openSSEClients.indexOf(es)
```
wborn: The caller still holds the `EventSource` originally returned by `newSSEConnection()`. After a reconnect, this replaces that object in `openSSEClients`, but the caller's reference is never updated. A later `SSEService.close(this.sseClient)` therefore looks up the old object, gets `-1`, and closes only that already-broken source while the replacement remains active.

There also seems to be a race when `close()` is called while a reconnect timeout is pending: closing the original source leaves it with `readyState === CLOSED`, so the pending timer can still create a replacement and take the `else` branch here, effectively reopening an explicitly closed connection.

Could we track a logical connection separately from the underlying `EventSource`, with a stable handle, the current `EventSource`, a closed flag, and the pending reconnect timer? Then `close()` could cancel any pending reconnect and always close the current underlying source.

A regression test covering both `reconnect -> close(original handle)` and `close -> pending reconnect timer fires` would also be useful here.

wborn (changes requested):
I'm not a maintainer in this repository; I'm just trying to help by having my AI take a first pass over the changes. A maintainer review is still needed.

Overall, the cleanup changes look sensible, but the AI found one issue in the SSE reconnect handling that seems important to address because it can preserve the kind of connection/component leak this PR is intended to fix.

I've left an inline comment with the details.

bundles/org.openhab.ui/web/patches/framework7+7.1.5.patch
```diff
@@ -1,3 +1,39 @@
+diff --git a/node_modules/framework7/components/tooltip/tooltip.js b/node_modules/framework7/components/tooltip/tooltip.js
```
jsjames, line 1: Do you know if these changes are included in later framework7 versions - and if so, are the changes 'aligned' with how they addressed in the framework?

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-layout.vue
```diff
@@ -234,9 +234,7 @@ export default {
     }
   },
   beforeUnmount() {
-    if (!this.context.editmode) {
-      window.removeEventListener('resize', this.setDimensions)
-    }
+    window.removeEventListener('resize', this.setDimensions)
```
jsjames: I don't see where setDimensions is defined?  Its not on this component or the mixin.

bundles/org.openhab.ui/web/patches/framework7+7.1.5.patch
florian-h05: Haven’t checked that as the upgrade path is more or less blocked due to their removal of the Aurora theme.
Should I check?

bundles/org.openhab.ui/web/src/components/widgets/layout/oh-canvas-layout.vue
florian-h05: I have to check that. I didn’t look that close at the full code of the components, only at the memory related parts.

bundles/org.openhab.ui/web/patches/framework7+7.1.5.patch
florian-h05: And just FYI; I've opened a PR to F7 icons nearly two years ago (the PR is just a variant of an existing icon with a slash) and haven't received any feedback until now, so my motivation to open a PR against F7 is near-zero: https://github.com/framework7io/framework7-icons/pull/49

jsjames (approved):
LGTM

PR #4448, opened 2026-08-12
[BasicUI] Confirmation dialog

mherwege:
Depends on https://github.com/openhab/openhab-core/pull/5769

This PR implements a confirmation dialog in BasicUI. It shows a confirmation dialog when changing the state through a widget with the `confirmCmd` parameter set.

See the core PR for a gif showing what it looks like.


mherwege:
Note this PR has a separate commit configuring the build to create sourcemaps for javascript. This made debugging a lot easier. It should only have a development impact. I can remove this if deemed inappropriate.

PR #4449, opened 2026-08-15
[GHA] Add new workflow for upgrading maven wrapper

holgerfriedrich:
This is the same workflow we use on other repos to upgrade mvnw and mvn version.
See openhab/openhab-addons#20593

Once the workflow is triggered (either by cron or manually from the actions tab), it will create a new PR, upgrading mwnw and mvn version from 3.9.13 to 3.9.16.

Project settings might need adaptation to allow the PR creation.

### DIFF
--- a/.github/workflows/update-maven.yml
+++ b/.github/workflows/update-maven.yml
@@ -1,0 +1,83 @@
+name: Update Maven Wrapper
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


holgerfriedrich:
obsoleted. dependabot can now upgrade mvnw and mvn...

PR #4451, opened 2026-08-17
Fix mvn signature, use sha256sum

holgerfriedrich:
dependabot now handles mvnw and mvn upgrades.
sha512 signature is anyway useless as it is ignored by mvnw.

This PR replaces sha512 checksum by sha256. It will then be updated by dependabot as well.


jsjames (approved):
LGTM

PR #4452, opened 2026-08-18
Charts: Apply series offset before rendering series components (fix oh-aggregate-series with offset being empty)

### LINKED ISSUE #4438: [MainUI] Offsets (Comparison) no longer work in oh-aggregate-series charts in openHAB 5.2.1
## The problem

Offsets (Comparison) no longer work in oh-aggregate-series charts in openHAB 5.2.1. Series with no offset are OK, but series with an offset are not shown anymore. "Tested" with infliuxdb and rrd4j.

Image

Temperature from yesterday is missing:

Image

## Expected behavior

This is a screenshot after downgrade to 5.2.0. All good:

Image

## Steps to reproduce

It can be reproduced on demo page: https://demo.openhab.org/page/chart_tabs/3

## Your environment
All good in 5.2.0. Problem started with patch release 5.2.1

Ltty:
Fixes #4438.

## Description

Series using *Offset Amount* / *Offset Unit* render empty since #4264.

`useChart` requests such a series for a shifted period, but passed the **unshifted** chart `startTime`/`endTime` to the series component's `get()`. For `oh-aggregate-series`, that meant every group was filtered out as being before the unshifted `startTime`:

```ts
.filter((_d, idx) => !groups[idx][0].isBefore(startTime))
```

For an offset series *all* groups are before that time — e.g. a `year` chart starting `2026-01-01` with `offsetAmount: 1, offsetUnit: year` only ever has `2025-xx` groups — so the entire series was dropped.

`useChart`'s `getSeriesPromises()` now shifts `startTime`/`endTime` for offset series once, and passes that shifted period both to the persistence query (as before) and to the series component's `get()` (new). This fixes the empty series above, and as a side effect also fixes `dimensionFromDate()` placing an offset series at the wrong index on multi-year charts, since it also relied on the unshifted `startTime`. Non-offset series are unaffected.

Reproducible on the demo instance: https://demo.openhab.org/page/chart_tabs/3

## Testing

Added a regression test to `useChart.test.ts` covering a series with an offset, exercising the whole flow (persistence query args + resulting series data); it fails on `main` (`expected [] to have a length of 2`) and passes with the fix.

`vitest`, `oxfmt --check`, `eslint` and `vue-tsc` are all clean.

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-aggregate-series.test.ts
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-aggregate-series.test.ts
@@ -210,5 +210,33 @@ describe('aggregateSeries', () => {
       expect(result.data[0][0]).toBe(0)
       expect(result.data[1][0]).toBe(1)
     })
+
+    it('should keep the data points of a series with an offset', () => {
+      const component = {
+        config: {
+          item: 'TestItem',
+          aggregationFunction: AggregationFunction.average,
+          offsetAmount: 1,
+          offsetUnit: 'day'
+        }
+      } as any
+      // The data is requested for the offset period, while get() receives the unshifted startTime
+      const offsetStartTime = startTime.subtract(1, 'day')
+      const points = [
+        {
+          name: 'TestItem',
+          data: [
+            { time: offsetStartTime.valueOf(), state: '10' }, // Hour 0
+            { time: offsetStartTime.add(1, 'hour').valueOf(), state: '20' } // Hour 1
+          ]
+        }
+      ] as any
+
+      const result = aggregateSeries.get(context, component, points, startTime, endTime)
+
+      expect(result.data).toHaveLength(2)
+      expect(result.data[0]).toEqual([0, '10'])
+      expect(result.data[1]).toEqual([1, '20'])
+    })
   })
 })
--- a/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-aggregate-series.test.ts
+++ b/bundles/org.openhab.ui/web/src/components/widgets/chart/series/oh-aggregate-series.test.ts
@@ -210,33 +210,5 @@ describe('aggregateSeries', () => {
       expect(result.data[0][0]).toBe(0)
       expect(result.data[1][0]).toBe(1)
     })
-
-    it('should keep the data points of a series with an offset', () => {
-      const component = {
-        config: {
-          item: 'TestItem',
-          aggregationFunction: AggregationFunction.average,
-          offsetAmount: 1,
-          offsetUnit: 'day'
-        }
-      } as any
-      // The data is requested for the offset period, while get() receives the unshifted startTime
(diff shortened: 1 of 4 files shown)


florian-h05:
@jsjames Please let me review this, I have worked a lot on charting in the past months

florian-h05 (commented):
Instead of applying the offset startTime inside oh-aggregate-series, I'd prefer doing this centrally in useChart.

Ltty:
@florian-h05 Done...

Moved the offset shift into `useChart`'s `getSeriesPromises()`. It now passes the already-shifted `seriesStartTime`/`seriesEndTime` to the series component's `get()`, instead of the displayed (unshifted) period, so `oh-aggregate-series` is offset-agnostic again.

As a side effect this also fixes `dimensionFromDate()` placing an offset series at the wrong index on multi-year charts (`twoYears`/`threeYears`/`fiveYears`), since it derives the category index from `startTime.year()`. That was still broken on `main` even with a correct filter, because `get()` never saw the shifted start.

Moved the regression test from `oh-aggregate-series.test.ts` to a new `useChart.test.ts`, since that's where the offset handling now lives — it exercises the whole flow (persistence query args + resulting series data) and fails on `main` for the reason above.

One thing I left out of scope: `boundary`/`itemState` (whether to fetch the current Item state at the query boundary) are still decided from the *unshifted* start/end. For a chart that includes "now", an offset series can therefore get the current Item state appended past the end of its (past) period. Happy to fix that here too, or as a separate follow-up. Let me know which you'd prefer.

florian-h05:
> Moved the regression test from oh-aggregate-series.test.ts to a new useChart.test.ts, since that's where the offset handling now lives — it exercises the whole flow (persistence query args + resulting series data) and fails on main for the reason above.

Do you have AI communicating on your behalf on GitHub? That sounds very much like AI ...

Ltty:
Actually, no. I tend to use Claude in the review process to make sure technical changes are accurate and complete, but I'm not using it to automate communication.

florian-h05:
Hmm okay, then sorry for the wrong allogations.

I just wanted to check that, because I am not willing to communicate with an AI - there are some cases among our GitHub repos where people used AI to communicate on issues what lead to very inefficient miscommunication and factually wrong stuff being posted.

Ltty:
Fully understood, and no hard feelings about it... I'm actually struggling with the same issue in other repositories.

florian-h05:
> One thing I left out of scope: boundary/itemState (whether to fetch the current Item state at the query boundary) are still decided from the unshifted start/end. For a chart that includes "now", an offset series can therefore get the current Item state appended past the end of its (past) period. Happy to fix that here too, or as a separate follow-up. Let me know which you'd prefer

I quickly fixed that myself inside your PR, had to push a few minor code changes to the regression test anyway.

florian-h05 (approved):
LGTM, thanks!
And congratulations to your first Main UI contribution!

florian-h05:
This will also ship with the next 5.2.x patch release.

PR #4453, opened 2026-08-19
oh-repeater: fix when sourceType is an array of objects

jimtng:
Fix a regression introduced in #4035 

When the source is an array of objects, e.g. `String([ { "foo":"bar" }])` results in `"[object Object]"`.

@jsjames I don't really understand the typing around SourceArray and just added the extra `| string` there - quite likely what I added is wrong there. I'd appreciate it if you could correct it.

This bug broke my page that's relying on oh-repeater with JSON source like this:

```yaml
- component: oh-repeater
  config:
    for: camera
    in: =JSON.parse(@'PhotoFrame_Cameras')
    sourceType: array
```

### DIFF
--- a/bundles/org.openhab.ui/web/src/components/widgets/system/oh-repeater.vue
+++ b/bundles/org.openhab.ui/web/src/components/widgets/system/oh-repeater.vue
@@ -36,7 +36,7 @@ import { ApiError } from '@/js/hey-api'
 import { simpleHash } from '@/js/openhab/utils'
 
 type SourceTypeElement = api.EnrichedItem | api.EnrichedGroupItem | api.EnrichedRule | api.StateOption | api.CommandOption | string | number
-type SourceArray = SourceTypeElement[] | null
+type SourceArray = SourceTypeElement[] | string | null
 
 // define*
 const props = defineProps<{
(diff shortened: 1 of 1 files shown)


jsjames:
Thx for the fix.
the string type is correct - i missed the comma delimited string as a valid option for sourceArray.
I believe the better fix here is to validate the cfg.in object is an array and if so, send it back vs. checking the configuration value.  If you could validate this still fixes the issue for you.

jimtng:
> I believe the better fix here is to validate the cfg.in object is an array and if so, send it back vs. checking the configuration value. If you could validate this still fixes the issue for you.

Yes indeed checking if it's an array is a better fix, thanks! Confirmed it still works.

jsjames (approved):
LGTM

jimtng:
Thanks for merging this so quickly!

> the comma delimited string as a valid option for sourceArray.

Is this documented somewhere?

jsjames:
> Thanks for merging this so quickly!
> 
> > the comma delimited string as a valid option for sourceArray.
> 
> Is this documented somewhere?

Not that I can find anywhere.

florian-h05:
Please backport to 5.2.x

PR #4460, opened 2026-08-23
Chat page: Update splash screen to be vendor-agnostic

florian-h05:


florian-h05:
image

lsiepel:
It crossed my mind that maybe we will have more LLM/HLI bindings in the future, like xAi and Antropic. Do we need to mention the individual ones we have now or make it more generic? 

Anyway, I would drop some of the technical wording. It currently mentions HLI/LLM several times, while there isn't actually an “LLM HLI” binding in the add-on store. If needed, we could add a “Learn more” link/button that explains the technical details.

Maybe something like:
Set up AI to use Chat
Chat needs an AI provider to understand your requests and interact with your openHAB home.

For the red button, I would also try to make it less technical. If possible, it could take you directly to the add-on store with the compatible bindings already filtered (filter by HLI?).

florian-h05:
> It crossed my mind that maybe we will have more LLM/HLI bindings in the future, like xAi and Antropic. Do we need to mention the individual ones we have now or make it more generic?

I tend to mention a few specific ones here, as this makes the whole text way more descriptive and helpful for end users.

I've just checked, both Anthropic and xAI provide OpenAI API-compatible APIs, so the ChatGPT binding can also connect to them. I don't think we should mention all supported providers there explicitly, that would be too many - just the most "relevant ones". Anthropic IMO is too expensive to be listed there, and xAI isn't that popular. 

I'll try to come up with less technical wording :+1:

florian-h05:
Better this way:

image

jsjames (approved):
LGTM

PR #4485, opened 2026-09-01
Harden Content-Security-Policy

florian-h05:
- Removes default-src 'unsafe-eval' as ECharts finally doesn't need this anymore.
- Removes script-src 'unsafe-inline' (by moving 'unsafe-inline' from default-src to style-src) as this doesn't cause violations anymore.
- Add object-src 'none' to block plugins such as Flash, Java applets, etc.
- Add base-uri 'self' to only allow same origin to set `<base href="...">`


jsjames (approved):
LGTM

PR #4502, opened 2026-09-07
group-box: Fix console error when title isn't used

jimtng:
Title was meant to be optional. This fixes an error where title wasn't used, e.g. in `tag-input.vue` or in button wrappers at the bottom of some pages.

image


jsjames (approved):
LGTM

Issue #4491, opened 2026-09-03
[MainUI] Long names of rules are never shown


dilyanpalauzov:
In openHAB 5.2.1 I create a rule with long name, description and UID:

> name = "A0123456789 B0123456789 C0123456789 D0123456789 E0123456789 F0123456789 G0123456789";
> uid="UID-0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVZW-0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVZW";
> description = "A0123456789 B0123456789 C0123456789 D0123456789 E0123456789 F0123456789 G0123456789 H0
123456789 I0123456789 J0123456789 K0123456789 L0123456789 M0123456789 N0123456789 O0123456789";

When I visit in Main UI `GET /settings/rules/` I see the UID and Description, but Name is shortened:

<img width="873" height="407" alt="Image" src="https://github.com/user-attachments/assets/ce07d8eb-ed6d-4103-be7e-5127be4ab467" />

When I load the individual rule, all of name, uid and description are shortened:

<img width="1054" height="690" alt="Image" src="https://github.com/user-attachments/assets/969d890c-d36a-4cb1-9f4d-0ce4af3efc46" />

* On the page with all rules `/settings/rules/` the names of the rules should be shown in their completeness, as is done for uid and description
* On the page showing the details of an individual rule, all of name⇔label, description and UID should be shown in their completeness
* It would be good if both view show UID, Label and Description in the same order.  This is currently not the case.

dilyanpalauzov:
I see now that the name of the rule is displayed twice: at the top of the page and after `Label`.  It should be displayed only once to avoid distraction.  For “not-editable” rules the name of the rule is nevertheless not shown in its completeness, as the rendering at the top puts horizontal ellipsis to shorten its name.

* The horizontal ellipsis happen too late, so that different texts overlap

<img width="1662" height="426" alt="Image" src="https://github.com/user-attachments/assets/0f142263-8c38-4968-a078-cb08ec37d51d" />

PR #4504, opened 2026-09-08
Channel Details: Wrap long Channel UID

jimtng:
## Summary
- Replaces `f7-list-input` (with input box) with a static display component for read‑only fields.
- The new static display:
  - Wraps long text (e.g., long UIDs, labels, descriptions).
  - Supports an optional “copy UID” icon.
  - Provides a consistent, centralized implementation across pages.
- Editable fields continue using the original `f7-list-input`.
- The component is now used for:
  - UID
  - Label
  - Description
- This approach removes duplicated styling logic and avoids global CSS overrides.

Forms affected:
- Item form
- Thing general settings
- Channel general settings
- Rule
- Transformation

## Motivation
Long UIDs and labels were previously truncated or clipped, making them difficult to read.  
This PR introduces a reusable component that ensures proper wrapping and consistent styling everywhere these read‑only fields appear.  
It also lays the groundwork for future refactoring of other pages that need similar treatment.

## Additional Notes
- This resolves the wrapping issues noted in #4491 (except navbar title, to be addressed separately).  
- The component avoids unnecessary global CSS bloat and keeps the implementation DRY.  
- The optional clipboard icon is integrated cleanly, with redundant styling removed.

Before:

image

--- 

After

image

### DIFF
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -581,7 +581,7 @@ html
 .toast .toast-content
   justify-content:  center
 
-.list ul li.uuid-list-input .item-input-wrap
-    white-space normal
-    word-break break-all
+.list ul li.uid-list-input .item-input-wrap
+    white-space normal !important
+    word-break break-all !important
 
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -581,7 +581,3 @@ html
 .toast .toast-content
   justify-content:  center
 
-.list ul li.uid-list-input .item-input-wrap
-    white-space normal !important
-    word-break break-all !important
-
--- a/bundles/org.openhab.ui/web/src/css/app.styl
+++ b/bundles/org.openhab.ui/web/src/css/app.styl
@@ -581,3 +581,7 @@ html
 .toast .toast-content
   justify-content:  center
 
+.list ul li.uuid-list-input .item-input-wrap
+    white-space normal
+    word-break break-all
+
--- a/bundles/org.openhab.ui/web/src/components/item/item-form.vue
+++ b/bundles/org.openhab.ui/web/src/components/item/item-form.vue
@@ -24,14 +24,7 @@
               @click="$oh.utils.normalizeInput('#input')" />
           </template>
         </f7-list-input>
-        <f7-list-input v-else label="Name" type="text" class="uid-list-input" :input="false">
-          <template #input>
-            <span>
-              {{ item.name }}
-              <clipboard-icon v-if="item.name" :value="item.name" tooltip="Copy UID" />
-            </span>
-          </template>
-        </f7-list-input>
+        <wrapped-list-output v-else label="Name" :value="item.name" clipboard />
         <f7-list-input
           label="Label"
           type="text"
@@ -195,7 +188,7 @@
 <script>
 import { f7 } from 'framework7-vue'
 
-import ClipboardIcon from '@/components/util/clipboard-icon.vue'
+import WrappedListOutput from '@/components/util/wrapped-list-output.vue'
 import SemanticsPicker from '@/components/tags/semantics-picker.vue'
 import ItemPicker from '@/components/config/controls/item-picker.vue'
 import GroupForm from '@/components/item/group-form.vue'
@@ -220,7 +213,7 @@ export default {
     stateDescription: String
   },
   components: {
-    ClipboardIcon,
+    WrappedListOutput,
     SemanticsPicker,
     ItemPicker,
     GroupForm,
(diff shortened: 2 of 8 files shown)


jsjames:
What do you think of adding a class style in app.style so this can be done globally vs. in each component?  I would think there are going to be many locations where UID should be wrapped (in addition to the 2 you have done)?

jimtng:
> What do you think of adding a class style in app.style so this can be done globally vs. in each component? I would think there are going to be many locations where UID should be wrapped (in addition to the 2 you have done)?

I was thinking more about making a component of some sort. So far only noticed these 3 places but I haven't been deliberately looking for them.

I'm not super familiar with how vue integrates the stylus. If we added a class in app.styl won't that cause a bloat when not used?

jimtng:
There are other places that could use some wrapping too, example: #4491

A generic structure needs to be devised to cover these individual use cases: some with copy uid, some without. Will be used a lot on uneditable fields (e.g. file based) too.

My current motivation was just to get what I noticed fixed quickly, then maybe later on circle back with a more thorough refactor / clean up but I'm not promising to do that soon, as I'm currently working on something else ATM.

jsjames:
take a look at this - https://github.com/jsjames/openhab-webui/tree/wrap-uid

I think its much simpler and works in those 2 cases (thing-general-settings and channel-general-settings).  I didn't test in the other cases you mentioned.

jsjames:
> There are other places that could use some wrapping too, example: #4491
> 
> A generic structure needs to be devised to cover these individual use cases: some with copy uid, some without. Will be used a lot on uneditable fields (e.g. file based) too.
> 
> My current motivation was just to get what I noticed fixed quickly, then maybe later on circle back with a more thorough refactor / clean up but I'm not promising to do that soon, as I'm currently working on something else ATM.

The app.style is loaded for the full app, but if its a style used in many places, it is better to include it once.  There are 1000's of styles loaded for the app (look at framework7 styles) so adding any is pretty insignificant.

jsjames:
> take a look at this - https://github.com/jsjames/openhab-webui/tree/wrap-uid
> 
> I think its much simpler and works in those 2 cases (thing-general-settings and channel-general-settings). I didn't test in the other cases you mentioned.

I just tested for rules-general-settings and assuming we convert to the same structure, it will work.  Specific to the rules-general-settings, we need to add the :input="false" - or have that dependent on whether the rule is editable or not.  Another issue with rule-general-setting, if I non-editable rule, it shows the type and description can be edited.

image

jimtng:
@jsjames Thanks for taking a look. Since you've got a global approach figured out, feel free to go ahead with your solution and close this PR along with #4501.

bundles/org.openhab.ui/web/src/css/app.styl
```diff
@@ -581,3 +581,7 @@ html
 .toast .toast-content
   justify-content:  center
 
+.list ul li.uuid-list-input .item-input-wrap
```
jimtng: I don't think it should be called `uuid`. A uuid usually looks like `123e4567-e89b-12d3-a456-426614174000`

We should rename this to `uid-list-input` instead

jsjames: If everything else looks good, I can merge since updates your PR.

jimtng:
In this PR:

image

Whereas in #4501

image

Which was why I was thinking a custom component might do it better and DRY the boilerplate of setting it up the same way across different pages.

jsjames:
Making the styles !important fixes that issue.

> Which was why I was thinking a custom component might do it better and DRY the boilerplate of setting it up the same way across different pages.

If you want to create a custom component, that would be a good way to go to keep things consistent.  Was just trying for a minimum update to your PRs.

jimtng:
> Making the styles !important fixes that issue.

Yes, it's wrapped around now, thanks.

> If you want to create a custom component, that would be a good way to go to keep things consistent. Was just trying for a minimum update to your PRs.

I appreciate your help.

What should we do from here: merge this then refactor into component in a separate PR, or would you like me to do it in this before we merge?

jsjames:
> > Making the styles !important fixes that issue.
> 
> Yes, it's wrapped around now, thanks.
> 
> > If you want to create a custom component, that would be a good way to go to keep things consistent. Was just trying for a minimum update to your PRs.
> 
> I appreciate your help.
> 
> What should we do from here: merge this then refactor into component in a separate PR, or would you like me to do it in this before we merge?

If you were going to update this now, let's do it in this PR and merge once.  When you create the custom component, you then wouldn't need the global style since you can put in the new component.

jimtng:
> If you were going to update this now, let's do it in this PR and merge once. When you create the custom component, you then wouldn't need the global style since you can put in the new component.

I'll work on this here.

jimtng:
This should now resolve #4491 too except for the navbar title being too long, but I think that should be addressed separately.

jimtng:
I'm going to see if we can avoid the v-else in non-clipboard cases

jimtng:
> I'm going to see if we can avoid the v-else in non-clipboard cases

I can't think of a better way to avoid the use of v-if / v-else. Every alternative I could think of has drawbacks. So for now I'll stick with the current solution. WDYT?

PR #4505, opened 2026-09-09
Inbox List: Fix a regression from #4498 causing truncation of dialog button texts 

jimtng:
A regression was introduced in #4498 when it changed the stylus to `scoped`. The Inbox Thing Add uses  `f7.dialog.create` which dynamically appends the dialog to the root of the app, hence its style mustn't be scoped.

This PR moved the `thing-inbox-approve-dialog` back into an unscoped stylus.

Problem caused:

image


florian-h05 (approved):
LGTM, thanks for the fix!

PR #4512, opened 2026-09-14
Navbar & sidebar navigation improvements

florian-h05:
Follow up for #4370.

- Hide navbar/left panel button on mobile (instead, navbar can be opened via swipe).
- Only enable swipe gesture for sidebar on mobile.
- Limit swipe gesture area to left screen border (https://v7.framework7.io/docs/panel#panel-parameters).


digitaldan:
Question, how will a user know to swipe?  Also b/c i use floorplan views often, swiping does not work for me as it moves the florrplan view..... so having a menu button somewhere is kinda important.

florian-h05:
The menu button is still present where it was present as of openHAB 5.1, so on all pages where you don't have a back button. If there is a back button, you normally don't have a menu button on iOS. We just remove the case where both back and menu button are shown on mobile. The swipe gesture only triggers on the screen border, I haven't tested it with floor plans though.
Maybe you just want to try it out, I guess that's the easiest way to get an impression of it ;-)

PR #4513, opened 2026-09-14
Consolidate breakpoints & DRY breakpoint handling

florian-h05:

PR #4514, opened 2026-09-14
parameter-text: Fix compiler warn after #4503

florian-h05:


bundles/org.openhab.ui/web/src/components/config/controls/parameter-text.vue
```diff
@@ -80,13 +80,13 @@
...
 
-<style lang="stylus">
+<style scoped lang="stylus">
 .parameter-text
   .network-address-text
     &:hover
       cursor pointer
 
-  :deep(.item-content)
+  ::deep(.item-content)
```
jsjames, line 89: Is the double colon needed (or valid).  I couldn't find where it is valid to use 2 colons on deep?

florian-h05: It only works with the double colon, not sure why though. Might be stylus ...
IIRC when only using a single colon, there still is a compiler error.

jsjames: hmm - I just tried to build with the double colon and actually get this warning: 
```
[lightningcss minify] 'deep' is not recognized as a valid pseudo-element. Did you mean ':deep' (pseudo-class) or is this a typo?
18 |    cursor: pointer;
19 |  }
20 |  .parameter-text[data-v-fa10a54c] ::deep(.item-content) {
   |                                          ^
21 |    align-items: flex-start !important;
22 |  }
```

However, I can't seem to reproduce the original warning you were seeing (with only a single colon)?

florian-h05: `::deep` works, `:deep` doesn't work.
I guess this has something to do with Stylus. Let's wait for merge of #4513 and then look into configuring lightningcss accordingly.

florian-h05: BTW, the original warning was:

```
[lightningcss minify] 'deep' is not recognized as a valid pseudo-class. Did you mean '::deep' (pseudo-element) or is this a typo?
18 |    cursor: pointer;
19 |  }
20 |  .parameter-text :deep(.item-content) {
   |                        ^
21 |    align-items: flex-start !important;
22 |  }
```

So they are quite contradicting ...

florian-h05: Okay, now the behavior gets even weirder after the rebase.
I can now remove the ::deep ... flex-start !important block and the layout stiil looks good. Can you please verify @jsjames?

jsjames: I still get this error when I build with the double colon:
```
[lightningcss minify] 'deep' is not recognized as a valid pseudo-element. Did you mean ':deep' (pseudo-class) or is this a typo?
18 |    cursor: pointer;
19 |  }
20 |  .parameter-text[data-v-fa10a54c] ::deep(.item-content) {
   |                                          ^
21 |    align-items: flex-start !important;
22 |  }
```

In fact, I verified (in Chrome) that the align-items style is not even applied to the .item-content (with a single colon, it does get applied). That may be why removing the !important didn't effect anything?

Not sure if there is a some nuance difference between our setups?  I did rm node_modules and ran an npm i just to make sure I was using the right versions of everything.

jsjames: Regarding whether the flex-start is needed or not, it vertically centers the "show/hide password" icon in the item-content div. Without it, the icon is at the top.  There are likely different ways to accomplish this.

jsjames: I did confirm my lightningcss is version 1.33.0 (which is the latest)

florian-h05: I also get this error when building with the double colon, so that's similar for us.

> Without it, the icon is at the top. There are likely different ways to accomplish this.

I can remove that full ::deep style without any visual change, don't understand what's going on. Do you need that style to center the icon?

jsjames: I think we can remove it at this point.  I just tried again and (not sure exactly what is happening), when the style is enabled, it does not center the show/hide icon.  Let's remove for now and see if it causes any issues down the road.

PR #4517, opened 2026-09-15
Revert patch for UIComponent and RootUIComponent typescript definitions

jsjames:
Only typescript changes
Reverted patch for UIComponent and RootUIComponent after core PRs #5821 and #5791
Regenerated rest client


florian-h05 (commented):
LGTM!

PR #4528, opened 2026-09-20
Sidebar: Let native apps add their entries

stamateviorel:
inside the android app the only way from main ui to the rest of the app (other servers, sitemaps, notifications, app settings) is the exit to app button on the home page. in openhab/openhab-android#4048 i tried a toolbar for that and maniac103 said he prefers the app adding its entries to main ui's own drawer, the same idea came up in openhab/openhab-android#3706. this is the main ui side of that.

it is three small calls, all optional: main ui calls `OHApp.menuReady()`, the app answers with `MainUI.setAppMenu({ title, items })` and main ui calls `OHApp.menuItemSelected(id)` when an entry is tapped. nothing needs a return value so it should also work for the ios app. the entries show up as an own section in the sidebar, an entry can have an icon, a small text under the title and one level of children (used for the server list). when an app provides a menu the exit to app button is hidden. apps that don't implement it see no change.

i have the android side working in a test build against this, screenshot below. i will open that pr once the interface here is ok for you.
Screenshot_2026-09-20-08-45-55-932_org openhab habdroid beta


florian-h05 (changes requested):
Before we review or merge this PR, there has to be a discussion with our iOS & Android app maintainers. The iOS app maintainers are also working on improving the app experience and have a completely different proposal.

stamateviorel:
didn't know about the ios work, thanks for pointing to it. to get everybody in one place like you asked: @maniac103 @mueller-ma for android, @timbms @DigiH @digitaldan for ios.

the problem is the same on both apps, inside main ui you can't reach the rest of the app (other servers, sitemaps, notifications, app settings). on android maniac103 asked for the app adding its entries to main ui's own drawer (openhab/openhab-android#4048, also openhab/openhab-android#3706), that is what this pr does on the main ui side. the ios app goes the other way with a native toolbar menu that also lists the main ui pages.

i'm fine with either direction, i have the android part working both ways in test builds so i can adapt it to whatever you agree on. would just be nice if both apps end up behaving the same.

maniac103:
> The iOS app maintainers are also working on improving the app experience and have a completely different proposal.

Got a reference/link for that? I discussed this with @digitaldan some time ago ([here](https://github.com/openhab/openhab-android/pull/3706)), and we noticed iOS is less likely to need this because their app drawer is on the right side, not the left one. That doesn't change the necessity on Android though, and I'm curious what other proposal they have.

florian-h05:
Yes, will provide this soon, probably create a central discussion for that here and link to it.

florian-h05:
See https://github.com/openhab/openhab-ios/discussions/1315#discussioncomment-18455690. Probably best to open a new discussion in the openhab-webui repo ...

stamateviorel:
read the ios discussion, makes sense. fine for me to move this onto the message based api instead of adding more single functions, i can test the android side when there is something to test.

